PR #376, opened 2015-01-27
Disallow access to unpublished events

RichieB2B:
When events are unpublished they are visible for all MISP users. We want to approve submissions before they are visible. This patch disallows access to unpublished events, except for users of the same organization or site admins.


iglocska:
This is actually intended functionality. The current best practice way to hide an event before it is published is to leave the event setting on "Your organisation only" and only change it to the final distribution setting before publishing it. This allows organisations to co-operate on ongoing unpublished analyses.

elhoim:
Both type of cases have a use depending on the way for the organisation(s)
to work.
Maybe isolating it as an instance-level option?

iglocska:
That does indeed sound like a good option, though having an optional switch on the access control can be a recipe for issues down the road (it definitely makes testing the access control more complex).

RichieB2B:
It shouldn't be too hard to make this behavior dependent on a configuration option. MISP.unpublishedprivate with a default of 'false' comes to mind. I'll try to add this in the coming week.

iglocska:
Excellent, thanks!

RichieB2B:
This branch now implements MISP.unpublishedprivate as well. It seems to work nicely on my test system.

iglocska:
Great, thanks a lot, will go through merging all of these bigger changes before the next release!

PR #377, opened 2015-01-27
Added INSTALL files for CentOS 6 and 7

RichieB2B:
Maybe after this we should rename the original INSTALL.txt to INSTALL.ubuntu.txt


iglocska:
Excellent, tried it out, worked perfectly.

Issue #261, opened 2014-05-15
Notification when a new comment is added to an event


adulau:
It would be great to have a notification mechanism when a comment is added to an event. The notification should be only sent to the owner of the event and the ORG(s) who did a comment.

RichieB2B:
Is this already in development? If not, I might work on a patch for this. We have a low volume MISP server and really require E-mail alerts when something happens.

iglocska:
That would be fantastic! It's on the "list" but it's a long list with a lot
of stuff having priority :)

RichieB2B:
I'm about to submit this patch. Should it respect the autoalert flag? So only send out E-mail if a use has this flag set? I am implementing this in the Post Model, so this covers both comments on Events and discussion Threads.

In my view it does not make sense to use the autoalert flag for Event owners, they should want to get notified when a comment on their Event is posted. I am torn on discussion threads. There it could be useful to not send alerts if someone doesn't want to be spammed.

Rafiot:
Pull request, merged, closing.

PR #379, opened 2015-01-28
Send E-mail notifications for new posts in discussion and event threads

RichieB2B:
This is a fix for issue #261 
This does not use autoalert, but I can add this if required.


iglocska:
Great, will merge this later on into develop before the next release!

Issue #323, opened 2014-11-18
Add MISP attribute comment as STIX indicator description 


RichieB2B:
Currently the MISP attribute comments are lost when exporting events to STIX. Please consider this patch that preserves them in the STIX indicator description field.
```
diff --git a/app/files/scripts/misp2stix.py b/app/files/scripts/misp2stix.py
index e6dacb3..b83adae 100644
--- a/app/files/scripts/misp2stix.py
+++ b/app/files/scripts/misp2stix.py
@@ -256,6 +256,8 @@ def generateThreatActor(attribute):
 def generateIndicator(attribute):
     indicator = Indicator()
     indicator.id_="example:indicator-" + attribute["uuid"]
+    if attribute["comment"] != "":
+        indicator.description = attribute["comment"]
     setTLP(indicator, attribute["distribution"])
     indicator.title = "MISP Attribute #" + attribute["id"] + " uuid: " + attribute["uuid"]
     confidence_description = "Derived from MISP's IDS flag. If an attribute is marked for IDS exports, the confidence will be high, otherwise none"
```

Btw, indicator.title is set twice to the same value, first in generateIndicator() and then again in handleIndicatorAttribute().

RichieB2B:
I think I have mapped all possible comments into STIX, except for the victim Identity which is created with resolveIdentityAttribute() in misp2ciq.py. The STIX/CIQ datamodel doesn't seem to have an appriate field in this case.

RichieB2B:
Fixed in c3615ef

PR #382, opened 2015-01-29
Add MISP attribute comments as STIX descriptions

RichieB2B:
This is a fix for #323 

It maps all possible attribute comments into STIX, except for the victim Identity which is created with resolveIdentityAttribute() in misp2ciq.py. The STIX/CIQ data model doesn't seem to have an appropriate field in this case.


RichieB2B:
Fixed in 98f65aae096a1866e0ca640af7a8663b5590020e

Issue #398, opened 2015-02-11
index.php not executing after installation


andrew-morris:
I'm installing MISP on Ubuntu 14.04. I followed the instructions perfectly. The only 

The only difference between my setup and the documentation was that my apache root was located at /var/www/html for some reason. I made sure to change that in each command I executed.

Now, after installation, attempting to access the site in my browser yields the plaintext PHP code, but it doesn't seem to be executing.

![image](https://cloud.githubusercontent.com/assets/3838808/6158731/71dd3d60-b21b-11e4-9c21-35a01cf910cc.png)

I have ensured that all permissions are set properly, but PHP does not appear to be executing.

elhoim:
Do you have *libapache2-mod-php5* package installed? 
And is the php module enabled by apache (you can check using *a2enmod*?

andrew-morris:
@elhoim that worked like a charm, thanks so much! don't know how I missed that...

elhoim:
Yeah I had the problem doing an install really from a scratch ubuntu 14.04 without anything installed.

PR #400, opened 2015-02-12
Added libapache2-mod-php5 as a dependency

elhoim:
Added libapache2-mod-php5 as a dependency, so that even installs from bare ubuntu 14.04 work.
Related to closed issue #398


iglocska:
Good idea will add it!

PR #412, opened 2015-02-24
Use attribute uuid/timestamps for STIX id's/timestamps

### LINKED ISSUE #397: Use constant id's for STIX observables and CYBOX objects
In the STIX exports, the id's of indicator:Observable and cybox:Object elements are generated by the create_id() helper function. This function creates id's containing a random uuid. This makes deduplication of attributes in a target system harder than it should be.

Since in the STIX export each stixCommon:Indicator contains a single indicator:Observable which contains a single cybox:Object, can't we just use the attribute['uuid'] for all of them?

RichieB2B:
I found a way to close #397 
Still needs to testing to see if it covers all corner cases

### DIFF
--- a/app/files/scripts/misp2cybox.py
+++ b/app/files/scripts/misp2cybox.py
@@ -187,13 +187,20 @@ def createArtifactObject(indicator, attribute):
 def returnAttachmentComposition(attribute):
     file_object = File()
     file_object.file_name = attribute["value"]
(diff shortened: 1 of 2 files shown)


RichieB2B:
The commits of today take care of the last corner cases and will prevent observable id clashes. The only random id now is the Package id in stix:STIX_Package.

PR #422, opened 2015-03-05
Changed regex to validate date and make sure the right dates are validated

stevengoossensB:
Changed regex to validate date and make sure the right dates are validated

Changes to be committed:
	modified:   AttributesController.php


adulau:
@iglocska Do you think it could be possible to merge this as a hotfix?

iglocska:
The date part definitely, the changes to the threat level would change the
meaning of a lot of the data in there though...

adulau:
Sorry my bad. I was only talking about the date verification. We can discard the other part, I just got the confirmation from @stevengoossensB Thanks a lot.

iglocska:
Sounds good, I am doing some more changes to it as other exports are
currently not validating the from / to field for the API at all, I'll have
a hotfix with this and some other things up later tonight!

iglocska:
Added a while back

PR #433, opened 2015-03-13
update to install howto

deralexxx:
remove of -p password in order to avoid having the password in your bash history and the install command for postfix


iglocska:
Makes sense!

Issue #368, opened 2015-01-20
Modify SyncTool to allow syncrhonisation from behind proxy


elhoim:
Allow SyncTool to support optional arguments for connecting to MISP instances for pulling data from behing a proxy.

Restrict the configuration of this to instance admins, and only a single global value.

RichieB2B:
I remember this being discussed before. Is this already in development? If not, I might take a look at this.

RichieB2B:
Looks like this implementing this will be complicated by issue https://github.com/cakephp/cakephp/issues/2057 that has not been fixed just yet.

elhoim:
I quickly checked the cakephp issue you mentionned, and it seems to be implemented in a commit mentioned in it. However I am a bit lost with the branches and I am not sure it is now present in the latest stable version of cakePHP.
If it is, can you please let me know, so that I can close this issue?

RichieB2B:
The SSL via proxy is fixed in CakePHP 2.6.4. MISP merge request #443 also includes this fix for CakePHP. This issue can closed.

elhoim:
Cool, thanks!

PR #438, opened 2015-03-18
Proxy support for SyncTool

RichieB2B:
Fix for #368. If you need HTTPS support via a proxy you will need https://github.com/cakephp/cakephp/pull/6098


RichieB2B:
Will redo on right branch

PR #464, opened 2015-04-17
Certificate Authentication plugin for MISP

capile:
This is a proposal for a new authentication plugin using client SSL certificates. I tried to be the least intrusive as possible at the MISP code (leaving everything configurable), but still I had to make a minor update to the AppController.

The plugin also has some options to fetch data from a REST API and even create users on demand — if you prefer, I can make these procedures as separate plugins.

I may also upload afterwards more detailed documentation, such as how to set up CACert (which issues free, public SSL certificates) as a alternative authentication.

Our main goal is to use FIRST.Org client SSL certificates for authentication on a beta MISP service.

Thanks in advance,

Guilherme Capilé


capile:
Removed session handling from the plugin and moved it to the AppController, so that the gpgkey could be properly unset. Also made that update on the token authentication — that was saving the GPG key with the session.

PR #484, opened 2015-05-11
Note salt key length requirement to INSTALL.md

molnarg:


elhoim:
According to [CakePHP doc for 2.x branch](http://book.cakephp.org/2.0/en/core-utility-libraries/security.html), it seems it needs to be at least 32 bytes, not exactly 32 bytes.

PR #516, opened 2015-06-04
CakePHP 2.6.7 requires the php mbstring extension

RichieB2B:
- on CentOS this is a separate package php-mbstring
- on Ubuntu this is part of libapache2-mod-php5


iglocska:
Nice catch!

PR #541, opened 2015-06-09
Adding specific cybox version when checking out

elhoim:


iglocska:
Already fixed a while back

PR #542, opened 2015-06-10
Only log exception for not found only if the debugging mode is enabled

elhoim:
Only log exception for not found shadowAttributes/proposals by EventUuid only if the debugging mode is enabled.

This would have save me 20.000 lines of log per month... :)


elhoim:
Feel free to delete the branch when merged or when pull request is declined.

elhoim:
Solved by https://github.com/MISP/MISP/commit/a93807439d987543aa4f1b5e1d5479f6e0c83b55 change to core.default.php

PR #547, opened 2015-06-15
Fix STIX export date ranges

RichieB2B:
The from/to/last parameters for STIX export did not work. Now they do. 

I also clarified the automation documentation a bit, since reading the code to figure out how to build the proper export URL should not be required. ;-)


iglocska:
Excellent, thanks, will merge it asap!

PR #567, opened 2015-07-13
Use setupHttpSocket for fetchPGPKey

RichieB2B:
Using setupHttpSocket will ensure the HttpSocket proxy configuration is properly set up.


iglocska:
My bad, thanks for the PR!

PR #569, opened 2015-07-16
Corrected the recursive permission setting

elhoim:
Corrected the recursive permission setting from root to www-data


RichieB2B:
Do we really want to advocate making all of the MISP source writeable for the uid that is running the web server? There are only a few directories where that is required. See the CentOS install files.

elhoim:
Indeed bad idea. Sorry.
Withdrawing/closing my pull request.

Issue #302, opened 2014-10-19
Few modifications on the INSTALL file


devnull-:
Hello all,

I propose these little modifications on the "INSTALL.txt" file in order to fix several errors during the installation of MISP from scratch (in particular when you want to install python-cybox and python-stix). 
The modifications are attach to this "issue" (git diff).

devnull-

```
diff --git a/INSTALL/INSTALL.txt b/INSTALL/INSTALL.txt
index 3f52d62..982ba22 100755
--- a/INSTALL/INSTALL.txt
+++ b/INSTALL/INSTALL.txt
@@ -12,6 +12,7 @@ Install a minimal ubuntu 12.04-server system with the software:
 - OpenSSH server
 - LAMP server (don't forget php5-mysql)
 - Mail server
+- gcc (mandatory to install python-cybox)
 
 You will get some questions, you will probably want to set:
 - Postfix Configuration: Satellite system
@@ -50,8 +51,11 @@ apt-get install python-dev python-pip libxml2-dev libxslt-dev zlib1g-dev
 cd /var/www/MISP/app/files/scripts
 git clone https://github.com/CybOXProject/python-cybox.git
 git clone https://github.com/STIXProject/python-stix.git
+cd /var/www/MISP/app/files/scripts/python-cybox/
 python /var/www/MISP/app/files/scripts/python-cybox/setup.py install
+cd /var/www/MISP/app/files/scripts/python-stix/
 python /var/www/MISP/app/files/scripts/python-stix/setup.py install
+cd /var/www/MISP/
 
 4/ CakePHP
 -----------
@@ -85,7 +89,7 @@ chown -R www-data:www-data /var/www/MISP
 chmod -R 750 /var/www/MISP
 chmod -R g+ws /var/www/MISP/app/tmp
 chmod -R g+ws /var/www/MISP/app/files
-chmod -R g+ws /var/www/MISP/app/files/scripts/tmp
+chmod -R g+ws /var/www/MISP/app/files/scripts/tmp (create it if the directory is not present: mkdir -p /var/www/MISP/app/files/scripts/tmp)
 
 Red Hat users: Keep in mind that your apache user is called apache, not www-data

```

PR #616, opened 2015-08-26
Added gcc as dependency to be installed for ubuntu1404 and centos7 install instructions

elhoim:
Related to https://github.com/MISP/MISP/issues/302
GCC is needed to install python-cybox and python-stix


Rafiot:
Fix bfeac5ffa4b8c0092648a6f1f91959b2dd9e1f98

PR #629, opened 2015-08-31
Export MISP tags as STIX journal entries

RichieB2B:
For MISP converting them to Journal Entries seems logical. I can do the same for other formats, but we need to agree how to represent the tags in those formats.


adulau:
Thank for the contribution. FYI in 2.4 (as tags will be exportable), this will be adapted to match exportable and local tags.

PR #663, opened 2015-09-16
Fix #654

### LINKED ISSUE #654: Upload API should return the event ID when creating a new event
It would fix that issue: https://github.com/Rafiot/viper/blob/mispadvanced/modules/misp.py#L237

Rafiot:
At least, I think so, please review :)


iglocska:
Woops, you are indeed correct!

PR #679, opened 2015-09-27
Debugging coverage

stevepeak:
Debugging coverage, please do not merge. Thank you!

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,12 @@
+[run]
+branch = True
+source = app
+
+[report]
+exclude_lines =
+    pragma: no cover
+    raise NotImplementedError
+    if __name__ == .__main__.:
+ignore_errors = True
+omit =
+    tests/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -94,5 +94,7 @@ script:
     - coverage run setup.py test
 
 after_success:
-    - coveralls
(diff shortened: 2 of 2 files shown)


stevepeak:
@Rafiot not sure how to produce a php `clover.xml` coverage report. I have no experience with php. With Codecov you can upload both php coverage and python coverage on the same repository.

Our php example is here: https://github.com/codecov/example-php/blob/master/.travis.yml#L15 (you can use the python package, so ignore the bash call)

stevepeak:
I think I'm confused at the issue. I'll await your reply to above to continue. Thank you

Rafiot:
Thank you for your work there, very much appreciated :)

Don't spend time on the coverage of the PHP code itself, as far as I know, there are no tests implemented (we may work on it at some point in the future, but not right now, especially because PHP is a pain).

The plan is to use the Python API: the highest the coverage of the python module is, the better (most of the code of the interface will also be tested)

stevepeak:
Ok. So you are seeking Python coverage. In that case I"m not sure why it is not being included in the `coverage xml`... Are you able to produce a report in your local dev? Can you run your tests then call `coverage report` what files show up?

Rafiot:
Merged in https://github.com/MISP/MISP/pull/840

PR #762, opened 2015-12-16
Use Ubuntu 14.04 instead of 12.04 for travis

Rafiot:
**DO NOT MERGE**


Rafiot:
@iglocska https://travis-ci.org/Rafiot/MISP/builds/97501379#L825
is `app/Console/cake userInit -q ` still supposed to work?

adulau:
Fixed in https://github.com/MISP/MISP/commit/a06da3e90b2fffe32a4f08710deb89ae1dcf8e90. Now for Travis to work, we need ensure that the assert test can be valid for 2.3 installs and 2.4 installs (which is the current error).  @iglocska proposed to maintain two branches in PyMISP (2.3 and 2.4).

Rafiot:
I'm not sure we want to maintain two versions, that sounds like a lot of unnecessary pain. Let's decide if/when we have an incompatible change (I'd rather use deprecated warnings for the 2.3 only methods).

I'll fix the travis thing asap.

Rafiot:
Clean commit => https://github.com/MISP/MISP/pull/807

Issue #799, opened 2015-12-29
New attribute types


FafnerKeyZee:
Hello,

Is it possible to create 2 new attribute types ?
Attribution -> Whois Creation Date
Attribution -> Whois Registar 

Thanks :)

advancedeng:
Indeed, it would be good to have more whois attributes. Cheers!

adulau:
Thank you for the contrib. The pull request has been merged.

PR #800, opened 2015-12-29
Adding more information about Whois

FafnerKeyZee:
A possible modification for ticket #799

We obtain 2 new attributes:

Attribution -> whois-registar
Attribution -> whois-creation-date

The second point can be interesting in case of Threat Intell. Feel free to give your opinion :)


adulau:
Thank you very much for your contribution.

PR #834, opened 2016-01-08
2.4

stevengoossensB:
Changes to also create attributes of types md5,shall and sha256 (without filename|) when adding a malware-sample.


stevengoossensB:
too many commits in pull request

PR #835, opened 2016-01-08
Changes to upload of malware-sample in order to also create attributes

stevengoossensB:
Changes to upload of malware-sample in order to also create attributes of the type MD5,SHA1 and SHA256

### DIFF
--- a/app/Controller/AttributesController.php
+++ b/app/Controller/AttributesController.php
@@ -321,7 +321,7 @@ class AttributesController extends AppController {
 	 */
 	public function add_attachment($eventId = null) {
 		if ($this->request->is('post')) {
-			$hashes = array('md5' => 'malware-sample', 'sha1' => 'filename|sha1', 'sha256' => 'filename|sha256');
+			$hashes = array('md5' => 'md5','sha1' => 'sha1','sha256'=>'sha256');
 			$this->loadModel('Event');
 			$this->Event->id = $this->request->data['Attribute']['event_id'];
 			$this->Event->recursive = -1;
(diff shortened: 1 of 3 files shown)


Rafiot:
Thanks for the pull request.

I'm not sure why you want to do that? The correlation is already working if there is a match on the filename OR on the hash. Adding extra attributes for the hashes without filenames will add noise.

Or do you have a specific usecase? such as exporting hashes and you miss the md5 if they come from a malware-sample?

stevengoossensB:
Hi Raphael,

The main reason is that all our integration works on the export of MD5 hashes (the md5 type). If I run the MD5 export, it doesn’t include the hashes attached to the 'filename|MD5’. It is true though that I could’ve changed the export scripts to include an export for filename|MD5, our the MISP project could include them in the export for the MD5 attribute type..

This however shouldn’t add ‘noise’ to the monitoring, as the hash is exactly the same and malicious, but will indeed fill up the database more. 

Best regards,

Steven

iglocska:
Hello Steven,

would using the /hids APIs be a solution to you? You can also specify MD5 there and it will fetch the hashes from all attributes with md5 hashes in them (md5, filename|md5, malware-sample).

Best regards,
Andras

stevengoossensB:
Hi Andras,

That would do the trick. 

Best regards,

Steven

elhoim:
@stevengoossensB OK if I close the PR?

stevengoossensB:
Yes, no problem.

Rafiot:
Thanks!

PR #854, opened 2016-01-15
Update CentOS documentation

RichieB2B:
Since CentOS 6 comes with php 5.3 and CentOS 7 comes with php 5.4 and MISP 2.4 requires php 5.5+ I had to resort to the RedHat Software Collections to install php 5.6. See https://www.softwarecollections.org/en/scls/rhscl/rh-php56/


iglocska:
Thanks a lot, that looks great!

PR #866, opened 2016-01-20
Updated version check for cybox to be consistent with documentation

elhoim:
Related to installation documentation update recommending to use 2.1.0.12 as cybox version (a23027e)


iglocska:
Thanks!

PR #869, opened 2016-01-20
Change default persistence setting value to true

elhoim:


adulau:
https://meta.wikimedia.org/wiki/Why_persistent_connections_are_bad not really recommended for a very small performance gain. I'll close it. Thank you.

PR #893, opened 2016-01-29
mention install howto in Documentation part

jaegeral:
Spent some minutes to find the documentation how tot install MISP (and it is not mentioned in the PDF btw)


adulau:
Thank you for the contribution.

This was indeed a good idea.

PR #894, opened 2016-01-29
Update INSTALL.ubuntu1404.txt

jaegeral:
change base url should be not optional but required for every installation to be changed


adulau:
Thank you for the update.

PR #897, opened 2016-02-01
Update INSTALL.ubuntu1404.txt

koenigswinter:
30: + gnupg-agent (apt-get install)
178: - su www-data -c 'bash /var/www/MISP/app/Console/worker/start.sh', + sudo -u www-data bash /var/www/MISP/app/Console/worker/start.sh (default Ubuntu installation fails to start the workers after a reboot --> "www-data" has no shell in "/etc/passwd").
220: + pip install redis


iglocska:
Thanks a lot, looks great!

PR #903, opened 2016-02-02
Create INSTALL.debian.txt

jaegeral:
I know there is already an ubuntu document, but still I found value to have it being mentioned that debian is supported as well

(There are also some minor changes to the ubuntu docu, would adjust the ubuntu doc as well)


iglocska:
Oh, excellent, thanks!

PR #908, opened 2016-02-03
Fixed a bug where adding attributes fails if the event has been creat…

sterue:
When creating a new event and shortly after adding attributes it happens that the creation time of the event `$existingEvent['Event']['timestamp']` equals the time of the modification data `$this->request->data['Event']['timestamp']` which is set to 

`$dateObj = new DateTime();`
`$date = dateObj->getTimestamp()`
`...`
`$this->request->data['Event']['timestamp'] = $date`

in which case the condition

`if ($this->request->data['Event']['timestamp'] > $existingEvent['Event']['timestamp']) {`

is false and the operation fails.

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -1191,7 +1191,7 @@ class EventsController extends AppController {
 					// user.org == event.org
 					// edit timestamp newer than existing event timestamp
 					if (!isset($this->request->data['Event']['timestamp'])) $this->request->data['Event']['timestamp'] = $date;
-					if ($this->request->data['Event']['timestamp'] > $existingEvent['Event']['timestamp']) {
+					if ($this->request->data['Event']['timestamp'] >= $existingEvent['Event']['timestamp']) {
 						// If the above is true, we have two more options:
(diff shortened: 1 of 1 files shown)


iglocska:
Just wondering what was the situation that caused this in the first place?

You've mentioned that it happens when you tried adding attributes, how are you adding them? 

There are 2 ways to do this:
1. POST an attribute to /attributes/add/event_id
2. POST an updated event to /events/edit/event_id

The latter can indeed cause issues if the edits are posted in < 1 second intervals, but it is also a mechanism that we use to ensure that we don't get into a loop of tools / MISPs editing the same event over and over (if it's the same age or older, we block it), so editing it might have some other consequences. Would it be an option for you to switch to adding attributes via the attributes/add API?

sterue:
Using pyMISP

``` python
# Get MISP instance that stores the IOCs
inst = self.misp_to_store_iocs

event = inst.new_event(distribution=0, threat_level_id=4, analysis=0, info=event_description)
resp = inst._send_attributes(event, attributes)
```

PR #909, opened 2016-02-03
Fixed a bug where adding attributes fails if the event has been creat…

sterue:
When creating a new event and shortly after adding attributes it happens that the creation time of the event `$existingEvent['Event']['timestamp']` equals the time of the modification data `$this->request->data['Event']['timestamp']` which is set to 

`$dateObj = new DateTime();`
`$date = dateObj->getTimestamp()`
`...`
`$this->request->data['Event']['timestamp'] = $date`

in which case the condition

`if ($this->request->data['Event']['timestamp'] > $existingEvent['Event']['timestamp']) {`

is false and the operation fails.


adulau:
This should already been fixed previously. Could you test a recent version and reopen if the issue is still present for you? Thank you very much.

PR #953, opened 2016-02-18
Changed text Find valid IOCs in search page

elhoim:
Using "Only find IOCs to use in IDS" instead since the IOCs where to_ids=0 are not invalid.
It was confusing to some users.


iglocska:
Thanks!

PR #954, opened 2016-02-18
Clarify documentation for API calls

elhoim:
Clarify which fields of events are used by **to**, **from** and **last** API calls parameters.


iglocska:
Thanks a lot!

adulau:
@elhoim Usually the source of the documentation should be misp-book? Could you update it there too? We will remove the doc from the MISP code base soon and only reference back to misp-book. Thanks a lot.

elhoim:
Oops. Error corrected.
Information added to misp-book

adulau:
Super. Thank you very much.

Issue #999, opened 2016-03-03
Backup script in tools


jaegeral:
I would like to have a backup script in the tools folder.

Based on https://github.com/daverstephens/The-SOC-Shop
i forked it to https://github.com/deralexxx/The-SOC-Shop/tree/master/MISP-Stuff

If it is interesting I am happy to make a PR to get it into the repository.

A

iglocska:
That would definitely be interesting

jaegeral:
awesome, will make a PR later today hopefully.
hope @daverstephens is fine with it

iglocska:
thanks :)

daverstephens:
I'm more than happy but you guys might want to check/test it first as I did that one quite a while back and can't recall if I ever actually finished it properly or not - If it works then brilliant and if not feel free to use it as a base for a script that does :)

jaegeral:
yep already working on that, guess with adding it to the official repo it will get more attention, but I will make sure to have a "warning" included for usage.

Thanks

iglocska:
Just a quick heads-up after a quick look at it, some things that would be handy to include in the exported package:
- sample files
- server certificates for sync connections
- organisation logos (though this isn't as important, but it's still annoying to lose them)

jaegeral:
@iglocska indeed, good idea, will add that, feel free to accept PR already, will do another PR, or just wait till I add the stuff you mentioned, planned to do it next week.

jaegeral:
@iglocska  have a look at https://github.com/MISP/MISP/pull/1001 hope you can merge that PR

PR #1001, opened 2016-03-03
misp backup script

jaegeral:
Adressing https://github.com/MISP/MISP/issues/999

will develop it in https://github.com/deralexxx/misp-backup and push it to the main misp repo on a need to do base

A


iglocska:
Thanks a lot!

PR #1063, opened 2016-04-07
Update UPDATE.txt

jaegeral:
solved an issue I had with "This account is currently not available."

Guess we can also change that in install guide


iglocska:
Thanks a lot!

PR #1069, opened 2016-04-11
Defang hostname attributes

RichieB2B:
IP addresses and domain names are _defanged_ (. replaced by [.]) but hostnames are not. Clicking on hostnames can be just as dangerous. This simple patch also defangs hostnames.


iglocska:
Thanks, good point!

PR #1084, opened 2016-04-13
Model/Attribute.php: remove obsolete HTML-linebreak

rotanid:
this change removes a HTML linebreak from an info pop-up on the add event page, that isn't evaluated but only displayed as text


iglocska:
Thanks a lot!

PR #1085, opened 2016-04-13
add margin between fileupload and submit button

rotanid:
forms that have a file upload input item at the end should have a margin after the upload-item.
(n the long-term, it might be better to use a more bootstrap-based approach for form-styling)


iglocska:
Thank you!

PR #1092, opened 2016-04-15
Change to the upgrade script

stevengoossensB:
Change to the upgrade script in order to create the lastfetchedid column in the servers table


adulau:
Thank you for the contrib but the field is not used any more.  So I assume we can close the PR.

PR #1121, opened 2016-05-05
Update UsersController.php

I-am-Sherlocked:
Grouping by Organization.name will throw a MySQL error 

> Syntax error or access violation: 1055 Expression #3 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'misp.Organisation.id' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by" 

in "Request URL: /users/memberslist" , since Organization.name is not a unique field. Grouping by Organization.id instead will fix the issue.


iglocska:
Thanks a lot!

PR #1123, opened 2016-05-09
Update MYSQL.sql

I-am-Sherlocked:
The job_id returned by CakeResque.php is a hex string, but it's declared as int in the 'tasks' table, causing it to throw an error 

> Error: [PDOException] SQLSTATE[01000]: Warning: 1265 Data truncated for column 'job_id' at row 1
> Request URL: /tasks/setTask

when scheduling tasks.


I-am-Sherlocked:
Actually the job_id should be an int. The value that is being returned and saved is the incorrect one (process_id instead of job_id)

PR #1129, opened 2016-05-17
Casting the sharing_group_id to int value

I-am-Sherlocked:
Saving the sharing_group_id as it is returned by CaptureSG results in 

> Error: [PDOException] SQLSTATE[HY000]: General error: 1366 Incorrect integer value: '' for column 'sharing_group_id' at row 1. 

Wrapping it in intval will insert the correct int value.

### DIFF
--- a/app/Model/Event.php
+++ b/app/Model/Event.php
@@ -1812,7 +1812,12 @@ class Event extends AppModel {
 		}
 		
 		if ($data['Event']['distribution'] == 4) {
-			$data['Event']['sharing_group_id'] = intval($this->SharingGroup->captureSG($data['Event']['SharingGroup'], $user));
+			$sg = $this->SharingGroup->captureSG($data['Event']['SharingGroup'], $user);
+			if ($sg===false){
+				$sg = 0;
+				$data['Event']['distribution'] = 0;
+			}
+			$data['Event']['sharing_group_id'] = $sg;
 			unset ($data['Event']['SharingGroup']);
 		}
 		if (isset($data['Event']['Attribute'])) {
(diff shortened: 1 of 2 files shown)


I-am-Sherlocked:
This is happening when CaptureSG in SharingGroup.pgp returns False, the value is converted to an empty string by php.

iglocska:
Interesting, need to look into this. If it returns false, we should block the event save process.

sfossen:
Looking at it, there are 3 cases:
1. unable to save the sharing group ( SharingGroup->save() )
2. sharing group without enough information to create one ( missing ['Organisation'] and ['SharingGroupOrg'] )
3. if the user is not authorized to see the group (SharingGroup->checkIfAuthorised() )

I'm experiencing 2 in a MISP import of a MISP export in XML, this fix allows the import to complete In the first branch of the SharingGroup->captureSG() it throws an exception when the user is not authorized to create a SharingGroup, perhaps  3 should as well.
And distribution should change to 0 for 2 ?

iglocska:
2 is tricky. Converting it to 0 gives it a different meaning, so I'm a bit torn whether we should do that or outright block it and throw an exception.

sfossen:
Looking at the source MISP where the data is coming from.

There is an internal MISP sharing to an external MISP through an internal sharing group. The external MISP created a passive sharing group, and both the export and sync operations create sharing group information that is not importable nor syncable.

The newly proposed fixes, allow the both operations to proceed to completion. Looking at the permissions of the passive sharing group, it is supposed to be accessible and syncable to all orgs on that MISP.

sfossen:
@iglocska please comment again, a fix is needed or different proposal.

iglocska:
Sorry will try to look into this later tonight.

iglocska:
Can't see anything wrong with it, if issues pop-up we can look at it again. Thanks a lot for the PR and sorry for the slow response.

Issue #749, opened 2015-12-08
Syntax error in query for "Events with proposals"


I-am-Sherlocked:
Error: [PDOException] SQLSTATE[42000]: Syntax error or access violation: 1055 Expression #2 of ORDER BY clause is not in GROUP BY clause and contains nonaggregated column 'misp.ShadowAttribute.type' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by
Request URL: /events/proposalEventIndex

Seems to be an issue with changes in MySQL 5.7.5+ .  Plans to fix?

I-am-Sherlocked:
Also happens when querying for "Members List"

iglocska:
Thanks for the heads up, looks like it's indeed a change with MYSQL 5.7+ (they made only_full_group_by default). I'll try to fix it soon.

PR #1131, opened 2016-05-18
Resolving the sql_mode=only_full_group_by error in Search Log

I-am-Sherlocked:
Similar to pull request #1121 and issue #749, the ID needs to be in group_by to solve this error in /admin/logs/search

> Error: [PDOException] SQLSTATE[42000]: Syntax error or access violation: 1055 Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'Log.id' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by


iglocska:
Thanks a lot

PR #1136, opened 2016-05-18
remove warnings when importing event attributes without distribution set

sfossen:
Warning (2): Illegal string offset 'distribution' [APP/Model/Event.php, line 1810]
Notice (8): Uninitialized string offset: 0 [APP/Model/Event.php, line 1810]
Warning (2): Illegal string offset 'distribution' [APP/Model/Event.php, line 1821]
Notice (8): Uninitialized string offset: 0 [APP/Model/Event.php, line 1821]


iglocska:
Looks good, thanks!

PR #1138, opened 2016-05-19
Resolve only_full_group_by error in filterEventIndex

I-am-Sherlocked:
Event.id required in group by, to resolve 

> Error: [PDOException] SQLSTATE[42000]: Syntax error or access violation: 1055 Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'Event.id' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by

in Request URL: /events/filterEventIndex


iglocska:
Thanks!

PR #1139, opened 2016-05-19
improvements for comments & a regex

rotanid:
- improved comments
- made an ipv4 regex more readable (if that is even possible) and added an explanation to it


iglocska:
Thanks, looks good!

PR #1140, opened 2016-05-19
deleteSelected: fix request method check

rotanid:
if i understood this part correctly, it should check if either POST or AJAX is used.
without the suggested change, the if-clause would always return true and hence the Exception would always be thrown, because the request can't be POST and AJAX at the same time, can it?

### DIFF
--- a/app/Controller/AttributesController.php
+++ b/app/Controller/AttributesController.php
@@ -954,8 +954,7 @@ class AttributesController extends AppController {
 	}
 	
 	public function deleteSelected($id) {
-		if (!$this->request->is('post') || !$this->request->is('ajax')) {
-		//if (!$this->request->is('post')) {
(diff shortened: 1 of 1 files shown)


iglocska:
Delete selected is for POSTing an AJAX request (it's when you check a few rows and click delete on top). The check says that if you've somehow arrived there via a request that is not a POST or not via AJAX then throw an exception.

rotanid:
- so request->is('post') and request->is('ajax') can be true for one single request?
- isn't the is('post') obsolete according to your explanation?
- at least the commented out line below is obsolete ;-)

iglocska:
Yup, both can and should be true! Basically an ajax request can be either get or post, in this case we want to ensure that it is a post request and it is arriving via ajax instead of the user manually loading up the action.

iglocska:
The commented line is definitely obsolete though :)

PR #1150, opened 2016-05-20
improve some text passages

rotanid:


iglocska:
Thanks a lot, sounds a lot less Engrish!

PR #1153, opened 2016-05-20
Handle error in getEventIdsFromServer better

sfossen:
both pull branches handle string error messages, and syncProposals wasn't handling errors.


iglocska:
Looks good, thanks.

PR #1154, opened 2016-05-20
reduce warnings in debug log

sfossen:
- don't query if we don't have the key
- set missing keys to null in foreach

Notice: Undefined index: uuid in [/misp_dir/MISP/app/Model/SharingGroup.php, line 327]
Notice: Undefined index: uuid in [/misp_dir/MISP/app/Model/SharingGroup.php, line 340]
Notice: Undefined index: organisation_uuid in [/misp_dir/MISP/app/Model/SharingGroup.php, line 340]
Notice: Undefined index: created in [/misp_dir/MISP/app/Model/SharingGroup.php, line 340]
Notice: Undefined index: modified in [/misp_dir/MISP/app/Model/SharingGroup.php, line 340]

### DIFF
--- a/app/Model/SharingGroup.php
+++ b/app/Model/SharingGroup.php
@@ -322,7 +322,7 @@ class SharingGroup extends AppModel {
 	}
 	
 	public function captureSG($sg, $user) {
-		$existingSG = $this->find('first', array(
+		$existingSG = !isset($sg['uuid']) ? null : $this->find('first', array(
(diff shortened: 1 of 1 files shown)


sfossen:
@iglocska this removes the query when the missing key would cause the query to return empty result set anyways and quiets the warnings logs.

iglocska:
Sorry for all the delays, going through them now! Thanks a lot for all the PRs!

PR #1163, opened 2016-05-23
misc cleanup round 2

rotanid:
please don't merge without checking and comment on lines that need more work or shouldn't be changed


rotanid:
failed to extend the PR, therefore a new one exists which includes these changes

PR #1165, opened 2016-05-23
misc cleanup round 2

rotanid:
please don't merge without checking and comment on lines that need more work or shouldn't be changed


rotanid:
so i did just rebase this PR:
- contains only clear things now
- split up into multiple commits with clear commit messages
  so the check should be easy, @iglocska 
  i'll open two other PRs with the remaining changes that were here before.

iglocska:
Looks great, thanks!

PR #1166, opened 2016-05-24
Fix for mod_proxy_fcgi + Apache 2.2 REST API authentication

RichieB2B:
When using rh-php56-php-fpm together with mod_proxy_fcgi and mod_fcgid under (the CentOS 6 default) Apache 2.2 all environment variables are stripped. This means the Authorization HTTP header is not passed to PHP as the HTTP_AUTHORIZATION environment variable.

This patch converts the Authorization HTTP header to the AUTHORIZATION CGI parameter and accepts this parameter as an alternative method for authenticating REST API calls.

### DIFF
--- a/INSTALL/apache.misp.centos6
+++ b/INSTALL/apache.misp.centos6
@@ -10,6 +10,7 @@
 	</Directory>
 
 	<IfModule !mod_php5.c>
+		SetEnvIfNoCase Authorization "(.*)" HTTP_AUTHORIZATION=$1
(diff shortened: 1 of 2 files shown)


RichieB2B:
Ok, I've found a better solution that only requires to change the Apache config. Revising this PR now.

RichieB2B:
All done. Using SetEnvIf is much cleaner than resorting to CGI parameters.

iglocska:
Looks good, thanks!

PR #1169, opened 2016-05-24
change default to match check and downloadEventFromServer

sfossen:
use null instead of false to match internal check.


iglocska:
Thank you!

Issue #1107, opened 2016-04-25
TLP:AMBER hardcoded in email subject


klingsor83:
Hi,

why is the TLP:AMBER hardcoded in the email subject for alerts? If you create an event with TLP:GREEN (for example), it could be confused if the email sended has the TLP:AMBER.

Could be possible to send in the subject the tlp tagged in the event? And if the event hasn't the tag, send the email without the TLP.

Regards

cvandeplas:
Idea: use the TLP marking of the event, or the most sensitive TLP of the attribute. 
Or fall back on the default of the platform.
Also the distribution flag, and choose the most restrictive one.

rotanid:
probably fixed by #1240

iglocska:
It's a first step, but let's keep it open for now and then improve it later. Right now it's a server setting, in the future it would be nice to read the tags and look for a TLP tag - if not found fall back to the default setting, as described by @klingsor83

PR #1172, opened 2016-05-25
TLP:AMBER hardcoded in email subject, issue #1107 on github.

cristianbell:
Added a new entry in the server settings set, updated necessary files.


RichieB2B:
This PR adds app/Plugin/CakeResque/Config/config.php which is a configure file that can be modified by the site administrator. Especially on servers running multiple MISP instances. Please remove this file from the PR.

iglocska:
Hello Cristian,

the idea with the request was to instead check the attached tags on the event and see if a TLP tag is found - if yes, use that as the e-mail's TLP, if not revert to a default (which should ideally be based on the event's distribution level). Let me know if you have any questions!

cristianbell:
Hi Andras, 
I know what the idea was, but I've spoken with @cvandeplas (who had the idea) before implementing any changes. 
He said that the best way forward is to add this to the server settings at the moment. Leave it like this for a while and see if a more automated approach is something that the user community really needs. 

:)

app/Controller/EventsController.php
```diff
@@ -2763,8 +2763,17 @@ public function freeTextImport($id) {
 		
 		if ($this->request->is('post')) {
 			App::uses('ComplexTypeTool', 'Tools');
+
+
+            Configure::write('debug', 2);
```
adulau: Enabling the debug is maybe something you would like to avoid in production ;-)

cristianbell: ;) sure, but this is not yet final. There was a part1 before, this would be part0, and part 2 and maybe part3 will follow. Atm I'm saving after each issue or when leaving the office.. :) 
I spoke w/ Andras several times and I'll let him know when it's in a state when it can be inspected..

was a quick fix for a local need, I'll do this in some local file which is never sent 2 github,:D

adulau: Thank you for the feedback. So the best would be separated small and separated PR. It's easier for testing and review. Thank you and enjoy your evening.

cristianbell: :) already spoke w/ Andras about this, I'll create separate branches for separate issues. 
And thanks, have a nice evening and weekend as well!

Rafiot:
First of all, the PR has merge conflict with 2.4, we won't be able to merge it.

Beside that, here are the generic requirements in order to contribute to MISP:
- One Pull Request per fix/feature/change/... to avoid polluting the history: I see you're trying to fix at least two things (email subject and free text import)
- I see you added your config file (containing GPG passwords and salt (!)). If we merge this branch, it will be in the history of the project and anyone can just grab it out, even if you deleted the file in a later commit.
- Please make sure Travis CI works on this request, update the test cases if needed

rotanid:
obsolete, replaced by #1240

Rafiot:
True, closing.

PR #1182, opened 2016-05-26
Allow empty events in pull since they are pushed and importable

sfossen:
Pulling events from a MISP instance didn't match the events imported from that same MISP instance export, nor did it match events published onto that MISP instance and viewable.

Events without attributes:
    1) can be pushed
    2) imported
    3) exported

This fix allows them to be pulled to allow consistency with all other actions.


sfossen:
@iglocska see event 52d64e01-8720-4643-be4f-467f950d2109

iglocska:
Sounds good thanks for the great PRs, it's really appreciated!

PR #1185, opened 2016-05-27
don't query every event for proposals, when you don't have permission…

sfossen:
… to get proposals

A little hacky, but without correct permission, the returning null causes the else case ( Fallback for < 2.4.7 instances )  which then queries every event, for proposals which it doesn't have permission for, so wastes resources on both side.


iglocska:
Thanks, well spotted. This was due to a recent reversal due to an incompatibility with some external tools, well spotted!

PR #1186, opened 2016-05-27
remove deprecation warning.

sfossen:
Deprecated (16384): Using key `action` is deprecated, use `url` directly instead. [APP/Lib/cakephp/lib/Cake/View/Helper/FormHelper.php, line 383]


iglocska:
Great, thanks!

PR #1192, opened 2016-05-31
Removal of obsolete code

rotanid:
this PR removes obsolete code lines and unused functions.
please check, perhaps i didn't see a usage of one of these!

this is based on #1165 , so please merge that before this one!

PR #1193, opened 2016-05-31
add some defaults

rotanid:
this PR adds some default values and return values
please check, i might chose wrong defaults.

this is based on #1165 and #1192 , so please merge these before this one!


rotanid:
@iglocska as you requested, i removed the "returns" from this PR

iglocska:
Thanks, looks good!

PR #1202, opened 2016-06-03
not local and no uuid, it's an invalid organisation

sfossen:
#### What does it do?

#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

sync fails with
[2016-06-01 21:04:26] main.ERROR: {"queue":"default","id":"99b7d5ef61e24389ea2edf8c3f209856","class":"ServerShell","args":[["pull","1","1","full","2075"]]} failed: SQLSTATE[HY000]: General error: 1364 Field 'uuid' doesn't have a default value {"type":"fail","log":"SQLSTATE[HY000]: General error: 1364 Field 'uuid' doesn't have a default value","job_id":"99b7d5ef61e24389ea2edf8c3f209856","time":55606,"worker":"misp:14872"} []

### DIFF
--- a/app/Model/Organisation.php
+++ b/app/Model/Organisation.php
@@ -88,6 +88,7 @@ class Organisation extends AppModel{
 		if (empty($this->data['Organisation']['uuid']) && (isset($this->data['Organisation']['local']) && $this->data['Organisation']['local'])) {
 			$this->data['Organisation']['uuid'] = $this->generateUuid();
 		}
+		if (empty($this->data['Organisation']['uuid'])) return false;
 		$date = date('Y-m-d H:i:s');
 		if (!isset($this->data['Organisation']['date_created']) || empty($this->data['Organisation']['date_created'])) $this->data['Organisation']['date_created'] = $date;
 		$this->data['Organisation']['date_modified'] = $date;


iglocska:
Thanks a lot!

iglocska:
Reverted this, there are valid reasons not to have a uuid (initial org, pull from a 2.3 instance).

sfossen:
this fixes an actual crash, which breaks syncing, because the generateUUID is not called unless it is a locally created organization, which why I added the check and return after the generate should have been called. The other way I saw was to remove that local check, but that seemed more wrong.

sfossen:
@iglocska this blocks the save which fails when the uuid for an org is not set, another fix would be to add a default null to the SQL.

iglocska:
Sorry, am currently travelling abroad so a bit strapped for time to dig too
deep into it, but will definitely try to get around to it soon. A local or
with no uuid should be valid, indeed making the field nullable on an SQL
level would probably be the best solution, but I need to check why you have
failures and why we're not seeing them. Could be that either the SQL
install scripts diverged from the upgrade scripts or something, so best to
figure out how we ended up in this situation.

sfossen:
@iglocska also at FIRST, maybe we can talk if you have free time.

sfossen:
@iglocska I think this was caused by the other patch I submitted where the related event was not adding the UUID to orgs, but if it's possible to not have the UUID because of older versions ,then the NULLable field would does make the most sense.

iglocska:
Indeed, makes sense. Yeah just nudge me whenever you run into me! I am
currently sitting next to the hall where the keynote was on the sofas to
the left when you exit it.

PR #1203, opened 2016-06-03
allow related events to send org uuid, since events send them already

sfossen:
#### What does it do?

#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

There is the potential, that an org shows up in the RelatedEvent before it shows up in an Event and causes sync to fail. Already submitted a pull request to fix the crash, but potential for incomplete data.


iglocska:
Makes sense, thanks!

PR #1205, opened 2016-06-03
Change job_id to varchar to resolve #1180

### LINKED ISSUE #1180: Error When Scheduling Tasks
As explained before in #1123 , task scheduling is currently throwing 

> Error: [PDOException] SQLSTATE[01000]: Warning: 1265 Data truncated for column 'job_id' at row 1
> Request URL: /tasks/setTask

Since the process_id returned by CakeResque.php is a hex string, but TasksController.php is trying to save it as the job_id field in tasks, which is an int.
The actual jobs (scheduled for each server) are created in ServerShell.php. That field save either needs to be dropped, or a separate job needs to be created for that process_id, and the job_id be put in the task.

I-am-Sherlocked:
As mentioned in #1180, every spot that task->job_id is being set:
- app/Model/Task.php
- app/Controller/TasksController.php

it's the returned value from CakeResque::enqueueAt which is the process_id (MD5).
And I think renaming the field to process_id might be more representative of what it is?
#### What does it do?

Fixes #1180 
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/INSTALL/MYSQL.sql
+++ b/INSTALL/MYSQL.sql
@@ -514,7 +514,7 @@ CREATE TABLE IF NOT EXISTS `tasks` (
   `type` varchar(100) CHARACTER SET utf8 COLLATE utf8_bin NOT NULL,
   `timer` int(11) NOT NULL,
   `scheduled_time` varchar(8) NOT NULL DEFAULT '6:00',
-  `job_id` int(11) NOT NULL,
+  `job_id` varchar(32) DEFAULT NULL,
   `description` varchar(255) NOT NULL,
   `next_execution_time` int(11) NOT NULL,
   `message` varchar(255) CHARACTER SET utf8 COLLATE utf8_bin NOT NULL,


iglocska:
I'll have a look into this but I think the solution would be rather to save the job ID correctly instead of the process id.

I-am-Sherlocked:
Yeah I thought about that as well, but that solution would require creating another job, to hold that process_id then saving the job_id in tasks. Seeing that the usual "job" fields (progress, etc) wouldn't be relevant in this case, I thought it would make sense to just store the process_id.

iglocska:
I'll try to look into it more this weekend :)

sfossen:
@iglocska as discussed, on new installs, scheduling new tasks is broken without this.

sfossen:
@iglocska any progress?

sfossen:
added MySQL/AppModel updates to pull #1372, since it would require a version change.

PR #1207, opened 2016-06-04
remove obsolete semicolon after closing curly bracket

rotanid:
command:
`grep -rlI "\s};" | grep "\.php" | while read file; do sed -i -r $'s|(\s});|\\1|' $file; done`


iglocska:
Thanks a lot!

PR #1212, opened 2016-06-04
Model/Event.php cleanup

rotanid:


app/Model/Event.php
```diff
@@ -2557,7 +2551,7 @@ public function stix($id, $tags, $attachments, $user, $returnType, $from = false
 		$tagIDs = $this->Attribute->dissectArgs($tags);
 		$idList = $this->getAccessibleEventIds($eventIDs[0], $eventIDs[1], $tagIDs[0], $tagIDs[1]);
 		if (empty($idList)) throw new Exception('No matching events found to export.');
-		$events = $this->fetchEvent($user, array('idList' => $idList, 'last' => $last, 'from' => $from, 'last' => $last));
```
iglocska, line 2560: This looks like a bug, the second last is a copy paste error and should be to, I'll correct it after the merge.

iglocska:
Looks good thanks.

PR #1213, opened 2016-06-04
Model/Event.php cleanup 2

rotanid:
removes methods and variables that are not in use from Model/Event.php


iglocska:
Looks good, thanks!

PR #1214, opened 2016-06-04
EventsController: optimise file handling

rotanid:
this PR adds error handling to fopen() calls in EventController.php, closes file handles after reading and changes variable names


iglocska:
Excellent, thanks!

PR #1215, opened 2016-06-04
replace exec unzip by ZipArchive class

rotanid:
as using exec() or similar functions that directly call shell commands always carries security problems, these should be replaced by native solutions.
as a first step, this PR replaces the exec("unzip...") in EventsController.php by using the ZipArchive class that is available when PHP is compiled with --enable-zip

this is untested, because i
a) don't have an example GFI sandbox file
b) don't have any experience with CentOS (to know if they have zip enabled in their PHP packages)

therefore, testing and feedback would be appreciated.

afterwards, i intend to also convert the other zip-execs

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -1853,12 +1853,17 @@ class EventsController extends AppController {
 			$result = $zipFile->write($zipData);
 			if (!$result) $this->Session->setFlash(__('Problem with writing the zip file. Please report to administrator.'));
 
-			// extract zip
-			$execRetval = '';
-			$execOutput = array();
-			exec("unzip " . $zipFile->path . ' -d ' . $rootDir, $execOutput, $execRetval);
-			if ($execRetval != 0) {	// not EXIT_SUCCESS
-				throw new Exception('An error has occured while attempting to unzip the GFI sandbox .zip file. We apologise for the inconvenience.');
+			// extract
+			$zip = new ZipArchive();
+			$res = $zip->open($zipFile->path);
+			if ($res === TRUE) {
+				$res2 = $zip->extractTo($rootDir);
+				if ($res2 === FALSE) {
+					throw new Exception('An error has occured while attempting to unzip the GFI sandbox .zip file.');
+				}
+				$zip->close();
+			} else {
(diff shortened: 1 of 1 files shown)


RichieB2B:
According to http://rpm.pbone.net/index.php3/stat/4/idpl/31973116/dir/centos_6/com/rh-php56-php-common-5.6.5-7.el6.x86_64.rpm.html zip.so is provided for CentOS.

iglocska:
We use zip in several parts of the application and sadly several of them cannot be achieved via ZipArchive. We were looking into this a while back but ended up using exec because there is still currently no way of creating password protected zip archives via PHP. 

This might also create an additional dependency and if we only end up replacing the zipping in one place I'd rather keep it as it is. The zipping functionality of MISP has been pen-tested and should be safe as it is (we took several steps to mitigate it).

rotanid:
starting with PHP 5.6, setting a password is possible:
http://php.net/manual/de/ziparchive.setpassword.php

rotanid:
starting with PHP 5.6, setting a password is possible:
http://php.net/manual/de/ziparchive.setpassword.php

although ubuntu 14.04, which seems to be supported, uses PHP 5.5, so my work would be for the trash can :D

perhaps using exec only if PHP < 5.6 and setting a password is needed (so not for unzip)

it might be true, that the zipping/unzipping was pen tested, but i know of many people that disable exec and similar functions completely so even injected code or so can't use it...

iglocska:
You'd think it would do that, if it wasn't PHP. It actually sets the password to be used to decompress the archive (I would like to congratulate the dev that came up with that naming convention).

Right now if you disable exec a lot of functionality in MISP won't work either way (STIX export, background workers, etc), it's a requirement atm.

rotanid:
ok, indeed @iglocska you're right, only unzipping with password is supported.
so zipping would still require exec() and as you want to use similar mechanisms this PR would have to be closed then

adulau:
Just a small note for the future. Maybe the upload of samples should be part of the new import/export modules then we can handle the functionality independently and more safely in the external modules.

iglocska:
That is indeed a good point, that would completely solve the issue.

PR #1216, opened 2016-06-06
chg: remove handling of unsupported OS Windows

rotanid:
as MISP doesn't support Windows Hosts at all (according to @iglocska ), this PR removes the remaining code handling windows hosts


iglocska:
Thanks a lot!

PR #1217, opened 2016-06-06
Zip exec error handling

rotanid:
errors with the exec of the zip command should be handled, not simply ignored


iglocska:
Thanks!

PR #1218, opened 2016-06-06
chg: add newline character before EOF to non-minified (text-)files

rotanid:
to comply with POSIX every (text-)file should end with a newline character.

i did this:
`for f in $(find . -regextype egrep -regex '.*\.(php|js|css|ctp|py|txt|sql|SH)'); do if [ -n "$(tail -c 1 <"$f")" ]; then echo >>"$f"; fi; done`
the only files i excluded after execution from the commit are the minified js/css files


iglocska:
Looks good thanks!

PR #1219, opened 2016-06-06
chg: update jquery to 2.2.4 & jquery-ui to 1.11.4

rotanid:
MISP uses some old versions of jquery/jquery-ui from 2014.
this PR updates these javascript libraries.

according to the release notes, there are no incompatible changes:
https://blog.jquery.com/2016/01/08/jquery-2-2-and-1-12-released/
https://blog.jqueryui.com/2014/06/jquery-ui-1-11-0/


Rafiot:
I see there is also the jquery css file : https://github.com/MISP/MISP/blob/2.4/app/webroot/css/jquery-ui-1.10.4.min.css

This one should probably be updated. 

Also, it might be good to create a bash script to update all the external dependencies at once.

app/View/Layouts/ajaxTemplate.ctp
```diff
@@ -17,9 +17,9 @@
  */
 ?>
 <?php
-//echo $this->Html->script('jquery-2.1.0.min');
-echo $this->Html->css('jquery-ui-1.10.4.min');
+//echo $this->Html->script('jquery-2.2.4.min');
+echo $this->Html->css('jquery-ui-1.11.4.min');
```
Rafiot: btw, this is the CSS, you definitely want to add that file in the commit :)

rotanid:
yes, of course...
i also changed the filename of the CSS so it reflects the theme name, as there are many other themes for jquery ui.

iglocska:
Thanks a lot!

PR #1222, opened 2016-06-06
chg: improve fopen/fread/fwrite handling via Lib

rotanid:
introduces better (if there even was any) error handling to fopen/fread/fwrite.

using a separate private function in two of the three files to reduce code redundancy.


rotanid:
wait with the merge, i'll try to move the functions to a Lib so they can be re-used

rotanid:
i updated this PR, it now uses an external Lib so the functions can be re-used.

as that is my first try at generating a Lib for cakephp, there might be problems with the way i did it.
for example, i was not sure if there are Exceptions allowed in a Lib, but it would have been more code if not (or the error messages would have been more general).

iglocska:
Thanks a lot, working great and very clean!

cristianbell:
great work, thanks!

PR #1223, opened 2016-06-06
fix: move unlink() to correct location

rotanid:
there were situations, where unlink() wasn't called after return or exception so the temporary files would stay in place "forever".

this PR moves them to other locations in the code.


iglocska:
Looks good, thanks

PR #1225, opened 2016-06-06
chg: use escapeshellarg() instead of addslashes() with exec()

rotanid:
there is a specific function escapeshellarg() in PHP for escaping arguments to shell commands.


iglocska:
Looks good, thanks!

PR #1229, opened 2016-06-06
fix: dont override type variable

rotanid:
the type variable is always overriden by "filename|" even if it was set to "other" before.

i'm not 100% sure if my PR achieves the intended behaviour, but at least it doesn't override something.
please comment if this isn't the intended behaviour.


iglocska:
Looks good, thanks!

PR #1233, opened 2016-06-06
chg: remove obsolete variables

rotanid:
removes variables that aren't used anymore after being set


iglocska:
Looks good thanks!

PR #1238, opened 2016-06-06
cleanup obsolete code

rotanid:
again, a PR that removes code that is obsolete as far i can tell


rotanid:
updated this PR to only contain things without any doubt at all to be able to merge it myself

PR #1239, opened 2016-06-07
Fix CertAuth plugin

RichieB2B:
#### What does it do?

The CertAuth plugin returned an user object that was missing the ['Role'], ['Organization'] and ['Server'] elements. This PR fixes that. It also adds documentation for the userDefaults feature.

In my testing the CertAuth configuration was never written to config.php and actually removed when present there. So I used the documentation from README.md and placed it in bootstrap.php instead. Perhaps the example CertAuth configuration in config.default.php should be moved to bootstrap.default.php
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Impeccable timing, thanks a lot!

PR #1243, opened 2016-06-07
Small, non breaking changes to the MISP look and feel

rhaist:
- Removed old school glass stuff from navbars (bye bye)
- Removed blue flame effect from MISP branding (srsly WHY?)
- Minor ajustments to flush the changes globally


iglocska:
Come on, the blue flame was an awesome way to raise your CPU usage!...

Thanks for the changes :)

PR #1244, opened 2016-06-07
fixed compatibility issues between the CertificateAuth plugin and Apache

capile:
#### What does it do?

Fixes compatibility issues between the plugin and Apache. Apache uses commas to separate the properties of the `$_SERVER['SSL_CLIENT_I_DN']` and `$_SERVER['SSL_CLIENT_S_DN']`, and whenever it's forwarded to an external FCGI process, it misses the `$_SERVER['SSL_CLIENT_VERIFY']`. However, the other variables are only filled with valid SSL credentials.
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Perfect, thanks a lot for all the work!

PR #1254, opened 2016-06-12
chg: use escapeshellarg() instead of addslashes() with exec()

rotanid:
there is a specific function escapeshellarg() in PHP for escaping arguments to shell commands.

the first PR #1225 broke zip creation, probably because there were now double quotation marks.
it got reverted with commit 6275e4fcd4640d051f1ce87be6d376c094d5eb13.

so this PR removes the quotation marks, that are obsolete when using escapeshellarg().

it also adds one additional escapeshellarg() to the params that were missing escaping before. that might not be needed at the moment, but it is safer as it would prevent problems if the code that leads to this variable is changed in the future.

as i couldn't find any information on what the problem and error message was that lead to the revert, @iglocska should probably test the upload himself on his dev system before merging this PR.

### DIFF
--- a/app/Model/Attribute.php
+++ b/app/Model/Attribute.php
@@ -1129,7 +1129,7 @@ class Attribute extends AppModel {
 		if ($malware) {
 			$execRetval = '';
 			$execOutput = array();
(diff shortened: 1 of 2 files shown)


iglocska:
Just tested it, seems to work fine, thanks for the fix!

PR #1256, opened 2016-06-12
misc cleanup round 3

rotanid:
this PR removes obsolete code lines and unused functions.
please check, perhaps i didn't see a usage of one of these!

### DIFF
--- a/app/Model/Warninglist.php
+++ b/app/Model/Warninglist.php
@@ -33,10 +33,6 @@ class Warninglist extends AppModel{
 		return true;
 	}
 
-	public function checkValidTypeJSON($check) {
-		return true;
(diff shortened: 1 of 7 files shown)


iglocska:
Makes sense, thanks!

app/Model/Warninglist.php
```diff
@@ -33,10 +33,6 @@ public function beforeValidate($options = array()) {
 		return true;
 	}
 
-	public function checkValidTypeJSON($check) {
```
Rafiot, line 36: Haha :p

iglocska, line 36: "I am sure I'll implement this one day!"

PR #1264, opened 2016-06-14
compress IPv6 addresses on import

### LINKED ISSUE #641: IPv6 addresses compression
It seems that MISP doesn't do IPv6 address compression:

~~~~
2a01:0000::1 
2a01::1 
~~~~

The above IP addresses are synonyms but they appear at two different attributes. The best would be to compress the address before adding them in MISP.

rotanid:
this compresses IPv6 addresses to their short form notation, fixes #641 

perhaps in a next step there should be some on-upgrade-job that compresses every address already in the database or some other way to solve the issue for existing data?

### DIFF
--- a/app/Model/Attribute.php
+++ b/app/Model/Attribute.php
@@ -920,6 +920,12 @@ class Attribute extends AppModel {
 			case 'email-src':
 			case 'email-dst':
 			case 'domain|ip':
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks, looks good!

Issue #1195, opened 2016-06-01
Installation, backup, and restore instructions for MISP on Ubuntu 16.04 with MariaDB


infde6:
Thought I would post this, in case it helps.I know it's not an issue, but didn't know where else to put it.

Would be willing to work with someone to get this polished and into the MISP documentation.

[MISP-Ubuntu-16.04_install.txt](https://github.com/MISP/MISP/files/293301/MISP-Ubuntu-16.04_install.txt)

INSTALLATION INSTRUCTIONS
------------------------- for ubuntu 16.04-server
### but, as always:

!USE AT YOUR OWN RISK!
## 1/ Ubuntu install with LAMP and dependencies

---

Install a minimal ubuntu server system with the LAMP software:

(Adapted from https://www.howtoforge.com/tutorial/perfect-server-ubuntu-16.04-with-apache-php-myqsl-pureftpd-bind-postfix-doveot-and-ispconfig/)
### Make sure your system is up2date:

apt upgrade
### Install depenencies

apt-get install **vim curl gnupg-agent git redis-server zip gcc make sudo binutils openssl imagemagick memcached mcrypt python python-dev python-pip libxml2-dev libxslt1-dev zlib1g-dev**
### Install Postfix

apt-get install **postfix**
You will get some questions, you will probably want to set:
- Postfix Configuration: Satellite system
- postconf -e 'relayhost = your.relay.here'

/etc/init.d/postfix restart
### Install MariaDB (MySQL alternative)

apt-get install **mariadb-client mariadb-server**
### Secure the MySQL installation

mysql_secure_installation
### Install Apache2

apt-get install **apache2 apache2-doc apache2-utils apache2-suexec-pristine libapache2-mod-fcgid libapache2-mod-fastcgi libapache2-modsecurity libapache2-mod-qos**
### Enable modules, settings, and default of SSL in Apache

a2enmod ssl
a2enmod suexec rewrite ssl actions include cgi
a2enmod dav_fs dav auth_digest headers
a2enmod actions fastcgi alias
a2dissite 000-default
a2ensite default-ssl
### Install PhP and dependencies

[...]

PR #1276, opened 2016-06-18
installation instructions for MISP on Ubuntu 16.04

rotanid:
this adds detailed installation instructions for MISP on Ubuntu 16.04 "Xenial"

the instructions are based on those for Ubuntu 14.04 and parts of it on #1195 by @infde6

improvements to both base-instruction-documents:
- use SSL by default
- don't install not-in-use php-packages
- don't depend on root shell, use sudo most of the time

instructions have been tested, although i obviously didn't test every other feature of MISP

### DIFF
--- a/INSTALL/INSTALL.ubuntu1604.txt
+++ b/INSTALL/INSTALL.ubuntu1604.txt
@@ -1,0 +1,269 @@
+INSTALLATION INSTRUCTIONS
+------------------------- for ubuntu 16.04-server
+
+1/ Minimal Ubuntu install
+-------------------------
+
+# Install a minimal Ubuntu 16.04-server system with the software:
+- OpenSSH server
+
+# install postfix, there will be some questions.
+sudo apt-get install postfix
+# Postfix Configuration: Satellite system
+# change the relay server later with:
+sudo postconf -e 'relayhost = example.com'
+sudo postfix reload
+
+# Make sure your system is up2date:
+sudo apt-get update
+sudo apt-get upgrade
+
+
+2/ Install LAMP & dependencies
+------------------------------
+Once the system is installed you can perform the following steps:
+
+# Install the dependencies: (some might already be installed)
+sudo apt-get install curl gcc git gnupg-agent make python openssl redis-server sudo vim zip
+
+# Install MariaDB (a MySQL fork/alternative)
+sudo apt-get install mariadb-client mariadb-server
+
+# Secure the MariaDB installation (especially by setting a strong root password)
+sudo mysql_secure_installation
+
+# Install Apache2
+sudo apt-get install apache2 apache2-doc apache2-utils
+
+# Enable modules, settings, and default of SSL in Apache
+sudo a2dismod status
+sudo a2enmod ssl rewrite
+sudo a2dissite 000-default
+sudo a2ensite default-ssl
+
+# Install PHP and dependencies
+sudo apt-get install libapache2-mod-php php php-cli php-crypt-gpg php-dev php-json php-mysql php-opcache php-readline php-redis
+
+# Apply all changes
+sudo systemctl restart apache2
+
+
+3/ MISP code
+------------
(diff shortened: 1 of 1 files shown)


cristianbell:
great work! 
Q: are we switching the installation guide to use mariadb now instead of mysql? 
I know that it is just a fork of mysql and the story behind, but lots of organizations who use MISP have lists with approved software and I'm afraid madiadb will not be on them. 
Providing at least: 
Option A: sudo apt-get install mariadb-client mariadb-server... and 
Option B: sudo apt-get install mysql-server..  
would be good I think. 

Also, is there something similar to mysql_secure_installation for mariadb? As this should be part of all our installation guides.

iglocska:
Cristian, the two are 100%compatible. If your organisation prevents you
from using mariadb, don't use it, it's fine. I'd pick Maria db over MySQL
though so i prefer this.

iglocska:
For example, for those using centos, there is a good chance they are using
using mariadb too.

adulau:
Thank you very much for the contribution. Regarding database back-ends, have you made some tests with PostgreSQL?  Maybe you want to have a chat with Michael K. who told us that they have an interest into it.

iglocska:
That would indeed be fantastic. Cakephp should support it out of the box,
but just in case a test to verify it would be helpful. If it works we could
change the default recommendation.

cristianbell:
Sure, that would be smth cool, though it’d pb not benefit my instances. :( 

But before we jump to Postgresql I think we should try to be little more consistent: all installation guides talk about mysql, except the one for Ubuntu 16.04. 

They are just installation guides and the app works just as good with mysql as with mariadb. Hence I’d recommend having one of them as the standard install commands, while the other at least mentioned as a comment: #if you want to install …. just run … instead. 

> From: Andras Iklody [mailto:notifications@github.com] 
> Sent: Sunday, 19 June, 2016 12:08 PM
> To: MISP/MISP MISP@noreply.github.com
> Cc: Cristian Bell cristian.bell@freenet.de; Comment comment@noreply.github.com
> Subject: Re: [MISP/MISP] installation instructions for MISP on Ubuntu 16.04 (#1276)
> 
> That would indeed be fantastic. Cakephp should support it out of the box, but just in case a test to verify it would be helpful. If it works we could change the default recommendation.

iglocska:
Check the centos ones again. They have been mariadb for ages, since that is
now the default db coming with the distro, whilst it is MySQL for Ubuntu
14.04. Since the two are compatible, I don't see an issue.

cristianbell:
:) 

I also have no issue w/ either of them, but: 
- we are not consistent
- we should really offer both options in the installation guide, and let the installer go for one or the other. 

As for mariadb in CentOS, I can’t find any mention of it here: https://github.com/MISP/MISP/blob/2.4/INSTALL/INSTALL.centos6.txt

rotanid:
@iglocska i'll check to update ubu14.04 and do a separate debian8 setup guide using mariadb and the other stuff from this one.

@cristianbell "mysql_secure_installation" is for both, mariadb and mysql, because they are compatible. also, mariadb still uses "mysql", "mysqldump", "mysqld" and so on, because it's a drop-in.
if you want the setup guides to contain both alternatives, it's just adding one line.
But i don't feel comfortable recommending MySQL for Open Source projects, especially since there is the more open and community based MariaDB for a while now.
At my previous job we switched dozens of servers from MySQL to MariaDB when upgrading from Debian 7 to 8 and we had not a single problem.

iglocska:
Spot on @rotanid. Also, we are actually very consistent with the default database installations that come with each distro. So mysql on Ubuntu + centos6 and mariadb on centos7.

I would also pick mariadb over mysql whenever anyone asks for a recommendation, and since most people installing misp will most likely know about mariadb being a drop in replacement, i wouldn't worry about changing the docs for a false sense of consistency.

R6OX (commented):
its not supported and will not intall to the 16.04

PR #1282, opened 2016-06-20
remove some obsolete FIXME notes

rotanid:
this PR removes some comments containing FIXME notes about checking filenames and paths.

i have checked the variables that are being combined and asked @iglocska about some of them, if they can be trusted (as in: they aren't user input), and there don't seem to be issues, so the comments are obsolete.


iglocska:
Thanks again for looking into it and all the cleanup!

PR #1283, opened 2016-06-21
Push events to server with push rules on non-exportable tags

RichieB2B:
#### What does it do?

It fixes a bug where a push to a server with a push rule based on a non-exportable tag would never happen. Using includeAllTags is safe because __updateEventForSync() called by restfulEventToServer() filters out the non-exportable tags just before the REST HTTP call.
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Nice catch, thanks!

PR #1284, opened 2016-06-21
install script for ubuntu14, based on da667 (deusexmachina667@gmail.c…

cristianbell:
Install script for ubuntu14, based on da667 (deusexmachina667@gmail.com) work. He has all the credit, I just changed the script a little and tested it on a Ubuntu 14.04 VM. 
#### What does it do?

If it adds a new feature, install script. 
#### Questions
#### Release Type:
- [ ] Major
- [x] Minor
- [ ] Patch


INSTALL/auto-MISP-ubuntu14.sh
```diff
@@ -0,0 +1,442 @@
...
+	echo "	SSLCipherSuite EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH:EECDH+ECDSA+AESGCM:EECDH+aRSA+AESGCM:EECDH+ECDSA+SHA384:EECDH+ECDSA+SHA256:EECDH+aRSA+SHA384:EECDH+aRSA+SHA256:EECDH:EDH+aRSA" >> /etc/apache2/sites-available/misp-ssl.conf
+	echo "	SSLProtocol All -SSLv2 -SSLv3" >> /etc/apache2/sites-available/misp-ssl.conf
+	echo "	SSLHonorCipherOrder On" >> /etc/apache2/sites-available/misp-ssl.conf
+	echo "	Header always set Strict-Transport-Security \"max-age=63072000; includeSubdomains; preload\"" >> /etc/apache2/sites-available/misp-ssl.conf
+	echo "	Header always set X-Frame-Options DENY" >> /etc/apache2/sites-available/misp-ssl.conf
+	echo "	Header always set X-Content-Type-Options nosniff" >> /etc/apache2/sites-available/misp-ssl.conf
+	echo "	SSLCompression off" >> /etc/apache2/sites-available/misp-ssl.conf
+#	echo "	SSLSessionTickets Off" >> /etc/apache2/sites-available/misp-ssl.conf
+	echo "	SSLUseStapling on" >> /etc/apache2/sites-available/misp-ssl.conf		
+	echo "	SSLOpenSSLConfCmd Curves secp384r1" >> /etc/apache2/sites-available/misp-ssl.conf
```
cristianbell: wait for this, now I restarted the VM and the apache doesn't start anymore.. :(

Rafiot:
When you get it to work, please update the travis file to use this script, it will be cleaner than my dirty set of 20+ commands :)

cristianbell:
So this is the _final version. (_Probably might change it after users feedback): 
- mysql passwords are asked from the command prompt;
- the user is asked if the http to https should be set up;
- the user is asked if https should be set up;
- CA variables (ctry, region, city, org) are also asked from the user input - only if the user said he wants the https to be set up. 
  more ideas for the future: 
- short components and app test script;
- an upgrade script (os+app+restarting apache&workers).

cristianbell:
Cool idea! 

(I won’t get to it very soon though, I’m a little backlogged w/ support atm..:| )

iglocska:
OK, after some testing here are the issues that I've found:
1. The script refers to autosnort in the on-screen instructions
2. It assumes that the MySQL installation has no root password set. If you have it set the installation will fail.
3. After restarting the server post-setup, I get the following message:
   python3: Can't open file'/opt/misp_mod/misp-modules/bin/misp-modules.py: No such file or directory
   Seems like the misp modules are not correctly installed
4. CakeResque seems to revert to Redisent, this is a sign that PHP-redis is not correctly installed (this will result in a bunch of error messages in the logs and the schedule worker not working)
5. After system restart I see 5 live workers and 5 dead ones started by a different user.

Let me know if you have any questions while trying to resolve these / if you cannot reproduce any of the issues!

Otherwise looks good!

rotanid:
if #1353 gets merged, this script should also only use the latest tagged release and not the latest commit

Rafiot:
16.04 is preferred for new install.

PR #1286, opened 2016-06-21
remove obsolete upload function from ShadowAttribute

rotanid:
this removes the uploadAttachment function from Model/ShadowAttribute.php

while digging through the code i thought that it might be unused, @iglocska wasn't sure but told me how to test for it.
i did a fresh setup, created an event as admin and created a new user with role "user" and in a different organisation.
then i logged in as the user and uploaded an attachment to said event which then showed up as a proposal for the site admin and can be downloaded again - although i already removed the code.

so it seems even for proposal uploads the uploadAttachment function from Model/Attribute.php is used, which looks almost the same.

it might be good though, if someone else tests this, too.

### DIFF
--- a/app/Model/ShadowAttribute.php
+++ b/app/Model/ShadowAttribute.php
@@ -398,66 +398,6 @@ class ShadowAttribute extends AppModel {
 		}
 	}
 
-/**
- * add_attachment method
- *
- * @return void
- */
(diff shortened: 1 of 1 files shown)


iglocska:
Tested the proposal attachments, they indeed go through the attribute attachment upload now and the code is obsolete, thanks!

PR #1288, opened 2016-06-22
updated CentOS 7 INSTALL guide

peasead:
Added a few clarifications for external browsing (regarding hostname), updated the PHP Redis installation instructions, updated from mysql to mariadb, and added the "homepage".


iglocska:
Looks good, thanks!

PR #1294, opened 2016-06-23
changing the loading bar update interval from 1000 to 3000 (as it is …

cristianbell:
#### What does it do?

In the event\export: it changes the loading bar update time from 1000 to 3000. This decreases the requests number. In the Jobs list it is already 3000. 
#### Questions
#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch

…also in the jobs list);


iglocska:
Looks good, thanks.

PR #1297, opened 2016-06-23
fix minor css and html issues

cristianbell:
#### What does it do?

fixes a cpl of css and html issues; 
#### Questions
#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch


iglocska:
Looks good, thanks.

Issue #1300, opened 2016-06-24
Pull and publish use the same worker


RichieB2B:
My MISP is syncing with a large remote server. Each pull cycle from this server takes more than 1 hour. During this time publishing event are queued until the pull completes. A publishing delay of 1 hour is not acceptable for us.

My proposal to fix this is to rename the _email_ worker to _prio_ worker and move the publishing of events to the _prio_ worker. If this approach is approved I can code up a PR for this.

iglocska:
This sounds like a great idea indeed. Moving the publishes to a separate
worker is a great idea!

iglocska:
Just one more idea: Maybe instead of renaming the email worker, creating an additional worker with its own queue for some prio tasks would make more sense. I think that in our case for example the e-mail worker is the one that tends to end up with the lognest queues during a sync.

RichieB2B:
Ok, I will add another worker. After this is implemented, would adding a prio flag to the servers make sense? So most servers use the default worker but some can use the prio worker. Or is doing 2 pulls in parallel a no-go?

rotanid:
this got fixed and merged by #1303 , please open a new issue if you want to discuss/suggest something else or similar.

PR #1303, opened 2016-06-24
Add prio worker

RichieB2B:
#### What does it do?

This PR is a fix for https://github.com/MISP/MISP/issues/1300 and adds a new prio queue for publishing events.
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Perfect, thanks a lot. Made some minor changes with a fallback in case the users don't restart the workers.

PR #1307, opened 2016-06-27
fix: move case statement and add break

### LINKED ISSUE #1306: Undefined offset: 1 in [/var/www/MISP/app/Model/Attribute.php, line 923]
### Work environment

| Questions | Answers |
| --- | --- |
| Type of issue | Bug |
| OS version (server) | CentOS 6 |
| OS version (client) | not applicable |
| PHP version | 5.6 |
| MISP version / git hash | 2.4.48, hash of the commit |
| Browser | not applicable |

I get a log of these lines in resque-worker-error.log :
`Notice: Undefined offset: 1 in [/var/www/MISP/app/Model/Attribute.php, line 923]`
Looking at the code inside the modifyBeforeValidation() function:

```
        public function modifyBeforeValidation($type, $value) {
                switch ($type) {
...
                        case 'email-dst':
                        case 'domain|ip':
                                $parts = explode('|', $value);
                                if (filter_var($parts[1], FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)) {
```

This seems to be caused by an empty attribute $value so $parts gets to be an empty array. Which is strange because value1 of the attributes table is set NOT NULL.

rotanid:
this PR fixes an issue in app/Model/Attribute.php that was introduced by #1264 / 5b390a0e3216b9d1d66327c0832ba3c92562566c and reported as a bug by @RichieB2B , fixes #1306

### DIFF
--- a/app/Model/Attribute.php
+++ b/app/Model/Attribute.php
@@ -918,6 +918,10 @@ class Attribute extends AppModel {
 			case 'tlsh':
 			case 'email-src':
 			case 'email-dst':
+			case 'target-email':
+			case 'whois-registrant-email':
+				$value = strtolower($value);
+				break;
 			case 'domain|ip':
 				$parts = explode('|', $value);
 				if (filter_var($parts[1], FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)) {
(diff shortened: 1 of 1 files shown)


iglocska:
Nice catch, thanks!

PR #1309, opened 2016-06-27
exceptions in _readGfiXml()

rotanid:
this PR makes sure that malformed (perhaps intentionally) xml files lead to exceptions where necessary and necessary default values are always set.

as valid gfi xml files seem to always use integers as process index ids, is_numeric() is called so there is nothing else - especially nothing like "../../../", which might lead to files being created in directories where they shouldn't be.


iglocska:
Looks good, thanks!

PR #1324, opened 2016-07-04
MYSQL.sql cleanup

rotanid:
this PR removes duplicate spaces, uses consistent lowercase for "int" and removes quotes from default values for integers (so no type conversion is necessary)


iglocska:
Looks good, thanks!

PR #1327, opened 2016-07-05
MYSQL.sql cleanup #2

rotanid:
as in MySQL/MariaDB, KEY/INDEX are synonyms, we should use the same everywhere, i went for INDEX

(doesn't apply to PRIMARY KEY, therefore left unchanged)

also, quotes are removed from three more integer defaults that i overlooked in the last PR


iglocska:
Looks good, thanks.

PR #1339, opened 2016-07-07
fix: check for correct event uuid and id

### LINKED ISSUE #1338: Display event using UUID not working if UUID's first numbers equals to an real ID
Hi,

I am using the "display event using UUID" features (https://github.com/MISP/MISP/issues/88)
but I think there is a bug : 

I have an event with uuid = 577a061ee-9158-4b5e-8a8e-aca8950d2109 and when I go to :
   https://myurl.com/events/view/577a061ee-9158-4b5e-8a8e-aca8950d2109
it displays : 
Event ID 577   with Uuid 5491454c-0308-4270-b33b-459d950d210b  

Easy to reproduce, just go to 
https://myurl.com/events/view/549withanyletters

==> it works because event ID 549 exists

If I go to https://myurl.com/events/view/5490withanyletters
==> it doesn't work because event ID 5490 doesn't exist

Do you need any more informations ?

rotanid:
this might fix #1338

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -852,12 +852,14 @@ class EventsController extends AppController {
 	 */
 
 	public function view($id = null, $continue=false, $fromEvent=null) {
-		// If the length of the id provided is 36 then it is most likely a Uuid - find the id of the event, change $id to it and proceed to read the event as if the ID was entered.
-		if (strlen($id) == 36) {
+		// find the id of the event, change $id to it and proceed to read the event as if the ID was entered.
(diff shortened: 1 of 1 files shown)


iglocska:
This is indeed cleaner than just checking the length, but it will not fix the issue in #1339. You are indeed correct that the UUID provided there is invalid and it would also fail Validation::uuid()

rotanid:
it sure would fix the issue, my bugfix does not only check the uuid but also is_numeric ;-)

iglocska:
You are indeed correct, it's reverted!

PR #1341, opened 2016-07-07
remove obsolete uuid() wrapper

rotanid:
#### What does it do?

this removes an ancient generateUuid() wrapper from AppModel.php, because this wrapper does nothing else than calling CakeText::uuid() for every case, because String::uuid() is only an alias for the other function (String class extends CakeText and changes nothing and is deprecated, too)
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot, if anyone runs into issues with the UUID generation, please update your CakePHP version.

PR #1342, opened 2016-07-07
fix: use different variable name in sub-loop

### LINKED ISSUE #1285: ShadowAttribute: double use of loop variable name in sub-loop
in Model/ShadowAttribute.php in function __afterSaveCorrelation there is a foreach loop with two enclosed foreach loops.
one of them uses the same variable name for the key as the parent loop.
https://github.com/MISP/MISP/blob/89e8307c1a920bacc1aa724bd3aefae272a9274e/app/Model/ShadowAttribute.php#L237
i suggest changing $k to $key here.
i didn't do a PR yet, as i'm not sure if line 240 in said file has to still use $k or also be changed to $key then.

please check, tell me and i'll change this.

rotanid:
#### What does it do?

should fix #1285 
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks!

PR #1350, opened 2016-07-10
fix: specify correct&specific branches in .gitmodules

### LINKED ISSUE #1345: Create Event Error
Hi all, 

Just have this problem
### Work environment

| Questions | Answers |
| --- | --- |
| Type of issue | Bug |
| OS version (server) | ubuntu |
| OS version (client) | Ubuntu |
| PHP version | PHP 5.5.9-1ubuntu4.17 (cli) (built: May 19 2016 19:05:57) |
| MISP version / git hash | 2.4.48 |

This is  what I get in my Google Chrome

Fatal Eror
Error: Class 'CakeText' not found  
/var/www/MISP/app/Model/Event.php 
Line: 365
Notice: If you want to customize this error message, create app/View/Errors/fatal_error.ctp

Cheers,
Andreas

rotanid:
as discussed with @iglocska on gitter:
this PR corrects the branch setting in the .gitmodules file.
the branch "2.7" doesn't exist (at least anymore) in cakephp repo, it has to be "2.x"
also, this adds the branches to  taxonomies and warninglist repos, just to be sure.

it might be possible, that no installation that was simply updated with git uses cake php 2.8 yet, because of this wrong branch setting.
also, one that got installed with 2.8.3 still uses that and not the current 2.8.5, because it wont update because of the wrong branch.

this might also fix #1345 and allow us to reintroduce the reverted #1341

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,7 @@
 [submodule "app/Lib/cakephp"]
 	path = app/Lib/cakephp
 	url = https://github.com/cakephp/cakephp.git
-	branch = 2.7
+	branch = 2.x
 [submodule "PyMISP"]
 	path = PyMISP
 	url = https://github.com/MISP/PyMISP.git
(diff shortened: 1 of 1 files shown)


adulau:
Indeed, this makes sense to me for the taxonomies and the warning-lists. It seems ok for Cake PHP.

iglocska:
Looks good, thanks a lot!

PR #1351, opened 2016-07-10
Sql bugfix & add to mysql.sql

rotanid:
this fixes a bug (no SET when using ADD!) from cd6fb5236c66091410906a87ebcbae8370f51b96 and adds the database table column perm_delegate also to MYSQL.sql to be consistent again


iglocska:
Looks good thanks.

PR #1352, opened 2016-07-10
Mysql updates cleanup

rotanid:
this goes on with the changes in cd6fb5236c66091410906a87ebcbae8370f51b96 to remove quotes where they aren't necessary (at least i assume so, because they were removed in the linked commit).

also, it removes obsolete spaces.

it adds semicolons where they were missing so the usage of semicolons is consistent.


iglocska:
Thanks for the cleanup.

PR #1353, opened 2016-07-11
Install/update doc improvements - releases instead of random commits

rotanid:
this updates the INSTALL and UPDATE documentation so it uses releases of MISP instead of the latest commit at the time of setting up or updateing the instance.

if someone willingly wants to use a specific commit between releases, he can still do that, but the default should be that only releases (tagged revisions) are used.

this is achieved in this commit by using the latest "tag" in the default branch (2.4.48 in branch 2.4 at the time of writing)


iglocska:
Looks good, thanks!

Issue #1363, opened 2016-07-11
support PostgreSQL database backend


rotanid:
as discussed many times, we might want to support PostgreSQL as an alternative to MySQL/MariaDB.

i've already worked on basic, experimental support which will soon be released either as a feature branch or as a pull request.

adulau:
That's a great news. Depending of the changes required, a feature branch might be useful.  Thank you for your work.

rotanid:
opened a PR #1364 for part of the necessary but not strictly postgres-dependent changes.

also created a feature branch https://github.com/MISP/MISP/tree/feature/postgresql (it already includes the PRs changes plus the 3 commits for postgres support itself)

rotanid:
the PR #1364 was tested by @iglocska and merged.
i updated the feature branch to reflect this merge.

feel free to test it :-)

rotanid:
update on the status:
- reduced 2.4 branches mysql dependency ( #1437 , #1442 , #1443 )
- feature branch is up to date again
- migration instructions from mysql/mariadb added to the install doc (tested with the misp training data)

testers needed, migration is easy!
note: no reverse migration documented, but you can always switch back to mysql easily (loosing changes you made in postgresql-db)

rotanid:
fixed in #1485

rotanid:
i noticed an issue:
if we do add new indexes via the update mechanism there might be a problem if the index already exists, because before PostgreSQL 9.5 (current upstream release) there was no easy way to say "create this index only if it doesn't exist yet"
the easy way wouldn't work for current Debian stable (8 / jessie) as it uses Postgres 9.4

iglocska:
That is fine. We currently push new indexes after running __dropIndex() which first drops the existing index then creates a new one. However, we might have to update __dropIndex() to include a path for Postgres.

rotanid:
i made a PR adding this path for Postgres to __dropIndex()

PR #1364, opened 2016-07-11
change of boolean datatype handling

rotanid:
the commits in this PR change the way, booleans and related things are handled by MISP.

because MySQL has no "bool" datatype, tinyint(1) is used and therefore booleans have to be casted to integers 1/0. in some parts of the MISP-code, the integers are used while in others, the booleans are used.
this can lead to problems with the cakephp ORM wrapper, which (sometimes?) casts a "false" to "" (empty string) instead of 0, which works with MySQL, but isn't clean and doesn't work with other DBMS like PostgreSQL.

therefore, this PR changes many true/false booleans used for sql queries to their 1/0 integer counterparts.
it also changes some $virtualFields sql statements to work with this.
third, it sets input type checkbox explicitly for some boolean input fields, so these input fields don't depend on type-guessing, because this won't work with PostgreSQL as it doesn't have the "tinyint(1)" datatype.

a second test of this by someone else before merging would be good.

these changes are also mandatory for the basic postgresql support ( #1363 )

### DIFF
--- a/app/Controller/AttributesController.php
+++ b/app/Controller/AttributesController.php
@@ -15,7 +15,7 @@ class AttributesController extends AppController {
 	public $paginate = array(
 			'limit' => 60,
 			'maxLimit' => 9999, // LATER we will bump here on a problem once we have more than 9999 events
-			'conditions' => array('AND' => array('Event.id >' => 0, 'Attribute.deleted' => false))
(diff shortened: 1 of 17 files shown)


iglocska:
Thanks a lot, tested and working as expected, great PR as always!

PR #1387, opened 2016-07-27
Using custom comments for module return values/attributes

3c7:
#### What does it do?

Enables custom comments (instead of `Enriched via...`) when using enrichment modules.

``` python
    r = {'results': [
                        {
                            'types': ['ip-src', 'ip-dst'],
                            'values': ['1.1.1.1', '2.2.2.2'],
                            'categories': ['Payload delivery', 'Network activity']
                        },
                        {
                            'types': ['filename'],
                            'values': ['test1.exe', 'test2.exe'],
                            'categories': ['Payload installation']
                        },
                    ],
         'comment': 'This is a custom comment.'}
```

If there's no comment set, the default comment is used. No changes in the template, because I think the comments should always be editable.
#### Questions
- [x] Using it on test server at the moment  
#### Release Type:
- [ ] Major
- [ ] Minor
- [x] Patch

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -3667,6 +3667,12 @@ class EventsController extends AppController {
 				}
 				$resultArray = array_merge($resultArray, $freetextResults);
 			}
+			if(isset($result['comment']) && $result['comment'] != "") {
+				$importComment = $result['comment'];
+			}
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks a lot, looks good

PR #1388, opened 2016-07-27
Create categories array, if only one category given

3c7:
#### What does it do?

Handles categories from returning module values as the values and types. If only one category given and it is not in an array, create an array with that value.
#### Questions
- [x] On our test environment
  #### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Good idea, thanks.

Issue #1013, opened 2016-03-09
redundant members list and organisations page


cvandeplas:
In the Global Actions menu there is a redundancy between the two pages "members list" and "organisations". 

One of the two (I guess members list) needs to be taken away or merged in the organisations page.

Additionally, the Organisation link could be grouped together in the Sharing Groups section of the menu. That way those two logical things are kept together visually.

PR #1391, opened 2016-07-28
chg: redundant members list and organisations page #1013

cristianbell:
#### What does it do?

It resolves `#<1013>`
#### Questions
#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch


app/View/Users/memberslist.ctp
```diff
@@ -1,4 +1,5 @@
-<div class="users index">
+<div class="users index"><?php
+/*
```
rotanid: please just remove the code if it isn't necessary anymore, we can always take it from the git history if we need it again :)

iglocska:
A few issues after a very brief check:
1. What Andreas pointed out already
2. The code in places if very hard to read, please use proper spacing
3. The code currently removes the members list but doesn't offer an alternative for users to have a quick overview over the organisations and the member count.
4. The PR removes the view portion of the members list but not the underlying look-ups. (in the Controller )
5. Please use the alternate if/for notation for code that envelops HTML

cristianbell:
updated the core, hope it's all ok now.

rotanid:
could you please fix the wrong indention of some lines in:
OrganisationsController.php
UsersController.php

app/Controller/UsersController.php
```diff
@@ -726,18 +726,8 @@ public function resetauthkey($id = null) {
...
-				'fields' => $fields,
-				'recursive' => -1,
-				'contain' => array('Organisation'),
-				'group' => array('Organisation.name', 'Organisation.id'),
-				'order' => array('UPPER(Organisation.name)'),
-		);
-		$orgs = $this->User->find('all', $params);
-		$this->set('orgs', $orgs);
+	public function attributehistogram() {
+	    //all code is called via JS
```
Rafiot, line 730: Why having the function then? (if it is a PHP weirdness, I apologize)

cristianbell, line 730: it's needed to generate the view. If you call a /attributehistogram in the URL but there is no function with the same name in the controller the page will not be displayed. 
MVC principle + CakePHP. (and actually other fw work the same way here)

app/View/Users/attributehistogram.ctp
```diff
@@ -0,0 +1,14 @@
+<div class="users index">
+	<div id = "histogram"></div>
+	<?php //echo $this->element('histogram');?>
```
rotanid, line 3: please remove this if it isn't used (as it's masked as a comment at the moment)

rotanid:
assuming you have tested it's working, i think the code is ok.
todo:
- remove commented codeline
- rebase to current misp 2.4 branch state
- rebase this into one commit only (http://stackoverflow.com/a/5189600) and git push -f update the branch of the PR
  perhaps wait with the latter until @iglocska also checked and gave his OK so you don't have to do it multiple times :)

iglocska:
I've accepted the commit and made some changes. Please review them for future reference. Some of it is more nice to have stuff (such as not breaking the layout with a long column name for a column that is going to be pretty small, indentation, etc) but some of it is crucial:
- correct use of contain
- not keeping obviously model level functions in the controller
- using the correct model when doing a lookup
- dealing with array indices if they are not set

Let me know if you have any questions about any of these.

PR #1410, opened 2016-08-04
SSO plugin (Shibboleth based). Instructions to enable and configure i…

ppanero:
#### What does it do?

Single Sing-On authentication integration via Plugin (following the same structure than the CertAuth plugin. 
As a summary, the plugin will take the values from the environment variables (given by apache after login-in in via SSO) and authenticate the user (creating it if it does not exists)
#### Questions
- [ ] Are you using it in production? No, development
#### Release Type:
- [ ] Major
- [X] Minor
- [] Patch


adulau:
Thank you very much for your contribution.

ppanero:
You're welcome. It is my pleasure!

PR #1417, opened 2016-08-08
Fix export for non md5/sha1/sha256 types

RichieB2B:
#### What does it do?

If fixes an issue where generating exports except md5/sha1/sha256 would actually generate an export for sha256.
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Nice catch, thanks!

PR #1426, opened 2016-08-10
Sso integration plugin

ppanero:
#### What does it do?

Minor changes in ShibbolethAuthentication plugin: an organization is created if the one given by the environment variables / default in config.php does not exists. Session and get parameter added in case it is needed to be disable upon login. 

Bug fix: Jobs were being displayed as queued on export. That was due to the query in "getProcess" comparing organization name to organization id. Also the creation of the process was taking org_id from the user ignoring if it is an admin or not (line 41 in Job.php) now it checks if isSiteAdmin puts org_id to 0, if not, it is put to org_id.
#### Questions
#### Release Type:
- [ ] Major
- [X ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/JobsController.php
+++ b/app/Controller/JobsController.php
@@ -68,8 +68,9 @@ class JobsController extends AppController {
 	}
 
 	public function getProgress($type) {
-		$org = $this->Auth->user('Organisation')['name'];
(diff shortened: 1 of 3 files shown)


iglocska:
Thanks a lot!

Issue #1428, opened 2016-08-11
GET misp2.4.49.js - 404 Not Found


cristianbell:
_If you would like to report a bug, please fill the template bellow_
### Work environment

| Questions | Answers |
| --- | --- |
| Type of issue | Bug |
| OS version (server) | Debian, ubuntu, CentOS, RedHat, all |
| OS version (client) | all |
| PHP version | 5.4, 5.5, 5.6, 7.0, 7.1... |
| MISP version / git hash | 2.4.50, hash of the commit |
| Browser | If applicable |
### Expected behavior

Line 48 in /app/Controller/AppController.php :  private $__jsVersion = '2.4.49';
but the JS file is called 2.4.50, see /app/webroot/js/misp2.4.50.js 
### Steps to reproduce the behavior

Start the app. 
### Logs, screenshots, configuration dump, ...

GET misp2.4.49.js - 404 Not Found

cristianbell:
fixed by https://github.com/MISP/MISP/commit/80e1c61fb533298763ad84cb050060fe90707753

PR #1429, opened 2016-08-11
GET misp2.4.49.js - 404 Not Found #1428

cristianbell:
#### What does it do?

resolves `#1428`
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

Issue #1149, opened 2016-05-20
Free text import does not recognize new international Top Level domains


cbakkers:
https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains
domains with ending .top, .xyz for example recognized as filenames instead of domain/hostname.
seen in MISP 2.4.44

rotanid:
i noticed some regex matches regarding domains in the code yesterday and also thought immediatly, that some of them don't cover the new TLDs.
the way i see it, that can't be worked around easily, as there are over a thousand TLDs which might overlap with filename extensions that people use.
perhaps we could use a list of the top100 or so TLDs (in terms of registered domain names) insted of using all of them to work around this...

adulau:
Indeed, good point. Not sure if we could directly use the TLDs list from the warning-list https://github.com/MISP/misp-warninglists/blob/master/lists/tlds/list.json which includes the IANA ones.

iglocska:
We could indeed go with the warninglist!

rotanid:
but as i said, there would be overlap with common file extensions like ZIP, which is also a TLD

iglocska:
yup, think we just need to offer both options to the user and let him/her decide

cristianbell:
talking with @cvandeplas about this resulted in a couple of ideas: 
- splitting the function __resolveType() in app/Lib/Tools/ComplexTypeTool.php into smaller functions;
- using the validate functions from Model/Attribute.php for some of the validation.

any input on this would be highly appreciated.

iglocska:
Splitting it into smaller functions - yes that can be handy.

Reusing the validation methods - I disagree. The idea is not to pass each value through every (sometimes overly loose) validation function, but instead to use heuristics to run as few of the validation scripts as possible to get to the (group of) best fitting type(s). I would definitely treat the two separately.

PR #1432, opened 2016-08-11
fix: "Free text import does not recognize new TLDs", issue #1149

cristianbell:
#### What does it do?

`fixes #<1149>`
#### Questions
#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -2802,15 +2802,12 @@ class EventsController extends AppController {
 		}
 
 		if ($this->request->is('post')) {
-            //getting the warninglists
-            App::uses('WarninglistsController', 'Controller');
-            $warningList = new WarninglistsController();
-            $IANAEntries = $warningList->Warninglist->getAllIANAEntries();
-
 			App::uses('ComplexTypeTool', 'Tools');
 			$complexTypeTool = new ComplexTypeTool();
-            //fix:: 1149, added the 3rd parameter - all entry values for warninglists:
-			$resultArray = $complexTypeTool->checkComplexRouter($this->request->data['Attribute']['value'], 'FreeText', $IANAEntries);
+			$this->loadModel('Warninglist');
+			$IANATLDentries = $this->Warninglist->getAllIANAEntries();
+
+            $resultArray = $complexTypeTool->checkComplexRouter($this->request->data['Attribute']['value'], 'FreeText', $IANATLDentries);
 			foreach ($resultArray as &$r) {
 				$temp = array();
 				foreach ($r['types'] as $type) {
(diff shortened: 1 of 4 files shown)


rotanid:
could you please check your indention, if you look at the diff here on github you notice that many changes aren't changes in the code but only in the indention. also part of the new code has wron indention.
(thanks, you also fixed some indention and obsolet spaces in this PR, which is OK in this case because these were only in the same parts of the code that got changed/added.)

app/Controller/EventsController.php
```diff
@@ -2802,9 +2802,15 @@ public function freeTextImport($id) {
 		}
 
 		if ($this->request->is('post')) {
+            //getting the warninglists
+            App::uses('WarninglistsController', 'Controller');
+            $warningList = new WarninglistsController();
```
iglocska: You should never have to call functions from a different controller, this breaks basic MVC principles.

cristianbell:
should be ok now, please take a look when u have time. thanks!

app/Controller/EventsController.php
```diff
@@ -2804,7 +2804,10 @@ public function freeTextImport($id) {
 		if ($this->request->is('post')) {
 			App::uses('ComplexTypeTool', 'Tools');
 			$complexTypeTool = new ComplexTypeTool();
-			$resultArray = $complexTypeTool->checkComplexRouter($this->request->data['Attribute']['value'], 'FreeText');
+			$this->loadModel('Warninglist');
+			$IANATLDentries = $this->Warninglist->getAllIANAEntries();
+
+            $resultArray = $complexTypeTool->checkComplexRouter($this->request->data['Attribute']['value'], 'FreeText', $IANATLDentries);
```
Rafiot: This line is wrongly indented, I think it is the last one.

cristianbell:
and that one fixed as well, thanks.

iglocska:
I just reviewed the latest version and here are the issues I found:

1 (major).
Currently if the warninglists are not updated on a blank installation this fails with:

Notice (8): Undefined index: WarninglistEntry [APP/Model/Warninglist.php, line 243]
Warning (2): array_map() [function.array-map]: Argument #2 should be an array [APP/Model/Warninglist.php, line 243]

It should check whether the warninglist exists instead of assuming that it does. Not finding the correct warninglist should make the freetext import fall back to the old functinality.

2 (critical). 
Once I enabled the warninglist, resolving a domain name correctly identifies it as network activity, yet the type field is broken/missing. I didn't go through the code what change causes this but this breaks the core functionality of the freetext import tool as well as most likely the enrichment and import modules as those rely on the same functionality internally. 

3 (cosmetic, worst case leave this I'll fix it if the rest works). 
Coding style issues: Spacing around conditionals, loops as well as non-brace-encapsulated contents thereof.

cristianbell:
will create a new pull request and work on it w/ @cvandeplas

Issue #1355, opened 2016-07-11
attribute list on view event page sort by date issue


cvandeplas:
### Work environment

| Questions | Answers |
| --- | --- |
| Type of issue | Bug, |
| MISP version / git hash | 2.4.48 |
### Expected behavior

On the events page, when sorting on date, the items are sorted by date.
### Actual behavior

Results are sorted by category, then only by date.
### Steps to reproduce the behavior

Click on the 'date' header of the attributes table in the view event page.

PR #1435, opened 2016-08-12
chg: attribute list on view event page sort by date issue #1355

cristianbell:
#### What does it do?

`It resolves issue #<1355>`
#### Questions
#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch

### DIFF
--- a/app/Model/Event.php
+++ b/app/Model/Event.php
@@ -2630,9 +2630,7 @@ class Event extends AppModel {
 		$correlatedAttributes = isset($event['RelatedAttribute']) ? array_keys($event['RelatedAttribute']) : array();
 		$correlatedShadowAttributes = isset($event['RelatedShadowAttribute']) ? array_keys($event['RelatedShadowAttribute']) : array();
 		foreach ($event['Attribute'] as $attribute) {
-		    //adding a date key so that attributes can be sorted on date - #1355
-		    $attribute['date'] = $attribute['timestamp'];
-			if ($filterType && !in_array($filterType, array('proposal', 'correlation', 'warning'))) if (!in_array($attribute['type'], $this->Attribute->typeGroupings[$filterType])) continue;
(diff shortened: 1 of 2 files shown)


iglocska:
There are 3 lines changed - 2 of them are not needed (one is a var_dump left in commented out, the other a comment about an issue - we don't need that, we have git for version control / issue tracking).

For the actual solution, instead of keeping the timestamp in the object twice, it's better to tell the paginator to use the timestamp for the pagination. The actual pagination happens in app/View/Elements/eventattribute.ctp

Line 148 there is this:
<th><?php echo $this->Paginator->sort('date');?></th>

Which can be changed to:

<th><?php echo $this->Paginator->sort('timestamp', 'Date');?></th>

This will tell the paginator to use the timestamp field and still label it Date and then we don't need to store extra junk in memory.

cristianbell:
ok, fixed.

iglocska:
Thanks a lot

app/Model/Event.php
```diff
@@ -2629,7 +2629,7 @@ public function rearrangeEventForView(&$event, $passedArgs = array(), $all = fal
 		$correlatedAttributes = isset($event['RelatedAttribute']) ? array_keys($event['RelatedAttribute']) : array();
 		$correlatedShadowAttributes = isset($event['RelatedShadowAttribute']) ? array_keys($event['RelatedShadowAttribute']) : array();
 		foreach ($event['Attribute'] as $attribute) {
-			if ($filterType && !in_array($filterType, array('proposal', 'correlation', 'warning'))) if (!in_array($attribute['type'], $this->Attribute->typeGroupings[$filterType])) continue;
+    			if ($filterType && !in_array($filterType, array('proposal', 'correlation', 'warning'))) if (!in_array($attribute['type'], $this->Attribute->typeGroupings[$filterType])) continue;
```
rotanid, line 2632: your IDE added some spaces here :D

rotanid:
in the future, please don't add commit after commit in a PR but do "commit --amend" oder "rebase -i" etc. instead to not pollute MIPSs commitlog with unnecessary commits, makes reading the history easier (and space and performance etc)

PR #1437, opened 2016-08-15
Less mysql dependency

rotanid:
#### What does it do?

this PR removes some dependency on MySQL-specifics
- remove backticks where no keyword-escaping is necessary
- remove mysql-only function TIMESTAMPADD from a query
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Looks great, thanks!

PR #1438, opened 2016-08-16
chg: add index for cake_sessions expires to MYSQL.sql

rotanid:
adds the INDEX that was introduced in 0ac613d4e502e0ec957a03d2fdef68a3cc39f8ea also to MYSQL.sql for fresh setups


iglocska:
Good point, thanks.

PR #1441, opened 2016-08-16
Cleanup

rotanid:
#### What does it do?

this PR removes
- stuff about database schema that hasn't been used and is obsolete
- upgrade documentation for very old MISP versions. they arent up-to-date and if there are such old installations still out there, they can look it up in the git history
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot for taking the trash out!

PR #1443, opened 2016-08-17
change of boolean datatype handling #2

rotanid:
#### What does it do?

this PR is a followup to #1364 , especially ff7a7de

let me quote the relevent part of the older PRs text:

> because MySQL has no "bool" datatype, tinyint(1) is used and therefore booleans have to be casted to integers 1/0. in some parts of the MISP-code, the integers are used while in others, the booleans are used.
> this can lead to problems with the cakephp ORM wrapper, which (sometimes?) casts a "false" to "" (empty string) instead of 0, which works with MySQL, but isn't clean and doesn't work with other DBMS like PostgreSQL.
> 
> therefore, this PR changes many true/false booleans used for sql queries to their 1/0 integer counterparts.
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Console/Command/Task/RolesTask.php
+++ b/app/Console/Command/Task/RolesTask.php
@@ -13,12 +13,12 @@ class RolesTask extends Shell {
 
 		$roles = ClassRegistry::init('Role');
 		$roles->create();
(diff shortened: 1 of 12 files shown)


iglocska:
Thanks a lot!

PR #1447, opened 2016-08-17
update cakephp to 2.8.6

rotanid:
#### What does it do?

updates the cakephp submodule to version 2.8.6 which was released on August 3rd
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot

Issue #607, opened 2015-08-19
REST API - Get openIOC output


elhoim:
Might be needed for integration with [openioc_scan (volatility plugin)](https://github.com/TakahiroHaruyama/openioc_scan), see https://github.com/TakahiroHaruyama/openioc_scan/issues/2

Both for individual events, and a global one.

iglocska:
This was already resolved in #1448 by @TheDr1ver

PR #1448, opened 2016-08-17
Add support to export an OpenIOC file via API

TheDr1ver:
#### What does it do?

Fixes `#607` .... partially. With these changes you can at least download a single event's OpenIOC file. It does not implement global exports or exports based on filters/tags.

Make a request with your auth key to https://<misp_host>/events/downloadOpenIOCEvent/download/<event_id> and you'll receive your OpenIOC file.
#### Questions
  No.
  Yes
  Kind of? It shouldn't break anything, it just allows for API requests to download OpenIOC files.
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/AppController.php
+++ b/app/Controller/AppController.php
@@ -52,7 +52,7 @@ class AppController extends Controller {
 	// Used for _isAutomation(), a check that returns true if the controller & action combo matches an action that is a non-xml and non-json automation method
 	// This is used to allow authentication via headers for methods not covered by _isRest() - as that only checks for JSON and XML formats
 	public $automationArray = array(
-		'events' => array('csv', 'nids', 'hids', 'xml', 'restSearch', 'stix', 'updateGraph'),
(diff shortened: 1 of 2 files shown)


rotanid:
thanks for your contribution.
could you please "repair" your indention? part of your Pull Request uses spaces instead of tabs, you can see it very clearly if you look at the "files changed" Tab of this Pull Request.

ideally you would do the change/repair, do "git commit --amend" to add the changes to the previous commit and after that "git push -f" to update the branch and therefore this pull request.

i haven't checked the functionality of this code, only the structure.

TheDr1ver:
I tried it the first way with amend/push, but for some reason it ignored the changes. Editing it directly on GitHub seems to have worked though, so it should be fixed now.

rotanid:
maybe you forgot to do "git add ..." before amending/committing?
nevermind, but now the indention is one "level" to little, as the code is inside the function.

TheDr1ver:
Oops! Fixed.

iglocska:
Looks good, thanks a lot.

PR #1451, opened 2016-08-18
fix: minor code issues:

cristianbell:
#### What does it do?

fixes a cpl of minor code issues;
#### Questions
#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch
- redundant var initialization
- for the automatically created organization the "created_by" is 0, which produces a Notice error in /View/Organization/view.ctp


iglocska:
Looks good, thanks!

PR #1454, opened 2016-08-19
Bug fixing on shibboleth auth. DB group loading and missing email bug…

ppanero:
#### What does it do?

It fixes two bugs in the shibboleth authentication plugin.  The first one concerns the loading of groups from database, previously it was always returning null due to bad query sintax. Also checking for missing email is implemented, if shibboleth does not return an email authentication will fail, previously it failed due to sql exception.
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Plugin/ShibbAuth/Controller/Component/Auth/ApacheShibbAuthenticate.php
+++ b/app/Plugin/ShibbAuth/Controller/Component/Auth/ApacheShibbAuthenticate.php
@@ -73,6 +73,8 @@ class ApacheShibbAuthenticate extends BaseAuthenticate {
         $groupRoleMatching = Configure::read('ApacheShibbAuth.GroupRoleMatching');
 
         // Get user values
+        if(!isset($_SERVER[$mailTag])) return false;
+
         $mispUsername = $_SERVER[$mailTag];
 
         //Change username column for email (username in shibboleth attributes corresponds to the email in MISPs DB)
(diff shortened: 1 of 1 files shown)


rotanid:
maybe not only checking for "isset" but also "empty":
`if(!isset($_SERVER[$mailTag]) || empty($_SERVER[$mailTag])) return false;`

or even better, check for mail syntax?
`if (!isset($_SERVER[$mailTag]) || filter_var($_SERVER[$mailTag], FILTER_VALIDATE_EMAIL) === FALSE) return false;`

iglocska:
Thanks a lot, just noticed @rotanid's comments after I merged it, he is indeed correct, could you do a PR with those changes?

app/Plugin/ShibbAuth/Controller/Component/Auth/ApacheShibbAuthenticate.php
```diff
@@ -194,7 +196,11 @@ public function getUserRoleFromGroup($groupTag, $groupRoleMatching, $roleId)
     public function checkOrganization($org, $user)
     {
         $orgModel = ClassRegistry::init('Organisation');
-        $orgAux = $orgModel->find($org);
+        $orgAux = $orgModel->find('first', array(
+                'fields' => array('Organisation.id'),
+                'conditions' => array('name' => $org),
+            )
+        );
```
iglocska, line 203: Are you sure that this works as intended? Wouldn't the orgId be in $orgAux['Organisation']['id'] ?

iglocska:
Just left a comment in the code, I am not 100% sure that it's correct.

ppanero:
@iglocska You are completely right it should be $orgAux['Organisation']['id'], I think there is a commit missing on this, Ill check it out and push on monday. I could also implment what @rotanid  suggested if there is not already a PR of it, I think it is in fact better.

iglocska:
Since we have the changes live already I just pushed a small fix for the $orgAux['Organisation']['id'] issue in 8f3ac0970d2d3ab484655ad74fe2fd0d9ff1b8bb. Indeed, @rotanid's suggestion would be great, nudge me when you're ready with it next week. Have a nice Sunday!

PR #1456, opened 2016-08-19
FileAccess cleanup/consistency

rotanid:
#### What does it do?

to make Tools consistent, this changes FileAccess tool:
- rename to FileAccessTool
- use instantiable class instead of static

where it is only used once, the method is directly called from a new object instad of assigning the object to a variable first. (supported since PHP 5.4)

( related to #1222 )
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #1457, opened 2016-08-21
better generation of random strings

rotanid:
#### What does it do?

so far, in many places of MISP random strings were generated by many functions in almost the same way.
first aim of this PR is to centralize that so code redundancy is reduced.

second, some generated random strings are used for authentication and therefore the used algorithm should be cryptographically secure. rand() or mt_rand() and many other functions aren't.
although there is "openssl_random_pseudo_bytes()", there are the newer functions "random_int()" (or also "random_bytes()") without openssl.
these are only available starting with PHP 7.0, but there is a compatibility library random_compat from paragonie which does exactly the same.
this PR uses this library (version 2.0.2) for compatibility as a git submodule.
also, this PR comes with the RandomTool class which has the "random_str()" method which is called with the security strength, the length and the character set you want and replaces every other random code generation present in MISP.

the default passwort length is raised to 40 (from 12) but can be overridden by a even higher password policy.

related links:
http://php.net/manual/en/function.random-int.php
https://paragonie.com/blog/2015/07/how-safely-generate-random-strings-and-integers-in-php
https://github.com/paragonie/random_compat
https://paragonie.com/blog/2015/07/common-uses-for-csprngs-cryptographically-secure-pseudo-random-number-generators

there would've been a nice ready-mode Random class called RandomLib by ircmaxell but as it doesn't support random_int() yet, an own class RandomTool is used here.
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Model/User.php
+++ b/app/Model/User.php
@@ -1,6 +1,7 @@
 <?php
 App::uses('AppModel', 'Model');
 App::uses('AuthComponent', 'Controller/Component');
+App::uses('RandomTool', 'Tools');
 
 class User extends AppModel {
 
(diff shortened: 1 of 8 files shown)


app/Model/User.php
```diff
@@ -243,33 +244,14 @@ class User extends AppModel {
...
 	public function beforeValidate($options = array()) {
 		if (!isset($this->data['User']['id'])) {
 			if ((isset($this->data['User']['enable_password']) && (!$this->data['User']['enable_password'])) || (empty($this->data['User']['password']) && empty($this->data['User']['confirm_password']))) {
-				$this->data['User']['password'] = $this->__generatePassword();
+				$this->data['User']['password'] = $this->generateRandomPassword();
 				$this->data['User']['confirm_password'] = $this->data['User']['password'];
 			}
 		}
-		if (!isset($this->data['User']['authkey']) || empty($this->data['User']['authkey'])) $this->data['User']['authkey'] = $this->generateAuthKey();
+		if (!isset($this->data['User']['authkey']) || empty($this->data['User']['authkey'])) $this->data['User']['authkey'] = $this->generateRandomPassword();
```
iglocska: This looks like a mistake, it should be generateAuthKey() still, right?

iglocska:
At first glance saw a small issue with a mixed up method call, looks great other than that. Will test it during the week.

rotanid:
yep, changed it - or rather reverted that changed line.

klingsor83:
What's the state of this pull request?

rotanid:
@klingsor83 still waiting for review by @iglocska - that's the status.

iglocska:
It's merged now, thanks a lot.

mariuskagc:
Hi there! When do you think you will have the next release with this pull?

iglocska:
Within the next few days, but feel free to just git pull the latest commit on the 2.4 branch if you want to fast-track it.

mariuskagc:
No problem, I'll wait. :)

rotanid:
attention: simply pulling will not give you the advantages if you don't run PHP7.
you would also need to init&update the submodules (see UPDATE.txt) to get the compat layer for PHP 5.5/5.6

iglocska:
Good point

PR #1458, opened 2016-08-21
chg: remove not used old plugin file

rotanid:
#### What does it do?

this removes a file that was never used, it was introduced in 2012 by 7bf0e2f882f2d0588cfc685040d0805d4148584d and with the same commit the correct version of this file was introduced
this correct file/plugin was removed in 2014 by 6dc75d6b220cbbee2af1f33e065ef323d53c5eb3
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Looks good, thanks!

rotanid:
for reference, @iglocska reverted the change and will introduce it again with 2.4.51

PR #1459, opened 2016-08-21
remove several usages of App::import() in favor of App::uses()

rotanid:
#### What does it do?

this PR replaces the usages of App::import() by App::uses(), which is recommended in cakephp 2.x

in app/Model/Event.php it even removes an import of the AttributesController as i couldn't find out why a AttributesCONTROLLER should be imported into the EventModel - and i didn't find any usage of it.

but maybe i overlooked something, so please have a close second look on this PR before merging.
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Console/Command/Task/RoleIdTask.php
+++ b/app/Console/Command/Task/RoleIdTask.php
@@ -1,5 +1,5 @@
 <?php
-App::import('Controller', 'Users');
+App::uses('UsersController', 'Controller');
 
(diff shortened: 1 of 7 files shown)


iglocska:
Woah, that's a nice catch with the controller being loaded. That crap has been there since 2012 apparently. Well spotted!

PR #1462, opened 2016-08-22
remove obsolete backups of config files

rotanid:
#### What does it do?

removes two files that seem to be only backup files created accidently in 2014 by commit caeba7fb913588e49b9334c5cf32563cd6d7e49a
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks, they're indeed unused now. We used to have 2 sets of config files, one with background workers disabled (the regular default config files) and one with them enabled by default (the .b default config files). However we now default to enabling them on all installations.

PR #1465, opened 2016-08-23
Bro export

ppanero:
#### What does it do?

Bro export finished:

4 spaces idention changed by tabs.
logic of the export is on Model, NidsBroExport and HidsBroExport.
Handlers are in AttributeController and EventShell.
Automation view documentation added.
Export view added.

Tested in qa environment.

*.iml added to .gitignore fro IDEA based IDEs
#### Questions
#### Release Type:
- [ ] Major
- [ X ] Minor
- [] Patch

### DIFF
--- a/app/Console/Command/EventShell.php
+++ b/app/Console/Command/EventShell.php
@@ -299,6 +299,53 @@ class EventShell extends AppShell
 		$this->Job->saveField('date_modified', date("y-m-d H:i:s"));
 	}
 
+	public function cachebro()
+	{
+		$broHeader = "#fields indicator\tindicator_type\tmeta.source\tmeta.url\tmeta.do_notice\tmeta.if_in\n";
+		$userId = $this->args[0];
+		$user = $this->User->getAuthUser($userId);
+		$id = $this->args[1];
+		$this->Job->id = $id;
+		$format = $this->args[2];
+		$this->Job->saveField('progress', 1);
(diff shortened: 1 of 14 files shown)


app/Console/Command/EventShell.php
```diff
@@ -299,6 +299,55 @@ public function cachenids() {
...
+			$file = new File($dir->pwd() . DS . $filename);
+			$file->write($broHeader);
+			foreach ($final as $attribute) {
+				$file->append($attribute . "\n");
+			}
+			$file->close();
+
+			$execRetval = '';
+			$execOutput = array();
+			var_dump('zip -gj  ' . $zip->path . ' ' . $dir->pwd() . '/' . $filename,
```
rotanid: this seems to be a debug output, has to be removed

app/Lib/Export/HidsBroExport.php
```diff
@@ -0,0 +1,50 @@
...
+{
+
+
+// below overwrite functions from HidsExport
+	public function export($items, $type = 'MD5', $continue = false)
+	{
+
+		$orgsName = array();
+		if (!empty($items)) {
+			foreach ($items as &$item) {
```
rotanid: is it really necessary to use the &$item here instead of simple $item  ? (same in line 37)

rotanid:
in NidsBroExport.php, every function has "&$sid" as a parameter. perhaps i don't understand why, but i don't see any usage of this variable in the code of the functions?

PR #1467, opened 2016-08-23
chg: only show API/authkey to user with API key rights, fixes #1311

### LINKED ISSUE #1311: only show API/authkey to user with API key rights
Today the api/authkey is shown to everyone, even if he has no API key rights.

1/ Either remove the API key permission and give everyone rights to use the API key. (why wouldn't we?) 
(I'm not talking about the special sync right that allows org-spoofing)

2/ Or remove all references of the API/authkey for a user that doesn't have the rights to use one. Replace the key by a message stating who (s)he should contact to get the additional permission.

Pages include (but it might be more than this simple list)
- users/view/me
- (reset key page)

It seems that the events/automation page is already being filtered (access denied) if the user doesn't have api key access.

cristianbell:
#### What does it do?

It changes showing the API key behavior, not shown for users where it is not enabled. 
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/UsersController.php
+++ b/app/Controller/UsersController.php
@@ -43,10 +43,7 @@ class UsersController extends AppController {
 	 */
 	public function view($id = null) {
 		if ("me" == $id) $id = $this->Auth->user('id');
-		$responsibleAdmin = [];
-		if (!$this->_isSiteAdmin()) {
-			$responsibleAdmin = $this->User->findAdminsResponsibleForUser($id);
-			if($this->Auth->user('id') != $id)
+		if (!$this->_isSiteAdmin() && ($this->Auth->user('id') != $id)) {
 				throw new NotFoundException(__('Invalid user or not authorised.'));
 		}
 		$this->User->id = $id;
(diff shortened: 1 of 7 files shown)


app/Controller/UsersController.php
```diff
@@ -34,33 +34,50 @@ public function beforeFilter() {
...
 
-/**
- * edit method
- *
- * @param string $id
- * @return void
- * @throws NotFoundException
- */
+	public function request_API(){
+		$responsibleAdmin = $this->User->findAdminsResponsibleForUser($this->Auth->user('id'));
```
iglocska: Why not pass $this->Auth->user() here? That way you wouldn't have to retrieve the user again in findAdminsResponsibleForUser()

app/Model/User.php
```diff
@@ -935,4 +935,45 @@ public function getMembersCount() {
 		}
 		return $usersPerOrg;
 	}
+
+	public function findAdminsResponsibleForUser($id){
+		$userOrg = $this->find('first', array(
+			'conditions' => array(
```
iglocska: No recursive => -1 means it will fetch all relations

app/Model/User.php
```diff
@@ -935,4 +935,45 @@ public function getMembersCount() {
...
+				'User.id' => $id,
+			),
+			'contain' => array(
+				'Organisation' => array('fields' => array('id')),
```
iglocska: You don't actually need Organisation here, you only want the org_id, which is contained in the User model itself (it's a foreign key)

app/Model/User.php
```diff
@@ -935,4 +935,45 @@ public function getMembersCount() {
...
+		$admin = $this->find('first', array(
+			'recursive' => -1,
+			'conditions' => array(
+				'Role.perm_site_admin' => 0,
+				'Role.perm_admin' => 1,
+				'User.disabled' => 0,
+				'User.org_id' => $userOrg['Organisation']['id']
+			),
+			'contain' => array(
+				'Role' => array('fields' => array('perm_admin'))
```
iglocska: You are also doing a lookup on Role.perm_site_admin

app/webroot/js/misp2.4.50.js
```diff
@@ -2489,3 +2494,23 @@ $(".queryPopover").click(function() {
 	});
 });
 
+function requestAPIAccess() {
+	var destination = 'users';
+	var action = 'request_API';
+	url = "/" + destination + "/" + action + "/";
```
iglocska: You can just hard code the URL here as

url = "/users/request_API";

app/Controller/UsersController.php
cristianbell: thanks, good catch!

app/webroot/js/misp2.4.50.js
cristianbell: fair enough. though I'm not a big fan of hardcoding, but I guess in this case.. Though maybe we should have only a cpl of functions who do async req to the server and get the url, container to update, extra functions to perform as params.

iglocska: Which is fine, if it was a parametrised generic function I'd be all for it. For single use functions such as this it doesn't make sense.

rotanid:
made obsolete / was replaced by #1498

PR #1469, opened 2016-08-23
Centralize CIDR checks

rotanid:
#### What does it do?

moves the check for valid ip/mask combinations to one single central function
(and doesn't use preg_match anymore for this)
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Good idea, looks much cleaner.

PR #1470, opened 2016-08-24
cleanup TemplatesController.php

rotanid:
#### What does it do?
- removes obsolete function
- removes obsolete variables
- sets default value for $count
- spacing/empty lines
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks for the cleanup!

PR #1471, opened 2016-08-24
filename regex changes

rotanid:
#### What does it do?

**this is NOT a finished PR yet, it's content should be the base for a discussion.**

it is about regexes that check for filenames.

what it does so far:
- fix obvious errors in regexes (double chars, not quoted hyphen, quote chars that dont have to be quoted, replace \s by single space, + instead of *
- it marks other filename regexes, that don't have obvious errors, with a comment to have them hilighted in this PR

the questions for discussion are:
- why are there many different regexes for a filename, couldn't we reduce that to one?
- why allow some underscores, some commas, one a colon?
- why have some restrictions regarding the file extensions, others not?
- why do some allow the filename even starting with a space?
- why does one limit the length, all others don't?

my suggestion for a common regex would be:
`@^[\w_.,:]+[\w_.,\-: ]*[\w_.,\-:]+$@`
that wouldn't allow files to start with a hyphen or space and to end with a space.
additionally, a check for <=255 characters?
what this suggested regex would allow _less_ for some of the current cases:
- no tabs
- no start with space or hyphen
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -1938,7 +1938,7 @@ class EventsController extends AppController {
 			$rootDir = APP . "files" . DS . $id . DS;
 			App::uses('Folder', 'Utility');
 			$dir = new Folder($rootDir, true);
-			if (!preg_match('@^[\w-,\s,\.]+\.[A-Za-z0-9_]{2,4}$@', $this->data['Event']['submittedgfi']['name'])) {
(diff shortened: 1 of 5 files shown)


iglocska:
Looks good, thanks!

PR #1475, opened 2016-08-24
chg: creator e-mail in the event details, fixes #1252

### LINKED ISSUE #1252: feature: creator e-mail in the event details
Having the creator e-mail in the event detail view (under Published for example) is needed.

cristianbell:
#### What does it do?

It changes showing the API key behavior, not shown for users where it is not enabled. 
#### Questions
#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch

### DIFF
--- a/app/Model/Event.php
+++ b/app/Model/Event.php
@@ -19,22 +19,22 @@ class Event extends AppModel {
 		'Containable',
 	);
 
-/**
- * Display field
- *
- * @var string
- */
+	/**
+	 * Display field
+	 *
+	 * @var string
+	 */
 	public $displayField = 'id';
 
 	public $virtualFields = array();
 
 	public $mispVersion = '2.4.0';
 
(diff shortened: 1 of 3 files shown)


app/Model/Event.php
```diff
@@ -1293,10 +1293,14 @@ public function fetchEvent($user, $options = array()) {
 				),
 			)
 		);
-		if ($user['Role']['perm_site_admin']) {
+		if ($user['Role']['perm_site_admin'] | $user['Role']['perm_admin']) {
```
rotanid: i think this has to be "||", double not only once

rotanid: was it intentional @cristianbell that you removed this change completely instead of fixing it?

iglocska: Yeah, we had a chat about it.

PR #1476, opened 2016-08-24
new filename regex & separate functions

rotanid:
#### What does it do?

following #1471 i discussed the topic on gitter with @iglocska 

the result are these regexes, one with slashes (if a whole path is checked) and one without.
we dont allow colons or commas anymore, but can be easily reintroduced if someone has trouble.

```
@^([a-z0-9_.]+[a-z0-9_.\- ]*[a-z0-9_.\-]|[a-z0-9_.])+$@i
@^\/?(([a-z0-9_.]+[a-z0-9_.\- ]*[a-z0-9_.\-]|[a-z0-9_.])+\/?)+$@i
```

also, the regex is now in only 3 places, one function for Model, one for Controller and the single path check. before it was in 10 places.
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/AppController.php
+++ b/app/Controller/AppController.php
@@ -667,4 +667,8 @@ class AppController extends Controller {
 		$this->Session->setFlash('Caches cleared.');
 		$this->redirect(array('controller' => 'servers', 'action' => 'serverSettings', 'diagnostics'));
 	}
+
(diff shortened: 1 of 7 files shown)


iglocska:
Looks great, thanks!

PR #1485, opened 2016-08-31
support PostgreSQL database backend

rotanid:
#### What does it do?

this adds experimental support for PostgreSQL to MISP as planned in #1363 

as discussed with @iglocska , we plan to merge it with an explicit note in the doc, that it is not well-tested - instead of leaving it in a feature-branch and having to rebase that all the time.

after merging, database-related changes (everything that's SQL, even if only a condition) have to be done with both DBMS in mind and changes to the structure have to be done in all of the .sql-files
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/INSTALL/INSTALL.postgresql.txt
+++ b/INSTALL/INSTALL.postgresql.txt
@@ -1,0 +1,100 @@
+INSTALLATION INSTRUCTIONS
+------------------------- for Debian 8 "jessie" server + PostgreSQL
+
+######### WARNING #########
+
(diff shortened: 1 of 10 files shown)


rotanid:
updated to include dd68eb0aae2644bb6c6673aa3bf7e9fb898da430

iglocska:
Thanks a lot for all the work on this! 

If anyone decides to give Postgresql a shot with MISP, please let us know if you run into any issues!

PR #1486, opened 2016-08-31
fix: update version number to 2.4.51 in MYSQL.sql

rotanid:
#### What does it do?

updates the version number of the sql setup file to the current version.
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #1493, opened 2016-09-01
change in SELtype, httpd_sys_content_rw_t does not exists

ppanero:
#### What does it do?

In the Installation manual for CentOS 7 there it is said to change the SELinux type to httpd_sys_content_rw_t. However I was getting some errors, checked on my system and that type does not exists, however httpd_sys_rw_content_t does, therefore I change the installation manual.
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #1497, opened 2016-09-02
Update INSTALL.centos7.txt

ppanero:
#### What does it do?

Some libraries and packages (Crypt_GPG and redis) were told to be install through pear and pecl, however if using php 5.6 they have to be installed opening a bash for php56, if not it will not recognized them, failing and giving warnings in some features of MISP (workers, diganostics, etc)
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot for documenting this!

PR #1498, opened 2016-09-02
chg: only show API/authkey to user with API key rights, fixes #1311

### LINKED ISSUE #1311: only show API/authkey to user with API key rights

cristianbell:
#### What does it do?

chg: only show API/authkey to user with API key rights, fixes #1311 
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/Component/ACLComponent.php
+++ b/app/Controller/Component/ACLComponent.php
@@ -328,6 +328,7 @@ class ACLComponent extends Component {
 					'logout' => array('*'),
 					'attributehistogram' => array('*'),
 					'resetauthkey' => array('*'),
+					'request_API' => array('*'),
 					'routeafterlogin' => array('*'),
 					'statistics' => array('*'),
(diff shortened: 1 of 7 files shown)


rotanid:
regarding the syntax and so on, this looks good. haven't checked logic/functionality.

iglocska:
Thanks a lot!

PR #1501, opened 2016-09-04
remove some references to variables

rotanid:
#### What does it do?

this reduces the usage of references by almost 60%
`grep -rIF "&$" app/ | grep -vF "Lib/cake" | grep -F ".php:" | wc -l`

references don't save space in PHP (-> copy-on-write!) and don't speed up anything, sometimes may be slower.
but they may lead to changes that you didn't want, because you may not know that some variable is actually a reference.

see this blogpost from Johannes Schlüter, for example:
http://schlueters.de/blog/archives/125-Do-not-use-PHP-references.html

please wait with merging until you have time for a second look on the changes - although i did look on them at least twice.
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/AppController.php
+++ b/app/Controller/AppController.php
@@ -363,7 +363,7 @@ class AppController extends Controller {
 	private function __convertEmailToName($email) {
 		$name = explode('@', $email);
 		$name = explode('.', $name[0]);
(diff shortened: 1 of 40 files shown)


iglocska:
Thanks a lot, looks good!

PR #1503, opened 2016-09-04
remove redundant className attributes

rotanid:
#### What does it do?

removes the "className" attribute from class relations where it is redundant because it's the same as the array key.
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/NewsController.php
+++ b/app/Controller/NewsController.php
@@ -17,7 +17,6 @@ class NewsController extends AppController {
 	);
 
 	public function index() {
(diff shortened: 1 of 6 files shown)


app/Controller/NewsController.php
```diff
@@ -17,7 +17,6 @@ public function beforeFilter() {
 	);
 
 	public function index() {
-		$this->News->bindModel(array('belongsTo' => array('User' => array('className' => 'User'))));
```
rotanid, line 20: removed because already done by "$belongsTo" in News.php

iglocska:
Thanks a lot!

PR #1512, opened 2016-09-06
Tag.php: use dependency instead of explicit deleteAll

rotanid:
#### What does it do?

as far as i understand the cakephp model relations, setting the hasMany array entry of EventTag to dependent results in the same as the current explicit call of deleteAll.

perhaps i might have forgot something, like the order of the deletion might be different (deleting EventTags before or after Tag deletion)

http://book.cakephp.org/2.0/en/models/associations-linking-models-together.html
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Looks good, thanks.

PR #1520, opened 2016-09-09
stringer checks on email and nids_sid of user calculated from max

ppanero:
#### What does it do?

Stronger checks on email. Nids_sid value of created users calculated with max from db.
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #1527, opened 2016-09-12
update cakephp to 2.8.7

rotanid:
#### What does it do?

updates the cakephp submodule to version 2.8.7 which was released on September 10th
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #1535, opened 2016-09-15
chg: creator e-mail in the event details, fixes #1252

### LINKED ISSUE #1252: feature: creator e-mail in the event details

cristianbell:
#### What does it do?

displays the creator e-mail in the event details for org admins, fixes #1252
#### Questions
#### Release Type:
- [ ] Major
- [x] Minor
- [ ] Patch

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -858,8 +858,8 @@ class EventsController extends AppController {
 		$results = $this->Event->fetchEvent($this->Auth->user(), $conditions);
 		if (empty($results)) throw new NotFoundException('Invalid event');
 		//if the current user is an org admin AND event belongs to his/her org, fetch also the event creator info
-		if ($this->userRole['perm_admin'] && !$this->_isSiteAdmin() && $this->Event->isOwnedByOrg($id, $this->Auth->user('org_id'))) {
-			$results[0]['User'] = $this->User->getUserMinimalInfo($results[0]['Event']['user_id']);
+		if ($this->userRole['perm_admin'] && !$this->_isSiteAdmin() && ($results[0]['Org']['id'] == $this->Auth->user('org_id'))) {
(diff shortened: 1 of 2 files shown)


cvandeplas (changes requested):
The current View/view.ctp already contains the feature to show the email adress to the siteadmin. For a normal user (or org_admin) this info is not shown.

So the view() function in this Controller already fetches that information somewhere. This same code should be extended to also allow visibility for the org_admin for events of his org. 
Hint: $this->Event->fetchEvent($this->Auth->User(), $conditions); 

Not relevant as this code should be removed:
- The isOwnedByOrg function has no value, except doing an additional DB lookup. In the $results variable you already have the event info. Use that var to check if the event org is the same as the user org. (or org_id)
- The getUserMinimalInfo function is also not needed, as it's covered by an earlier query.

cvandeplas (approved):

cvandeplas:
Ok, sorry about the review, ignore what's inside... I tried a new feature of Github, and failed using / removing / updating it.
The remarks I have are: 

The `isOwnedByOrg` function has no value, except doing an additional DB lookup. In the `$results`variable you already have the event info. Use that var to check if the event org is the same as the logged in user org. (or org_id)

For the `getUserMinimalInfo($id)` function there are 2 options: 
1. Remove the function and search for the data immediately `$this->User->field('email', array('id' , $results[0]['Event']['user_id']))` 
2. Rename the function to something that says what it does: for example: `getUserEmailById($id)` that returns a string with the email address.    (and there is no need to query/return the id in the fields as you already have it, and it's not used in the existing code)

cristianbell:
- removed use of isOwnedByOrg()
- removed getUserMinimalInfo(), replaced with $this->User->field('email'..)

PR #1541, opened 2016-09-16
fix: removing some unused code.

cristianbell:
#### What does it do?

removes some unused code. 
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Makes sense, thanks a lot.

PR #1550, opened 2016-09-20
chg: Default roles all have API access

rotanid:
#### What does it do?

same as aa0383064345d24e1ceb32621457ec156c2cd809
but for postgres
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch
- changed the default role set to have api access enabled
- existing installations not affected
- if a community wants to restrict API access for certain users they're free to do it

(same as aa0383064345d24e1ceb32621457ec156c2cd809 but for postgres this time)


iglocska:
Thanks a lot!

PR #1511, opened 2016-09-06
Cleanup - remove some obsolete code

rotanid:
#### What does it do?

this PR - again - removes probably obsolete code or even files.

please check before merge!
#### Questions
- [ ] Does it require a DB change?
- [ ] Are you using it in production?
- [ ] Does it require a change in the API (PyMISP for example)?
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

PR #1555, opened 2016-09-22
remove view function that was broken for months

rotanid:
#### What does it do?

this PR removes the view() method from AttributesController.php
the method was broken since 2016-02-19, commit 52ddfb17ea9d029ca908cf8cbdf928907093bb8f
this commit removed View/Elements/eventattributerow.ctp which was used by View/Attributes/view.ctp
this view.ctp was removed two weeks ago because of its missing dependency in 3c7117b122bd6d2c539787d4487c37d304603f8c (#1511)
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/AttributesController.php
+++ b/app/Controller/AttributesController.php
@@ -797,22 +797,6 @@ class AttributesController extends AppController {
 		}
 	}
 
-	public function view($id) {
-		$this->Attribute->id = $id;
-		if (!$this->Attribute->exists()) {
-			throw new NotFoundException('Invalid attribute');
-		}
-		if ($this->_isRest()) {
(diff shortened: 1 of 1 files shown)


iglocska:
view() works fine. It is meant to be for API users only, which correctly returns the JSON of the attribute. 

When testing API functionality make sure you pass along the correct headers or set a type extension:

`Authorization: mykey`
`Accept: application/json`
`Content-type: application/json`
`GET: myMISP/attributes/view/1`

OR

`Authorization: mykey`
`GET: myMISP/attributes/view/1.json`

Normal users will be redirected to the event view.

rotanid:
my bad...

PR #1567, opened 2016-09-23
default org changed to id instead of name

ppanero:
#### What does it do?

Default organisation changed to id (numeric) instead of name, just in case an organisation changes the name, to be able to preserve the consistency with the numeric id.
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


adulau:
Thank you very much.

PR #1570, opened 2016-09-24
cleanup

rotanid:
#### What does it do?
1. removes some files from INSTALL that are either very outdated or otherwise obsolete.
2. removes a duplicate array key
3. simplifies handling of empty/temporary arrays
4. fixes a comment typo
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot.

PR #1571, opened 2016-09-25
fix: php warning on buildAlertEmailBody in Event.php

rotanid:
#### What does it do?

if an attributes type was longer than $appendlen-2 a php warning was logged.
str_repeat()'s 2nd parameter, an integer, must not be smaller than 0.
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Looks great, thanks!

PR #1574, opened 2016-09-26
fix: missing new TLDs in free text import, solves #1149

cristianbell:
#### What does it do?

fix: missing new TLDs in free text import, solves #1149 
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -2795,7 +2795,9 @@ class EventsController extends AppController {
 		if ($this->request->is('post')) {
 			App::uses('ComplexTypeTool', 'Tools');
 			$complexTypeTool = new ComplexTypeTool();
-			$resultArray = $complexTypeTool->checkComplexRouter($this->request->data['Attribute']['value'], 'FreeText');
+			$this->loadModel('Warninglist');
+			$IANATLDentries = $this->Warninglist->getAllIANAEntries();
+			$resultArray = $complexTypeTool->checkComplexRouter($this->request->data['Attribute']['value'], 'FreeText', $IANATLDentries);
 			foreach ($resultArray as $key => $r) {
 				$temp = array();
 				foreach ($r['types'] as $type) {
--- a/app/Lib/Tools/ComplexTypeTool.php
+++ b/app/Lib/Tools/ComplexTypeTool.php
@@ -10,7 +10,7 @@ class ComplexTypeTool {
 		'/\.+/' => '.'
 	);
 
-	public function checkComplexRouter($input, $type) {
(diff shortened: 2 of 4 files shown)


cvandeplas (commented):
Minor remark: for the future we would probably want to include other TLD sources from warning lists. We'll then need to refactor the string IANATLD to something more generic. 

Related to the merging of type/categories : please double check if the behaviour should be corrected on current types.

cristianbell:
1. more generic IANATLD string in function name, can be refactored once more sources are added and the function is changed. From `'conditions' => array('Warninglist.name' => 'TLDs as known by IANA')...` to something else. 
2. in __resolveType() function (ComplexTypeTool.php) there are several places where multiple types are returned;
3. **line 139**: email-src (Payload delivery), email-dst (Payload delivery, Network activity). Default category is email-src. If the user chooses email-dst, **Network activity can not be selected.** 
4. line 149: ip-dst, ip-src, ip-src/ip-dst - all same categories, ok
5. line 153: ip-dst, ip-src, ip-src/ip-dst - all same categories, ok
6. line 174: hostname, domain, url - all same categories, ok

iglocska:
1. Indeed, though the list of TLDs from IANA is pretty comprehensive, not sure if there's any need for this. Also, the lookup method has to be changed (currently it's 'first' instead of all) and the result arrays need to be merged/mapped.
2. This is correct.
3. This is a current weakness of the system, with the current mappings there wasn't a great need to refactor it so it's pretty low priority at the moment.

cristianbell:
> Indeed, though the list of TLDs from IANA is pretty comprehensive, not sure if there's any need for this. 

The ICANN list contains around 500 more entries than the IANA. One example is .shop. 

> Also, the lookup method has to be changed (currently it's 'first' instead of all) and the result arrays need to be merged/mapped.

Using 'all' here would return almost the same data str: 
**first** returns 2 arrays one of ["Warninglist"] => (id => VAL, name => VAL, ...) and ["WarninglistEntry"] array with 1290 arrays of [] => (value => **VAL we need**, warninglist_id => 12)
**all** returns an array[0] with 2 elements, each an array exactly as above ["Warninglist"] and ["WarninglistEntry"] arrays. 

It is already mapped, see Warninlist.php line 251.

iglocska:
Sure, if you find a more comprehensive list, just create a warninglist for it and add it to the passed parameters.

cvandeplas:
It seems that some issues were still remaining:
- allows invalid combination of categories / types
- default option for google.com is filename and not domain

PR #1575, opened 2016-09-26
Two small fixes

RichieB2B:
#### What does it do?

2 small fixes for bugs found while testing the delegation feature:
- Log target org instead of requesting org for the delegation
- Only show publish links for site admins (instead of org admins)
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Nice finds, thanks!

PR #1578, opened 2016-09-28
Cleanup

rotanid:
#### What does it do?

```
chg: remove obsolete dropIndex() - not needed only for reference, as there's a duplicate in AppModel.php (& in git)

chg: remove obsolete variables (i could not find any usage of those two)

chg: remove obsolete getEnrichmentSettings() - seems to have been replaced by Module.php getModuleSettings, please check!
```
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/ServersController.php
+++ b/app/Controller/ServersController.php
@@ -60,8 +60,6 @@ class ServersController extends AppController {
 	}
 
 	public function previewIndex($id) {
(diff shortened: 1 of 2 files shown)


iglocska:
Thanks a lot, spot on!

PR #1585, opened 2016-09-29
Make STIX validate

RichieB2B:
After this patch the STIX XML export passes the stix_validator

```
- correct namespace order
- stix set_id_namespace calls cybox set_id_namespace
```


iglocska:
Thanks a lot!

PR #1588, opened 2016-09-30
Two small STIX fixes (again)

RichieB2B:
- Use the correct CIQ namespace to avoid these errors:

```
    [!] stix-one-new.xml:13093:0:ERROR:SCHEMASV:SCHEMAV_CVC_DATATYPE_VALID_1_2_1: Element '{http://stix.mitre.org/Incident-1}Victim', attribute '{http://www.w3.org/2001/XMLSchema-instance}type': The QName value 'ciqIdentity:CIQIdentity3.0InstanceType' has no corresponding namespace declaration in scope.
    [!] stix-one-new.xml:13095:0:ERROR:SCHEMASV:SCHEMAV_ELEMENT_CONTENT: Element '{http://stix.mitre.org/extensions/Identity#CIQIdentity3.0-1}Specification': This element is not expected. Expected is ( {http://stix.mitre.org/common-1}Related_Identities ).
```
- Use consistent STIX/CyBOX IDs for domain|ip entries

### DIFF
--- a/app/files/scripts/misp2cybox.py
+++ b/app/files/scripts/misp2cybox.py
@@ -142,16 +142,22 @@ def resolveIPType(attribute_value, attribute_type):
 
 def generateDomainIPObservable(indicator, attribute):
     indicator.add_indicator_type("Domain Watchlist")
-    compositeObject = ObservableComposition()
-    compositeObject.operator = "AND"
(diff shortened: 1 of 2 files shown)


iglocska:
Thanks a lot!

PR #1589, opened 2016-09-30
Added python-setuptools for ubuntu install howto

thomai:
#### What does it do?

The installation Ubuntu installation readme file did not contain the package python-setuptools. I added it.
#### Questions
- [  ] Does it require a DB change?
- [  ] Are you using it in production?
- [  ] Does it require a change in the API (PyMISP for example)?
#### Release Type:
- [  ] Major
- [  ] Minor
- [X] Patch

You need to install the package python-setuptools on Ubuntu 16.04/Mint 18 to use the setup.py for the STIX installation.


adulau:
Thank you for the contribution.

PR #1591, opened 2016-09-30
sudo is needed to create new ssl key.

thomai:
#### What does it do?

I just added sudo to one command in the Ubuntu readme file.
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

A normal user does not have permissions to the used folder /etc/ssl/private in Ubuntu 16.04/Mint 18 by default.


iglocska:
Makes sense, thanks!

PR #1592, opened 2016-09-30
Put tag info from event in E-mail subject

### LINKED ISSUE #1107: TLP:AMBER hardcoded in email subject

RichieB2B:
#### What does it do?

This patch introduces 2 new settings:
- _email_subject_tag_ to hold the name of the tag you want in the E-mail subject line
- _email_subject_include_tag_name_ to indicate if you want the name of the tag or only the value of the tag in the E-mail subject

The default is "tlp" and "true" so it fixes #1107 
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Model/Event.php
+++ b/app/Model/Event.php
@@ -1521,8 +1521,21 @@ class Event extends AppModel {
 		} else {
 			$subject = '';
 		}
-		$tplColorString = !empty(Configure::read('MISP.email_subject_TLP_string')) ? Configure::read('MISP.email_subject_TLP_string') : "TLP Amber";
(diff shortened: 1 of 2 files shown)


iglocska:
Thanks a lot, much cleaner than a fixed string!

PR #1593, opened 2016-10-01
Allow merging of event attributes

RichieB2B:
#### What does it do?

It created a new item in de side menu "Merge attributes from..." which allows for attributes from an existing events to be copied to the current event. The use case is to keep all attributes from a certain threat in a single event. This reduces clutter in highly structured MISP instances.
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot, also added a way to merge attachments/samples.

PR #1594, opened 2016-10-02
Fix indication of new attributes in E-mail alerts

### LINKED ISSUE #1521: Updated event alert E-mail do not indicate new attributes
### Work environment

| Questions | Answers |
| --- | --- |
| Type of issue | Bug |
| OS version (server) | CentOS |
| OS version (client) | Seven |
| PHP version | 5.6 |
| MISP version / git hash | 2.4.51 |
| Browser | Not applicable |
### Expected behavior

The E-mail body says "Attributes (\* indicates a new or modified attribute)" so it should do that.
### Actual behavior

None of the attributes have a \* indicator.

Update an attribute, publish the event, observe the alert E-mail body

None

RichieB2B:
#### What does it do?

This patch fixes a race condition with the E-mail alerts by keeping the old publish_timestamp as a parameter so new attributes can accurately be identified by the E-mail alerting process. Fixes #1521
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Console/Command/EventShell.php
+++ b/app/Console/Command/EventShell.php
@@ -357,8 +357,9 @@ class EventShell extends AppShell
 		$processId = $this->args[1];
 		$job = $this->Job->read(null, $processId);
 		$eventId = $this->args[2];
+		$oldpublish = $this->args[3];
 		$user = $this->User->getAuthUser($userId);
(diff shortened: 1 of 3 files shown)


iglocska:
Thanks a lot, much better than what we had!

PR #1595, opened 2016-10-03
Fix STIX XML and JSON exports

RichieB2B:
- Make STIX XML framing orgname consistent with XML data
- Remove XML from STIX JSON format
- Avoid duplicate key-sequence to satisfy STIX unique identity-constraint


iglocska:
Thanks a lot, I keep forgetting that we have that stix-ish json format.

PR #1600, opened 2016-10-07
Add missing tags from pushed events

RichieB2B:
#### What does it do?

When the same event exists in two MISP instances additional tags placed on the event are not synced. This patch fixes that.

I'm not sure if both ['Event']['Tag'] and ['Event']['EventTag'] should be supported but I observed that pushed events use ['Event']['EventTag']. 
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Model/Event.php
+++ b/app/Model/Event.php
@@ -2059,9 +2059,9 @@ class Event extends AppModel {
 					}
 				}
 			}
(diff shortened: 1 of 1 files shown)


iglocska:
Nice catch, thanks a lot!

PR #1610, opened 2016-10-14
Add rh-php56-bcmath as a requirement for CentOS

RichieB2B:
#### What does it do?

CentOS does not install the bcmath php module by default. MISP uses the bcmod() function when validating IBAN numbers. So rh-php56-bcmath needs to be installed to avoid this error message:

```
2016-10-12 15:33:43 Error: Fatal Error (1): Call to undefined function bcmod() in [/var/www/MISP/app/Lib/Tools/FinancialTool.php, line 112]
2016-10-12 15:33:43 Error: [InternalErrorException] Internal Server Error
```
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #1617, opened 2016-10-19
fix: missing new TLDs in free text import, resolves #1149

### LINKED ISSUE #1149: Free text import does not recognize new international Top Level domains

cristianbell:
#### What does it do?

fixes `#1149`: 
- adds the dynamic category -> changes type functionality in Freetext Import Attributes list;
- IANAknown TLDs are only checked if the warninglists are present and enabled.
  
  NOTEs:
- the warninglist "TLDs as known by IANA" must be not empty and enabled for this functionality to work;
- a limitation of the current behaviour is using the IANA list which has around 500 domains less than the ICANN list;
- for the moment special characters - like 在线 or .الجزائر‎ - are not translated to their equivalent IANA code (xn--3ds443g and xn--lgbbat1ad8j).
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -2899,7 +2899,9 @@ class EventsController extends AppController {
 		if ($this->request->is('post')) {
 			App::uses('ComplexTypeTool', 'Tools');
 			$complexTypeTool = new ComplexTypeTool();
-			$resultArray = $complexTypeTool->checkComplexRouter($this->request->data['Attribute']['value'], 'freetext');
+			$this->loadModel('Warninglist');
+			$IANATLDEntries = $this->Warninglist->getAllIANAEntries();
+			$resultArray = $complexTypeTool->checkComplexRouter($this->request->data['Attribute']['value'], 'freetext', [], $IANATLDEntries);
 			foreach ($resultArray as $key => $r) {
 				$temp = array();
 				foreach ($r['types'] as $type) {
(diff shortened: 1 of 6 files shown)


iglocska:
Just tested it, with the following test case:

Freetext import with the following values:
- google.com
- host.google.com
- http://www.google.com
- filename.exe

Result in 2.4 (what I'd expect): 



Switched to the PR's branch and redid the same test with this result:

tld-branch

I still haven't checked whether the new category/type restrictions follow the ones imposed by the modules but until this part is fixed there is no real point in that anyway. I suggest that you don't try to override the mechanism that is used for the heuristics and instead simply replace the hard-coded list with the list that you provide via the warninglist as it 1. breaks the existing logic, 2. duplicates an existing mechanism.

cristianbell:
fair. 
and just adding the IANA List of TLDs string to the regexp was my initial thought. 
But I was unhappy that in some cases (for example a .zip - which can be a file ending or domain) will be seen as a domain. 
This is why I choose to change the types with types[] = filename, then domain if in IANA list then url (if multi part); 
I changed the code so that the default type is dependent on the "detected" types. This way the current behavior should be maintained. But is a little arbitrary as there can be multiple (different) possible types and only one default_type.

cristianbell:
fixed by c2fc803

PR #1630, opened 2016-10-25
minor changes and fixes, made the tags search more visible.

cristianbell:
#### What does it do?

Minor changes and fixes, removed /input, made the tags search more visible, added cursor:pointer on the tags and close div.
#### Questions
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


cristianbell:
main changes moved to #1641 other visual changes can be applied via the local custom css.

PR #1641, opened 2016-11-02
search input for "select tag" is fixed and does not scroll up anymore

cristianbell:
#### What does it do?

few minimal changes: 
- search input for "select tag" is fixed and does not scroll up anymore
- removed input end tag
- added css classes to select tag and select tag source popups - so they can be easily changed via the local extra css.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot.

PR #1651, opened 2016-11-04
Block alert e-mails based on tag

RichieB2B:
#### What does it do?

It adds the option to block e-mail alerts for events tags with a certain tag. This can be used to explicitly  block sending out alert e-mail for certain class of events.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot.

PR #1652, opened 2016-11-05
replaced ip address with * for the virtualHost

cristianbell:
#### What does it do?

replaced ip address with * for the virtualHost.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot.

PR #1653, opened 2016-11-05
sorts the "Attributes per organization" array by the total number of …

cristianbell:
#### What does it do?

sorts the "Attributes per organization" array (and histogram) by the total number of attr, highest on top.

#### Questions

#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch


iglocska:
Thanks a lot.

PR #1684, opened 2016-11-22
added chcon rw to selftest.php and a note about installing firewall manually when using the minimal ISO

cristianbell:
#### What does it do?

- added chcon rw for selftest.php
- added a note about installing firewall manually when using the minimal ISO

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


cristianbell:
removed the chcon on "selftest.php". Only change now is the note about installing firewalld in case it is not installed. 

Is this worth adding or should I just close the PR?

Rafiot:
For such small changes, please squash all the commits into one.

PR #1728, opened 2016-12-07
Some adjustments to misp-backup.sh:

RichieB2B:
#### What does it do?

- allow setting MISPPath in misp-backup.conf
- use MySQL username/password from database.php by default
- use machine sortable date for output file
 - do not store TmdDir name in tar
- use tar non-verbosely

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Looks good, thanks!

PR #1755, opened 2016-12-13
Add fulltext indexes from AppModel.php to MYSQL.sql

RichieB2B:
#### What does it do?

Adds missing fulltext indexes to INSTALL/MYSQL.sql

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot, I keep forgetting to keep MYSQL.sql up to date!

PR #1757, opened 2016-12-13
Use 24 hour clock

RichieB2B:
#### What does it do?

Several places in MISP used a 12-hour notation without AM/PM indicators. This leads to odd (off-by-12) results.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Makes sense, thanks!

PR #1766, opened 2016-12-15
Speed up MISP by factor 10

RichieB2B:
After applying these indexes viewing a large event the time spent on MySQL queries went down from 2 seconds to 200 ms.

I'll leave adding these indexes to existing MISP installations as an exercise for the developers. :wink:


adulau:
Indeed very good catch.

iglocska:
Woah, nice find. Will also include it in the upgrade script for the next version.

PR #1767, opened 2016-12-15
Don't let misp-backup.sh fill up /tmp

RichieB2B:
Large MISP installations can easily fill up /tmp when using misp-backup.sh, so use `$OutputDirName` instead.


iglocska:
Thanks a lot!

PR #1769, opened 2016-12-16
Make threat level in E-mail subject optional

RichieB2B:
The threat level is not a real reliable attribute of an event. We rather not put it in the E-mail subject of the notifications MISP sends out. This patch makes it optional but will put it in the E-mail subjects by default.


iglocska:
Looks good, thanks a lot!

PR #1774, opened 2016-12-20
Added support for creating users into different roles depending on ld…

enemarke:
#### What does it do?
*Added support for creating users into roles depending on their ldap group membership.
*Also added support for only allowing users in a specific group to access MISP.

#### Questions

Yes, with FreeIPA

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot looks very clean.

PR #1779, opened 2016-12-21
Use Fuzzy_Hash_Value for ssdeep

### LINKED ISSUE #1771: Export ssdeep hashed as STIX/CyBOX Fuzzy_Hash_Value
When exporting an SSDEEP hash to STIX, this is encoded as a cyboxCommon:Simple_Hash_Value element of cyboxCommon:Type _other_. It would make more sense to use a Fuzzy_Hash_Value with cyboxCommon:Type _ssdeep_.

See also http://stixproject.github.io/data-model/1.1.1/cyboxCommon/HashType/

RichieB2B:
It might be ugly but it fixes #1771

### DIFF
--- a/app/files/scripts/misp2cybox.py
+++ b/app/files/scripts/misp2cybox.py
@@ -85,21 +85,27 @@ def generateObservable(indicator, attribute):
 def resolveFileObservable(indicator, attribute):
     hashValue = ""
     filenameValue = ""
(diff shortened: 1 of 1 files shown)


iglocska:
Disgusting#comma# but that's not your fault. Thanks for fixing this!

PR #1786, opened 2016-12-23
Fix STIX exports for malware-sample attributes

RichieB2B:
Fix a bug introduced by #1779 that causes STIX exports of events with a malware-sample attribute to fail.


iglocska:
Thanks a lot!

PR #1833, opened 2017-01-10
Fixed typo in xINSTALL.centos7.txt

BenDrysdale:
xINTALL.centos7.txt refers to config.php when setting the timezone. This should be core.php.


iglocska:
Nice catch, thanks a lot.

PR #1834, opened 2017-01-10
Fixed typo in dependency installs

mdtro:
rh-php56-bcmath should be rh-php56-php-bcmath

#### What does it do?

Just a minor edit correcting a typo when running through the install document in the dependency section.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

Issue #1246, opened 2016-06-08
Run test on remote Server: enableCrypto operation on CakeSocket


jaegeral:
Trying to set up connection to a new server.

I have two MISP instances that verified the issue.

MISP1 is productive and already connected to a remote MISP (so connection with proxy works)
MISP2 is a VM (the test VM) with updated OS and latest version Currently installed version..... v2.4.47 

RMISP (the remote MISP) is  Currently installed version..... v2.4.26 

Added the RMISP as remote connection, added the auth key, uploaded the pem file.

Now running a test it gives:

```
Error: Connection test failed. Reason: "Unable to perform enableCrypto operation on CakeSocket"      
```

Doing a curl 
curl -k https://RMISP/events/xml/download.json
-->

`{"name":"Authentication failed. Please make sure you pass the API key of an API enabled user along in the Authorization header.","message":"Authentication failed. Please make sure you pass the API key of an API enabled user along in the Authorization header.","url":"\/events\/xml\/download.json"}`

I saw https://github.com/MISP/MISP/issues/740
But I updated MISP according to the update guide already, so I assume my cakephp is up2date as well.

I see that traffic is going to my proxy, but no data is on the receiver site. However if I browse the MISP instance, it works and curl also pops up in RMISP logs.

Don't use Proxy authentication.

Any ideas how to fix that?
### Work environment

| Questions | Answers |
| --- | --- |
| Type of issue | Bug |
| OS version (server) | ubuntu |
| OS version (client) |  |
| PHP version | 5.4, 5.5, 5.6, 7.0, 7.1... |
| MISP version / git hash | 2.4.47, hash of the commit |
| Browser | If applicable |
### Actual behavior

"
Error: Connection test failed. Reason: "Unable to perform enableCrypto operation on CakeSocket"  
"

RichieB2B:
When troubleshooting don't assume anything. ;-) Check your CakePHP version with:

```
git submodule status
```

When this is ok (running 2.8.x) check what MISP is actually sending to your proxy using tcpdump and/or wireshark.

Rafiot:
Assuming it's fixed.

jaegeral:
Nope, still an issue:

`root@misp:/var/www/MISP/app# git submodule status
 e7b22b70e0389c39b3ae1e863121d3859a061b62 ../PyMISP (v2.4.53-22-ge7b22b7)
 73dcbf2aa33b79512373baa63b39adf3fb838608 Lib/cakephp (2.8.9)
[...]

PR #1856, opened 2017-01-20
Small UI patch to make users aware to upload *.pem files

jaegeral:
#### What does it do?

hopefully reduces risk to run again in something like: #1246

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot, good idea!

PR #1860, opened 2017-01-23
Truncate bro cached export files

RichieB2B:
MISP only appends to bro cached export files, making them grow larger and larger until they fill up the entire partition, crashing MISP and making MISP admin very unhappy. This patch fixes all of that. In a single line.


adulau:
Nice finding. Thank you.

iglocska:
Woah nice find. Thanks a lot!

Issue #1691, opened 2016-11-29
Query user from uuid by API


truckydev:
Hi team,

Is there an easy way to get user mail from an uuid attribut by API ?

Exemple : An attribut with bad regex
actual workflow is : 
- get event id and copy
- go to audit > search log
- select "Event" in model and pst your ID

The page indicates the e-mail address of the user who modified or added
### Expected behavior

exemple : with  python, my script search for attribut with regex (with pymisp) an test the value.
If test wrong I would like to request for e-mail via pymisp.

Of course, the email address would be returned only if I have the right rights and in the same org

I propose to return this information only when the search is done with the uuid.
Here a sample : 
```python
{"request": {"uuid":"57575757-0000-4242-8c13-0a0b0c0d0e0f"}}
{"response": [
        {
            "Event": { .....
                 "Attribute": [
                         {
                        "category": "External analysis", 
                        "comment": "ProcessName.", 
                        "uuid": "57575757-0000-4242-8c13-0a0b0c0d0e0f", 
                        "event_id": "42", 
                        "timestamp": "1460000000", 
                        "to_ids": false, 
                        "deleted": false, 
                        "value": "mystest", 
                        "sharing_group_id": "0", 
                        "ShadowAttribute": [], 
                        "SharingGroup": [], 
                        "distribution": "0", 
                        "type": "pattern-in-memory", 
                        "id": "2016",
                        **"user_add" : "test1@ok.com",
                        "user_last_edit" : "test2@ok.com"**
                    }, ]}}]}

```

truckydev:
maybe better like this : 

{"request": {"uuid":"57575757-0000-4242-8c13-0a0b0c0d0e0f", "user_info":"true"}}

PR #1881, opened 2017-01-31
auditor user can see user creator email

truckydev:
## Why this pull request ?

Now you can create a user with auditor role.

add auditor roles with SQL or by UI
SQL : 
```SQL
INSERT INTO roles VALUES (7,"Auditor","2017-01-27","2017-01-27",0,0,0,0,0,0, 0,1,0,1, 0,0,0,0,0,0,0);
```
UI : 
- Name : Auditor
- Permissions : Read only
- [x] Perm Audit
- [x] Perm Auth

Create new user and assign Auditor role.
This user can search event and have access to the creator of the event if it is in the right organization.

#### sample with python  : 
```python
misp = PyMISP('http://mymisp.com/', 'AuditorApiKey')
print(misp.search(eventid='42')['response'][0]['Event']['event_creator_email'])
```
Result : 
```
test@test.com
```
#### Restriction : 
I just added the code for JSON. Look later if having this information in XML is necessary ^^

#### What does it do?

This proposal is an alternative solution for this issue : https://github.com/MISP/MISP/issues/1691

#### Questions

- [x] Does it require a DB change? (if you use sql ether not ^^)

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Model/Event.php
+++ b/app/Model/Event.php
@@ -1437,6 +1437,11 @@ class Event extends AppModel {
 				}
 				$event['ShadowAttribute'] = array_values($event['ShadowAttribute']);
 			}
+			// for auditor user : 
+			// insert into roles values (7,"Auditor","2017-01-27","2017-01-27",0,0,0,0,0,0, 0,1,0,1, 0,0,0,0,0,0,0);
+			if ($user['Role']['perm_audit']) {
+				$event['Event']['event_creator_id'] = $event['Event']['user_id'];
+			}
 		}
 		return $results;
 	}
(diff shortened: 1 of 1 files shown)


iglocska:
Wouldn't this also allow an auditor to see the e-mail address of events of other orgs?

truckydev:
I don't do this check :(

I add this soon

truckydev:
Ok, with this commits, it's ok :)

Return False if user not in the good org else, email creator
https://github.com/MISP/MISP/pull/1881/commits/8a7fbd25f2343d62c4aed770221ad48edec8db53

PR #1882, opened 2017-01-31
adds the option to add required fields and a validation pattern to the import

cristianbell:
adds the possibility to mark required fields and a validation pattern to the import. 

in the import module the following need to be set -  'required': True, 'pattern': '[a-z][0-9]{4,8}'}

#### What does it do?

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
We already have validation on the back-end, front-end validation is not something that we want for this. Please open a github issue if you run into issues with the current validation system.

PR #1886, opened 2017-02-03
attributes/text - optionally export attributes from not published events

cvandeplas:
#### What does it do?

This commit adds the option to download attributes from not published events. The new parameter is optional and should not affect the current functionality. 

This is not tested on a running MISP installation, thank you for your review and patience.

#### Questions

- [X] Does it require a change in the API (PyMISP for example)?  Yes and no: This change is transparent, however PyMISP will need a minor change to be able to use this feature. (upcoming PR)

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/AttributesController.php
+++ b/app/Controller/AttributesController.php
@@ -1966,8 +1966,8 @@ class AttributesController extends AppController {
 		$this->__downloadAttachment($this->Attribute->data['Attribute']);
 	}
 
-	public function text($key='download', $type = 'all', $tags = false, $eventId = false, $allowNonIDS = false, $from = false, $to = false, $last = false, $enforceWarninglist = false) {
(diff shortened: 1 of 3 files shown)


iglocska:
Thanks a lot, looks good!

RichieB2B:
Does this play nicely with MISP. unpublishedprivate = true ? In that case the allowNotPublished flag should be ignored.

iglocska:
Good point, will have a look today.

iglocska:
Just had a quick look, it should be fine. Attribute->text() i routed through Attribute->fetchAttributes(), the attribute level centralised fetching algorithm that is unpublishedprivate aware. :)

RichieB2B:
Great! Just making sure..

iglocska:
Glad you're keeping an eye on it, I wasn't sure myself. I remember this particular API was the last one to get migrated to the centralised fetcher.

Issue #1658, opened 2016-11-10
Url suricata Export problem


dewiestr:
Issue in https://github.com/MISP/MISP/blob/2.4/app/Lib/Export/NidsSuricataExport.php

When exporting a "url" IOC in MISP to a suricata rule, the generated suricata-rule will check the IOC as being a http_uri. According to the https://www.ietf.org/rfc/rfc1738.txt, a url has the format <scheme>//<user>:<password>@<host>:<port>/<url-path>, which I see a lot of people using as well. 

Unfortunately, the generated suricata rule will never trigger if the full format is used ( ref. https://redmine.openinfosecfoundation.org/projects/suricata/wiki/HTTP-keywords section on http_uri ). 

For example, on an HTTP GET request to the malicious url http://blabla.com/example/malware.exe : 

alert http $HOME_NET any -> $EXTERNAL_NET any (msg: "MISP e3524 Outgoing HTTP URL: http|3a|//blabla.com/example/malware.exe"; flow:to_server,established; content:"http|3a|//blabla.com/example/malware.exe"; fast_pattern; nocase; http_uri; tag:session,600,seconds; classtype:trojan-activity; sid:7854111; rev:1; priority:3;) 

I think this could be solved by the following:
 - Use an ioc url_path instead of url to export as http_uri ( however, current full scheme URLs will still not be handled correctly).
 - If a full url scheme is used, MISP needs a parser that will act accordingly

PR #1912, opened 2017-02-08
NidsSuricataExport refactoring for attribute *URL*

deloittem:
# What does it do ?

## The problem
The current export function for the attribute URL does not work as expected. The problem was already mentioned in issue #1658 and is the following::

1. The function processes the attribute as a whole and looks for the full pattern in the **HTTP-request-URI**. This is a problem since the HTTP-request-URI has multiples forms (see [RFC2616](https://www.ietf.org/rfc/rfc2616.txt)): 

        Request-URI    = "*" | absoluteURI | abs_path | authority
        
    This will only works if the **absoluteURI** is used but not when the **abs_path**. Example:
        
        attribute[URL]: "http://example.com/path1/path2/resource"
        [A] HTTP-request-URI with absoluteURI: "http://example.com/path1/path2/resource" **OK**
        [B] HTTP-request-URI with abs_path: "/path1/path2/resource" **PROBLEM**
        
    In example B, the HTTP-request-URI will never contain the attribute[URL] and we won't get any Suricata alert.
    
2. The function does not process the attribute URL as an URL. As explained in (1), the processing looks for the attribute in the HTTP request and assume that the URL is HTTP which is not always the case. We should process the attribute URL as defined in [RFC1738](https://www.ietf.org/rfc/rfc1738.txt):

        <scheme>//<user>:<password>@<host>:<port>/<url-path>
        
## The refactoring

1. We assume the user wrote an URL (RFC1738) and we look for the scheme.
2. Two cases
    * **there is a scheme**: we process the URL on the basis of this scheme
    * **no scheme**: this is not a valid URL and we can't do anything except processing the URL like before. This means looking for the attribute[URL] in the **HTTP-Request-URI** which was the default behavior. Note that this prevents changing how are processed old attributes.
3. Depending of the scheme
    1. **HTTP**: parsing the URL and create a rule which combines the information from the **url-path* and the **host**
    2. **HTTPS**: parsing the URL and looking for the host
        * host is a Domain Name: create a rule which looks for a match between the certificate subject coming from the server and the Domain Name
        * host is an IP address: create a new TCP rule on port 443 (default) with layer-protocol-detection=TLS since the certificate cannot be used. Note that with layer protocol specified the alert will only trigger when a packet containing TLS protocol is sent
    
    3. **FTP**: parsing the URL and looking for the host
    
        * host is a Domain Name: we cannot create a rule for FTP. The only option is to create a DNS rule on port 53 for this Domain Name but we leave this to the user who could create a domain attribute
        * host is an IP address: create a new TCP rule on port 20,21 (default) with layer-protocol-detection=FTP. Note that with layer protocol specified the alert will only trigger when a packet containing FTP data is sent (e.g. username data) but not at the initial TCP connection on port 20,21
    
    4. **SSH**: similar to FTP

## Examples

Here is a list of examples with different kind of URL and what is checked by Suricata for the generated rules:

#### HTTP rules
http

http_ip

http_port

#### HTTPS rules
https

https_ip

https_port

#### SSH rules
ssh

#### FTP rules
ftp

# Questions

#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch

### DIFF
--- a/app/Lib/Export/NidsExport.php
+++ b/app/Lib/Export/NidsExport.php
@@ -496,4 +496,32 @@ class NidsExport {
 		}
 		return false;
 	}
+
+    public static function getProtocolPort($protocol, $customPort) {
+        if($customPort == null) {
+            switch ($protocol) {
+                case "http":
+                    return '$HTTP_PORTS';
+                case "https":
+                    return '443';
+                case "ssh":
+                    return '22';
+                case "ftp":
+                    return '[20,21]';
+                default:
+                    return 'any';
+            }
+        } else {
+            return $customPort;
+        }
+    }
+
+    public static function getCustomIP($customIP) {
+        if(filter_var($customIP, FILTER_VALIDATE_IP)) {
+            return $customIP;
+        }
+        else {
+            return '$EXTERNAL_NET';
+        }
+    }
 }
--- a/app/Lib/Export/NidsSuricataExport.php
+++ b/app/Lib/Export/NidsSuricataExport.php
@@ -88,28 +88,124 @@ class NidsSuricataExport extends NidsExport {
 	}
 
 	public function urlRule($ruleFormat, $attribute, &$sid) {
-		// TODO in hindsight, an url should not be excluded given a host or domain name.
-		//$hostpart = parse_url($attribute['value'], PHP_URL_HOST);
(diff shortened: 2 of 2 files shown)


iglocska:
Great, thanks a lot!

PR #1924, opened 2017-02-12
Add sudo for cp logrotate

RichieB2B:
As suggested by @imifos in #1920


adulau:
Thank you @RichieB2B and @imifos.

PR #1936, opened 2017-02-15
fixed install instructions

Deventual:
Added php-xml, without it this issue can rise:
Class 'DOMDocument' not found in [/var/www/MISP/app/Lib/cakephp/lib/Cake/Utility/Xml.php

- [ x ] Are you using it in production?

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Surprised this isn't included by default, thanks!

PR #1937, opened 2017-02-15
Fixed install instructions

Deventual:
Added php-xml, without it this issue can rise:
Class 'DOMDocument' not found in [/var/www/MISP/app/Lib/cakephp/lib/Cake/Utility/Xml.php

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #1969, opened 2017-02-21
Add a option to sign GPG emails

devnull-:
Hello all,

Quick PR to add a option to sign GPG emails. This option might be useful in several specific/downgrade cases. For example avoiding to save/use a private GPG key on insecure/untrusted MISP instance (connected to Internet, on the cloud solutions, etc.). By default, all the GPG email will be signed (no changes on the actual situation/conduct)

#### What does it do?

Add a option to enable the signing of GPG emails. By default, GPG emails are signed.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


adulau:
@iglocska any specific reason why we don't merge this one?

iglocska:
None, never looked into it, but looks clean. Will merge it now.

PR #1971, opened 2017-02-21
feature: Add support for user creation with authkey

restanrm:
#### What does it do?

This allow PyMISP to create a user with a specified authkey. If "authkey" isn't specified, a value is autogenerated by conventional way.

A similar pull requested has been pushed to PyMISP repo.

#### Questions

#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch


iglocska:
Thanks a lot.

PR #1972, opened 2017-02-21
new: Add reverse proxy support for test of baseurl

restanrm:
#### What does it do?

This patch update testBaseUrl to check if MISP is behind a reverse proxy and conventional headers have been setted.

2 functions are added in order to return originally requested protocol or hostname.

Old parsing of $_SERVER values are kept in case the headers variable are not setted.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Model/Server.php
+++ b/app/Model/Server.php
@@ -2126,10 +2126,40 @@ class Server extends AppModel {
 		return true;
 	}
 
+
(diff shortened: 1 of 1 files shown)


iglocska:
Looks good, thanks a lot!

PR #1974, opened 2017-02-22
Update README.md

ppanero:
Updated readme with apache config for API/Syncs filtering from SSO

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot, that looks great and something we should look into for the LDAP auth too.

PR #1980, opened 2017-02-23
Minor update to start.sh

SteveClement:
#### What does it do?
- Added root check to make sure UID 0 does not execute start script
- Changed "-i" to "--interval"
- Added comment about bash quirk **cd "${0%/*}"** 

#### Questions
Yes

#### Release Type:
- [X] Minor


iglocska:
Awesome idea, thanks!

PR #1982, opened 2017-02-23
Update Server.php

ppanero:
Duplicate entry of property

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot, nice catch!

Issue #1677, opened 2016-11-18
Configuration Issues


jimbolya0607:
I'm sure this is super easy.  

I believe I have the files installed correctly.  What I'm trying to accomplish is to run MISP from the laptop that I installed it on.  I'm unsure as to what to make the config files for MISP.  I've tried putting in something simple like 192.1.1.1 in the baseurl and then try that from browser but that didn't seem to work.  Any help with config to get this up and running would be great. 

# This template is meant for bug reports, if you have a feature request, please be as descriptive as possible and delete the template

*If you would like to report a bug, please fill the template bellow*

### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug, Question, Feature Request, support...
| OS version (server)       | Debian, ubuntu, CentOS, RedHat, ...
| OS version (client)       | XP, Seven, 10, Ubuntu, ...
| PHP version               | 5.4, 5.5, 5.6, 7.0, 7.1...
| MISP version / git hash   | 2.4.XX, hash of the commit
| Browser                   | If applicable

iglocska:
Can you try with localhost or 127.0.0.1?

jimbolya0607:
yea - will try that and report back.

jimbolya0607:
local host resolves to the apache default page "it works!"

CheYenBzh:
Are you sure your DocumentRoot is /var/www/MISP/app/webroot in your apache conf?

Have you enable this conf with: 
`sudo a2dissite default-ssl`
`sudo a2ensite misp-ssl`
`sudo systemctl restart apache2`

Can you copy here your apache configuration?

jimbolya0607:
apache2.conf as follows.  I guess, also keep in mind that I have thehive project on this machine as well.  trying to run them both locally.  thehive is also local host.  I'm sure this is also part of the issue.  

Mutex file:${APACHE_LOCK_DIR} default

PidFile ${APACHE_PID_FILE}

Timeout 300

KeepAlive On

MaxKeepAliveRequests 100

KeepAliveTimeout 5

User ${APACHE_RUN_USER}
Group ${APACHE_RUN_GROUP}

HostnameLookups Off

ErrorLog ${APACHE_LOG_DIR}/error.log

LogLevel warn

IncludeOptional mods-enabled/_.load
IncludeOptional mods-enabled/_.conf

Include ports.conf

<Directory />
    Options FollowSymLinks
    AllowOverride None
    Require all denied
</Directory>

[...]

PR #1988, opened 2017-02-24
Script to wipe (reset) a MISP installation

RichieB2B:
As requested in #1677 

For now, the script will reset a MISP installation to same state as a fresh install. It leaves the first 3 users though, as this is useful for the MISP training instance.

It would be nice if the taxonomies/warninglists/galaxies could be automatically be repopulated by the script in the future.


iglocska:
Awesome, thanks!

PR #1990, opened 2017-02-24
Account for broken apache_request_headers()

RichieB2B:
On Ubuntu Xenial appreantly I have apache_request_headers() but it does not return the required headers. This patch makes testBaseURL() work regardless.

### DIFF
--- a/app/Model/Server.php
+++ b/app/Model/Server.php
@@ -2119,13 +2119,16 @@ class Server extends AppModel {
 
 
 	public function getHost() {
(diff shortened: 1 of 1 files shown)


iglocska:
Didn't see your PR and pushed a fix the other day, does that work for you?

RichieB2B:
Yep, a2c9cfb18fdc0ac46f8a2905e4d1c3180d4ef926 works fine.

PR #1994, opened 2017-02-26
Added -H flag to the sudo that running as www-data

Deventual:
Added -H flag to the sudo that running as www-data,
It eliminates this error: "warning: unable to access '/root/.config/git/attributes': Permission denied".
It can fix the problem when the home variable is not set and the script is trying to write to the root folder instead.
It possible that this issue will only manifest in container environment.

- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/INSTALL/UPDATE.txt
+++ b/INSTALL/UPDATE.txt
@@ -38,13 +38,13 @@ git submodule update --init --force
 cd /var/www/MISP/app/files/scripts/
 rm -rf python-cybox
 rm -rf python-stix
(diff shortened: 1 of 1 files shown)


SteveClement:
Old, PR, for reference, man sudo:

>      -H, --set-home
>                  Request that the security policy set the HOME environment variable to the home directory specified by the target
>                  user's password database entry.  Depending on the policy, this may be the default behavior.

Currently I deem this to be not necessary.

PR #1996, opened 2017-02-26
Updated comment for enrichment modules to reference value used for enri…

kx499-zz:
…chment for added context

#### What does it do?
It solves an issue I was coming across with identifying what attribute the enrichment came from. Passive Total whois details is a prime example, dns/reverse dns resolution is another. Ideally it would be nice if we could associate attributes to each other, but this is a quick fix for eye-balling what a piece of enrichment is related to within the event. 
capture

#### Questions

- No DB change?
- Not yet being used in production, but tested
- No change in the API.

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -3936,7 +3936,7 @@ class EventsController extends AppController {
 				$importComment = $result['comment'];
 			}
 			else {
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks a lot, looks great!

PR #2000, opened 2017-02-28
Bulk email

devnull-:
Hello all,

Tiny PR to add the option "Precedence : bulk" in email headers in order to filter auto-replies when you publish an event with email.

#### What does it do?

Add the option "Precedence : bulk" in email headers in order to filter auto-replies when you publish an event with email.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #2002, opened 2017-03-01
bro to_IDS and published flags fix on query. Now supporting block_eve…

ppanero:
#### What does it do?

Bug fix on BRO export. It was not taking into account unplublished or to_ids conditions in th equery. Therefore the export was not inline with "block_proposal_attirbutes"

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #2031, opened 2017-03-09
Suricata export update

deloittem:
#### What does it do?

- Fix bug when an URL is formed incorrectly (the host was not initialized)
- Add attribute tags in MSG field of the generated Suricata rule. This allows to sort directly and easier the raised alerts using these tags

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Neat, thanks a lot!

PR #2033, opened 2017-03-09
minor bugfix on TagsController.php

rmarsollier:
#### What does it do?

Fix a typo due to copy-paste i guess. 
This typo broke tag() and untag() pymisp calls.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Nice catch, thanks!

PR #2041, opened 2017-03-10
Update install instructions

seanthegeek:
- Configure Apache to:
- Only use strong, modern cyphers
- Enable client-side attack protections

- Add instructions for Debain 9 "stretch"
- Use php 7.0 on Ubuntu 16.04

### DIFF
--- a/INSTALL/INSTALL.debian8.txt
+++ b/INSTALL/INSTALL.debian8.txt
@@ -168,7 +168,17 @@ openssl req -newkey rsa:4096 -days 365 -nodes -x509 \
         SSLCertificateFile /etc/ssl/private/misp.local.crt
         SSLCertificateKeyFile /etc/ssl/private/misp.local.key
 #        SSLCertificateChainFile /etc/ssl/private/misp-chain.crt
+        SSLProtocol             all -SSLv3 -TLSv1 -TLSv1.1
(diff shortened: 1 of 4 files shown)


RichieB2B:
Thanks for the additions. Is only supporting TLS 1.2 really feasible yet? Can you show us a result from https://www.ssllabs.com/ssltest/ of a server running with this config?

seanthegeek:
@RichieB2B Yep. The updated TLS config comes from the "modern" preset of the [Mozilla SSL Configuration Generator][1].

> Oldest compatible clients: Firefox 27, Chrome 30, IE 11 on Windows 7, Edge, Opera 17, Safari 9, Android 5.0, and Java 8

Unfortunately, none of my servers are directly public facing, so I can't run an outside test, but that config does work. 

[1]: https://mozilla.github.io/server-side-tls/ssl-config-generator/?server=apache-2.4.18&openssl=1.0.1e&hsts=yes&profile=modern

RichieB2B:
That should indeed be no issue for MISP. Personal/commercial sites is another issue. Lots of old browsers out there..

seanthegeek:
I would hope security personnel would not be using unpatched/unsupported browsers. I'll include the older set in the instructions  as a fallback.

RichieB2B:
Until #2080 is fixed it is not a good idea to make MISP only accept TLS 1.2 connections.

SteveClement:
Dear @seanthegeek 

This is a fairly old PR and some of the INSTALL documents already include some of your proposed changes.

If you do not mind I will go through your changes in this PR by hand and integrate them in a new PR.

In case you prefer to do it yourself, that is no problem either.

In the meantime I close this PR. Thanks for the contrib.

PR #2056, opened 2017-03-17
Add perm_sighting to initial database and roles

RichieB2B:
perm_sighting was missing from the initial database and default roles


iglocska:
Woops, completely forgot about this. Thanks! :)

PR #2057, opened 2017-03-17
misp-wipe.sh fixes

RichieB2B:
* misp-wipe.sh does not backup, so no outputdir is needed
* clear data model cache upon wiping misp


FloatingGhost:
Line 42: you should probably use `read -s` to avoid echoing your password to the term

RichieB2B:
@FloatingGhost Absolutely! I stole this code from https://github.com/MISP/MISP/blob/2.4/tools/misp-backup/misp-backup.sh#L42 but never used (or tested) it, since I just the .conf file and script gets the password from the database.php file.

RichieB2B:
Wait, I thought @adulau merged this with c94d117d86a59feedb1f6f5594efc44d23a40950 but it is a reverse merge or something? I still would like this to be applied.

adulau:
Thank you for the update. I just merged with the current HEAD nothing extravagant ;-)

PR #2068, opened 2017-03-21
testForPath regex fixed in Server.php

ppanero:
#### What does it do?

BugFix in app/Model/Server.php testForPath Function. It was not undestandin concatenated paths such as the one in rh_shell_fix_path config default value.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

Issue #2050, opened 2017-03-15
Question on tag restriction


deloittem:
Hello,

I have a question about the feature:  Tag restriction to restrict the usage of a tag to a single instance.

What are the rules during the sync with other misp instances for the tags ? What are the definitions of tag exportable switch and the tag restriction when syncing ?

Regards

iglocska:
exportable simply means that the tag will be included in the sync (and API access for that matter). I think what you are after (and correct me if I am wrong) is that you have some tag that you don't want others to be able to set besides your own organisation (for example a tag that you in your community automate on). 

What you can do: You can restrict a tag to your organisation, that way others pushing events to you with that tag applied will get the tag filtered out.

deloittem:
Thank you for your help.

The exact problem I encounter is the following: I have a tag I do not want to include in the sync but if I set the exportable switch to FALSE, the tag is not included in export functions such as suricata rules

deloittem:
any way to achieve this ? keeping the tag in local export functions such as rules but not exporting it when syncing ?

Thanks

iglocska:
I think the solution is to add a method to include it in the exports even if it's set to non exportable. I'll look into adding something for this (the internal fetcher already uses the "includeAllTags" keyword)

deloittem:
Should "includeAllTags" keyword be added when calling fetchAttributes() in function nids() ?

iglocska:
Best would be to have an includeAllTags parameter for the nids function and then pass that along indeed to fetchAttributes()

deloittem:
a new parameter to function nids() like what has been done for $enforceWarninglist ?

iglocska:
Yup! I can do it in a bit, but it's a bit of a crazy week, we're giving a training at the moment. If you would like to implement it and do a PR that would be awesome, but if not, don't worry about it :)

deloittem:
Ok I will do the implementation and a PR. last question : the value of includeAllTags should be set to true in cachenids() then ?

iglocska:
Ha good question. I'd say no, I'd keep the non exportable tags from being
included by default and let users override it via the API.

PR #2069, opened 2017-03-22
All tag cannot be included in export functions such as suricata rules

deloittem:
# All tag cannot be included in export functions such as suricata rules

#### What does it do?

If fixes issue: #2050 where it was mandatory to set EXPORTABLE TAG to true in order to get the tag displayed in Snort/Suricata rule.

#### How ?
includeAllTags parameter was added to the API for the nids function and then pass that along to all function calls

#### Questions

#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -1869,7 +1869,7 @@ class EventsController extends AppController {
 		return $results;
 	}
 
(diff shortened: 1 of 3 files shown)


iglocska:
Looks good thanks a lot, and a special thank you for also updating the documentations ;)

Issue #2070, opened 2017-03-22
Attribute fields: value and value1,value2


deloittem:
Hello,

I have a question related to attribute fields: value and value1,value2.

I'd like to add NIDS rules for the new attribute[type] = IP-SRC|PORT and IP-DST|PORT but how should these fields be managed in NidsExport() since we currently only receive the virtual field = attribute[value] ?

I could split the virtual value again to value1 and value2 in NidsExport() but what about updating the fetchAttributes() function to directly retrieve value1,value2 instead of value ?

Regards

iglocska:
Changing fetchAttributes to retrieve value1 and value2 separately would break all existing functionality that expects value (such as the synchronisation) so it's definitely out of the question - however, just running explode('|', $attribute['value']) on composite attribute types should do the trick for you.

Edit: I was wrong about the synchronisation, it only relies on fetchAttributes(). However, other exports will rely on the same functionality, so running the explode is still highly prefered.

deloittem:
We could still retrieve value+value1+value2 (3 fields) in fetchAttributes() to prevent breaking functionalities but ok I will do the split again using explode()

iglocska:
Indeed, retrieving all 3 would work, though keep in mind it would inflate the returned data size for all attribute based exports with a bunch of duplication. CPU time is cheap, I'd just go with that ;)

PR #2073, opened 2017-03-23
Manage attributes IP-SRC|PORT and IP-DST|PORT when exporting nids rules

deloittem:
# Manage attributes IP-SRC|PORT and IP-DST|PORT when exporting nids rules

#### What does it do?

It fixes issue #2070

#### How ?
Value1:IP and Value2:PORT are exploded and the Suricate function for IP rule generation is called with value1. A port parameter has been added.

#### Questions

#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch


iglocska:
Looks good, thanks a lot!

PR #2091, opened 2017-03-28
Update xINSTALL.centos6.txt

jhopp1e:
#### What does it do?

Addresses PHP package install error: "No package rh-php56-bcmath available."
Change package name on line 36: 
From: "rh-php56-bcmath" 
To: "rh-php56-php-bcmath"

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


adulau:
Thank you very much.

PR #2112, opened 2017-04-05
Misp-wipe: Ensure root, don't echo password

FloatingGhost:
#### What does it do?

When running misp-wipe, ensure we're running as root.
Also, don't echo MySQL password to STDOUT.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #2128, opened 2017-04-11
Snort attribute generation rule now contains the initial msg field

deloittem:
#### What does it do?

When generating IDS rules for attribute "snort" the generated msg field is the following:
`msg: "MISP eXXX [tag] snort-rule"`

This value overwrites the initial msg value of the snort attribute.

The current PR will keep the snort attribube msg value. We will now get:
`msg: "MISP eXXX [tag] snort-rule | the initial snort msg"`

#### Questions

#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch


iglocska:
Thanks a lot!

PR #2137, opened 2017-04-16
Remove default value for column comment

juju4:
#### What does it do?

```
ERROR 1101 (42000) at line 20: BLOB, TEXT, GEOMETRY or JSON column 'comment' can't have a default value
```
https://travis-ci.org/juju4/ansible-MISP/jobs/222624828#L7561
(ubuntu xenial, mysql 5.7)
https://dev.mysql.com/doc/refman/5.7/en/blob.html

Strangely, this does not affect centos7 and mariadb 5.5 even if corresponding documentation states the same.
https://travis-ci.org/juju4/ansible-MISP/jobs/222624827#L4862

Coming from recent https://github.com/MISP/MISP/blame/2.4/INSTALL/MYSQL.sql#L32

#### Questions

- [ ] Does it require a DB change? default Mysql template
- [ ] Are you using it in production? no. from travis CI cron test

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/INSTALL/MYSQL.sql
+++ b/INSTALL/MYSQL.sql
@@ -29,7 +29,7 @@ CREATE TABLE IF NOT EXISTS `attributes` (
   `timestamp` int(11) NOT NULL DEFAULT 0,
   `distribution` tinyint(4) NOT NULL DEFAULT 0,
   `sharing_group_id` int(11) NOT NULL,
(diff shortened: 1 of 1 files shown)


iglocska:
Indeed, thanks a lot!

PR #2154, opened 2017-05-02
Add filename key for import modules

truckydev:
This pull/request add filename to all misp import module.

With this module, you can use the name file you try to import to misp like this : 
```python
q = json.loads(q)
print(q.get('filename'))
```

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #2155, opened 2017-05-02
misp.search_index returns now a same format than misp.search

PaulSec:
#### What does it do?

This fixes the opened bug on [PyMISP](https://github.com/CIRCL/PyMISP/issues/78) about: "misp.search_index returns a different format than misp.search"

```php
unset($events[$key]['Event']);
```
Has been commented in order to keep the "Event" structure.

#### Questions

Yes, it's now generic, whether you use misp.search and/or mis.search_index. 

#### Release Type:
- [X] Major
- [ ] Minor
- [ ] Patch

This might break code that people already developed so I put it as "Major".

### DIFF
--- a/app/View/Events/json/index.ctp
+++ b/app/View/Events/json/index.ctp
@@ -1,15 +1,15 @@
 <?php
 foreach ($events as $key => $event) {
     // rearrange things to be compatible with the Xml::fromArray()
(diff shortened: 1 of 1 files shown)


SteveClement:
If this is still needed please re-open.

PR #2161, opened 2017-05-04
Change feedback about email notification on sending proposals

### LINKED ISSUE #2158: Proposal notification failures for "at least one recipient"
When making a proposal to another org, MISP tries to alert its members by email.
However, it may find it is not able to notify everyone, in which case the proposer receives this message:
«The proposed Attribute has been saved but sending out the alert e-mails has failed for at least one recipient.»

However, I am more concerned on whether *anyone* received the notification than if *everybody* did. If someone received it, it will (hopefully) be handled internally, but if nobody did, I may need to bug them through different means.

Incidentally, ShadowAttribute::sendProposalAlertEmail() should be returning a boolean, but if the event has an email lock, it calls "return;", with no parameter, which will return a null, and then the above message would be shown as it is weakly compared (==) to false. So it ends up doing the right thing for the wrong reasons. An explicit return false would be in order imho.

Keisial:
Move from a “Failed for at least one recipient” warning notification
to warn when it was not sent to anyone, which is more interesting for
the user sending the proposal.

Fixes #2158

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/ShadowAttributesController.php
+++ b/app/Controller/ShadowAttributesController.php
@@ -374,7 +374,9 @@ class ShadowAttributesController extends AppController {
 					if ($successes) {
 						// list the ones that succeeded
 						$emailResult = "";
-						if (!$this->ShadowAttribute->sendProposalAlertEmail($eventId) == false) $emailResult = " but sending out the alert e-mails has failed for at least one recipient.";
+						if (!$this->ShadowAttribute->sendProposalAlertEmail($eventId) === false) {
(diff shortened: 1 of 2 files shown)


Keisial:
I later amended the commit to change from "organization" to "organisation", which seems preferred on the interface. Feel free to pick any of them.

iglocska:
Looks good, thanks a lot!

PR #2179, opened 2017-05-10
add possibility to define tags for import module.

truckydev:
Add possibility to define tags for import module.
in you use it like this : 
create input in your 
```python
userConfig = {	'default tag': {
				'type': 'String',
				'message': 'Add tags spaced by a comma (tlp:white,misp:threat-level="no-risk")',
				'validation' : '0'
			}
		}
```
When "validation" is 0 you can leave this field empty.

Add tag in your result : 
```python
    if q['config'].get('default tag') is not None:
        r["results"]["tags"] = q['config']['default tag'].split(",")
return r
```

The result page display tag field.
If the tags exist, it will be automatically added to your attribute.
Otherwise, if you have the right to create it, it will be created and added.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -3137,7 +3137,19 @@ class EventsController extends AppController {
 							$attribute['email'] = $this->Auth->user('email');
 							$attribute['event_uuid'] = $event['Event']['uuid'];
 						}
-						if ($this->Event->$objectType->save($attribute)) {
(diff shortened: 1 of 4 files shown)


iglocska:
Great idea and looks very clean, thanks a lot!

truckydev:
Thanks :)

PR #2182, opened 2017-05-11
newsread attribute fixed for user registration via sso

ppanero:
#### What does it do?

Fixes SSO regitration
#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #2183, opened 2017-05-11
Check bad links on attachments

aparriel:
#### What does it do?

Adding small diagnostic on Server Settings > Diagnostics Page to check if some attachments referenced in Database doesn't exist on filesystem.

- Initial state :
verifybadlinkinit

- Let's rename/delete one file on storage directory (app/files/<id_event>/<id_attribute>):
#/var/www/MISP/app/files/1# ls
2  ioc
#/var/www/MISP/app/files/1# mv 2 3
#/var/www/MISP/app/files/1# ls
3  ioc

- Now run the test :
verifybadlinkerror

- Let's rename the file and test again :
#/var/www/MISP/app/files/1# ls
3  ioc
#/var/www/MISP/app/files/1# mv 3 2
#/var/www/MISP/app/files/1# ls
2  ioc
verifybadlinkok

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

#### Improvements
Could be easily improved to list files and database entries in order to help correct anomalies.

### DIFF
--- a/app/Controller/Component/ACLComponent.php
+++ b/app/Controller/Component/ACLComponent.php
@@ -368,6 +368,7 @@ class ACLComponent extends Component {
 					'attributehistogram' => array('*'),
 					'change_pw' => array('*'),
 					'checkAndCorrectPgps' => array(),
+					'checkAttachments' => array(),
 					'dashboard' => array('*'),
 					'delete' => array('perm_admin'),
 					'downloadTerms' => array('*'),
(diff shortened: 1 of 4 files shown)


truckydev:
Great :)

aparriel:
I just realised it would make more sense to put it into AttributesController rather than UsersController.

iglocska:
Great idea, I've moved it to AttributesController as it indeed makes more sense.

Thanks a lot!

Issue #2181, opened 2017-05-11
STIX download returning An Internal Error Has Occurred.


jdubp:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug, Question, support
| OS version (server)       | RedHat 7.2
| OS version (client)       | Windows
| PHP version               | 7.0
| MISP version / git hash   | 2.4.72
| Browser                   | If applicable

### Expected behavior
Download event as STIX XML/STIX JSON

### Actual behavior
When attempting to generate a STIX download from a MISP event within the GUI, I receive

This XML file does not appear to have any style information associated with it. The document tree is shown below.
      <response><name>An Internal Error Has Occurred.</name><message>An Internal Error Has Occurred.</message><url>/events/stix/download/1138.xml</url></response>

In looking at /var/log/messages on the server, below is the error returned

May 11 05:29:17: detected unhandled Python exception in '/var/www/MISP/app/files/scripts/misp2stix_framing.py'
May 11 05:29:18: Executable '/var/www/MISP/app/files/scripts/misp2stix_framing.py' doesn't belong to any package and ProcessUnpackaged is set to 'no'
May 11 05:29:18: 'post-create' on '/var/spool/abrt/Python-2017-05-11-05:29:17-22335' exited with 1
May 11 05:29:18: Deleting problem directory '/var/spool/abrt/Python-2017-05-11-05:29:17-22335'

In looking at /var/www/MISP/app/tmp/logs/error.log, below is the error returned
2017-05-11 05:33:59 Error: [Exception] There was an issue generating the STIX export.
Request URL: /events/stix/download/1138.xml
Stack Trace:
#0 [internal function]: EventsController->stix('download', '1138')
#1 /var/www/MISP/app/Lib/cakephp/lib/Cake/Controller/Controller.php(491): ReflectionMethod->invokeArgs(Object(EventsController), Array)
#2 /var/www/MISP/app/Lib/cakephp/lib/Cake/Routing/Dispatcher.php(193): Controller->invokeAction(Object(CakeRequest))
#3 /var/www/MISP/app/Lib/cakephp/lib/Cake/Routing/Dispatcher.php(167): Dispatcher->_invoke(Object(EventsController), Object(CakeRequest))
#4 /var/www/MISP/app/webroot/index.php(92): Dispatcher->dispatch(Object(CakeRequest), Object(CakeResponse))
#5 {main}

[...]

PR #2186, opened 2017-05-12
fix: Move idgen call to mixbox

FloatingGhost:
STIX moved their idgen calls to mixbox a while back, and now backwards-compat doesn't work. Fixes that. #2181 

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot

PR #2190, opened 2017-05-15
Deal with the stupid errors STIX thinks it's ok to just throw

FloatingGhost:
#2181 
**SHOULD** fix all instances of exporting under different STIX versions. 

Uses stix.utils if possible, if not, tries to use mixbox. If mixbox works, which it sometimes doesn't.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/files/scripts/misp2stix_framing.py
+++ b/app/files/scripts/misp2stix_framing.py
@@ -18,6 +18,7 @@ from stix.common.confidence import Confidence
 from stix.common.vocabs import IncidentStatus
 from cybox.utils import Namespace
 from mixbox import idgen
+from stix import __version__ as STIXVER
 
 NS_DICT = {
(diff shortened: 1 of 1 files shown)


adulau:
Thank you for taking care of the most well-engineered XML format in the cyberspace.

RichieB2B:
That is such a mean thing to say, knowing they gave it up for JSON! :-P

Love those comments! <3

iglocska:
It was a great move for them. Finally we won't ever need libraries for STIX anymore, because it's just JSON and that obviously doesn't need a library.

adulau:
To be on the positive side concerning STIX 2, if you are interested on our ongoing discussions, it's all there:

* https://github.com/MISP/MISP/wiki/NotesMISP-STIX2

Including a proposal for an event SDO. Feedback and support highly appreciated.

PR #2193, opened 2017-05-16
Separate attachments into their own directory

itsnotapt:
#### What does it do?

Separate attachments into their own directory. This allows you to use a different mount point for this directory (in my case s3fuse).

Users will need to move their existing attachments from `/files/<eventid>` to `/files/attachments/<eventid>`. Proposed shadow attachments are still stored locally (this could be changed). Zipping and hashing occurs from the `tmp` directory.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/AttributesController.php
+++ b/app/Controller/AttributesController.php
@@ -292,7 +292,7 @@ class AttributesController extends AppController {
 	}
 
 	private function __downloadAttachment($attribute) {
-		$path = "files" . DS . $attribute['event_id'] . DS;
+		$path = "files" . DS . "attachments" . DS . $attribute['event_id'] . DS;
 		$file = $attribute['id'];
 		if ('attachment' == $attribute['type']) {
 			$filename = $attribute['value'];
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -2152,7 +2152,7 @@ class EventsController extends AppController {
 			$iocData = $fileAccessTool->readFromFile($this->data['Event']['submittedioc']['tmp_name'], $this->data['Event']['submittedioc']['size']);
 
 			// write
-			$rootDir = APP . "files" . DS . $id . DS;
+			$rootDir = APP . "files" . DS . "attachments" . DS . $id . DS;
 			App::uses('Folder', 'Utility');
 			$dir = new Folder($rootDir . 'ioc', true);
 			$destPath = $rootDir . 'ioc';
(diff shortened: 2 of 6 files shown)


iglocska:
I really like the idea but will wait with merging it until the next release. We'll have to move all the existing attachments to the same directory in an upgrade script otherwise all existing attachments become broken. Will take care of it once 2.4.74 is ready!

Thanks a lot!

kallix:
How about having a setting in config.php for the attachments directory ?
With something like: 
```
if attachments_dir is undefined:
   set attachments_dir to '/app/files' # current path
```
That would mean that unless the user does change this setting, there is nothing to migrate, and nothing is broken. 
As an added bonus: Those who want would be able to put their attachments dir anywhere (in particular, outside the /var/www/MISP hierarchy)

iglocska:
Would work, but we also need a way to migrate the data then.

If I already have a MISP instance and changed the setting, I'd expect all
my existing data to still be available.

kallix:
Not allowing this setting being changed from the interface would do the trick :-)
Not something that users would do everyday...

Migrating the data from MISP itself would be tricky. Not sure that it can be done simply AND reliably. 
In my mind, not even mentioning this setting in the UI, but having a paragraph about that in the doc that says it should only be done while MISP is shut down would be acceptable.

RichieB2B:
I agree with Kevin: changing this in a config.php only (or with a big warning in the GUI) will put the burden of the mv on the admin changing it.

iglocska:
Neat idea, indeed!

kallix:
Here is a quick first try: e1981d52f4aa67fd771f9a967a58ed338389047f
It has received only very minimal testing (some parts I don't even know how to exercise).
From what I managed to test, it seems to work™: 
- If you don't add the `attachments_dir` setting in config.php, it is set to `app/files` (i.e. the place attachments have always been). 
- If you do define it (to a valid dir): Attachments will be read from/written to there.

Does that seem like a reasonable approach ?

Also: I have no idea of the cost of calling `$this->loadModel('Server');` Should I only call `loadModel('Server')` when `$this->Server == null` ?

With this commit, this setting is modifiable in the GUI. Any clever idea on how to change that ? I mean: more clever than `if ($setting_name == 'attachments_dir') continue` ...

iglocska:
Closing this for now due to the other implementation that allows you to define an attachment directory through the settings. Thanks for starting the process!

PR #2194, opened 2017-05-17
Added support for attribute tags in the restSearch API

itsnotapt:
#### What does it do?

Allows you to search attributes using attribute tags, using the restSearch API.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/AttributesController.php
+++ b/app/Controller/AttributesController.php
@@ -1649,7 +1649,7 @@ class AttributesController extends AppController {
 	// the last 4 fields accept the following operators:
 	// && - you can use && between two search values to put a logical OR between them. for value, 1.1.1.1&&2.2.2.2 would find attributes with the value being either of the two.
 	// ! - you can negate a search term. For example: google.com&&!mail would search for all attributes with value google.com but not ones that include mail. www.google.com would get returned, mail.google.com wouldn't.
-	public function restSearch($key = 'download', $value = false, $type = false, $category = false, $org = false, $tags = false, $from = false, $to = false, $last = false, $eventid = false, $withAttachments = false, $uuid = false, $publish_timestamp = false, $published = false, $timestamp = false, $enforceWarninglist = false, $to_ids = false, $deleted = false) {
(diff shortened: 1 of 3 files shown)


iglocska:
Whilst we definitely need to allow users to use attribute level tags in their filters, I see two major issues with this PR:

1. It introduces a new parameter in the middle of the parameter list, which generally is not an issue for those using the (much preferred!) POST method of using the API, but will break existing scripts that rely on url parameter order (which sadly is quite common).

2. This is a more conceptual question that we've had some internal discussions about, but maybe the best approach would be to reuse the existing tag filter on both an event and an attribute level for the lookup. Basically, any attribute in an event carrying a tag implies that the event's tag is also applied to the attribute. This of course raises some questions about exclusivity which is not handled by the tagging system yet (tlp:white event with a tlp:amber attribute for example will have both tags matching).

I think the best approch would be to alter fetchAttributes() in Model/Attribute.php and fetchEvent() in Model/Event.php to simply filter on both levels.

itsnotapt:
Okay, that makes sense. It might be up to the user to do post processing on the results if they are in the scenario you mentioned. Alternately, they might be required to create new event tags to distinguish from attribute tags.
I'll look to update the PR next week as per your approach.

Keisial:
> This of course raises some questions about exclusivity which is not handled by the tagging system yet (tlp:white event with a tlp:amber attribute for example will have both tags matching)

There are basically two ways to treat them. As a replace or as an add. The tags themselves could be marked as exclusive or not (an item can only carry a single tlp tag, but could be on multiple circl:incident-classification tags).
The problem though, is how to handle an attribute **not** having a tag of his parent
 (for instance an incident could relate to phishing and malware, but we may want a specific attribute to have just one of those).

adulau:
@Keisial Indeed the "inherit behaviour" in tagging raised many questions regarding tag exclusivity especially when tags are mutually exclusive (tlp:amber at event level and tlp:white at attribute level). How would you see the see the tag exclusiveness in the taxonomy format? We could add a specific model in the format and it's up to the MISP instance/site-admin to enforce it or not.

adulau:
@Keisial @crruthe FYI, we recently introduced in the taxonomy, the exclusive flag to allow a predicate or a set of values to be exclusive. https://github.com/MISP/misp-taxonomies/blob/master/tlp/machinetag.json#L40 

This might solve partially some of the issues already discussed.

SteveClement:
Closed due to no activity and the general consent of it being not an ideal solution.

PR #2207, opened 2017-05-19
Also test for mixbox version

RichieB2B:
One of my servers did not have the mix box python package. STIX exports in MISP just failed without being able to easily trace the problem. This patch adds the mixbox version to the diagnostics page.


adulau:
Thank you. It was a good idea.

PR #2209, opened 2017-05-19
Fix invalid newsread

ghost:
#### What does it do?

While using Apache Secure Auth it will try to insert a broken field "newsread".
Looked around and found other code fragments which are using newsread => 0.
May be someone who is more familar with the code can comment on this if int(zero) isn't the best choice.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

Issue #2192, opened 2017-05-16
Baseurl / rewrite Problems


t-sec-cons:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | Ubuntu 16.04.2 LTS
| OS version (client)       | Seven
| PHP version               | 7.0
| MISP version / git hash   | v2.4.71
| Browser                   | Chrome

### Expected behavior
Misp should use the 'baseurl' config to calculate all links.

### Actual behavior
If you try to use a subdirectory of a domain for baseurl, or the relative path (let the 'baseurl' variable empty) it works for some urls and links, but on some urls the subdirectory is missing and not rewritten.
I have an apache setup where the MISP-instance is running as an alias ( /misp/ pointing to /var/www/xyz/) on my virtual host (using /var/www/html as root directory).
For example if you try to use a filter on the event list, you get "**Not Found
The requested URL /events/index/searchall:wannacry was not found on this server.**".
Correct URL is /**misp**/events/index/searchall:wannacry , which is working if you try it manually.

### Steps to reproduce the behavior
- add an alias directory to the apache vhost (ie: Alias /misp /var/www/html/MISP/app/webroot )
- set the baseurl to the alias ( 'baseurl' => 'https://xyz.com/misp/' ) or to relative ('baseurl' => '')
- try to add a filter to the event list

t-sec-cons:
I looked into it and it seems the baseurl is not writen into every view.
For example the quicksearch button in View/Events/index.ctp Line 43 is hardcodet as /events/index.
You can add <?php echo $baseurl;?> before, to fix this issue. Looks like this is the case in multiple files.

t-sec-cons:
In addition baseurl is completely missing in webroot/js/misp.js, so no javascript buttons are working.

t-sec-cons:
Found the same issue in #1431, sorry for the duplicate. Its not only in javascript, so maybe you take this as an addition to #1431.

1inkman:
Good day! I have this problem too, is it not fixed?

SteveClement:
bump @iglocska

SteveClement:
Dear all, if possible just leave the Base URL empty for the time being.

iglocska:
[...]

Issue #1431, opened 2016-08-11
js not honoring baseurl ?


rmarsollier:
### Work environment

| Questions | Answers |
| --- | --- |
| Type of issue | Bug |
| OS version (server) | Ubuntu srv 14.04 |
| OS version (client) | Debian 8 |
| PHP version | 5.5.9 |
| MISP version / git hash | 2.4.50, 751bad6 |
| Browser | Firefox |

Some additions:
Reverse proxy : apache2 with mod_proxy.
MISP is behind the RP, the URL to access it is `example.com/misp/`
MISP.baseurl is set accordingly
### Expected behavior

Every MISP features working after some RP configuration.

Current RP configuration:

```
<Location /misp/>
         Order deny,allow
         Allow from all
         SSLRequireSSL On
         ProxyPass http://10.10.10.10/
         ProxyPassReverse http://10.10.10.10/
         Header edit Location ^http://(.*)$ https://$1

         SetOutputFilter proxy-html
         ProxyHTMLExtended On
         ProxyHTMLURLMap "^/" "/misp/" "R"
         ProxyHTMLURLMap "https?://([^/]*)(/misp.*)$" "$2/" "R"
         ProxyHTMLURLMap "\"/events/" "\"/misp/events/" "R"
         ProxyHTMLURLMap "\"/jobs/" "\"/misp/jobs/" "R"
         RequestHeader unset Accept-Encoding
```
### Actual behavior

Many feature are not working, mainly the one depending on the misp2.4.XX.js functions to work.
I.e : editing server setting from webUI doesn't work, the url tried is `example.com/servers/serverSettingsEdit/[...]` instead of `example.com/misp/servers/serverSettingsEdit/[...]`.
#### My so dirty workaround

After editing the app/webroot/js/misp2.4.XX.js file to modify the functions (for example `function serverSettingsActivateField(setting, id)`) to include  the missing `/misp/` to the urls, it works. However, this doesn't seem to be the clean fix :)
### Steps to reproduce the behavior

The impacted fonctionnalities are , at least, the following (with the corresponding urls):
- seting edition : /servers/serverSettingsEdit/[...]
- filtering of MyEvents/OrgEvents : /events/filterEventIndex?_= [...]
- event publishing : /events/alert/[...]
- event downloading : /events/exportChoice/[...]
- freetext import : /events/freeTextImport/[...]
- add tag : /tags/selectTaxonomy/[...]
- enrichment /Attributes/hoverEnrichment/[...]
[...]

PR #2210, opened 2017-05-19
Add baseurl to serverSettings Ajax

ghost:
#### What does it do?

Using of BaseURL is missing in some cases. Here I tried to fix it in the part of editing server settings. 

Refer also: #1431 and #2192 
But this small adjustment will probably NOT catch all potentially missing baseurls.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


SteveClement:
Closing for now, needs re-work.

PR #2211, opened 2017-05-20
FIX #2205 attachTagToObject permissions so that tagger role are able …

### LINKED ISSUE #2205: attachTagToObject missing check for perm_tagger
*Expected behavior*

when using the API call "tags/attachTagToObject" through PyMISP, a user with permission tagger can tag objects

*Actual behavior*

due to this condition in TagsController.php line 583
`!$this->_isSiteAdmin() && $object['Event']['orgc_id'] != $this->Auth->user('org_id')`

**cannot tag an object with different orgc_id even though the user has permission tagger**

unlike the API call used by the MISP interface which calls "events/addTag"
in EventsController.php line 2873 where the condition is 
`!$this->userRole['perm_tagger'] || ($this->Auth->user('org_id') != $event['Event']['org_id'] && $this->Auth->user('org_id') !== $event['Event']['orgc_id'])`

checks for perm_tagger

*Steps to reproduce the behavior*

tag an event using a user while event.orgc_id != user.org_id using the PyMisp tag() call (should not work and get "invalid target")

tag an event through the MISP interface with a user while event.orgc_id != user.org_id (should work)

*note*: am i confusing something? or this is actually a bug?

kajogo777:
…to tag objects where obj.orgc_id != user.org_id

#### What does it do?

`#2205`

using "tags/attachTagToObject" users with permission tagger cannot tag objects where obj.orgc_id != user.org_id

however using "events/addTag" allows it

therefore PyMISP tag() call will not work, however the deprecated addTag() and the MISP interface will work

this fixs add a condition to check first if a user has the tagger permission before sending back "invalid target" error

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/TagsController.php
+++ b/app/Controller/TagsController.php
@@ -580,7 +580,7 @@ class TagsController extends AppController {
 		));
 		$type = 'Event';
 		if (!empty($object)) {
-			if (!$this->_isSiteAdmin() && $object['Event']['orgc_id'] != $this->Auth->user('org_id')) {
+			if (!$this->_isSiteAdmin() && !$this->userRole['perm_tagger'] && $object['Event']['orgc_id'] != $this->Auth->user('org_id')) {
 					throw new MethodNotAllowedException('Invalid Target.');
 			}
 		} else {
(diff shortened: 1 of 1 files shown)


iglocska:
Very good point, thanks a lot!

PR #2212, opened 2017-05-22
Few Additional edits

neok0:
#### What does it do?
Attribute level tag search: behaves exactly like the Event Tag search, setting for enable/disable this feature included

Attribute level tagging included into sync process: behaves exactly like the Event Tag syncing, setting for enable/disable this feature included

Related events view modified. If there are more then ten related events in the Events view list, they will be replaced by a '+' placeholder. If clicked it will then show all events. 

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


SteveClement:
@iglocska this is a very old PR. Are we interested in following up?

Issue #2188, opened 2017-05-14
New OSINT Feed by inThreat


gbossert:
inThreat generates an OSINT feed for communities, aggregating massive amounts of raw threat information. This data is mostly uncontextualized but updated in near real-time. This feed provides an extensive number of indicators and should be ingested by appropriate tools only. For this reason, inThreat releases a MISP version of this feed. Events have a 7day time-to-leave and are classified with as many tags, attributes and galaxies possible.

This feed is freely available here : https://feeds.inthreat.com/osint/misp/

I ⋅ being part of inThreat ⋅ believe this feed could be useful for MISP users. For this reason, I would like to register it as a « Default Feed » in MISP. Would it be possible to have your opinion on this ?

PR #2213, opened 2017-05-22
inThreat OSINT Feed as a default Feed

gbossert:
See Issue #2188 for more details.

This PR registers the feed in the MySQL (`INSTALL.sql`) and PostgreSQL (`POSTGRESQL-data-initial.sql`) install scripts. The MISP update process is also modified (`AppModel.php`) to insert the feed when upgrading to v. 2.4.74.

**Feed parameters**

- Provider: inThreat
- Name: inThreat OSINT Feed
- URL: https://feeds.inthreat.com/osint/misp

#### Questions

- [ ] Are you using it in production? _(not this patch, but the feed yes)_

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


adulau:
Already added. So we can safely close it ;-)

Issue #2238, opened 2017-06-01
STIX Export Error After MISP Update (2.4.74)


iamshleya:
This issue is related to issue 2181

We updated our instance to STIX v1.2.0.4. After attempting to export an event in STIX format, we received "An internal error has occurred."

After doing some troubleshooting locally on the machine, we received the following errors. We have mixbox installed and there are no errors on any of the imports.

>>> from mixbox import idgen
>>> from cybox.utils import Namespace
>>> from mixbox import idgen
>>> stix.utils.idgen.set_id_namespace({'test' : 'test2' })
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
**AttributeError: 'module' object has no attribute 'idgen'**

Looking at the code under the misp2stix_framing.py, it does not look like it accounts for an attribute error...

    try:
        stix.utils.idgen.set_id_namespace({baseURL: orgname})
    **except ValueError:**
        # Some weird stix error that sometimes occurs if the stars
        # align and Mixbox is being mean to us
        # Glory to STIX, peace and good xmlns be upon it
        try:
            idgen.set_id_namespace(Namespace(baseURL, orgname))
        **except TypeError:**
            # Ok this only occurs if the script is being run under py3
            # and if we're running a REALLY weird version of stix
            # May as well catch it
            idgen.set_id_namespace(Namespace(baseURL, orgname, "MISP"))

Even though the code failed prior to reaching the exceptions, we tested the logic, and it also looks like that would fail as well.

>>> idgen.set_id_namespace(Namespace('test1', 'test2'))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/site-packages/mixbox-1.0.2-py2.7.egg/mixbox/idgen.py", line 97, in set_id_namespace
    _get_generator().namespace = namespace
  File "/usr/lib/python2.7/site-packages/mixbox-1.0.2-py2.7.egg/mixbox/idgen.py", line 46, in namespace
    raise ValueError("Must be a Namespace object")
**ValueError: Must be a Namespace object**
>>> idgen.set_id_namespace(Namespace('test1', 'test2', "MISP"))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/site-packages/mixbox-1.0.2-py2.7.egg/mixbox/idgen.py", line 97, in set_id_namespace
    _get_generator().namespace = namespace
[...]

PR #2240, opened 2017-06-01
fix: Use IDGen from literally any module that has it

FloatingGhost:
#### What does it do?
 
Tries to fix the godforsaken idgen in #2238 

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


adulau:
Thank you very much.

PR #2243, opened 2017-06-01
Fix: misp-backup explanation for meaning of variables

SHSauler:
#### What does it do?

misp-backup's default OutputDirName (current dir) led to error on Ubuntu 16.04, tar 1.28. Provided works and look neater.

#### Questions

#### Release Type:
- [ ] Major
- [X] Minor
- [X] Patch


iglocska:
Good idea, thanks a lot!

PR #2244, opened 2017-06-02
fix: forgot to catch for weird STIX version

FloatingGhost:
I forgot that some versions of STIX are so strange as to not work

May as well patch it

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


adulau:
Your dedication is impressive ;-) Thank you for the PR.

PR #2251, opened 2017-06-06
Feature/apache auth allow update

stinnux:
#### What does it do?

* Correct ldap group membership checking when using group-to-role mapping
* Allow Updating of User data with ldap data (group membership, email)
* Disable users when no longer member of any required ldap group
* Reenable user when added to a required ldap group

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot, good idea!

At a first glance it all makes sense and looks clean, we don't use LDAP ourselves, so if anyone gets issues with it let us know!

PR #2255, opened 2017-06-09
Feature/allow proxy config

### LINKED ISSUE #1134: Bypass of proxy for syncing to specific instances
this is more of a feature request / partial patch. 

Sometimes MISP architectures connect via internal networks and Internet (w/ proxy) to different instances.

I implemented a dirty hack since I was in a pickle and I didn't want to touch the database model, and do not connect to instances with self signed certificates. Actually, I don't know how to touch the model as I am not proficient in PHP let alone Cake.

Instead of a "bypass_proxy" flag, I am using the "self_signed" flag from the database. Would it be possible to add such a field to the db?

For my hack, I added the following in **app/Lib/Tools/SyncTool.php** (copying below to showcase a possible solution path, for your consideration):

```
<?php

class SyncTool {
        // take a server as parameter and return a HttpSocket object using the ssl options defined in the server settings
        public function setupHttpSocket($server = null) {
                $params = array();
                App::uses('HttpSocket', 'Network/Http');
                if(!empty($server)) {
                        if ($server['Server']['cert_file']) $params['ssl_cafile'] = APP . "files" . DS . "certs" . DS . $server['Server']['id'] . '.pem';
                        if ($server['Server']['self_signed']) $params['ssl_allow_self_signed'] = $server['Server']['self_signed'];
                }
                $HttpSocket = new HttpSocket($params);

                // Dirty, dirty hack
                if (!$server['Server']['self_signed']) {
                        $proxy = Configure::read('Proxy');
                        if ($proxy) $HttpSocket->configProxy($proxy['host'], $proxy['port'], $proxy['method'], $proxy['user'], $proxy['password']);
                } // Dirty, dirty hack ends.

                return $HttpSocket;
        }

        public function setupHttpSocketFeed($feed = null) {
                App::uses('HttpSocket', 'Network/Http');
                $HttpSocket = new HttpSocket();
                $proxy = Configure::read('Proxy');
                if (isset($proxy['host']) && !empty($proxy['host'])) $HttpSocket->configProxy($proxy['host'], $proxy['port'], $proxy['method'], $proxy['user'], $proxy['password']);
        }
}
```

stinnux:
#### What does it do?

This PR allows disabling the proxy for a specific feed or server. Therefore a new field is introduced in Sync Actions->Feeds and Sync Actions->Servers. This field is used in app/Lib/Tools/SyncTool.php to disable the proxy for a specific feed or server. 

This fixes #1134.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/FeedsController.php
+++ b/app/Controller/FeedsController.php
@@ -209,7 +209,7 @@ class FeedsController extends AppController {
 				$this->request->data['Feed']['settings']['delimiter'] = ',';
 			}
 			$this->request->data['Feed']['settings'] = json_encode($this->request->data['Feed']['settings']);
-			$fields = array('id', 'name', 'provider', 'enabled', 'rules', 'url', 'distribution', 'sharing_group_id', 'tag_id', 'fixed_event', 'event_id', 'publish', 'delta_merge', 'source_format', 'override_ids', 'settings', 'input_source', 'delete_local_file', 'lookup_visible');
+			$fields = array('id', 'name', 'provider', 'enabled', 'rules', 'url', 'distribution', 'sharing_group_id', 'tag_id', 'fixed_event', 'event_id', 'publish', 'delta_merge', 'source_format', 'override_ids', 'settings', 'input_source', 'delete_local_file', 'lookup_visible', 'disable_proxy');
 			$feed = array();
 			foreach ($fields as $field) {
 				if (isset($this->request->data['Feed'][$field])) {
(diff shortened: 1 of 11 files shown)


wzr:
This does actually solve an issue in deployments where MISP needs to connect to both external (proxied) and internal (non-proxied) instances.

Would be a great addition to support enterprisey deployments, provided it doesn't break anything ;)

SHSauler:
Was this solved via another means? We had to manually patch our MISP just now :(

SteveClement:
Closing this for now, seems too old and would need a re-work.

PR #2260, opened 2017-06-13
Fix: validation for port attribute

MattCarothers:
The logical check for a valid port attribute was backwards.  It looked for an integer outside the range of 1-65535 rather than inside.

#### What does it do?

Adding a port attribute to an event resulted in the error "Port numbers have to be positive integers under 65536."  The logic to check for a valid port was backwards.  This update fixes it.

#### Questions

- [Yes ] Are you using it in production?
- [No ] Does it require a change in the API (PyMISP for example)?

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Model/Attribute.php
+++ b/app/Model/Attribute.php
@@ -806,7 +806,7 @@ class Attribute extends AppModel {
 				}
 				break;
 			case 'port':
-				if (!is_numeric($value) || $value > 1 || $value < 65536) {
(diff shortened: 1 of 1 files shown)


MattCarothers:
Also, now that I look at it again, the max should be 65535 and not 65536.

iglocska:
Thanks a lot, indeed, with the fixed condition it should be 65535. Just pushed a fix for it.

PR #2267, opened 2017-06-16
Keep misp2stix Python 2.6 compatible

RichieB2B:
Even though we all want to live in a perfect world, some of us are still running RHEL 6 which ships with 2.6.6. So please, when possible and not too much of a nuisance, keep MISP compatible with this ancient technology a little while longer.


iglocska:
Thanks a lot!

PR #2280, opened 2017-06-22
Fix: Galaxy Cluster JSON encoding broke schema

FloatingGhost:
#### What does it do?

In the `getCluster` method, 

```php
$elements = array();
foreach ($cluster['GalaxyElement'] as $element) {
    if (!isset($elements[$element['key']])) {
        $elements[$element['key']] = array($element['value']);
    } else {
        $elements[$element['key']][] = $element['value'];
   }
}
```

In the case that no items were specified in `$cluster['GalaxyElement']`, this would json_encode as an empty array `[ ]`, the schema specifies that this element should be an object. Also if we *had* elements, it'd json_encode as `{"key":value}`. 
 
#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Model/GalaxyCluster.php
+++ b/app/Model/GalaxyCluster.php
@@ -144,6 +144,13 @@ class GalaxyCluster extends AppModel{
 					$elements[$element['key']][] = $element['value'];
 				}
 			}
+
+            // if $elements is still empty, it will json_encode as [], when
+            // the schema requires it to be {}
+            if (count($elements) === 0) {
+                $elements = json_decode("{}");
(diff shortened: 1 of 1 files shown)


FloatingGhost:
If you want a proof of bug, MispPriv:8188 exports a cluster that causes the JSON export to fail JSON validation

iglocska:
Just had a look and it broke the event view for me for the affected events :<

I've changed the JSON converter tool that we use instead, so it should now play nice with the expected format.

PR #2281, opened 2017-06-24
Change shebang to /usr/bin/env xxx for better portability

kallix:
#### What does it do?

This small patch fixes non-portable shebang:
- #!/bin/bash --> #!/usr/bin/env bash
- #!/usr/bin/python --> #!/usr/bin/env python

Note: 
- This patch does NOT change shebang when it's #!/bin/sh (I consider that portable enough)
- This patch does NOT change shebang in INSTALL/ansible (which is only for Debian/Ubuntu)
- This patch does NOT change shebang in app/Console/cake.php because it uses php -q (and the documented way of using it is through the app/Console/cake shell script anyway)

Tested on : 
- Debian 9
- FreeBSD 11
- The latest ubuntu MISP VM (found at https://www.circl.lu/services/misp-training-materials/#misp-virtual-machine)

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/INSTALL/get-misp-automation.py
+++ b/INSTALL/get-misp-automation.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
 
 '''
 Example file on how to get the exported IDS data from MISP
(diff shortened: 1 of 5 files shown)


iglocska:
Thanks a lot!

PR #2283, opened 2017-06-25
Use posix_getpgid to check whether a pid is running

kallix:
#### What does it do?

To check whether or not a process is running, the current code base uses: 
- `(substr_count(trim(shell_exec('ps -p ' . $pid)), PHP_EOL) > 0 ? true : false)`
- or a `file_exists('/proc/' . $pid)` without testing that `/proc` is mounted

This patch replaces all those with a call to [`posix_getpgid($pid)`](http://php.net/manual/en/function.posix-getpgid.php). 
It fixes the workers status page on systems where `/proc` does not exist or where `/proc` is not mounted. 
Since this patch removes the dependence on `/proc`, it also removes the error message "MISP cannot access your /proc directory[...]"

Tested OK on Debian 9, Ubuntu Training VM and FreeBSD 11.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Lib/Tools/PubSubTool.php
+++ b/app/Lib/Tools/PubSubTool.php
@@ -46,8 +46,7 @@ class PubSubTool {
 		$pid = $pidFile->read(true, 'r');
 		if ($pid === false || $pid === '') return false;
 		if (!is_numeric($pid)) throw new Exception('Internal error (invalid PID file for the MISP zmq script)');
-		$result = trim(shell_exec('ps aux | awk \'{print $2}\' | grep "^' . $pid . '$"'));
-		if (empty($result)) return false;
+		if (!posix_getpgid($pid)) return false;
 		return $pid;
 	}
 
--- a/app/Model/Server.php
+++ b/app/Model/Server.php
@@ -3045,18 +3045,13 @@ class Server extends AppModel {
 				'prio' => array('ok' => true),
 				'scheduler' => array('ok' => true)
 		);
-		$procAccessible = file_exists('/proc');
 		foreach ($workers as $pid => $worker) {
 			$entry = ($worker['type'] == 'regular') ? $worker['queue'] : $worker['type'];
 			$correct_user = ($currentUser === $worker['user']);
 			if (!is_numeric($pid)) throw new MethodNotAllowedException('Non numeric PID found.');
-			if ($procAccessible) {
-				$alive = $correct_user ? (file_exists('/proc/' . addslashes($pid))) : false;
-			} else {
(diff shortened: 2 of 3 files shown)


iglocska:
AFAIK posix_ functions are not included in RHEL/CentOS builds so we couldn't use that command without causing a fatal error on those OSes (or introducing new dependencies), hence why we're stuck with the crappy check for a running process. Anyone with CentOS/RHEL willing to give it a shot?

RichieB2B:
I think you're right: http://www.bigsoft.co.uk/blog/index.php/2014/12/08/posix-php-commands-not-working-under-centos-7

kallix:
Arf...
If I can find enough time, I'll try to install CentOS and RHEL tomorrow....

BTW, app/Vendor/kamisama/php-resque-ex/lib/Resque/Worker.php has one posix_kill (Don't know if that code is used though).

iglocska:
"If I can find enough time, I'll try to install CentOS and RHEL tomorrow...."

There goes your sanity, safe journey!

kallix:
@iglocska I've never claimed to have any sanity at all :-)

As a proof, I managed to obtain a rhel 7.3, install it in a Virtual Machine, write a small php script that does posix_getpgpid, drop this script in /var/www/html and call it from a browser. 
It does work. Without doing anything special (Just checked Web Server during Install).
posix_getpgpid does what it should. PHP (5.4.16) doesn't even complain.

iglocska:
Wow, excellent, thanks a lot for going through the trouble to test it!

adulau:
@kallix  Did you install the POSIX extension for PHP? Just tested on another RedHat without POSIX extension this doesn't work. The "old approach" works on both cases. 

@iglocska I think we should roll-back to be sure that people using RedHat without POSIX extension PHP can still manage/monitor their workers.

kallix:
@adulau I did not use yum at any point (I tried, but my Red Hat subscription is not valid anyway)
All I did is to select the option Web Server during install process. 

I'll check this afternoon if php-posix (or php-process or whatever) was installed. 

Of course, if you test on a server WITHOUT the posix extension, it won't work :-) 
The question is: Is it possible and simple to get the posix ext installed on Red Hat ?

iglocska:
Sadly it's not too easy when it comes to installations with SCL - but even with that in mind, it would be a change in the requirements which we try to avoid unless absolutely unavoidable (the idea is to keep the upgrade of MISP versions to a git pull only). 

The instance we've tested it on is an operational 7.3 instance running rh-php56 via SCL

PR #2292, opened 2017-06-27
fix: cleanup/sync of installation guides

SHSauler:
#### What does it do?

I diffed the installation guides and tried to unify them somewhat.

Only thing I cannot verify: Package `python-setuptools` is in Ubuntu 16.04 install guide. It exists in Debian stretch, so I added it to sync them.

#### Questions

#### Release Type:
- [ ] Major
- [X] Minor
- [X] Patch


iglocska:
Looking very good, thanks a lot!

Issue #2036, opened 2017-03-10
CertAuth Plugin - No possibility to map the Certificate's Subject Alternative Names


obert01:
Hello,

I am trying to authenticate users with client certificates, based on the e-mail address present in their certificate and mapping it to the email property of the User model in MISP. 

The CertAuth plugin seems to only support mapping the client certificate's subject (or DN) properties to user properties in MISP. 

But, in some cases, the email address of the user is only present in the certificate's Subject Alternative Names (SAN). 

I have patched the code in a very dirty way on my side, to do what I am looking for. I have not tried to submit a pull request because I don't know how to do that properly (I am not a developper, and not familiar with the MISP's code). 

Here follows what I have done:

Just after:
```
if (self::$ca && isset($_SERVER['SSL_CLIENT_S_DN'])) {
   self::$client = self::parse($_SERVER['SSL_CLIENT_S_DN'], Configure::read('CertAuth.map'));
}
```
I have added:
```
$map = Configure::read('CertAuth.map');
    if (isset($map["emailAddress"]) && !empty($_SERVER["SSL_CLIENT_SAN_Email_0"]) && empty(self::$client["email"]))
       self::$client["email"] = $_SERVER["SSL_CLIENT_SAN_Email_0"];
```

ANy idea for fixing this properly is welcome.

Issue #1768, opened 2016-12-16
CertAuth plugin does not update last login


RichieB2B:
When users are authenticated using the CertAuth plugin MISP does not update the last_login of the user.

PR #2293, opened 2017-06-27
Fixed empty user creation and user updates when org changes

capile:
#### What does it do?

- CertAuth plugin was creating some empty records when a new Organisation was created (specifically missing the User id in line 190 below);

- It was also creating an empty user if it's to synchronize to an API and it's not found in it (case when certificates were revoked but still worked in MISP): enabled the mapping of other parameters, like the serial number, subject Alt Email etc (also resolves issue #2036):
  * On issue #2036: add a map to SSL_CLIENT_SAN_Email_0 pointing to email in config.php, like this:
  ```
  'CertAuth' => array (
    'map' => array (
        // ...
      'SSL_CLIENT_SAN_Email_0'=>'email',
    ),
  ),
  ```        

- Also resolved bug #1768, the $write was missing to update user properties in line 166;

- There is an update to the behavior of this plugin: when a user changes its Organization, the default values are applied to the user (lines 148-155). This is to prevent a user from gaining an unauthorized role in the new Org.

#### Questions

- [ ] Does it require a DB change? _no_
- [ ] Are you using it in production? _yes_
- [ ] Does it require a change in the API (PyMISP for example)? _no_

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a million for the quick fix!

iglocska:
I still seem to have the issue sadly. I am still locked out and empty lines are being generated with my account.

PR #2294, opened 2017-06-28
Show the welcome_text in tab title

garanews:
#### What does it do?

Show MISP.welcome_text_top value also in the tab title.
Useful when managing many MISP instances.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


lucamemini:
Thx @garanews  for the submit ;-)

iglocska:
Thanks a lot.

iglocska:
Ah just realised it's using the welcome text. The problem with that is that it's a field that has been used for a while and quite a few instances I've seen have quite long texts stored in there. I might move this to its own new field soon that is meant to be a short string.

PR #2306, opened 2017-06-29
Fixing crash on Event Tag delete+refresh on recent MySQL version

edhoedt:
The bug can be reproduced on MISP VM v2.4.76
After deleting a tag from _/events/view/[id]_ , the AJAX query from loadEventTags to _/tags/showEventTag/[id]_ produced an error 500


iglocska:
Thanks a lot!

PR #2307, opened 2017-06-29
Attribute tags: fixing automatic refresh after deleting/adding a tag

edhoedt:
Error could be reproduced on MISP VM v2.4.76.
Adding/deleting a tag to an Attribute previously required manual refreshing


iglocska:
Nice catch, thanks!

PR #2314, opened 2017-07-03
Allow Redis to be password-protected

kallix:
#### What does it do?

This patch allows MISP to work with a password-protected Redis server. 
Every time a redis connection is made, it checks whether `MISP.redis_password` is set in `config.php` and if so, issues a `$redis->auth($redis_pwd)`. 
This fixes the last components of MISP that don't already support a password-protected Redis. 

If `MISP.redis_password` is not set in `config.php`, this patch should have no impact.

I'm unsure about how to handle this setting though: Should an "official" `redis_password` option be added to `app/Model/Server.php` ? 
I think changing/setting the redis password from the gui is a terrible idea. If you use only one redis server like I expect 99% of people do, the password should also be set in many other places (php.ini, CakeResque config, Plugin.ZeroMQ_redis_password, app/config/bootstrap.php)
Any thought on that ?

A word of caution: Given how poorly some libraries are handling the redis password (CakeResque, I'm looking at you!), several "special" characters will break things if used in the redis password. In particular, ' (single quote), " (double quote) and $ will create problems. (The list is probably much longer.) 
CakeResque-ex will pass the Redis password to workers through env variables on the command line (!), without proper escaping (!!) and in such a way that bash will evaluate the content (!!!). 
One other side effect of that: the redis password is visible with `ps`

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/ServersController.php
+++ b/app/Controller/ServersController.php
@@ -1013,7 +1013,11 @@ class ServersController extends AppController {
 			if ($found['type'] == 'numeric') {
 				$this->request->data['Server']['value'] = intval($this->request->data['Server']['value']);
 			}
-			$testResult = $this->Server->{$found['test']}($this->request->data['Server']['value']);
+			if  (!empty($leafValue['test'])) {
+				$testResult = $this->Server->{$found['test']}($this->request->data['Server']['value']);
+			} else  {
+				$testResult = true;  # No test defined for this setting: cannot fail
+			}
 			if (!$forceSave && $testResult !== true) {
 				if ($testResult === false) $errorMessage = $found['errorMessage'];
 				else $errorMessage = $testResult;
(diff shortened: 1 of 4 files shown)


adulau:
It's a good idea. Could you also add it in the MISP settings? Thanks a lot.

kallix:
The way this option is added with this commit, it will show red if unset, and I don't know of any way to set it to an empty string with the gui to make it stop complain. 
I thought about setting a different 'level' value, but then the `redis_password` option would end up far away from the other redis_ parameters. 

BTW, I tried to change the `redis_password` from the gui (to an invalid password), and to my surprise, nothing obviously catastrophic happened.

iglocska:
Did you try 

```
'null' => true
```
? 

Should do the trick.

kallix:
Just tested the following:
- `'null' => true`
- `'null' => false`
- nothing

I didn't see any difference: I always get the unhappy red if the value is empty.

iglocska:
After a quick test, it looks good, thanks a lot!

PR #2327, opened 2017-07-07
Add an optional setting attachments_dir, and adapt existing code to use this setting

kallix:
#### What does it do?

Currently, attachments are stored in the hard-coded directory `app/files/`. This PR adds an optional setting `attachments_dir`. (The discussion started in #2193)
If this setting is not set by user, it defaults to `app/file` (in other words: It should not break existing MISP installations). 
If this setting is set in `config.php`, every attachment (including attachment proposals + GFI files) will be read from/written to the location defined in the setting.

This setting will appear in the Server Settings -> MISP settings page, with a comment indicating that: 
1. This setting should only be changed when MISP is not running. 
2. MISP will not migrate the data by itself. User has to do it.
3. MISP will not even create the directory. User has to do it.

#### Testing
The current path (`app/files`) was hard-coded in 15 different functions through the codebase. 
I managed to test this PR on 12 of those modified functions (Thanks @iglocska for the help!).
I have not been able to properly exercise the 3 remaining functions, though I'm reasonably confident the changes should work OK (in those 3 cases, The exact same modification was tested OK elsewhere in the same class).

controller `EventsController` `_addGfiZip` TESTED OK (doesn't actually need a GFI file to test the modified part)
controller `EventsController` `_addIOCFile` TESTED OK
controller `EventsController` `_readGfiXML` **NOT TESTED**

controller `AttributesController` `__downloadAttachment` TESTED OK
controller `AttributesController` `checkAttachments` TESTED OK

model `Attribute` `base64EncodeAttachment` TESTED OK
model `Attribute` `saveBase64EncodeAttachment` TESTED OK
model `Attribute` `handleMaliciousBase64` TESTED OK
model `Attribute` `uploadAttachment`  **NOT TESTED**
model `Attribute` `beforeDelete` TESTED OK

model `Event` `beforeDelete` TESTED OK

model `ShadowAttribute` `afterSave` TESTED OK
model `ShadowAttribute` `beforeDelete` **NOT TESTED**
model `ShadowAttribute` `base64EncodeAttachment` TESTED OK
model `ShadowAttribute` `saveBase64EncodeAttachment` TESTED OK

TESTED OK means that I successfully tested both with the `attachments_dir` setting set in `config.php` AND without.

If someone is willing to provide me with (or tell me where I can find) a GFI sandbox file (whatever that is), that would allow me to test 2 untested cases. 
The last untested case is -- it seems -- code that is never called (but that should be)

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/AttributesController.php
+++ b/app/Controller/AttributesController.php
@@ -295,7 +295,7 @@ class AttributesController extends AppController {
 		$attachments_dir = Configure::read('MISP.attachments_dir');
 		if (empty($attachments_dir)) {
 			$this->loadModel('Server');
-			$attachments_dir = $this->Server->serverSettings['MISP']['attachments_dir']['value'];
+			$attachments_dir = $this->Server->getDefaultAttachments_dir();
 		}
 		$path = $attachments_dir . DS . $attribute['event_id'] . DS;
 		$file = $attribute['id'];
@@ -2820,7 +2820,7 @@ class AttributesController extends AppController {
 			$attachments_dir = Configure::read('MISP.attachments_dir');
 			if (empty($attachments_dir)) {
 				$this->loadModel('Server');
-				$attachments_dir = $this->Server->serverSettings['MISP']['attachments_dir']['value'];
+				$attachments_dir = $this->Server->getDefaultAttachments_dir();
 			}
 			foreach ($attributes as $attribute) {
 					$path = $attachments_dir . DS . $attribute['Attribute']['event_id'] . DS;
--- a/app/Controller/AttributesController.php
+++ b/app/Controller/AttributesController.php
@@ -292,7 +292,12 @@ class AttributesController extends AppController {
 	}
 
 	private function __downloadAttachment($attribute) {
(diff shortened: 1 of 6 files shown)


kallix:
For information, the travis test sees an error in app/Model/Server.php line 354, which is the `'value'` line in : 
```php
'attachments_dir' => array(
    'level' => 2,
    'description' => 'Directory where attachments are[...]',
    'value' => APP . 'files',
    'errorMessage' => '',
    'null' => false,
    'test' => 'testForWritableDir',
    'type' => 'string',
),
```
It seems the PHP in travis' test somehow expects a `)`

I am unable to reproduce the issue on any of my MISP instances, and I am not knowledgeable enough about PHP to detect the problem at hand.

Any hint on why php is unhappy would be appreciated.

kallix:
I've run my set of tests on this modified PR: Testing Status is the same as described in the 1st comment of this PR.
In addition, it should now work on older PHP as well, and it makes travis happy.

iglocska:
A happy travis makes us happy too!

Issue #2347, opened 2017-07-21
Add cookie data type


dspruell:
Is _cookie_ a supported attribute type today? Wanted to suggest ability to track HTTP cookie names explicitly could be very useful (for tracking malicious traffic distribution, etc.). Would use `text` or `pattern-in-traffic` today but don't feel that either is quite right.

adulau:
Good idea. We will add it.

truckydev:
Hi team,

It would be better to have the distinction between "cookie-value" and "cookie-name".

I can make the PR.

dspruell:
Can this (if not already) also be added to Payload Delivery? Cookies play
significantly in a lot of traffic distribution as well.

PR #2355, opened 2017-07-25
Proposal for amendment #2347

truckydev:
Proposal to split the cookie value into two parts.
The name, which is the key and its value.

sample from HTTP header : 
GET /sample.html HTTP/1.1
Host: www.sample.org
Cookie: sessionToken=abc123

Will become in MISP : 
cookie-name => sessionToken
cookie-value => abc123

#### What does it do?

https://github.com/MISP/MISP/issues/2347

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


adulau:
@truckydev Thank you for the PR and the idea. I did a cookie object https://github.com/MISP/misp-objects/blob/master/objects/cookie/definition.json (as complex attributes will now be as MISP object, in the next release of MISP). Could you review the MISP cookie object template and tell me if this fine for your use-case? Thanks.

truckydev:
Yes, all good for me :)

truckydev:
thx

Issue #2359, opened 2017-07-26
add "referer" type


truckydev:
It will be cool to add a "referer" type.

referer (or referrer) as an url form but it's not the same information.
referer can be add as rule in some IDS like snort or bro ...

PR #2360, opened 2017-07-26
add referer type

truckydev:
https://github.com/MISP/MISP/issues/2359

#### What does it do?

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


adulau:
Interesting but we could create a generic MISP http header object template where referer or other could fit. I'll look it.

truckydev:
refere is in http-request object
https://github.com/MISP/misp-objects/blob/master/objects/http-request/definition.json#L63

I hope misp-object come soon :)
but in json "misp-attribute": "other"

maybe change by referer after this PR no ?

adulau:
currently the misp attribute should match an existing one. `other` was the most appropriate.

adulau:
I suppose we can close this PR as misp-objects are implemented and a complete HTTP object template exists in https://github.com/MISP/misp-objects/blob/master/objects/http-request/definition.json. @truckydev let me know if there is any issue with the object. Thank you.

PR #2385, opened 2017-08-10
Fixed error: 'Value is not a boolean, make sure that you convert 'tru…

cedricbonhomme:
…e' to true for example.' when enabling/disabling MISP with the command line tool.

#### What does it do?

It fixes an issue.

When enabling/disabling MISP with the command line tool, the _Error Message_ column of the _server settings_ table was displaying "Value is not a boolean, make sure that you convert 'true' to true for example.".
The strange thing is that the command line tool seemed to really change the state of MISP (enabled/disabled). So I do not exactly know why there was this error message.

#### Questions

- [ No] Does it require a DB change?
- [ No] Are you using it in production?
- [ No] Does it require a change in the API (PyMISP for example)?

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Console/Command/LiveShell.php
+++ b/app/Console/Command/LiveShell.php
@@ -1,8 +1,8 @@
 <?php
 /*
- * Reset a password
+ * Enable/disable misp
  *
- * arg0 = baseurl
+ * arg0 = [0|1]
  */
 class LiveShell extends AppShell {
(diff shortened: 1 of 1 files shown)


iglocska:
Nice catch, thanks!

PR #2389, opened 2017-08-10
Expose galaxies lit to api

truckydev:
Because in different instances, id of the galaxies in not the same, we need to check them.

This PR return the list of galaxies

sample with python : 
```python
#!/usr/bin/env python3
from pymisp import PyMISP
import json
import requests
from urllib.parse import urljoin

session = requests.Session()
session.headers.update(
            {'Authorization': 'XXXXxxxXXXxxxXXX',
             'Accept': 'application/json',
             'content-type': 'application/json'})
url = urljoin('http://misp.local/', 'galaxies')
r = session.post(url)
print(json.dumps(r.json(), indent=4))
```

or directly : 
misp.local/galaxies.json


iglocska:
Thanks a lot!

PR #2392, opened 2017-08-11
Make all tag with disable taxonomy to hidden

truckydev:
You can restrict the use of taxonomy to your instance from the list of taxonomies.

but this is not applicable with old tag. This old tag can be selectable by user ether if they are not in an enable taxonomy.

this function make all tags who are not in an enable taxonomy selectable.
It passes them in hidden.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


SteveClement:
I suppose this can be closed.

adulau:
This was solved in different ways a bit later on. Thanks for the pull-request, it gave some ideas for improvements.

Issue #2401, opened 2017-08-16
Users cannot login after upgrading to MISP 2.4.78


RichieB2B:
After upgrading to 2.4.78 (from 2.4.75) users with the old password system cannot log in. Only after resetting their password (and thus upgrading them to Blowfish) they can log in, but then they can not change their password, see #2394 

This issue is that the old 40 bytes hash is fed into the cakephp BlowfishPasswordHasher::check() function which failes because those 40 bytes are not a valid blowfish format.
```
2017-08-16 10:10:01 Warning: Warning (512): Invalid salt: {40xHEX} for blowfish Please visit http://www.php.net/crypt and read the appropriate section for building blowfish salts. in [/var/www/MISP/app/Lib/cakephp/lib/Cake/Utility/Security.php, line 323]
Trace:
ErrorHandler::handleError() - APP/Lib/cakephp/lib/Cake/Error/ErrorHandler.php, line 230
Security::_crypt() - APP/Lib/cakephp/lib/Cake/Utility/Security.php, line 323
Security::hash() - APP/Lib/cakephp/lib/Cake/Utility/Security.php, line 115
BlowfishPasswordHasher::check() - APP/Lib/cakephp/lib/Cake/Controller/Component/Auth/BlowfishPasswordHasher.php, line 45
BaseAuthenticate::_findUser() - APP/Lib/cakephp/lib/Cake/Controller/Component/Auth/BaseAuthenticate.php, line 138
FormAuthenticate::authenticate() - APP/Lib/cakephp/lib/Cake/Controller/Component/Auth/FormAuthenticate.php, line 79
AuthComponent::identify() - APP/Lib/cakephp/lib/Cake/Controller/Component/AuthComponent.php, line 771
AuthComponent::login() - APP/Lib/cakephp/lib/Cake/Controller/Component/AuthComponent.php, line 611
UsersController::login() - APP/Controller/UsersController.php, line 769
ReflectionMethod::invokeArgs() - [internal], line ??
Controller::invokeAction() - APP/Lib/cakephp/lib/Cake/Controller/Controller.php, line 491
Dispatcher::_invoke() - APP/Lib/cakephp/lib/Cake/Routing/Dispatcher.php, line 193
Dispatcher::dispatch() - APP/Lib/cakephp/lib/Cake/Routing/Dispatcher.php, line 167
[main] - APP/webroot/index.php, line 92
```

iglocska:
[...]

PR #2402, opened 2017-08-16
Rebuild _authenticateObjects cache in mixed authentication setups

RichieB2B:
When CertAuth is mixed with normal FormAuthentication the upgrade from Simple to Blowfish did not happen because of the internal _authenticateObjects cache. Calling constructAuthenticate() rebuilds this cache.


iglocska:
Thanks a lot, fantastic find!

RichieB2B:
Closes #2401

PR #2412, opened 2017-08-22
Vagrant dev environment

cedricbonhomme:
#### What does it do?

This simply set better default values for OpenSSL and GPG.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


adulau:
Thank you very much for the fixes.

PR #2415, opened 2017-08-22
Baseurl miss in events filter

CheYenBzh:
Minor change, adding baseurl for events search

#### What does it do?

Fix failed searches when a baseurl is configured.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #2422, opened 2017-08-25
Added a button to fetch all enabled feeds

panzertime:
This PR adds a button to the Feeds page.  When this button is clicked, all enabled feeds will be fetched.

### DIFF
--- a/app/Controller/FeedsController.php
+++ b/app/Controller/FeedsController.php
@@ -304,6 +304,59 @@ class FeedsController extends AppController {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


adulau:
@iglocska IMHO, this PR makes sense as we currently expose in the [API only the fetch&cache of the feeds](https://www.circl.lu/doc/misp/automation/#enable-disable-and-fetching-feeds-via-the-api) and not the import into MISP. We just need to be very clear on the difference between fetch&cache versus fetch&import into MISP.

iglocska:
Sounds good thanks!

PR #2428, opened 2017-08-30
Make vagrant a submodule

cedricbonhomme:
#### What does it do?

Vagrant is now a submodule of MISP. This is less intrusive and reduces code duplication between the repositories.

It is still possible to only clone the repository of misp-vagrant and to deploy a demo environment by setting an environment variable ``export MISP_ENV='demo'`` before launching vagrant. But in this case it is advised to use [misp-packer](https://github.com/MISP/misp-packer) which can generate a clean VirtualBox image from a standard ISO file.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


cedricbonhomme:
thanks ;-)

PR #2474, opened 2017-09-15
Improved the accessibility of the "Scheduled tasks" page for screen readers

obert01:
The "aria-label" of the buttons for each tasks (frequency, time, date) should be set to their value rather than their meaning. 
In fact, the meaning of the value is provided by the header of the column, which is already perfectly read by all screen reader I have tested.


iglocska:
Thanks a lot!

PR #2480, opened 2017-09-15
Return empty STIX when no data

### LINKED ISSUE #2478: MISP returns error when empty STIX update is requested
When retrieving data in STIX format (either XML or JSON) when there is no (new) data to export, MISP will throw an `HTTP 500: Internal Server Error`. One example is:
```
http://misp.local/events/stix/download/false/false/false/false/false/1d
```
when there are no event less than 1 day old.

The error.log shows:
```
2017-09-15 13:58:35 Error: [Exception] No matching events found to export.
Request URL: /events/stix/download/false/false/false/false/false/1d
Stack Trace:
#0 [internal function]: EventsController->stix('download', false, false, false, false, false, 1505390315)
#1 /var/www/MISP/app/Lib/cakephp/lib/Cake/Controller/Controller.php(491): ReflectionMethod->invokeArgs(Object(EventsController), Array)
#2 /var/www/MISP/app/Lib/cakephp/lib/Cake/Routing/Dispatcher.php(193): Controller->invokeAction(Object(CakeRequest))
#3 /var/www/MISP/app/Lib/cakephp/lib/Cake/Routing/Dispatcher.php(167): Dispatcher->_invoke(Object(EventsController), Object(CakeRequest))
#4 /var/www/MISP/app/webroot/index.php(92): Dispatcher->dispatch(Object(CakeRequest), Object(CakeResponse))
#5 {main}
```

The reason is that [in](https://github.com/MISP/MISP/blob/2.4/app/Model/Event.php#L2889) [several](https://github.com/MISP/MISP/blob/2.4/app/Model/Event.php#L2894) [places](https://github.com/MISP/MISP/blob/2.4/app/Model/Event.php#L2976) the stix() function returns with success=0. Setting success=1 fixes the issue and returns HTTP 200 with a 0 bytes body. I'm not sure if that is the correct CakePHP way to handle this though.

RichieB2B:
Return an empty STIX package when no events match an export query.
Fixes #2478

### DIFF
--- a/app/Model/Event.php
+++ b/app/Model/Event.php
@@ -2885,13 +2885,9 @@ class Event extends AppModel {
 		$eventIDs = $this->Attribute->dissectArgs($id);
 		$tagIDs = $this->Attribute->dissectArgs($tags);
 		$idList = $this->getAccessibleEventIds($eventIDs[0], $eventIDs[1], $tagIDs[0], $tagIDs[1]);
-		if (empty($idList)) {
-			return array('success' => 0, 'message' => 'No matching events found to export.');
-		}
-		$event_ids = $this->fetchEventIds($user, $from, $to, $last, true);
-		$event_ids = array_intersect($event_ids, $idList);
-		if (empty($event_ids)) {
-			return array('success' => 0, 'message' => 'No matching events found to export.');
+		if (!empty($idList)) {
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks a bunch!

PR #2483, opened 2017-09-15
Accessibility improvement: ARIA properties for the "Add new cluster" button - events/view

obert01:
Given the "button" role and appropriate aria-label to the "Add new cluster" button in the "galaxy quick preview" on an events/view page.


iglocska:
Thanks a lot for fixing these!

PR #2489, opened 2017-09-19
bugfix for freetextimport and email

truckydev:
Correction for a bug when you add an email in freeTextImport.

When you select 'whois-registrant-email' attribut never created and an error is displayed.

because : 
'whois-registrant-email' not in 'Social network'  and 'Payload delivery' but only in 'Attribution'.

This PR add the type 'whois-registrant-email' in 'Social network'  and 'Payload delivery' category.

#### What does it do?

no issue has been created.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #2493, opened 2017-09-19
Use sanitized orgname in STIX header

RichieB2B:
The STIX header used the plain orgname where the STIX elements used the sanitized version. This makes the STIX document not validate (undefined namespace used).


iglocska:
Thanks a lot.

PR #2495, opened 2017-09-19
new: added alternate nameserver option to rpzexport

arnydo:
Added option to add an alternate nameserver to RPZ export.


iglocska:
Looks good, just a minor thing: Could you put the new parameter for the API to the end of the list? Otherwise we have a change in the order and it might break tools relying on it.

Thanks in advance!

iglocska:
Welp, that was fast, thanks a million!

arnydo:
No problem! Thank you.

PR #2502, opened 2017-09-22
Fix Tag json format

aparriel:

Fix json format from PR #2500

Json structure returned is now similar to structure returned by view function.


iglocska:
Thanks a lot!

PR #2509, opened 2017-09-25
added convertCompoundToObject.py tool

infosec-intern:

#### What does it do?

Converts MISP attributes that are compound-types (domain|ip, ip-src|port, etc.) into corresponding MISP objects. Also serves as an example for building objects via PyMISP

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/View/Elements/Events/View/value_field.ctp
+++ b/app/View/Elements/Events/View/value_field.ctp
@@ -20,9 +20,12 @@
     }
   } else if (strpos($object['type'], '|') !== false) {
     $separator = in_array($object['type'], array('ip-dst|port', 'ip-src|port')) ? ':' : '<br />';
-    $separator_pos = strpos('|', $object['value']);
-    $final_value = h($object['value']);
-    echo substr_replace(h($object['value']), $separator, $separator_pos, strlen($separator));
+    $value_pieces = explode('|', $object['value']);
+    foreach ($value_pieces as $k => $v) {
+      $value_pieces[$k] = h($v);
+    }
+    $object['value'] = implode($separator, $value_pieces);
+    echo ($object['value']);
   } else if ('vulnerability' == $object['type']) {
     $cveUrl = (is_null(Configure::read('MISP.cveurl'))) ? "http://www.google.com/search?q=" : Configure::read('MISP.cveurl');
     echo $this->Html->link($sigDisplay, $cveUrl . $sigDisplay, array('target' => '_blank', 'class' => $linkClass));


iglocska:
Woah, this is super dangerous. If you "convert" an event you basically generate objects with new uuids. Running the same tool on the same event on another instance will naturally generate new uuids and you'll end up with a clash and duplicate uuids.

infosec-intern:
Whew. Glad I don't have direct write access to the repo then. I'm a little unclear on what you mean by "you'll end up with a clash and duplicate uuids" though. If I'm generating new uuids on the same event on different servers, why wouldn't that be different uuids instead of duplicate uuids?

iglocska:
Basically imagine this:

Event1 is distributed across several instances.

Event1 has 1 attribute: domain|ip bla.bla|1.1.1.1

Instance1 upgrades it to an object, having two uuids for the attributes: uuid1, uuid2 the object has uuid3.

Instance2 upgrades it to an object, having two uuids for the attributes: uuid4, uuid5 the object has uuid6.

On the next sync, instance1 pushes to instance2, updating event1 with a new object, uuid3 along with 2 attributes uuid1, uuid2. Event1 now has 2 objects on instance2.

iglocska:
There are some long term solutions that could work if we really wanted to make a conversion, however at this point I don't think it would make too big of a difference.

1. Manually force UUIDs to resolve the issue: object could become (original-attribute:uuid-1), attribute 1 would become original-attribute:uuid, attribute 2 would become (original-attribute:uuid+1)

2. Allow blowing up composites into objects directly in MISP. An API such as /attributes/edit_composite/[id] would work and people wouldn't get the idea that it's an upgrade script but rather what it actually is: Editing of existing data. Same rules would apply as when you mess with someone else's event using privileged accounts.

The pitfalls to consider:

1. Which version of the template would be used to generate any objects? Would a new version of the template means that different users would create different objects altogether from the same attribute?

2. How would we restrict site-admins from running a script like that on data they don't own, inadvertently causing desynced objects which after a synchronisation would result in duplicate objects?

3. How do we deal with malware-samples? We currently only have one way of storing samples, via the malware-samples attribute type which is a composite type.

So let's keep your PR in store for now, the idea in the long term is something that we should probably do, but we definitely need to resolve some underlying challenges first! It's definitely a tricky issue to resolve.

infosec-intern:
Ok cool, thanks for the explanation. I didn't think much about the UUIDs but I'll be sure to in future projects. In any case, it was good practice for making new objects via the API

iglocska:
It might still be something we can do once we make some changes though so let's keep it in the pipe ;)

PR #2517, opened 2017-09-27
user right update

truckydev:
Make all user access to /attributes/describeTypes.json

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Nice catch, thanks!

PR #2529, opened 2017-10-03
Fix: Removed duplicates from $categoryDefinitions

SHSauler:
#### What does it do?
Removed duplicates from `$categoryDefinitions` arrays. I stumbled on this while  converting `$categoryDefinitions` to JSON.

```
Payload delivery/ip-dst|port
Payload delivery/ip-src|port
Support Tool/text
```

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a bunch!

PR #2533, opened 2017-10-04
Add ip-src|port and ip-dst|port attributes to STIX export

RichieB2B:
The _ip-src|port_ and _ip-dst|port_ attributes were missing in STIX exports. 
This patch also correctly sets the _is_source_ and _is_destination_ STIX attributes and leaves those out when unknown (like for _domain|ip_).


iglocska:
Looks good thanks a lot!

PR #2535, opened 2017-10-04
Add an imput for search on all attributes in an event.

truckydev:
## What does it do?
this PR add possibility to filter on attributes in an event.

animation

check you conf file ti add requierement field and add this in the MISP array : 
```php
'Attributes_Values_Filter_In_Event' => 'id, uuid, value, comment, type, category, Tag.name',
```
the option can be change in the administration page.

#### Questions

#### Release Type:

- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot, great idea!

PR #2536, opened 2017-10-04
Add MISP objects to STIX export

RichieB2B:
A very simple way to add MISP objects to the STIX export. It runs each object through the existing Attributes export mechanism and moves the resulting Observables into a new ObservableComposition with an AND condition.


iglocska:
Woaaah, thanks a million. That's awesome!

PR #2539, opened 2017-10-05
Allow creating users with CertAuth via userDefaults

### LINKED ISSUE #2538: Populating new user with CertAuth from userDefaults fields broken
We create MISP users on the fly using the CertAuth plugin. We use userDefaults to populate the fields and not the more popular restApi method. Since the [security fix](https://github.com/MISP/MISP/commit/be111a470204a974c50682054c9c7d4b94396ed9 ) this no longer works because getRestUser() is not allowed to return NULL anymore.

RichieB2B:
Only call getRestUser() when a URL is configured for it. Closes #2538


iglocska:
I just asked about a PR like 3 seconds ago and here it is, you rock!

PR #2563, opened 2017-10-13
Use MISP TLP tags to set STIX tlpMarking

RichieB2B:
Use the MISP event TLP tags and attribute TLP tags for the tlpMarking in STIX exports.

There is a fallback to the old way of determining to tlpMarking from the MISP distribution field when no TLP tags are present.

For MISP objects all attribute TLP tags in the object are considered and the highest wins.


iglocska:
Sweet, thanks a lot!

PR #2582, opened 2017-10-20
Fix: Attribute type list when editing should be the category's one if…

ppanero:
#### What does it do?

[No issue has been open for this]
When editing an attribute even if the category was already set the type list will be the full list. It should only display the corresponding list to the category.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/webroot/js/misp.js
+++ b/app/webroot/js/misp.js
@@ -2741,6 +2741,7 @@ function initPopoverContent(context) {
 			});
 		}
 	}
+	formCategoryChanged('Attribute');
 }
 
 function getFormInfoContent(property, field) {


iglocska:
Thanks a lot!

jurg:
This change makes it impossible for me to add any attributes with a non-default type (ie. selecting anything from the dropdown "Type" doesn't work).

The information shown in the infobox is correct for the type I just clicked, but the default type (ie. "md5" for Category "Artifacts dropped") is still selected, and when submitting I get an error on the value (ie. it expected an MD5).

Edit: I have narrowed it down to this specific commit - tried it with and without the single line change.

ppanero:
@jurg Could you lead me to reproduce the error please. I have tried the following:
- Take an attribute (ip-src) and change it to an md5 (so change both categories and type) and put the md5 value -->It works.
- Add an attribute with "Other" as a category and a few types (added 3 attributes) --> Works

Could you give me an exact example so I can debug it? Thanks!

EDIT: Sorry, you mean changing the type without changing the category right? I checked and that doesnt work. WIP to fix

jurg:
1) Click to add an attribute to an existing event
2) Choose any category (ie. "External analysis")
3) Try to select any Type (default is "md5", I chose "comment")

The selected value in the dropbox never changes from "md5", and no matter what I select as type the form tries to validate the value as an md5.

If I remove the change to `app/webroot/js/misp.js` everything works fine.

ppanero:
@jurg the mentioned PR above should fix it. Sorry for the inconvenience

jurg:
@ppanero Didn't fix it for me, but after spending a few minutes on testing the PR above fixed my issues.

PR #2585, opened 2017-10-20
Beautify edit object validation

ppanero:
#### What does it do?

Beautify the object edition validation. Before the attributes where in a list, now they all fit in a table that is more user friendly and human readable. Example:
screen shot 2017-10-20 at 15 12 06

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a million ;)

PR #2587, opened 2017-10-20
Add Reporter and Producer fields to STIX

RichieB2B:
Add Reporter to STIX Indicent
Add Producer to STIX Indicator


iglocska:
Thanks for that##comma## cool fixes today ;)

PR #2588, opened 2017-10-20
bugfix for listing types when editing non object attrs

ppanero:
#### What does it do?

Fix https://github.com/MISP/MISP/pull/2582 in a proper way

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot! ;)

PR #2600, opened 2017-10-27
Custom csv export

cedricbonhomme:
#### What does it do?

Enables the user to select the attributes to be included in the CSV export. It is possible to specify attributes of events and attributes of objects.

##### Get all attributes (legacy)

```bash
GET https://<misp-instance>/events/csv/download/<event-id>
```

##### Select some event attributes

```bash
GET https://<misp-instance>/events/csv/download/<event-id>?attributes=timestamp,type,uuid,value
```
(the order of the attributes wil be honored, by default attributes of the object won't be returned)

##### Select some event attributes with object attributes

```bash
GET https://<misp-instance>/events/csv/download/<event-id>?attributes=timestamp,type,uuid,value&obj_attributes=uuid,name
```

Columns of the CSV file that will be returned:
`timestamp,type,uuid,value,object-uuid,object-name`

The string _object-_ is prepended to the object attributes names.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -2251,12 +2251,32 @@ class EventsController extends AppController {
 				}
 			}
 		}
+		$requested_attributes = ['uuid', 'event_id', 'category', 'type',
+								'value', 'comment', 'to_ids', 'timestamp'];
+		$requested_obj_attributes = ['uuid', 'name', 'meta-category'];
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks a million!

PR #2632, opened 2017-11-11
Create INSTALL.ubuntu1604.with.webmin.txt

PaoloVecchi:
#### What does it do?

Some, maybe a friend, can't be asked to configure and manage all the services on an Ubuntu 16.04 so Webmin could be useful.

Tested with:
MISP 2.4.82
Webmin 1.860

#### Questions

#### Release Type:
- [ ] Major
- [x] Minor
- [ ] Patch


adulau:
Nice! Thank you for the contribution.

PR #2633, opened 2017-11-12
Download terms redirect fix

dawid-czarnecki:
When server setting MISP.terms_download=true and MISP.terms_file exists under MISP/app/files/terms directory user wasn't able to download terms and conditions before accepting it.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #2639, opened 2017-11-14
update args.sleep on typeError

truckydev:
Convert string to int for time.sleep when sub.py use with -t

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


adulau:
Good point. Maybe could you enforce it at argparse level? It will be cleaner? `type=int` Could you update the PR? Then I can merge it. Thanks a lot.

truckydev:
thanks to you.

adulau:
Perfect. Thank you too.

PR #2651, opened 2017-11-17
Added possibility to use always default org for new users

ppanero:
#### What does it do?

SSO shibboleth plugin edited to allow new users to be created with the default organization (specified via config.php) instead of the one that the provider sends. This feature can be disabled via config.php (default false)

#### Questions

#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch


iglocska:
Thanks a lot!

PR #2668, opened 2017-11-22
fix: refresh rows in settings editor

JanSkalny:
Fixed following problems connected to settings editor:
 - unable to open row editor after changing item,
 - old value is displayed after item is changed,
 - fast modifications to configuration might get lost. (if user changed more items within 3 seconds)


iglocska:
Thanks a million, this has been annoying for ages ;)

PR #2669, opened 2017-11-22
fix: settings editor not working on touch devices

milankowww:
#### What does it do?

Fix: double click in settings editor was not working on touch devices.  This patch sends synthetic double-click event on long touch. 

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Neat, thanks a lot!

PR #2670, opened 2017-11-22
fix: self signed cert verification

milankowww:
#### What does it do?

This changes the behavior of certificate verification for remote server connections, if (and only if) "Self Signed" check box is checked:
 * server name will not be verified (because very often DNS records are missing and IP addresses, or different server names are used in URLs)
* if sever certificate is not uploaded to MISP, verification would normally fail, so it is disabled.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Makes sense, thanks!

PR #2671, opened 2017-11-22
change behavior of login page to return to original page after authen…

milankowww:
#### What does it do?

fix: change behavior of login page to return to original page after authentication

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
I totally didn't see what was going on in here *cough* ;)

Issue #2678, opened 2017-11-23
internal error in /users/view/me with API


truckydev:
When I try somthing like : 
```python
r = session.get('http://misp.local/users/view/me')
print(r)
```

response is : 
```python
{"name":"An Internal Error Has Occurred.","message":"An Internal Error Has Occurred.","url":"\/users\/view\/me"}
```

### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | Debian
| OS version (client)       | Debian
| PHP version               | 7.0,
| MISP version / git hash   | last

### Expected behavior
return json with user information 

### Steps to reproduce the behavior
curl -XGET http://misp.local/users/view/me -H "Accept: application/json" -H "content-type: application/json" -H "Authorization: XXXxxxXXX"

### Logs, screenshots, configuration dump, ...
**error.log : **
```bash
Error: [MissingViewException] View file "Users/json/view.ctp" is missing.
Exception Attributes: array (
  'file' => 'Users/json/view.ctp',
)
Request URL: /users/view/me
```

truckydev:
https://github.com/MISP/MISP/commit/f9de35ddf656bf62d43794602552d8494c186844

PR #2679, opened 2017-11-23
proposal for json on user query

truckydev:
refer to this issue : https://github.com/MISP/MISP/issues/2678


iglocska:
Just copy pasted the same way of showing it as we do with the admin view, it's a bit cleaner ;)

truckydev:
ok ^^

PR #2684, opened 2017-11-24
fix: remove trailing slash from MISP.baseurl

JanSkalny:
#### What does it do?

Automatic removal of trailing slash from MISP.baseurl setting is stored in configuration as well.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #2701, opened 2017-12-01
Fix STIX export format

RichieB2B:
This PR restores the STIX export format to the same content it had before the internal MISP JSON changes:
- Bring back the attribute level TLP markings
- Use threat level name instead of id


iglocska:
<3

PR #2719, opened 2017-12-07
basic support for misp-modules via API

cvandeplas:
Please review the code and if possible complete the missing elements before full merge.

#### What does it do?

- mini cleanup of FileAccessTool that's not needed
- basic support for file uploads with misp-modules via API (malware-samples not supported yet)

The file upload should happen with the following http post json
```
{"Event": {"data": data_base64_encoded}}
```

Malware samples generated by the misp-module are not yet supported and result in an error "Composite type, but value not explodable". This is because the ```__pushFreetext()``` function does not have support for this. There are some challenges with the ```saveFreeText()``` function in relation to redundant code if malware-sample support is added in the ```__pushFreetext()``` function.

#### Questions

- [x] Does it require a change in the API (PyMISP for example)?  Yes, adding a new function to provide access to the misp-modules API.

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -3345,7 +3345,12 @@ class EventsController extends AppController {
 			$result['type'] = $result['default_type'];
 			unset($result['default_type']);
 			unset($result['types']);
-			$result['category'] = $this->Event->Attribute->defaultCategories[$result['type']];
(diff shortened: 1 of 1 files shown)


iglocska:
Thought I've merged this already, thanks a lot!

Issue #2749, opened 2017-12-19
Allow disabling PHP LDAP referrals


ghost:
In a scenario where the LDAP queries are done against a Microsoft AD server, the referrals may not be chased correctly in the PHP LDAP.

Situation refers to this: https://support.microsoft.com/en-ph/help/2561166/ldap-paged-queries-with-subordinate-referrals-are-not-chased-properly

Suggestion: Allow enabling and disabling LDAP referrals in the MISP ```config.php```. For backwards compatibility, keep the default setting as ```true```. Parameter to ```LDAP_OPT_REFERRALS``` can be either a boolean or an integer based on PHP LDAP documentation.

PR #2750, opened 2017-12-19
new: Allow configuring whether to chase LDAP referrals

ghost:
#### What does it do?

This pull request addresses #2749 by allowing enabling/disabling PHP LDAP referrals via ```config.php``` file. It will also keep the default as ```true``` to keep the original behavior prior to the latest version of MISP to keep backwards compatibility and avoid requirements for users to update all configuration files.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #2751, opened 2017-12-19
Creation of install documentation for Red Hat Enterprise Linux (RHEL) 7.x

jezkerwin:
#### What does it do?

Provides installation instructions of MISP on Red Hat Enterprise Linux 7.x

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot, should make a lot of users happy!

PR #2782, opened 2018-01-04
i18n prep - small commits…

SteveClement:

 app/View/Attributes/add_attachment.ctp
 app/View/Attributes/search.ctp

Added __(''); where needed.


iglocska:
Thanks a lot!

Issue #2804, opened 2018-01-11
cache_feeds job fails - "Field 'job_input' doesn't have a default value."


MattCarothers:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | Ubuntu 16.04
| OS version (client)       | Ubuntu
| PHP version               | 7.0.22
| MISP version / git hash   | 2.4.85
| Browser                   | If applicable

### Expected behavior

cache_feeds job should cache feeds

### Actual behavior

The worker reports an error:

```
{
  "type": "fail",
  "log": "SQLSTATE[HY000]: General error: 1364 Field 'job_input' doesn't have a default value",
  "job_id": "7545e3968391e065afa89f941c2ab52d",
  "time": 63,
  "worker": "df8499f58f05:9998"
}
```
### Steps to reproduce the behavior

Enable feed caching.  This MISP server was deployed prior to feed caching being a feature, so I suspect something was missed in an upgrade.  I did try clearing the model cache and restarting the workers.

### Logs, screenshots, configuration dump, ...

```
[2018-01-11 14:35:05] main.ERROR: {"queue":"default","id":"7545e3968391e065afa89f941c2ab52d","class":"ServerShell","args":[{"0":"enqueueFeedCache","1":1515681300,"2":"4","3":"12","s_time":1515677705}]} failed: SQLSTATE[HY000]: General error: 1364 Field 'job_input' doesn't have a default value {"type":"fail","log":"SQLSTATE[HY000]: General error: 1364 Field 'job_input' doesn't have a default value","job_id":"7545e3968391e065afa89f941c2ab52d","time":63,"worker":"df8499f58f05:9998"} []
```

StefanKelm:
this may be related to #2503 ?

PR #2831, opened 2018-01-16
Set job_input explicitly to an empty string for cache feed jobs

MattCarothers:
#### What does it do?

Older MISP deployments may interpret a missing field as a null value instead
of an empty string, which causes the NOT NULL restriction on the jobs.job_input
field to raise an error.  Fixes issue #2804.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #2899, opened 2018-02-01
Wipe objects & update lists after wipe

RichieB2B:
-  wipe tables related to MISP objects
-  update taxonomies, warninglists, galaxies and objectTemplates after wipe


iglocska:
Nice catch, thanks a lot!

Issue #2880, opened 2018-01-25
[Feature request] Make search bar available in the "Select Tag Source" modal


Res260:
When adding an event I like to search for tags by typing instead of scrolling. Right now, I need to click on "All tags" before typing:
![image](https://user-images.githubusercontent.com/14599855/35411828-3a9edba4-01e8-11e8-8dc2-6c506be10bb3.png)
So I want this search bar:
![image](https://user-images.githubusercontent.com/14599855/35411945-a5d57766-01e8-11e8-98b6-96da6ebebc18.png)
to be available in the first modal (first image).

I probably can do that in my free time, I just want to make sure nobody else is working on that!

PR #2902, opened 2018-02-02
Make search bar available in the "Select Tag Source" modal

Res260:
#### What does it do?

peek 2018-02-09 14-50

See issue: #2880

#### Questions

- [No. ] Does it require a DB change?
- [No, tested only locally] Are you using it in production?
- [No.] Does it require a change in the API (PyMISP for example)?

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

Note: I failed to gain focus on the input field using `$(document).ready()`, hence the `new Promise` in `taxonomy_choice.ctp`. Any better solution is welcome!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /vendors
 /app/Vendor
+/app/vendor
 /app/composer*
 !/app/composer.phar
 !/app/composer.license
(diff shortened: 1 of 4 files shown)


iglocska:
Awesome stuff, thanks a lot!

PR #2922, opened 2018-02-13
ModulesQuery controller

juancmontes:
#### What does it do?

Allow use the MISP API to can query misp_modules.

#### Questions

- [X] Does it require a change in the API (PyMISP for example)? It's a new feature for API.

#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch


juancmontes:
Change in ACL needed.

PR #2924, opened 2018-02-13
feature:ModulesController

juancmontes:
#### What does it do?
Allow use the MISP API to can query misp_modules.

#### Questions

- [X] Does it require a change in the API (PyMISP for example)? it's a new feature for API

#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch


app/Controller/ModulesQueryController.php
```diff
@@ -0,0 +1,64 @@
...
+                if (isset($temp['meta']['config'])) {
+                    foreach ($temp['meta']['config'] as $conf) $options[$conf] = Configure::read('Plugin.Enrichment_' . $modname . '_' . $conf);
+                }
+                break;
+            }
+        }
+        if (!empty($options)) $data['config'] = $options;
+
+        // Query
+        $result = $this->Module->queryModuleServer('/query', json_encode($data), true);
```
iglocska: This allows users that do not have access to a module to query them and is in general not clean. Since we already have all modules via getEnabledModules(), we should check whether the requested module was found in the first place and if not let the user know about it.

app/Config/routes.php
```diff
@@ -41,6 +41,10 @@
 	Router::connectNamed(
 		array('attributesPage' => array('controller' => 'events', 'action' => 'view'))
 	);
+    
+    // ModulesQuery    
+    Router::connect('/modulesQuery/list', array('controller' => 'modulesQuery', 'action' => 'index'));
```
iglocska: Why not just call the function index? Having a route to index under the list alias seems a bit pointless to me.

app/Config/routes.php
```diff
@@ -41,6 +41,10 @@
...
+    Router::connect('/modulesQuery/query', array('controller' => 'modulesQuery', 'action' => 'query'));
```
iglocska: This one doesn't actually do anything, by default, controller actions are exposed via /[Controller]/[action] - this creates a route for an already existing endpoint.

app/Controller/Component/ACLComponent.php
```diff
@@ -191,6 +191,10 @@ class ACLComponent extends Component {
 					'returnDates' => array('*'),
 					'pruneUpdateLogs' => array()
 			),
+            'modulesQuery' => array(
+                'index' => array(),
```
iglocska: These will restrict the new functionalities to site admins only. Why?

app/Controller/ModulesQueryController.php
```diff
@@ -0,0 +1,64 @@
+<?php
+App::uses('AppController', 'Controller');
+class ModulesQueryController extends AppController {
```
iglocska: This should be called ModulesController.

app/Controller/ModulesQueryController.php
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        // Set output
+        $this->set('results', json_encode($result));
+        $this->layout = 'ajax';
+    };
+
+	public function index() {
+        // Initialize models
+		$this->loadModel('Module');
```
iglocska: Just make sure that you use ModulesController as the controller and that the default table is modules - then the Module model will always be loaded by default.

app/Controller/ModulesQueryController.php
```diff
@@ -0,0 +1,64 @@
...
+		$modules = $this->Module->getEnabledModules($this->Auth->user());
+        
+        // Query
+        $result = $this->Module->queryModuleServer('/modules', false, true);
+        if ($result) {
+            $result = json_decode('{"error": "Something went wrong, no response from misp_modules."}');
```
iglocska: Why not `$result = array('error' => 'Something went wrong, no response form misp_modules.'));` ?

app/Controller/ModulesQueryController.php
```diff
@@ -0,0 +1,64 @@
...
+        }
+        
+        // Set output
+        $this->set('results', json_encode($result));
+		$this->layout = 'ajax';
```
iglocska: You can pass JSON responses using the RestResponse component - that way you don't have to create any views

iglocska (changes requested):
Thanks a lot, good idea to allow for direct queries to the modules remotely. Could you address the issues in the comments? Let me know if you need any clarification/help with any of them.

app/Controller/ModulesQueryController.php
juancmontes: Yes, you are right. I will fix it.

app/Config/routes.php
juancmontes: This was the first time working with cakephp, so I didnt know if that kind of declarations was mandatory or not. 
I will remove both lines from routes.php and I will rename the list function to index.

app/Controller/Component/ACLComponent.php
juancmontes: My idea was allow the access only to users with API, and I thought that way to do it.

I will fix it.

app/Controller/ModulesQueryController.php
juancmontes: Ok. That makes sense.

app/Controller/ModulesQueryController.php
juancmontes: I did it to avoid errors on json_encode before send the json. I will try it your way.

app/Controller/ModulesQueryController.php
juancmontes: I didnt know it. 

I will use the component and delete the Views.

app/Controller/Component/ACLComponent.php
iglocska: For users with API access just use this:

`array('perm_auth')`

app/Controller/ModulesQueryController.php
iglocska: Perfect, thanks!

app/Controller/Component/ACLComponent.php
juancmontes: Ok, thanks

juancmontes:
Fix Views: Removed all Views.
Fix routes.php: not use the file
Fix name: Name changed to ModulesController
Modify query function: Use the method "getEnabledMethod" to validate if the user can query the module.
Modified index function: Only return enabled modules.

juancmontes (commented):
Fix Views: Removed all Views.
Fix routes.php: not use the file
Fix name: Name changed to ModulesController
Modified query function: Use the method "getEnabledMethod" to validate if the user can query the module.
Modified index function: Only return enabled modules.
Modified returns: Use RestResponse to return json.

iglocska:
Made some final changes and merged. Changes: 

- query function renamed to query enrichment
- added check for disabled modules and for modules that the current user is not allowed to use
- removed the module config from the index function to avoid exposing API keys / credentials to users
- some formating fixes

PR #2934, opened 2018-02-14
fix - allows upload of files using the misp-modules API

cvandeplas:
#### What does it do?

This PR fixes an incomplete implementation of the REST api when talking to MISP modules.
While the original implementation ( #2719 ) allowed files to be uploaded, the misp-module could not create attachment/malware-sample attributes. 

It will allow a tool to populate an existing event with data from a misp-module. Warninglists will be applied in the 'soft' mode: attributes are added, but the `to_ids` flag is set to `0`

Please see #2719 for more details on how to talk to the REST API.

#### Questions

- [x] Does it require a change in the API (PyMISP for example)? ==> optionally yes, to allow PyMISP to use this functionality

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -3486,52 +3486,157 @@ class EventsController extends AppController {
 		}
 	}
 
-	public function __pushFreetext($resultArray, $eventId, $distribution = false, $sg = false, $adhereToWarninglists = false) {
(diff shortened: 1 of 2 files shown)


iglocska:
Completely missed this PR at first, thanks a lot! :)

PR #2944, opened 2018-02-15
Add the value in the field when filled in.

truckydev:
add the value in the field when filled in on event view.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thank you!

PR #2945, opened 2018-02-15
don't exlude attributes with non-exportable tag

truckydev:
exclude filter on attributes when tag is non-exportable

This fixe a probleme when you clic on filters header table.

When attributes have non-exportable table, he is not display when filters apply.

#### What does it do?

I d'ont think there is an issue on this ...

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot

PR #2950, opened 2018-02-17
Update start.sh

eCrimeLabs:
Fixed bug in scheduler line

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Weird, for me both ways of launching the schedule worker seem to work fine - but if it solves an issue for you, why not ;)

Thanks!

eCrimeLabs:
Also thank you ;) actually I experienced issues with the scheduler on RHEL 7.4 so this was why I noticed it

PR #2962, opened 2018-02-22
Small keyboard shortcuts changes

Res260:
1: The `a` shortcut in `event_view` now triggers the popup instead of changing page
2: The bottom right triangle has `cursor: pointer;`

(minor stuff)


iglocska:
Thanks a lot!

PR #2979, opened 2018-02-27
Added install step to make sure submodule permissions are ignored

SteveClement:
Added install step to make sure all the submodules ignore permissions


adulau:
Thank you!

Issue #2971, opened 2018-02-26
[Bug] Cant use freetext import to import filename|ssdeep attributes


Res260:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | Ubuntu
| OS version (client)       | 16
| PHP version               | Unknown
| MISP version / git hash   | 2.4.82, But tested on my local instance 2.4.88 as well
| Browser                   | Chrome

### Expected behavior

`filename|ssdeep` Should get imported properly using freetext import.

### Actual behavior

Everything else get saved, but not the `filename|ssdeep` attributes. This message is shown:
"`X` attributes created. `Y` attributes could not be saved. This may be due to attributes with similar values already existing."

### Steps to reproduce the behavior

1. Create an event
2. Open freetext import
3. Paste an IOC of format `filename|ssdeep`
4. Submit
5. See the error

### Logs, screenshots, configuration dump, ...

![image](https://user-images.githubusercontent.com/14599855/36697119-1f1ebef4-1b14-11e8-8c4e-27752e216959.png)

![image](https://user-images.githubusercontent.com/14599855/36697076-00b0b9cc-1b14-11e8-9cfb-4a1501e4da2a.png)

EDIT: In case it's not clear: when adding attributes normally, it works. FreeText Import seems to be the problem.

Res260:
So I've almost found the problem:
![image](https://user-images.githubusercontent.com/14599855/36818569-9bb9e466-1cb3-11e8-9743-49afcfdb1bd4.png)

(its for personal reference, I'll do a pull request once I nail it)

iglocska:
Awesome, thanks for finding and fixing it!

PR #2985, opened 2018-02-28
Fixed a bug regarding filename|ssdeep attributes importing using FreeTextImport

Res260:
See Issue #2971

#### What does it do?

Fixes the FreeTextImport `filename|ssdeep` import which was broken.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #2990, opened 2018-03-01
Move keyboard shortcuts from misp.js to its own file (to regain compatibility with IE11)

Res260:
#### What does it do?

Moves the keyboard shortcuts code from `misp.js` to its own file.

**Q: Why?**
**A:** So using ECMAScript6 for keyboard shortcuts broke the whole `misp.js` script which was not parsable by IE11 (it doesnt support ECMAScript6). Since a whole team at my office couldn't use MISP at all (since the front-end relies a lot on `misp.js`), I moved the code to its own file.

**Q: What does that change when using MISP?**
**A:** If using anything else than IE, nothing. If using IE, the app is going to work again, beside the keyboard shortcuts part.

Sorry for that, completely forgot IE (fuck IE).

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/View/Layouts/default.ctp
+++ b/app/View/Layouts/default.ctp
@@ -89,6 +89,7 @@
 	echo $this->Html->script('bootstrap-colorpicker');
 	if ($me) {
 		echo $this->Html->script('misp.js?' . $queryVersion);
+		echo $this->Html->script('keyboard-shortcuts.js?' . $queryVersion);
 	}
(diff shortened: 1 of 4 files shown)


iglocska:
Thanks a lot!

iglocska:
Also, agreed: Fuck IE.

PR #2997, opened 2018-03-02
Validate suricata rules

0xmilkmix:
#### What does it do?

This class allows to validate Suricata rules format. HTTP and DNS rules are also checked for keywords order (sticky vs modifiers).

Not called in the code anywhere. We only discussed the idea during MISP Hackathon in Zurich.

Known limitations:
- rules using list of src/dst (ex: [a.b.c.d, w.x.y.z]) shouldn't contain whitespaces
- this could be corrected using custom decoder instead of regexp. This is something from the devel branch of jasonish project.

#### Questions

#### Release Type:
- [ ] Major
- [X] Minor
- [] Patch


iglocska:
That looks great, we'll have to tie this into the system for sure. Thanks a million!

Issue #3001, opened 2018-03-05
[Feature request] Choose tag using keyboard


Res260:
When choosing tags, it would be cool to be able to choose a tag using the `up`/`down` arrow and `enter`.

I'll work on it (and this time f\*cking test it on IE *before* doing the pull request)

iglocska:
Hahaha, sounds good ;)

PR #3012, opened 2018-03-08
Add keyboard navigation when choosing tags for an event

Res260:
#### What does it do?

Adds tag navigation with arrow keys (`Up`/`Down`) and `PageUp`/`PageDown` and tag selection with `Enter`:

peek 2018-03-08 13-25

See #3001 

**Q: Does this work on every selection modal?**
**A:** No. This only works for tag selection. Because of the way modals are built, it is very hard to do significant changes to all of them at once :frowning_face: 

**Also, minor:** Added a delay before doing the request when searching for tags in the taxonomy choice. This reduces the possibility of losing characters when typing fast:

peek 2018-03-08 13-27

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

<sub><sup>tested on IE11, doesnt work but doesnt break anything</sup></sub>

### DIFF
--- a/app/View/Tags/ajax/select_tag.ctp
+++ b/app/View/Tags/ajax/select_tag.ctp
@@ -18,7 +18,7 @@
 	<div class="popover_choice_main" id ="popover_choice_main">
 		<table style="width:100%;">
 		<?php foreach ($options as $k => &$option): ?>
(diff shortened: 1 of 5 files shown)


Res260:
Man coveralls is drunk today

iglocska:
Awesome stuff, thanks. Also, coveralls, go home, you're drunk.

PR #3017, opened 2018-03-12
Fixes display of <select> (dropdown) in import module, closes #3005

### LINKED ISSUE #3005: Incomplete support for select (dropdowns) for modules?
### Work environment

| Type of issue             | Bug/Question
| OS version (server)       | Debian
| PHP version               | 7.1
| MISP version / git hash   | 7554af8edb683d6ffd2f508239af832af64d2570

### Expected behavior
Import modules that specify an input like so

    userConfig = {
      "Test": {
        "type": "Select",
        "message": "Description",
        "options": ["A", "B", "C"]
      }
    }

should result in creating a dropdown on the page at `/events/importModule/<module>` with the options "A", "B", "C".

### Actual behavior
The dropdown created does not have any options. I'm not sure if support for Select is already built in and I'm using it wrongly, but it appears that the implementation is half done. I'm presenting the way I got Select to work on my instance, but do correct me if its already fully implemented.

The dropdown not having options is because `options` is not passed to `$settings` [when creating the input with CakePHP](https://github.com/MISP/MISP/blob/2.4/app/View/Events/import_module.ctp#L39). This can be fixed by passing `options` if its trying to render a Select through adding the following code block [here](https://github.com/MISP/MISP/blob/2.4/app/View/Events/import_module.ctp#L17).

    if ($settings['type'] == 'select') {
      if (isset($config['options'])) {
        $settings['options'] = $config['options'];
      }
    }

The next issue after that is there is a [reference to a validation function defined for Selects](https://github.com/MISP/MISP/blob/77c6788a82d34bd02f08b4c35e7acc05f8ad847f/app/Model/Module.php#L43) but the actual function is not defined. I added 

    public function validateSelectField($value) { return true; }

to provide an always valid validation function since the Select shoudn't need validation. Removing the reference to validateSelectField throws errors in EventController.

AJohnDoe:
#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

To use, define your module like so:

    def introspection():
      modulesetup = {}
      modulesetup["inputSource"] = []
      modulesetup["userConfig"] = {
        "What day is it today": {
          "type": "Select",
          "message": "Select day of the week",
          "options": ["Monday", "Tuesday", "Wednesday", "Thursday", "Friday"]
        }
      }

The value returned in `handler` is the index of the selected option.

### DIFF
--- a/app/Model/Module.php
+++ b/app/Model/Module.php
@@ -71,6 +71,10 @@ class Module extends AppModel {
 		return 'Value has to be a boolean.';
 	}
 
+	public function validateSelectField($value) {
+		return true;
+	}
+
 
 	public function getModules($type = false, $moduleFamily = 'Enrichment', &$exception = false) {
 		$modules = $this->queryModuleServer('/modules', false, false, $moduleFamily, $exception);
--- a/app/View/Events/import_module.ctp
+++ b/app/View/Events/import_module.ctp
@@ -15,6 +15,11 @@
 					if (isset($configTypes[$config['type']]['field'])) {
 						$settings['type'] = $configTypes[$config['type']]['field'];
 					}
+					if ($settings['type'] == 'select') {
+						if (isset($config['options'])) {
+							$settings['options'] = $config['options'];
+						}
+					}
 					?>
 					<span class="bold">
 						<?php


iglocska:
Thanks a lot!

PR #3057, opened 2018-03-22
Fixed spelling errors for mysql command and php version.

jezkerwin:
#### What does it do?

Fixed a few spelling errors with mysql_secure_installation command
Fixed commands referring to wrong version of PHP
Change git clone command for LIEF installation that was stopping compiling.


adulau:
Thank you! It makes sense.

PR #3064, opened 2018-03-23
Added URLhaus (http://urlhaus.abuse.ch) malware urls as feed.

3c7:
#### What does it do?
Added the urlhaus malware url list as feed.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


adulau:
Good idea! Thank you.

Issue #3075, opened 2018-03-26
Performance issue on attribute filter


whitesheep:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Performance
| OS version (server)       | Gentoo
| OS version (client)       | Debian
| PHP version               | 7.0
| MISP version / git hash   | 2.4.88
| Attributes count | > 1M

### Expected behavior
Response time < 30s

### Actual behavior
Response time > 1m

### Steps to reproduce the behavior
Event Action -> Search Attributes
Search anything

PR #3076, opened 2018-03-26
Drop "LOWER" sql command in "WHERE" clause

whitesheep:
This is a performance fix relative to #3075 .
MySQL "LOWER" doesn't use index.


SteveClement:
@iglocska is this PR to be closed?

PR #3100, opened 2018-03-27
Use GnuPG consistently 

StefanKelm:
#### What does it do?

Uses _GnuPG_ consistently within the GUI

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot! o/

PR #3109, opened 2018-03-28
Configuration for attachments automatically downloaded when loaded in a browser

AJohnDoe:
#### What does it do?

Adds the configuration option "MISP.download_attachments_on_load" to set whether attachments are always downloaded when loaded or to let the browser try and display the file (eg .pdf) then download it if it cannot be displayed (eg .zip)

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/AttributesController.php
+++ b/app/Controller/AttributesController.php
@@ -409,7 +409,7 @@ class AttributesController extends AppController {
 		}
 		$this->autoRender = false;
 		$this->response->type($fileExt);
-		$this->response->file($path . $file, array('download' => true, 'name' => $filename . '.' . $fileExt));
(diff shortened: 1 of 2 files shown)


AJohnDoe:
Is more work on my part needed for this to be merged in?

iglocska:
There is a minor issue with it, but I will fix it, thanks!

iglocska:
OK, just merged it, the only issue was that 'value' was set to true by default for the setting in app/Model/Server.php - however, MISP will not use that field in reality, it is only used to show the default behaviour. Changed it so that it's also actually adhered to if the user did not make any changes to the setting.

PR #3119, opened 2018-03-28
Different category in typeDefinition / defaultCategory

3c7:
Assigned "Network activity" as default category for url in $typeDefinitions as defined in $defaultCategories. There were two different "default categories" used. "External analysis" is skipped by external tools because of normally referencing to reports etc. by external parties.

- Used on testing env.
- Patch

### DIFF
--- a/app/Model/Attribute.php
+++ b/app/Model/Attribute.php
@@ -118,7 +118,7 @@ class Attribute extends AppModel {
 			'email-attachment' => array('desc' => "File name of the email attachment.", 'default_category' => 'Payload delivery', 'to_ids' => 1),
 			'email-body' => array('desc' => 'Email body', 'default_category' => 'Payload delivery', 'to_ids' => 0),
 			'float' => array('desc' => "A floating point value.", 'default_category' => 'Other', 'to_ids' => 0),
-			'url' => array('desc' => 'url', 'default_category' => 'External analysis', 'to_ids' => 1),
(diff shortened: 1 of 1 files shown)


3c7:
Ouh shit. Breaking tests. So `External analysis` is expected to be the default value for category in the tests. That results in Feeds using `External analysis` -> urls in e.g. URLHaus feed not properly blocked by external tools. Lots of OSINT events also use `External analysis->url` for referring to reports instead of using the `link` datatype and because of that all `External analysis` attributes are getting skipped in order not to block websites of analysts and so on.

Views & opinions on this?

iglocska:
Ah ok, completely misunderstood this. Basically $typeDefinitions / $defaultCategories isn't always used when it comes to setting defaults. What you're after is the freetext parser's algorithm - just search for complextypetool (it's in /var/www/MISP/app/Lib/Tools/ComplexTypeTool.php IIRC)

3c7:
I'm using this in production right now to be able to automatially use some feeds for enhancing some internal blocklists. Changing `$typeDefinitions['url']['default_category']` fixed the problem for us. Basically I just wanted to know, why the default for url is `External Analysis` in MISP and maybe start a discussion to change that as this makes not too much sense to me. But maybe I'm not getting the idea behind that. ;)

iglocska:
indeed, the default for url should be Network activity. External analysis is just plain wrong - it should be that for "link", but definitely not url.

3c7:
Okay, so I create a PR for PyMISP to change it there as well?

PR #3134, opened 2018-04-04
fix: Fix wrong object's deletion buttons title

chkp-aliaksandrt:
#### What does it do?

It fixes wrong object deletion buttons title in the `Event` view (`Soft` have to be `Permanently`  and vice versa)

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Nice catch, thanks!

PR #3139, opened 2018-04-06
fix: Event graph stalling when object has unknown-category

mokaddem:
Unknown meta-category in object template do not longer raise an exception (use a default value instead)


iglocska:
Thanks a lot!

PR #3151, opened 2018-04-13
Tools/misp whitelist

TheDr1ver:
#### What does it do?

While using the built-in whitelisting function works great for pretty much every 
export method MISP can handle, ensuring the proper To_IDS values are set is important 
when ingesting indicators directly from the MySQL database (e.g. Splunk DBConnect).

This script checks a designated "whitelist event" for indicators and makes sure 
to_ids is set to True or False for each indicator across all events in the 
MISP database.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Just had a quick look at it and I am a bit confused: 

1. Why do you go through MySQL instead of the API? The problem with this approach is that you're avoiding a lot of the mechanisms that MISP has internally to facilitate synchronisation (timestamps / locking / publishing) as well as ACL (your script will blindly update any event owned by 3rd parties)

2. The use of a special event for whitelisting feels awkward, I'm trying to understand what you're missing in terms of functionalities in order to use one of the standard functionalities that serve this purpose (warninglists/whitelists).

Could you describe your exact use-case? We can hopefully help you achieve it in a cleaner / community friendlier way :)

TheDr1ver:
Basically we're trying to populate Splunk with various lookup tables based on the data we have in MISP, the largest of which would be one table with all observed attributes, and another with all observed events. 

Since our Splunk instance is locked down to prevent us from running arbitrary scripts on the search heads, we decided it was easier (though probably not as efficient) to connect to the database directly and make SQL queries than it would be to get the Splunk admins to write and implement a script that makes multiple REST API queries to essentially dump the database every X hour(s). 

Unfortunately, doing it that way bypasses the whitelist functionality as we can only query attributes we want based on to_ids=True. The alternative is we either make sure everything we want whitelisted is appended to the Splunk queries, or we toggle the to_ids values as needed across the MISP instance.

Since the guys in charge of whitelisting don't necessarily have the proper level of access to make those modifications in Splunk, I figured it would be easier to manage if they simply modified the attributes in a custom "whitelist event" living in MISP.

I agree it feels quite "hacky", and I can see your point for not wanting to modify 3rd party events. I just figured I'd share this script in case anyone else was running into a similar fringe case where they were required to export directly from the database and to_ids was the only mechanism of filtering noise.

iglocska:
Interesting problem indeed. I wonder if we can't just open the warninglists up to a standalone tool that would allow you to run queries against it to filter stuff out on the fly, avoiding having to tamper with the mysql stuff (fetching it internally is fine though). 

Another thing that might be interesting btw is to use ZMQ to feed Splunk, but then you still need to mess with removing data you want to whitelist.

TheDr1ver:
Wow, I can't believe I didn't think about checking the whitelist table from the SQL query. 

We could probably just add a statement to SELECT * from the whitelist table and create one big regex to exclude whitelist-matching indicators from the resulting Splunk tables.

Thanks!

iglocska:
Yeah that would work - the only downside with that we could think of internally here is that normal users can't add whitelist entries, so you need admins to add them. If that's fine for you then it would indeed work. Another alternative is to use warninglists as it's a bit more flexible and you can avoid running regexes (depends on who you want to create entries for exclusion, regexes can be scary ;))

TheDr1ver:
Yeah, the only tricky part there would be getting the SQL query to match on CIDR blocks, but I'm sure it's doable. Thanks again.

iglocska:
You could use this:

https://github.com/MISP/PyMISPWarningLists

Afaik it has algorithms for matching for each of the types :)

SteveClement:
This seems resolved.

Issue #2723, opened 2017-12-08
Reverse sort order in Event List view


keram79:
Currently clicking on the table header e.g. in the Event List view creates an ascending view, i.e. when clicking on "date" the oldest event is on top, when clicking on event ID the event with the lowest ID is on top.

I propose to switch the behaviour, so that by default the newest events are displayed first, because they are usually the most interesting.

SparkyNZL:
I think it would be nice to be able to set a sort order, at the moment when ever im in the event list view i have to click the date field twice (every time) and if i edit an event, i have to click it twice again. Is there any way to set this to sort on date by default ?

It seems super trivial , but DAMN its annoying ! :)

adulau:
It's indeed a very good point. I'm sure our Q&A engineer @rommelfs will agree on this? ;-)

iglocska:
Yeah, it was frustrating as hell.

adulau:
Fixed in https://github.com/MISP/MISP/commit/3d8e8039fa7920a96f65a89de2dd141382e22fe0

SparkyNZL:
Youre awesome ! :)

PR #3152, opened 2018-04-13
Default sort order for id / date reversed on click for Server preview index

StefanKelm:
#### What does it do?

It copies the fix applied in #2723 for the Server preview page.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #3157, opened 2018-04-16
add diff to pubsub

truckydev:
saving changes on when an attributes is edited will make it easier to update with the use of pubsub

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


SteveClement:
@iglocska do we want to fix this PR?

PR #3158, opened 2018-04-16
save what is edited for pubsub module

truckydev:
This PR  come with https://github.com/MISP/MISP/pull/3157 :)
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/AttributesController.php
+++ b/app/Controller/AttributesController.php
@@ -788,6 +788,8 @@ class AttributesController extends AppController {
 				$this->request->data = array('Attribute' => $this->request->data);
 			}
 			$existingAttribute = $this->Attribute->findByUuid($this->Attribute->data['Attribute']['uuid']);
+			// save diff to make life with pubsub easyer ^^
+			$attributDiff = array();
 			// check if the attribute has a timestamp already set (from a previous instance that is trying to edit via synchronisation)
 			// check which attribute is newer
 			if (count($existingAttribute) && !$existingAttribute['Attribute']['deleted']) {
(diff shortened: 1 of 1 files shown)


iglocska:
This may not be the best approach to signaling what has changed for the pubsub channel - keep in mind, if you go through the attribute edit function this will work - if you go through any of the other functionalities that could edit an attribute (such as event edit, sync, feed pull, etc) - it will not. A much better approach would be to do this on a model beforesave level for example.

truckydev:
that's exactly what I wanted to do at first but I didn't know how to get back what was changing during the modification. 
If I make something like that in the model.beforeSave : 
```php
CakeLog::write('debug', print_r($this->data, true));
```

That's why, I added this information in the controller, and next update what go to the pubsub
https://github.com/MISP/MISP/pull/3157/files#diff-cd568c0d060e48659582fb1bc5f06d5bR621

Next,  was not clear about the purpose : It is for this PR : 
https://github.com/MISP/misp-modules/pull/178

If I don't have this information, I can not update the hashstore : 
https://github.com/MISP/misp-modules/pull/178/files#diff-298f77667c2e74d0e60844b5baece509R57

I try another way latter, but if you have any idea what approach I'm taking.

truckydev:
Ok,
Thinking about it for two seconds is actually pretty easy, lol.
I have updated the PR to take into account your remark.
https://github.com/MISP/MISP/pull/3157/files

I think you can close this one ^^

truckydev:
cf https://github.com/MISP/MISP/pull/3157

PR #3170, opened 2018-04-20
Extended event support and tag filtergin in the event graph

mokaddem:
#### What does it do?
- Visual support of extended event (draw a hull outside of nodes belonging to extended event)
- Possibility to filter on tag presence in event graph
- General improvements
#### Questions

#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch


iglocska:
Thanks a lot Monsieur Mokaddem!

PR #3172, opened 2018-04-20
Correlation graph integration

mokaddem:
#### What does it do?

- Add the correlation graph in the event view
- Like the event graph, add a full-screen button


mokaddem:
A bug has been introduced in the event graph.
Closing until fixed.

PR #3183, opened 2018-04-23
Update Log.php

StefanKelm:
#### What does it do?

It alphabetically sorts the list of _Actions_ pull-down menu within "Search Logs"

#### Questions

#### Release Type:
- [ ] Major
- [X] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #3188, opened 2018-04-24
quickfix: editing an attribute was resetting its distribution level

mokaddem:
#### What does it do?

Temporary fix:

Editing an attribute was not setting the distribution level to the previously saved value. It was alway forcing the value to be equals to ``5`` or ``MISP.default_attribute_distribution``.
This issue was occurring because the ``AttributeController/edit`` was using the ``add`` template instead of the ``edit`` (which wasn't setting the saved distribution level).

We may consider to use again the ``edit``  template.

#### Questions

#### Release Type:
- [ ] Major
- [X] Minor
- [X] Patch

### DIFF
--- a/app/View/Attributes/add.ctp
+++ b/app/View/Attributes/add.ctp
@@ -31,12 +31,16 @@
 			?>
 				<div class="input clear"></div>
 			<?php
-
-			echo $this->Form->input('distribution', array(
+			$distArray = array(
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks, nice catch! Guess the reuse of the same template had some nasty unforseen consequences ;)

PR #3205, opened 2018-04-30
cleanup travis and move to requirements.txt

stephengroat:

#### What does it do?

Cleanup travis and move to requirements.txt, easier to install packages for local testing

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,9 @@ addons:
         packages:
             - python3-dev
             - python3-pip
+            - python3-nose
             - libxml2-dev
+            - libzmq3-dev
             - zlib1g-dev
             - apache2
             - curl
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -37,8 +38,6 @@ before_install:
     - git config --global user.name "TravisCI"
 
 install:
-    - sudo apt-get install python3-pip python3-nose
-    - git pull --recurse-submodules && git submodule update
```
adulau, line 41: The PR makes sense to me. Thank you for the clean-up. 

Just one little thing, is there a specific reason why you want to remove the `git submodule update`?

stephengroat: @adulau sorry about that, travis already runs the `git submodule update --init --recursive`, which provides the same support

adulau:
Looks good! Thank you for your contribution.

PR #3212, opened 2018-05-03
Update event-graph.js

StefanKelm:
#### What does it do?

If fixes a super-tiny typo...

#### Questions

#### Release Type:
- [ ] Major
- [X] Minor
- [X] Patch


adulau:
Good catch. Thank you!

PR #3214, opened 2018-05-03
Distribution graph

mokaddem:
#### What does it do?
It displays a summarized view of how information contained in an event is distributed among different actors allowing to:
    - quickly verify that the distribution of all elements matches the intended distribution level
    - quickly verify distribution inconsistencies in the event (attributes being *all communities* with an event being *org only*)

#### New feature
- Add a popover next to the distribution level with:
    - a donut chart showing the repartition of the distribution for the event, the attribute and the object's attribute
    - a progressbar showing a summary of the event distribution
    - a progressbar showing the number of involved sharing group
    - additional information about distribution including all concerned parties (organization and connected server)
- Possibility to filter elements in the event based on their distribution

#### MISP core changes
- **[EventController.php]** ``__valueInFieldAttribute`` can be used with or (``|`` char)
- **[EventController.php]** ``viewEventAttributes`` can overwrite the default ``filterColumns`` in order to bypass ``validFilters``, so that we can filter on other fields (e.g. distribution).
- **[misp.js]** ``filterAttributes`` supports filtering with specified columns

#### New dependencies
- ``Chart.js`` 

#### Release Type:
- [ ] Major
- [X] Minor
- [X] Patch

# Picture
image


mokaddem:
Feedback welcome! :)

adulau:
As already discussed, you implemented the changes and it seems incredibility good. I just need a final go before merging from the Master Chief of the MVC @iglocska to be sure that the approach of getting list of sync servers, orgs and sharing groups is fine. Thank you!

iglocska:
Awesome sauce, thanks a lot!

PR #3215, opened 2018-05-03
Honour `MISP.completely_disable_correlation` on attribute/event save/delete action

g7:
#### What does it do?

Even if `MISP.completely_disable_correlation` is enabled, correlations are still saved to the database.
This prevents that.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Looks good to me, thanks a lot!

PR #3233, opened 2018-05-07
chg: [Controllers] sets the ajax variable globally

mokaddem:
As well as removing useless set in controllers and accessing it instead
of passing through the request.

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #3237, opened 2018-05-08
Update attributeConfirmationForm.ctp

StefanKelm:
Upon soft-delete: match message text with what is being displayed at event view

#### Questions

#### Release Type:
- [ ] Major
- [X] Minor
- [X] Patch


iglocska:
Good point, thanks a lot!

PR #3250, opened 2018-05-15
Add config mapping for 'gpgconf' option in Crypt_GPG library.

WaryWolf:
This commit creates a mapping for an extra variable, 'gpgconf', in the config in config.php to parameters passed to the Crypt_GPG object which is used to perform GPG operations in MISP.

This option not only sets the location of the gpgconf binary, but
if set to false, disables behaviour that shuts down running agents
when a Crypt_GPG object is destroyed. This behaviour would also
kill any long-running or daemonised agents that are running and
configured in the gpg.homedir directory.

We recently experienced issues with our GPG agents dying while browsing to the diagnostics page in admin settings, which this commit fixes.

I have not made any changes to the default config - this allows the existing behaviour to continue by default, but setting the 'gpgconf' variable to 'false' explicitly in the config will enable the new behaviour (no killing of agents). I have left it this way so as to create minimal issues for current users/installs.

#### What does it do?

#### Questions

- [X] Are you using it in production? Yes.

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Model/Server.php
+++ b/app/Model/Server.php
@@ -3187,7 +3187,7 @@ class Server extends AppModel {
 			$continue = true;
 			try {
 				require_once 'Crypt/GPG.php';
-				$gpg = new Crypt_GPG(array('homedir' => Configure::read('GnuPG.homedir'), 'binary' => (Configure::read('GnuPG.binary') ? Configure::read('GnuPG.binary') : '/usr/bin/gpg')));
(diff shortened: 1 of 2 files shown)


WaryWolf:
The config option in Crypt_GPG:

https://github.com/pear/Crypt_GPG/blob/master/Crypt/GPGAbstract.php#L248

The commit that enabled this behaviour in Crypt_GPG:

https://github.com/pear/Crypt_GPG/commit/8d269a4d3e5b65f0f96e2a9ff5b22f3611f70b0c

adulau:
Thank you very much for your contribution!

Issue #3260, opened 2018-05-18
STIX export requires python 3.x


RichieB2B:
Because of d1c23720b4dda56e7d06c94a12a1d568c5066a1a the STIX export now uses pymisp and thus requires python 3.x. Since CentOS 6 comes with python 2.6 this needs some work to use SCL python 3.x instead.

The first indication of this is this message in `app/tmp/logs/error.log`:

```
2018-05-18 14:06:38 Error: [Exception] There was an issue generating the STIX export.
```

and the following in `app/tmp/logs/exec-errors.log`:
```
Traceback (most recent call last):
  File "/srv/www/ro.octomon/html/MISP/app/files/scripts/misp2stix_framing.py", line 152, in <module>
    main(sys.argv)
  File "/srv/www/ro.octomon/html/MISP/app/files/scripts/misp2stix_framing.py", line 137, in main
    stix_header.title="Export from {} MISP".format(orgname)
ValueError: zero length field name in format
```

RichieB2B:
The same goes for Ubuntu 16.04.4 LTS where Python 2.7.12 is the default.
`cat app/tmp/logs/exec-errors.log` gives:
```
WARNING [abstract.py:19 - <module>() ] You're using python 2, it is strongly recommended to use python >=3.5
WARNING [mispevent.py:26 - <module>() ] You're using python 2, it is strongly recommended to use python >=3.5
WARNING [api.py:31 - <module>() ] You're using python 2, it is strongly recommended to use python >=3.5
Traceback (most recent call last):
  File "/var/www/MISP/app/files/scripts/misp2stix.py", line 1211, in <module>
    main(sys.argv)
  File "/var/www/MISP/app/files/scripts/misp2stix.py", line 1206, in main
    stix_builder.generateEventPackage()
  File "/var/www/MISP/app/files/scripts/misp2stix.py", line 169, in generateEventPackage
    incident = self.generate_stix_objects()
  File "/var/www/MISP/app/files/scripts/misp2stix.py", line 216, in generate_stix_objects
    self.resolve_attributes(incident, Tags)
  File "/var/www/MISP/app/files/scripts/misp2stix.py", line 241, in resolve_attributes
    attribute_type, attribute.value)
UnicodeEncodeError: 'ascii' codec can't encode character u'\u201c' in position 54: ordinal not in range(128)
```

PR #3262, opened 2018-05-18
Use python3 interpreter for STIX exports

RichieB2B:
See #3260 : since pymisp is now used, Python 3.5 is a minimum requirement.


RichieB2B:
Event.php is already fixed by https://github.com/MISP/MISP/commit/5dd0753abf2dca4fdd43e170d2ff0863b03f0fea

chrisr3d:
Thanks a lot

PR #3269, opened 2018-05-22
chg: allow symlinks for public keys in footer

Lastpixl:
Allows replacing public GPG & SMIME keys (gpg.asc &
public_certificate.pem) with symbolic links, to store the actual files
in another format. This allows clean separation of MISP code (in
webroot) from configuration data.

Our use case: run MISP on top of kubernetes, storing configurations and
secrets in dedicated volumes, rather than in the Docker image.

#### Questions

- [x] Does it require a DB change? -> **no**
- [x] Are you using it in production? -> **not yet**
- [x] Does it require a change in the API (PyMISP for example)? **-> no**

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


SteveClement (commented):
This seems sane and would include a rather common setup

iglocska:
Thanks a lot!

Lastpixl:
Thanks for having merged this!

PR #3277, opened 2018-05-25
Fix STIX export corner cases

RichieB2B:
We had a lot of problems getting the new STIX exporter to work on real data. With these changes all MISP events in https://www.circl.lu/doc/misp/feed-osint can be exported without errors.


adulau:
Looks very good to me. @chrisr3d all ok for you?

adulau:
Thank you very much for the contribution.

chrisr3d:
Yep it is
Thank you

PR #3279, opened 2018-05-25
Add timestamp to outer STIX_Package

RichieB2B:
Currently there is no place where the time of export is saved. This used to be the outer STIX_Package timestamp. This patch brings it back.


adulau:
Indeed, good point. Thank you!

PR #3280, opened 2018-05-25
Remove leading tab

0xiso:
#### What does it do?

Remove leading tab character which will cause `Headers already sent` error.

The error occurs when output_buffering is set to "0" or no value  in `php.ini`.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #3283, opened 2018-05-28
Very small change to give the user a hint that multiple attachments can be uploaded

SteveClement:
See title.


iglocska:
Thanks a lot!

PR #3310, opened 2018-06-05
Remove contact details, they don't really need to be in there

jezkerwin:
Removed my contact details from the RHEL 7.x install file.
I think it doesn't need to be in there.


iglocska:
Hehe, ok ;-)

PR #3316, opened 2018-06-06
Quoted scl commands to properly execute python3 + cwd for Cake Install

jezkerwin:
Installing Cybox and STIX libraries, the SCL command to install won't properly run unless being quoted.
Added command to change working directory to /var/www/MISP before installing Cake


iglocska:
Thanks a lot!

PR #3323, opened 2018-06-08
Enable python3 for php-fpm for RHEL/CentOS

RichieB2B:
The documentation was missing a crucial step for RHEL/CentOS: enable the use of python3 from within php.


adulau:
Thank you!

PR #3324, opened 2018-06-08
fix: [EventView] Still allows object edition event if the event hasn't

mokaddem:
**fix**: ``EventView`` Still allows object edition even if the event hasn't been published


iglocska:
Good catch, thanks a lot!

Just a small modification, by moving the edit button completely of the if branch it also makes it available for users that don't have editing rights ;) 

Fixed it and made the code a bit less fugly.

PR #3330, opened 2018-06-11
new: Add email field autofocus on login page

dawid-czarnecki:
#### What does it do?
Add email field autofocus on login page

#### Questions

#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch


iglocska:
Thanks, good idea. btw, does IE even support that? ;)))

dawid-czarnecki:
It works on IE 11 :)

PR #3368, opened 2018-06-22
ATT&CK Tactic Matrix at the top!

mokaddem:
- ATT&CK Tactic is put at the top when picking galaxies
- Is also shown in ``All namespace`` mode


adulau:
Yeah!

PR #3372, opened 2018-06-22
Multiple pick in ATT&CK matrix

mokaddem:
Users can pick multiple clusters to be attached and submit their selection.


iglocska:
Thanks a lot B-)

Issue #2010, opened 2017-03-02
Internally exportable tags


RichieB2B:
Introduce an additional flag for tags that allow them to be exportable to internal MISP instances only. So that there will be three export states:

1) exportable: every MISP instance will see them
2) internally exportable: only internal MISP instances see them
3) non-exportable: no other MISP instances see them

iglocska:
Cool idea.

cristianbell:
this sounds like something from the CIICS M.O., :)

truckydev:
I also adhere.

PR #3373, opened 2018-06-22
Improvements tag syncing for "internal instance" server

lucamemini:
hello all,
I implemented two improvements for "tang syncing" when servers are configured as "internal instance".
1) For a tag, with exportable flag disabled, this is exported anyway during syncing.
2) when push an event update to remote server remote tags are removed and resynced from scratch


RichieB2B:
Thanks for the pull request. Always syncing non-exportable tags to internal instances is not a good solutions for everyone. We have a MISP server for internal use only and an external facing MISP instance for external parties to log on to. We do not want out internal tags to show up on the external MISP instance. What we need is ad additional _internally exportable_ flag for tags as described in #2010

lucamemini:
Hey @RichieB2B,
Your idea is more complete... (in my company have 3 misp server but is like was only one ... ) the complexity is related to change to db schema or change "field" logic :-)

@iglocska do you think about it?

Add new column for table  "tags" like this:
`internal_exportable` tinyint(1) NOT NULL,
 
Or incrase value of current "exportable" like this:

0 = not exportable
1 = exportable
2 = internal exportable

iglocska:
Generally we never delete data remotely ever in MISP - so let's not go down that route fully. However, what would be interesting is "soft deleting" tags - this could also apply to non internal instances (so basically having a deleted = 0|1 flag on the EventTag and AttributeTag objects) that is synchronised.

lucamemini:
Ok Andras! 
Next week try to implement this logic for "remote soft deletion", for "internal exportable tag" still waiting your news =)

iglocska:
Thanks you sir! B-)

lucamemini:
@iglocska my patch... :-)

Need to implement auto-upgrade database schema with this new columns:

ALTER TABLE `attribute_tags` ADD `deleted` tinyint(1) NOT NULL DEFAULT 0;
ALTER TABLE `event_tags` ADD `deleted` tinyint(1) NOT NULL DEFAULT 0;

SteveClement:
> @iglocska my patch... :-)
> 
> Need to implement auto-upgrade database schema with this new columns:
> 
> ALTER TABLE `attribute_tags` ADD `deleted` tinyint(1) NOT NULL DEFAULT 0;
> ALTER TABLE `event_tags` ADD `deleted` tinyint(1) NOT NULL DEFAULT 0;

@iglocska any updates on this one?

PR #3384, opened 2018-06-27
Makes more sense.

Rafiot:


adulau:
Indeed, very good point and much clearer.

PR #3399, opened 2018-06-28
Default sort order for timestamp in attribute view

StefanKelm:
#### What does it do?

It copies the fix applied in #2723 for the listing of attributes within an event

#### Questions

#### Release Type:
- [ ] Major
- [X] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #3417, opened 2018-06-30
Added initial internationalization for: French (6%), Japanese (21%)

SteveClement:
Many thanks to @ninoseki for his teams contributions on jpn.
Thanks to @treyka and @cvandeplas

-- Please support our translation teams: https://crowdin.com/project/misp
-- Other Languages in progress: Italian (9%), Korean (1%), Portuguese (1%)


adulau:
@SteveClement should we already merge? IMHO, I think it's better to merge because reaching 99% coverage will be difficult and having a partial coverage might trigger other contributors if they test their language.

PR #3435, opened 2018-07-04
OpenBSD and FreeBSD Install instructions updated

SteveClement:
Still some updates to be done until this can be considered stable.

At this stage basic MISP functionality is given.


iglocska:
Thanks a lot

PR #3439, opened 2018-07-05
chg: Case insensitive sort of organisation list

dawid-czarnecki:
#### What does it do?

In organisation list default view sort organisations case insensitive.

#### Questions

#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch


iglocska:
Thanks a lot!

PR #3440, opened 2018-07-05
fix: [bug] Set correct org id for logo path

dawid-czarnecki:
#### What does it do?
When adding organization, id for logo path was taken from request data instead of last inserted id.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Even better you could do 

`$this->Organisation->id`

PR #3449, opened 2018-07-06
EventGraph history

mokaddem:
#### EventGraph export and history

- EventGraph can now be exported in ``json``, ``png``, ``jpeg``, ``dot language``
- Possibility to save network history per ORG
        - only user can delete their network (or the siteAdmin)
        - every user from the same ORG can load the saved eventGraph
- Fixed some bugs

#### Questions

#### Release Type:
- [X] Major


app/Controller/EventGraphController.php
```diff
@@ -0,0 +1,153 @@
...
+	public $components = array(
+			'Security',
+			'RequestHandler'
+	);
+
+	public function beforeFilter() {
+		parent::beforeFilter();
+	}
+
+	public function get($event_id = false) {
```
iglocska: Could we change this to view? ;)

app/Controller/EventGraphController.php
```diff
@@ -0,0 +1,153 @@
...
+		if (!$this->request->is('get')) {
```
iglocska: Generally there's no need to block POST requests for GET actions, only the other way around. Consider that you might want to inject some parameters at some point via a JSON payload

app/Controller/EventGraphController.php
```diff
@@ -0,0 +1,153 @@
...
+			throw new MethodNotAllowedException(__('Invalid method.'));
+		}
+		if ($event_id === false) throw new MethodNotAllowedException(__('No event ID set.'));
+
+		// retreive current org_id
+		$org_id = $this->_checkOrg();
+
+		// validate event
+		$this->loadModel('Event');
+		if (!is_numeric($event_id)) {
```
iglocska: Perhaps also allow using UUIDs in addition to event IDs. Always consider the API use-case - do I ever want to be able to fetch the graph JSON's via the API?

app/Controller/EventGraphController.php
```diff
@@ -0,0 +1,153 @@
...
+		// fetch eventGraphs
+		$eventGraphs = $this->EventGraph->find('all', array(
+			'order' => 'EventGraph.timestamp DESC',
+			'conditions' => array(
+				'EventGraph.event_id' => $event_id,
+				'EventGraph.org_id' => $org_id
+			),
+			'contain' => array(
+				'User' => array(
+					'conditions' => array(
```
iglocska: No need for this if the correct model linking is set up.

app/Controller/EventGraphController.php
```diff
@@ -0,0 +1,153 @@
...
+				)
+			)
+		));
+		return $this->RestResponse->viewData($eventGraphs, $this->response->type());
+	}
+
+	public function add($event_id = false) {
+		if ($this->request->is('get') && $this->_isRest()) {
+			return $this->RestResponse->describe('EventGraph', 'add', false, $this->response->type());
+		} else if ($this->request->is('get')) { // retreive form
```
iglocska: Perhaps a structure like would be a bit nicer:

```
if ($this->request->is('get')) {
    if ($this->_isRest()) {
        #describe API and return
    }
    # retrieve form
} else {
    #do post action
}

File: app/Controller/EventGraphController.php [hunk 6]
```diff
@@ -0,0 +1,153 @@
...
+					throw new NotFoundException(__('Invalid event'));
+				}
+			}
+
+			$this->set('action', 'add');
+			$this->set('event_id', $event_id);
+			$this->render('ajax/' . $formURL);
+
+		} else {
+			if ($event_id === false) throw new MethodNotAllowedException(__('No event ID set.'));
```
iglocska: if (!empty($event_id))

app/Controller/EventGraphController.php
```diff
@@ -0,0 +1,153 @@
...
+
+			$this->loadModel('Event');
+			$event = $this->Event->fetchEvent($this->Auth->user(), array('eventid' => $event_id));
+			if (empty($event)) throw new NotFoundException('Invalid event');
+
+			$eventGraph = array();
+			if (!$this->_isSiteAdmin() && ($event['Event']['orgc_id'] != $this->_checkOrg() && !$this->userRole['perm_modify'])) {
```
iglocska: $this->Auth->user('org_id')

app/Controller/EventGraphController.php
```diff
@@ -0,0 +1,153 @@
...
+				$eventGraph['EventGraph']['network_name'] = null;
+			} else {
+				$eventGraph['EventGraph']['network_name'] = $this->request->data['EventGraph']['network_name'];
+			}
+
+			if (isset($this->request->data['EventGraph']['preview_img'])) {
+				$eventGraph['EventGraph']['preview_img'] = $this->request->data['EventGraph']['preview_img'];
+			}
+
+			$eventGraph['EventGraph']['timestamp'] = $date->getTimestamp();
```
iglocska: Move this to beforeValidate() in the model

app/Controller/EventGraphController.php
```diff
@@ -0,0 +1,153 @@
...
+
+			// Network pushed will be the owner of the authentication key
+			$eventGraph['EventGraph']['user_id'] = $this->Auth->user('id');
+			$eventGraph['EventGraph']['org_id'] = $this->Auth->user('org_id');
+
```
iglocska, line 99: $this->EventGraph->create();

app/Controller/EventGraphController.php
```diff
@@ -0,0 +1,153 @@
...
+		}
+	}
+
+	public function delete($id) {
+		if (!$this->request->is('post')) {
+			$this->set('id', $id);
+			$this->render('ajax/eventGraph_delete_form');
+		} else {
+			$this->set('id', $id);
+			$conditions = array('id' => $id);
```
iglocska, line 124: Add to the conditions org_id == user's org_id
```
$conditions = array(
   'id' => $id
);
if (!$this->_isSiteAdmin()) {
    $conditions['org_id'] = $this->Auth->user('org_id');
}
```

app/Controller/EventGraphController.php
```diff
@@ -0,0 +1,153 @@
...
+			$eventGraph = $this->EventGraph->find('first', array(
+					'conditions' => $conditions,
+					'recursive' => -1,
+					'fields' => array('id', 'event_id', 'user_id'),
+			));
+			if (empty($eventGraph)) throw new NotFoundException('Invalid EventGraph');
+			if ($this->request->is('ajax')) {
```
iglocska: Not needed - API users should be able to delete them too

app/Model/AppModel.php
```diff
@@ -958,9 +958,25 @@ public function updateDatabase($command) {
...
+					`user_id` int(11) NOT NULL,
+					`org_id` int(11) NOT NULL,
+					`timestamp` int(11) NOT NULL DEFAULT 0,
+					`network_name` varchar(255) CHARACTER SET utf8 COLLATE utf8_unicode_ci,
+					`network_json` MEDIUMTEXT NOT NULL,
+					`preview_img` MEDIUMTEXT,
+					PRIMARY KEY (id),
+					INDEX `event_id` (`event_id`),
+					INDEX `user_id` (`user_id`),
+					INDEX `org_id` (`org_id`)
```
iglocska, line 979:
```
INDEX `timestamp` (`timestamp`)
```

app/Model/EventGraph.php
```diff
@@ -0,0 +1,52 @@
...
+			'foreignKey' => 'user_id',
+			'conditions' => '',
+			'fields' => '',
+			'order' => ''
+		)
+	);
+
+
+	public $validate = array(
+		'is_json' => array(
```
iglocska: Add the field name

app/Model/EventGraph.php
```diff
@@ -0,0 +1,52 @@
...
+		),
+	);
+
+	public function beforeValidate($options = array()) {
+		parent::beforeValidate();
+		return true;
+	}
+
+	public function isValidJson($text) {
+		$check = json_decode($text);
```
iglocska: try catch (maybe, not sure)

app/Controller/EventGraphController.php
mokaddem:
```
/Perhaps//
/would be a bit/is/
```

app/Model/EventGraph.php
mokaddem: Apparently, this is not a true validation... see 
https://stackoverflow.com/questions/7841670/validate-json-in-php/47227564

Do I keep it like this?

mokaddem:
PR updated, feel free to go through once again ;)

iglocska:
Thanks a lot!

Issue #3423, opened 2018-07-02
ATT&CK matrix usage


xme:
I just upgraded my instance to 2.4.93, updated galaxies too.
I can see the matrix but:
- Can't select elements in the matrix
- No "Submit" button displayed
(I compared with the blog's video ;-)
Any idea?

SteveClement:
There is currently a bug being discussed in Gitter, this might relate. @iglocska will let us know once fixed.

iglocska:
That is not related. The one on gitter has to do with incorrect namespacing on galaxies and the attack matrix not having any elements.

iglocska:
@xme do you see anyhting funky in your browsers dev console?

iglocska:
Also, it looks like the submit button moved to the bottom of the screen, just above the cancel button ;)

xme:
Nothing fancy in the web console...
And no button at all (cfr pic)
<img width="1513" alt="screen shot 2018-07-02 at 13 10 03" src="https://user-images.githubusercontent.com/480944/42160959-61ccda90-7df9-11e8-8b7e-144d8685f90b.png">

adulau:
This is the view only. To add an att&ck technique, you need to add a galaxy via the add galaxy ;-)

xme:
OMG! Found it now... Tx! ;)

adulau:
@xme It's good feedback. We might refactor a bit the UI to avoid the confusion ;-)

iglocska:
Yeah, we just got the exact same feedback a few minutes ago on gitter from another experienced community member - definitely something we should address! :)

PR #3455, opened 2018-07-09
Update/attack matrix

mokaddem:
#### Slight UX improvement

When clicking on a cell in the ATT&CK matrix at the event level, 2 icons are proposed:
- Filter the event
- Or attach the cluster to the event

Also, fixes issue, #3423 

#### Release Type:
- [X] Patch


iglocska:
Thanks a lot!

PR #3460, opened 2018-07-09
fix: export events csv with CR (fix #3458)

### LINKED ISSUE #3458: Export events csv with CR
I use the events automation for exporting past event in csv.
When there is a CR in the event name, the field is not sanitize before export and the csv is broken.

kalyparker:
Export using automation functionality for ids does not clean the special char like CRLF.
When there is a carriage return in the event info, the csv is broken.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #3473, opened 2018-07-11
new: restSearch: add parameter value_exact to improve performance.

Lastpixl:
#### What does it do?
In order to take advantage of the index on the value1 & value2 fields:
* Get rid of the complex virtualField, filter on both value1 and value2
  instead.
* Get rid of the leading wildcard '%', which enables mysqld to use the
  index.

Observed improvements: on a typical SHA256 search, search performed
using PyMISP goes from 8s to 0.5s. The actual SELECT queries that
fetches results now executes in ~0.01s.

If you accept this commit, I will add support in PyMISP and open a pull request there as well.

#### Questions

- [x] Are you using it in production? pre-prod
- [x] Does it require a change in the API (PyMISP for example)? Required to take advantage of the new feature only, everything else is unmodified (no breaking change)

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/AttributesController.php
+++ b/app/Controller/AttributesController.php
@@ -2134,7 +2134,7 @@ class AttributesController extends AppController
             if (!isset($data['request'])) {
                 $data['request'] = $data;
             }
-            $paramArray = array('value', 'type', 'category', 'org', 'tags', 'from', 'to', 'last', 'eventid', 'uuid', 'published', 'publish_timestamp', 'timestamp', 'enforceWarninglist', 'to_ids', 'deleted', 'includeEventUuid', 'event_timestamp', 'threat_level_id');
+            $paramArray = array('value', 'value_exact', 'type', 'category', 'org', 'tags', 'from', 'to', 'last', 'eventid', 'uuid', 'published', 'publish_timestamp', 'timestamp', 'enforceWarninglist', 'to_ids', 'deleted', 'includeEventUuid', 'event_timestamp', 'threat_level_id');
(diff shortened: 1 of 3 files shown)


Lastpixl:
I have rebased this pull request against 2.4.
We are running this in pre-production, with great performance gains for our usage patterns.

iglocska:
I am currently refactoring the APIs completely and we've decided to treat
the default behaviour as a bug - the goal is to simply have the exact match
as the default behaviour and startng/ending/encapsulating the search
pattern in '%' will trigger a substring search.

Lastpixl:
That is great, thanks for the answer!
Will you still use the complex "value" virtualField, which result in an SQL `CONCAT`, and so prevents MySQL from using its index, resulting in a slow query?

iglocska:
Nope, it will simply use value1 and value2 separately.

To see the work in progress (still wonky as hell, but getting there):

https://github.com/MISP/MISP/blob/feature/api_rework/app/Model/Event.php#L2045
https://github.com/MISP/MISP/blob/feature/api_rework/app/Model/AppModel.php#L1664

Lastpixl:
Perfect, this PR can be closed then. We'll keep using this code until api_rework is ready.
Let me know if/when you want us to test api_rework!

iglocska:
sure will do! If at any time you want to play with it and find issues (we know of some internally thanks to @Rafiot's new test cases) that would ofc be great. Just don't run it in production yet ;)

PR #3478, opened 2018-07-12
chg: [deps] Set the correct and working version of Cybox in diagnostics

SteveClement:
chg: [doc] More kali tweaks


RichieB2B (commented):
We should also change the install instructions for cybox. Instead of using the git version we should probably use “pip3 install cybox” instead.

PR #3486, opened 2018-07-16
Fix event graph db update

mokaddem:
Fix the missing comma in the eventGraph SQL update query.

#### Release Type:
- [X] Patch


iglocska:
Thanks a lot!

PR #3517, opened 2018-07-31
Use original orgname at stix-header:title

RichieB2B:
Since f4ff165e367833c727aeda453e55eca6757da3ae the stix-header:title uses to mangled orgname. This patch preserves the original orgname as set in MISP.


chrisr3d:
Ah yes missed that one, thank you !

PR #3530, opened 2018-08-03
chg: [bug] Fixed wrong event lookup in case the uuid is passed as an eventId.

chkp-aliaksandrt:
#### What does it do?

It fixes the code that had two mutually exclusive conditions `Event.id = uuid and Event.uuid = uuid` so we were getting `Invalid event.` error

#### Questions

#### Release Type:
- [X] Patch


iglocska:
Makes sense, thanks a lot!

PR #3536, opened 2018-08-07
Default sort order for Id and Date

StefanKelm:
#### What does it do?

It changes the default sort order for Id and Date to "desc" on proposal views

#### Questions

#### Release Type:
- [ ] Major
- [X] Minor
- [X] Patch


iglocska:
Thanks a lot!

Issue #3544, opened 2018-08-09
Feature Request: option to GnuPG sign outgoing event emails (alerts)


WaryWolf:
Hi,

at AusCERT we use GPG to sign the majority of our outgoing emails. MISP has GPG support built in, but only offers encryption of outgoing mails if users have provided their GPG pubkey (as far as I know - I haven't tested this feature).

Is it possible to add an option to plaintext-sign outgoing emails with the MISP server's GPG key? I think it would just require calling [this function](http://pear.php.net/package/Crypt_GPG/docs/latest/Crypt_GPG/Crypt_GPG.html#methodsign) on the email body before sending it out.

Please let me know what you think.

adulau:
Did you enable and test the option "Optional | GnuPG.sign" ? Just wondering.

WaryWolf:
Hi adulau,

Yes - we've had that turned on for a while, but I think it only takes effect if the email is going to be encrypted - plaintext emails aren't signed.

I've had a look at the code, and it looks like all of the logic and GPG operations happen in app/Model/User.php:

GPG signing occurs when GnuPG.sign is true, on line 854: https://github.com/MISP/MISP/blob/7e3ac1f69e834bbfc5f9105fbbb3040bdbb55519/app/Model/User.php#L854-L857

However, this code is inside the `__encryptUsingGPG` function, which is only entered if `$canEncryptGPG` is set: 
https://github.com/MISP/MISP/blob/7e3ac1f69e834bbfc5f9105fbbb3040bdbb55519/app/Model/User.php#L755

https://github.com/MISP/MISP/blob/7e3ac1f69e834bbfc5f9105fbbb3040bdbb55519/app/Model/User.php#L771-L772

As `$canEncryptGPG` is only set for users who have added a pubkey, MISP will never send out clear-signed emails.

I think it shouldn't be too hard to split out the signing into a separate function. I can try submitting an implementation tomorrow but it won't be that good. 

Regards, Anthony

WaryWolf:
This has been resolved with #3546 . Thanks!

PR #3546, opened 2018-08-10
Split GPG signing and encrypting of outgoing emails into separate operations

WaryWolf:
Split GPG signing and encrypting of outgoing emails into separate operations
Allows for plaintext signing of outgoing emails.

#### What does it do?

Fixes the issue mentioned in #3544 

I've tested this on our internal dev/sandbox server - signing now works when gnupg.sign is set to true, and doesn't happen when gnupg.sign is set to false. Encrypted mails also work - they're signed when gnupg.sign is set to true, and not signed when gnupg.sign is set to false.

However I'm not that familiar with PHP so please check my code carefully.

Thanks!

#### Questions

- [ ] Are you using it in production? Not yet.

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Model/User.php
+++ b/app/Model/User.php
@@ -751,7 +751,8 @@ class User extends AppModel
         }
         $failed = false;
         $failureReason = "";
-        // check if the e-mail can be encrypted
+        // check if the e-mail can be signed & encrypted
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks a lot!

PR #3547, opened 2018-08-10
Update INSTALL.ubuntu1804.with.webmin.txt

PaoloVecchi:
systemd sucks!
;-)

#### What does it do?

Just telling systemd that /etc/rc.local is alive and kicking and it needs to run some workers

#### Release Type:
- [ ] Major
- [x] Minor
- [ ] Patch


iglocska:
Woah, thanks - and yeah it does ;)

PR #3549, opened 2018-08-12
Update report_validation_issues_events.ctp

PaoloVecchi:
Speling mistake? 'V' of validation outside php tag.

#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch


adulau:
Nice catch. Thank you!

PR #3553, opened 2018-08-13
Update EventsController.php

PaoloVecchi:
Just a ) missing.

#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch


adulau:
Thank you! good catch.

PR #3555, opened 2018-08-13
Add a permission check to the change password page.

### LINKED ISSUE #3554: Security bug: users can change passwords even when disableUserSelfManagement is set
### Work environment

| OS version (server)       | Ubuntu 16.04
| OS version (client)       | Ubuntu 18.04
| PHP version               | 7.0
| MISP version / git hash   | 2.4.94

### Expected behavior

If the "MISP.disableUserSelfManagement" config variable is set, the "Edit My Profile" and "Change Password" pages should not be accessible by users without admin privileges.

### Actual behavior

The rendering code for the  "Edit my profile" page checks to see if the user has permissions to view the page, but the "Change Password" page does not. If a user can (trivially) guess the URL of the change password page, they can change their passwords even when the "MISP.disableUserSelfManagement" config variable is set.

Links to code below:
"Edit" rendering code:
https://github.com/MISP/MISP/blob/b04f93800cffebb85d35dd5ebcc00f52afbfc6ed/app/Controller/UsersController.php#L90-L94

"Change Password" rendering code:

https://github.com/MISP/MISP/blob/b04f93800cffebb85d35dd5ebcc00f52afbfc6ed/app/Controller/UsersController.php#L165-L168

1. Log into a MISP instance with "MISP.disableUserSelfManagement" set to true, with a non-admin user account
2. visit https://MISP_URL/users/change_pw
3. you should be able to change your password.

WaryWolf:
Fixes #3554.

The 'MISP.disableUserSelfManagement' config variable is checked
when rendering the link to the change password page, but is not checked
when rendering the page itself. This could lead to unauthorised
password changes by users with existing accounts on the MISP
instance.

#### What does it do?

#### Questions

- [X] Are you using it in production? Yes, on our internal/sandbox server.

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/UsersController.php
+++ b/app/Controller/UsersController.php
@@ -164,6 +164,9 @@ class UsersController extends AppController
 
     public function change_pw()
     {
+        if (!$this->_isAdmin() && Configure::read('MISP.disableUserSelfManagement')) {
+            throw new MethodNotAllowedException('User self-management has been disabled on this instance.');
+        }
         $id = $this->Auth->user('id');
         $user = $this->User->find('first', array(
             'conditions' => array('User.id' => $id),


adulau:
Good catch! Thank you.

PR #3560, opened 2018-08-15
Use AWS S3 as an attachment storage

FloatingGhost:
#### What does it do?

Adds a setting to offload `attachments_dir` to an S3 bucket

Shouldn't interfere with standard non-s3 installations 

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


RichieB2B:
Wow, MISP is now officially cloud based! 8-)

iglocska:
Woah, this is awesome! Thanks a lot!

PR #3570, opened 2018-08-17
fix: use configured attachments_dir instead of app/files/ in ShadowAt…

Lastpixl:
#### What does it do?

ShadowAttributesController currently writes to app/files/ instead of writing to the configured attachments_dir. This PR fixes this.

#### Questions

- [x] Are you using it in production? very soon
- [x] Does it require a change in the API (PyMISP for example)? no

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Woah, quite the oversight that this wasn't the case already. Thanks a lot, completely missed this!

PR #3585, opened 2018-08-23
Control CakeResque Redis connection

FloatingGhost:
Adds `CakeResque.Redis.*` settings under `MISP` in server settings and maintenance. Currently has to do a bit of array flattening to deal with nested array settings, but seems to not break everything 

Also a misc fix in which attachments_dir was tested for writable failing when it was set to s3:// 

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Should this be closed? Or would you like to have it merged in the end? Am lagging super far behind with merging stuff but it seems to make sense to have this.

FloatingGhost:
oh, uh, I may have accidentally had to `--force` my forked repo to sync it with what I was running locally 

I still have a copy of it in the office, I'll re-open next time I'm in

PR #3587, opened 2018-08-23
Fix broken timestamps by using 24 hour clock and ISO 8601 date format

droe:
#### What does it do?

The event view shows a wrong "Last change", e.g. "2018/08/23 06:01:45"
for "2018/08/23 18:01:45".  The same problem affects the timestamp in
the XML generated by IOCExportTool.php.  Fix by correcting the PHP
date() code "h" to "H".

While here, also switch to a clearer ISO 8601 date representation for
"Last change", using dashes instead of slashes for separation of year,
month and day.

#### Questions

#### Release Type:

- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Lib/Tools/IOCExportTool.php
+++ b/app/Lib/Tools/IOCExportTool.php
@@ -55,7 +55,7 @@ class IOCExportTool
     {
         $temp = '';
         // We will start adding all the components that will be in the xml file here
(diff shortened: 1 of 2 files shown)


droe:
Reference: http://php.net/manual/en/function.date.php

iglocska:
Thanks a lot!

PR #3589, opened 2018-08-24
Add email field option for kerberos authentication

iwitz:
#### What do these changes do ?
This PR introduces a new configuration option that allows administrators to specify a field in which the Kerberos authentication component will look for users' email addresses. It also adds an example use of that option in the default configuration file.

#### How is this useful ?
I needed to authenticate users of different AD domains on MISP. One of these domains had email addresses in a field that wasn't named "mail", hence preventing MISP from finding email addresses for that domain's users. This PR allows administrators who'd want to authenticate users from multiple AD domains that have email addresses stored in different fields to do so.

#### Backward-compatibility
This new option is not required. When it is not set, it looks for email addresses in the "mail" field like it currently does. Therefore, this PR does not break existing configuration files.

#### Questions

- [ ] Does it require a DB change? => It doesn't
- [X] Are you using it in production? => Yes
- [ ] Does it require a change in the API (PyMISP for example)? => It doesn't

#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch


SteveClement:
Looks ok, thanks for the contribution.

PR #3623, opened 2018-09-06
chg: [doc] Guides in the main code base are obosolete

SteveClement:
As discussed, I removed the global menu reference from the internal Guide to the main `misp-book` link on CIRCL.lu

Added Categories and Types to Global Menu for easy access.


iglocska:
Thanks a lot!

PR #3654, opened 2018-09-09
fix: [doc] Add an option to checkout submodules recursively

0xiso:
#### What does it do?

Some submodules such as PyMISP have submodule in it. We need to pass `--recursive` option to recursively init and update submodules in order to make those submodules work properly.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #3660, opened 2018-09-10
Enables tag batch editing (exportable/hidden)

ancailliau:
The exportable and hidden flag can be edited in batch, from the tag list page.

**Release Type:** Minor


app/Controller/TagsController.php
```diff
@@ -885,4 +885,44 @@ public function viewGraph($id)
         $this->set('id', $id);
         $this->render('/Events/view_graph');
     }
+
+    public function massEditSelected()
+    {
+        if (!$this->_isSiteAdmin()) {
+            throw new NotFoundException('You don\'t have permission to do that.');
```
iglocska, line 1041: Best would be to instead set the restriction in the ACL to () instead of ('perm_tagger')

app/Controller/TagsController.php
```diff
@@ -885,4 +885,44 @@ public function viewGraph($id)
...
+        }
+        $this->render('ajax/tag_mass_edit');
+    }
+
+    public function editSelected()
+    {
+        if (!$this->_isSiteAdmin()) {
```
iglocska, line 1048: best to force this via the ACL component, just set it to () instead of ('perm_tagger') there

app/Controller/TagsController.php
```diff
@@ -885,4 +885,44 @@ public function viewGraph($id)
...
+            throw new NotFoundException('You don\'t have permission to do that.');
+        }
+        $tag_ids = json_decode($this->request->data['Tag']['tag_ids']);
+        $tags = $this->Tag->find('all', array(
```
iglocska, line 1056: best to introduce a check here for empty $tag_ids, just in case so it doesn't find every exisitng tag.

app/Controller/TagsController.php
```diff
@@ -885,4 +885,44 @@ public function viewGraph($id)
...
+            'conditions' => array(
+                'Tag.id' => $tag_ids
+            ),
+            'recursive' => -1,
+        ));
+
+        $hidden = $this->request->data['Tag']['hidden'];
```
iglocska: Check for empty, user might be coming from the API and not set that field

iglocska (changes requested):
Looks very good, could you change the few points noted above?

ancailliau:
The PR is updated according the comments.

ancailliau:
@iglocska Up, as discussed yesterday :)

SteveClement:
Tested NOT working.

You select tags, toggle them in the new UI and nothing happens.

Perhaps some things have changed since September 2018.

Will take a quick look later on.

SteveClement:
@ancailliau any upates on this PR?

ancailliau:
Sorry, I was abroad. I will check and update asap. Thx for the reminder.

PR #3664, opened 2018-09-11
chg: [doc] Moved INSTALL files around to reflect a more accurate support landscape.

SteveClement:
chg: [doc] Update README.md to explain some of the folders/files.
chg: [doc] Added zmq, redis, maec python module installations.


iglocska:
Thanks a lot!

PR #3725, opened 2018-09-28
Unify url for modules, make them usable behind proxies

lhirlimann:
#### What does it do?

Let's people using git behind proxies, get all the submodules.

#### Questions

- [ NO] Does it require a DB change?
- [ TRYING] Are you using it in production?
- [ NO] Does it require a change in the API (PyMISP for example)?

#### Release Type:
- [X] Patch


adulau:
Good catch! Thank you.

PR #3729, opened 2018-10-02
Trim spaces from source_id in merge form

RichieB2B:
Spaces get added when copy-pasting event ID's, it's very annoying to have to remove them by hand.


iglocska:
Thanks a lot!

PR #3730, opened 2018-10-02
fix: Customauth Headers are case-sentitive, do not strtoupper

FloatingGhost:
In the CustomAuth plugin, headers are auto-uppercase'd

Some authentication systems, like [mod_auth_openidc](https://github.com/zmartzone/mod_auth_openidc) store their claims in lower-case header names. Auto-uppercasing breaks compatibility in this case.

#### What does it do?

Removes a statement forcing Authentication headers in CustomAuth to be uppercase

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Wow, good to know, thanks.

drehak:
Thank you Hannah, very cool!

PR #3731, opened 2018-10-02
Only display "Show 2 more" and up

RichieB2B:
The message "Show 1 more..." takes up more space than actually showing it. So this patch only shows the message when there are at least 2 more attribute correlations to show in the event view.


iglocska:
Nice fix for the fail UX, thanks!

PR #3737, opened 2018-10-04
Control Cakeresque redis with settings

FloatingGhost:
#### What does it do?

Re-open of #3585 with not completely broken forked repos!

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


SteveClement:
@iglocska are you happy with those changes?

PR #3743, opened 2018-10-08
Add "manage workers" option.

WaryWolf:
#### What does it do?

Adds a boolean option to toggle MISP managing its own worker processes.

This is enabled by default, which replicates the current behaviour of having controls to start, stop and restart workers in the server settings page.
When set to disabled, these controls are hidden, which allows server administrators to manage the worker processes externally, e.g. via systemd.

A sample systemd unit file has also been included into the INSTALL directory.

#### Questions

- [ ] Are you using it in production? Yes, on our sandbox/development server
- [ ] Does it require a change in the API (PyMISP for example)? Not as far as I know.

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/INSTALL/misp-workers.service
+++ b/INSTALL/misp-workers.service
@@ -1,0 +1,17 @@
+[Unit]
+Description=misp-workers
+After=network.target apache2.service
+
+[Service]
+Type=forking
+KillMode=none
+User=www-data
+Group=misp-server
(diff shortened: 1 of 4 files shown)


iglocska:
Lovely idea, digging it, thanks!

PR #3751, opened 2018-10-10
Fixes a typo in an error message (control_workers -> manage_workers)

ancailliau:
The warning displayed in the Worker page does not refer to the correct setting key. This is very confusing but easily fixed :)


iglocska:
Thanks! Managing workers also sounds much less demeaning than controlling them ;)))

PR #3771, opened 2018-10-16
Update INSTALL.rhel7.txt

P4rs3R:
Added instruction (at line 109) and updated line 8. Pull type: patch.
Tested on RHEL 7.5 and CentOS 7.5.


adulau:
Thank you very much!

Issue #3633, opened 2018-09-07
Issue with event "Download as... MISP XML" that returns json instead of XML


l3m0ntr33:
Issue with event "Download as... MISP XML" that returns json 

### Work environment
misppriv.circl.lu

### Expected behavior
XML File format expected

### Actual behavior
JSON File format returned

### Steps to reproduce the behavior
Select any event, click download as on the left menu and select "MISP XML (metadata + all attributes) " and it returns JSON formatted output

iglocska:
Oh wow o.O. That's annoying. Will have a look.

vmapps:
XML content is correctly returned on previous MISP versions (tests made on 2.4.90 and 2.4.95)

l3m0ntr33:
@vmapps you mean 2.4.90 and 2.4."**93**" ;)

vmapps:
Update: XML content is correctly returned on previous MISP versions (tests made on 2.4.90 and 2.4.93)

vmapps:
Update: XML content is correctly returned on MISP version 2.4.94

PR #3778, opened 2018-10-19
Fixes Issue #3633 - Returned XML has application/json Content-Type header

tomking2:
#### Questions

#### Release Type:
- [X] Patch

### General
Created a patch to fix issue #3633, which affects the MISP XML output via "Download as" on an Event. 

Issue appears to stem from the fact that _sendResponse was checking the $format against "application/xml". it appears the RestSearches obtain the format via:

https://github.com/MISP/MISP/blob/6b60f3fa5c9900c82c9a966811ceba4580a9679d/app/Controller/AttributesController.php#L1725
https://github.com/MISP/MISP/blob/6b60f3fa5c9900c82c9a966811ceba4580a9679d/app/Controller/AttributesController.php#L1744

Where index 0 is simply 'xml':
https://github.com/MISP/MISP/blob/7a01de535928b52ebc7c219dcbba7fd82edb6b7b/app/Model/Attribute.php#L120-L130

However, other functions that use _sendResponse I believe do pass "application/xml", so both are necessary.

### DIFF
--- a/app/Controller/Component/RestResponseComponent.php
+++ b/app/Controller/Component/RestResponseComponent.php
@@ -342,7 +342,7 @@ class RestResponseComponent extends Component
 
     private function __sendResponse($response, $code, $format = false, $raw = false, $download = false)
     {
-        if (strtolower($format) === 'application/xml') {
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks a lot!

PR #3808, opened 2018-10-26
Fix: AWS Errors when given empty array of elements

FloatingGhost:
#### What does it do?

Fixes an error on deleting an event when S3 is enabled - if given `[]` to delete, AWS will just kill itself instead of deleting nothing. Funny, huh?

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


adulau:
Thank you!

PR #3826, opened 2018-10-31
chg: [eventview] changed default attribute sorting to timestamp->desc

mokaddem:
Changed default behavior in ``events/view`` UI's attribute table.
Default sorting is done on ``timestamp->desc`` instead of ``category->asc``


SteveClement:
Works as expected, thanks.

Issue #3834, opened 2018-11-01
PyMISP crafted objects with references failed to add via API


andurin:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | Ubuntu
| OS version (client)       | Ubuntu
| PHP version               | 7.0
| MISP version / git hash   | 2.4.XX, 1bf6e57118528e6de7a602196f1d06abc56b6635
| Browser                   | Firefox

### Expected behavior
Result should be a new created event with File-Objects and references.

### Actual behavior
Event is created but references are missing.

### Steps to reproduce the behavior
Tried to `add_event()` with the following json data.

references were created like this:
```
parent_uuid = 'real_life_uuid'
misp_object.add_reference(parent_uuid, 'extracted-from','')
```

JSON:
```
{
  "Event": {
    "Attribute": [],
    "Object": [
      {
        "Attribute": [
          {
            "category": "Payload delivery",
            "disable_correlation": true,
            "object_relation": "filename",
            "to_ids": false,
            "type": "filename",
            "uuid": "a5aaf110-9dfd-4d05-bcf2-0616beea1210",
            "value": "testfile.txt"
          },
          {
            "category": "Payload delivery",
            "disable_correlation": false,
            "object_relation": "md5",
            "to_ids": true,
            "type": "md5",
            "uuid": "c2b8136d-ced4-44c1-8181-e65c6db1646c",
            "value": "2205e48de5f93c784733ffcca841d2b5"
          },
          {
            "category": "Payload delivery",
            "disable_correlation": false,
            "object_relation": "sha1",
            "to_ids": true,
            "type": "sha1",
            "uuid": "ef12e254-21bc-4794-8147-213abac471ea",
            "value": "1c68ea370b40c06fcaf7f26c8b1dba9d9caf5dea"
          },
          {
            "category": "Payload delivery",
            "disable_correlation": false,
            "object_relation": "sha256",
            "to_ids": true,
            "type": "sha256",
            "uuid": "3e80690b-419e-4d52-8a27-971b9635d57c",
            "value": "c9d04c9565fc665c80681fb1d829938026871f66e14f501e08531df66938a789"
          },
          {
            "category": "Payload delivery",
            "disable_correlation": false,
            "object_relation": "sha512",
            "to_ids": true,
            "type": "sha512",
            "uuid": "4d538e8e-d9db-4069-9a2d-94f4fd6d584f",
[...]

PR #3857, opened 2018-11-09
Fix missed object references using API

ghost:
referenced_type should be an integer based on the named objectTypes from array

Issue #3834

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


ghost:
Guess this one is for @iglocska

Issue #3859, opened 2018-11-09
'Contact User(s)' should contain an option to contact users of an Org


rommelfs:
When in the context of an Org, the 'contact everyone' is misleading. Someone (*cough*) might believe (*cough cough*) he's going to contact everyone at that Org.

A feature request is: add an additional option to contact everyone at an Org.

PR #3861, opened 2018-11-09
Contact user sanity check

### LINKED ISSUE #3858: Sanity check for 'Contact User(s)'
When using 'Contact User(s)' add a sanity check by showing an acknowledgement message displaying the number of recipients.

mokaddem:
## What does it do
fix #3858 and #3859 
- Allows admin users to preview which users will receive a mail after form submission.
- Allows admin users to send a mail to all members of an organization

## Dev note
Performs the same SQL conditions building as the standard form handling function. But if the preview parameter is set to true, will only show the selected users.

### DIFF
--- a/app/View/Users/admin_email.ctp
+++ b/app/View/Users/admin_email.ctp
@@ -76,7 +76,7 @@ $(document).ready(function() {
 
     // Confirm before submit
     $('#UserAdminEmailForm').submit(function(e) {
-        var url = 'http://127.0.0.1:8085/admin/users/email_confirm?';
+        var url = 'http://127.0.0.1:8085/admin/users/email/true?';
         url += 'recipient=' + $('#recipient').val();
         url += '&recipientEmailList=' + $('#UserRecipientEmailList').val();
         url += '&orgNameList=' + $('#UserOrgNameList').val();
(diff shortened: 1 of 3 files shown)


SteveClement:
Tried it in my test-rig but the submit button did not respond...
Need to investigate further.
Perhaps also another cross-check would be nice...

Console message:

Empty string passed to getElementById().

app/View/Users/admin_email.ctp
```diff
@@ -64,13 +67,33 @@ $("#customMessage").change(setAll);
...
 $(document).ready(function() {
     var org = "<?php echo $org;?>";
     subjects = ["", "[" + org + " MISP] " + "<?php echo __('New user registration');?>" , "[" + org + " MISP] " + "<?php echo __('Password reset');?>"];
     standardTexts = ['', '<?php echo h($newUserText); ?>', '<?php echo h($passwordResetText); ?>'];
-    //setAll();
+    setAll();
+
+    // Confirm before submit
+    $('#UserAdminEmailForm').submit(function(e) {
+        var url = 'http://127.0.0.1:8085/admin/users/email/true?';
```
SteveClement: Hmm, not sure about this line...

mokaddem: Replaced with $baseurl...

PR #3880, opened 2018-11-15
fix: [tools] misp-restore.sh incorrectly validating 'BackupFile' from…

crford:
… the command line

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #3892, opened 2018-11-21
fix: fixes variable initialization inconsistency in Server push

cvandeplas:
Fixes issues like: Warning (2): count(): Parameter must be an array or an object that implements Countable in [/var/www/MISP/app/Model/Server.php, line 2353]

#### Release Type:
- [X] Patch


iglocska:
Thanks a million!

PR #3908, opened 2018-11-27
fix: Return attributes & proposals in fetchEvent

dawid-czarnecki:
Returns events with attributes and proposals when metadata = True

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
This is working as intended - the metadata flag allows the return of meta data only. The PR would do the opposite meaning that we always unset the attributes unless we ask for the metadata.

PR #3912, opened 2018-11-29
fix: [model-data] Network activity category: add x509-...-md5 and x509-...-sha256

Sh3idan:
#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Makes sense, nice catch. Thanks!

PR #3913, opened 2018-11-30
make the filter checks for the event index API call case-insensitive

### LINKED ISSUE #3855: Bug: datefrom ignored in PyMISP request
### Work environment

| OS version (client)       | macOS / pyMISP
| MISP version / git hash   | 2.4.96
| Browser                   | PyMISP access (no browser)

### Expected behavior
If `datefrom` parameter is set to a datetime one hour ago in the PyMISP api, only events which are not older than one hour are selected

### Actual behavior
`Datefrom` parameter has no effect. All events are selected, no matter what is in the variable `datefrom`

```
misp_instance = pymisp.api.PyMISP(url=misp_url, key=misp_key, ssl=misp_verifycert, out_type='json',
                                  debug=None, proxies=None, cert=None, asynch=False)
events = misp_instance.search_index(datefrom=(datetime.datetime.now() - datetime.timedelta(hours=1)))
```

bildschirmfoto 2018-11-08 um 13 31 15
bildschirmfoto 2018-11-08 um 13 31 26

WaryWolf:
#### What does it do?

Fixes #3855 as well as making all filters sent to this api call case-insensitive.
There is already a `strtolower()` call on filters in this method , but it gets called after filters are checked against a whitelist. I fiddled with the checking so filters are first changed to lowercase, then compared to the whitelist.

I've tested the code by calling the API with a bunch of different filters, some non-existent, some existent but with capital letters... it works, but please double-check my code. Thanks!

#### Questions

- [ ] Are you using it in production? Yes (Sandbox server)

No, but it does fix a bug in the API that was introduced with b4f084f9fe66547d65bb1d9187aa56bd8c9b2dc3 .

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -275,16 +275,25 @@ class EventsController extends AppController
         $urlparams = "";
         $overrideAbleParams = array('all', 'attribute', 'published', 'eventid', 'datefrom', 'dateuntil', 'org', 'eventinfo', 'tag', 'tags', 'distribution', 'sharinggroup', 'analysis', 'threatlevel', 'email', 'hasproposal', 'timestamp', 'publishtimestamp', 'publish_timestamp', 'minimal');
         $passedArgs = $this->passedArgs;
+
         if (isset($this->request->data)) {
             if (isset($this->request->data['request'])) {
                 $this->request->data = $this->request->data['request'];
             }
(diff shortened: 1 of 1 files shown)


SteveClement:
This seems to be fixed as per https://github.com/MISP/MISP/issues/3855#issuecomment-469157025

PR #3924, opened 2018-12-06
Fix error handling in Server.php

keram79:
according to http://php.net/manual/de/exception.getmessage.php , the parenthesis are required

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Argh, epic fail. Nice catch, thanks ;)

Issue #3921, opened 2018-12-04
allow access to Object definitions via API [feature]


cvandeplas:
Considering each MISP instance can have its own Object definitions, these should be made available through the API.
(all at once, and/or each separately)

iglocska:
Like this?

/object_templates/index

cvandeplas:
preferably using the same structure as the events, attributes, objects, galaxies and taxonomies :-)
But in the end I guess I'll just use the PyMISP implementation

iglocska:
But we can't provide it if we don't know about it ;) MISP just loads whatever is in the appropriate directories - we have no way of telling remote vs local apart.

cvandeplas:
Could you also make `objectTemplates/view/688c46fb-5edb-40a3-8273-1af7923e2215.json` available?
As the Object within an event only refers to the uuid?

cvandeplas:
It seems there's still a bug with the objectTemplates via UUID. 

Comparing `/objectTemplates/view/8b3228ad-6d82-4fe6-b2ae-05426308f1d5.json` 
`/objectTemplates/view/71.json` gives me different versions of the template. The one with the `uuid` in the url gives an older version.

cvandeplas:
It looks like there can be double versions of object templates in the database:
```
MariaDB [misp]> select * from object_templates where uuid = '8b3228ad-6d82-4fe6-b2ae-05426308f1d5';
+----+---------+--------+--------------------------------------+--------------+---------------+----------------------------------------------------------------------------------------------+---------+-----------------------------------------+-------+--------+
| id | user_id | org_id | uuid                                 | name         | meta-category | description                                                                                  | version | requirements                            | fixed | active |
+----+---------+--------+--------------------------------------+--------------+---------------+----------------------------------------------------------------------------------------------+---------+-----------------------------------------+-------+--------+
| 31 |       1 |      1 | 8b3228ad-6d82-4fe6-b2ae-05426308f1d5 | registry-key | file          | Registry key object describing a Windows registry key with value and last-modified timestamp |       2 | {"requiredOneOf":["key","name","data"]} |     1 |      0 |
| 71 |       1 |      1 | 8b3228ad-6d82-4fe6-b2ae-05426308f1d5 | registry-key | file          | Registry key object describing a Windows registry key with value and last-modified timestamp |       4 | {"requiredOneOf":["key","name","data"]} |     1 |      1 |
[...]

PR #3927, opened 2018-12-09
fix: [objecttemplates] fixes issue #3921

cvandeplas:
#### What does it do?

If it fixes an existing issue, please use github syntax: `#3921`

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot, nice catch

PR #3932, opened 2018-12-11
Mod headers to be enabled

davidonzo:
Apache configuration needs to load the headers module.
Added the line 103


adulau:
Good point. Thank you for the contribution!

Issue #3939, opened 2018-12-12
Still supporting IE11?


tomking2:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | Ubuntu
| OS version (client)       | Seven
| PHP version               | 7.1
| MISP version / git hash   | 2.4.99, 083beae50271efc8ab960f07ae7563f3eff3bd94
| Browser                   | IE11

I'm wondering if MISP is looking to phase out the support of IE11, or is it something we are looking to persist?

If so, spotted an issue with the latest code from a commit made 2018/11/24, specifically this line:
https://github.com/MISP/MISP/blob/260f75d8c38f2ce8a95e3b6d869e5cfe9b83244b/app/webroot/js/misp.js#L301

This line is using default parameters supported in ES2015. Default params unfortunately aren't supported in IE11 :(. As a result, this makes causes an error when parsing the misp.js file, and all javascript features (such as Download As... on an event) no longer work.

### Steps to reproduce the behavior
Update to at least commit 28cd0322641f930f0f286977581ffeae1a1524ab. Go to IE11 and visit an event. You'll see that an error is raised and unable to run javascript funtions such as showPopup as it's not defined.

If the decision has been made to scrap IE11 then so be it :), but thought I'd validate first.

iglocska:
Nah, definitely no such decision has been made - we just don't use any windows machines so IE11 support is mostly based on users notifying us if we break something :) Will fix it!

PR #3941, opened 2018-12-13
fix: Make IE11 compliant by removing default param value from ES2015

tomking2:
#### What does it do?
More fixed for issue `#3939`, relating to ES2015 default parameter support on IE11. 

I should have checked that default params weren't in place elsewhere, whoops!

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

tomking2:
No problem!

PR #3942, opened 2018-12-13
fix: Make IE11 compliant by adding startsWith polyfill

tomking2:
#### What does it do?

Oh and one more fix for `#3939`. You happy to polyfill the startsWith prototype? Otherwise we'll have to switch startsWith for indexOf in the code on this line:
https://github.com/MISP/MISP/blob/3f07d8367eb8dbd64fe2ffbaa8ff69ce1e0c95b6/app/webroot/js/misp.js#L3541

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Yep that looks good to me.

PR #3956, opened 2018-12-18
Fix disabling correlations for single attributes

dawid-czarnecki:
#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #3965, opened 2018-12-19
Add cdhash attribute for macOS code signature CDHash

droe:
#### What does it do?

Add `cdhash` attribute to describe Code Directory Hashes on signed Apple Mach-O executables, app bundles, kext bundles etc.

The CDHash is computed as a cryptographical hash over the hashes of each memory page of the executable (or bundle) plus additional data, such as the requirements string, the entitlements and resources. The hash function used was initially SHA-1, but is now SHA-256 truncated to SHA-1 length for userspace compatilibity reasons. Many APIs and tools do not differentiate between SHA-1 or SHA-256/160 based CDHashes. Also note that the CDHash is a very different thing than the flat file hash of the executable file. The CDHash of a Mach-O or bundle code signature is not generally reproducible, i.e. you cannot verify the CDHash of an unsigned executable or bundle contents without knowing all the different factors that go into it in addition to the code. However, many security mechanisms in macOS use the CDHash as the identifier for (signed) code, and as such it is a useful indicator in macOS incident response and threat intelligence. Especially so since there is no established, useful and stable way to compute a flat file hash over a bundle.

This pull request is just a small step towards better support of Mach-O executables and code signatures, perhaps with respective object definitions. In any case I think we need a base attribute type for CDHashes, because none of the existing ones would do the job very well.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

/cc @iglocska @cvandeplas following our discussion on gitter

### DIFF
--- a/app/Lib/Export/BroExport.php
+++ b/app/Lib/Export/BroExport.php
@@ -46,6 +46,7 @@ class BroExport
         'sha512/224' => array('brotype' => 'FILE_HASH'),
         'sha512/256' => array('brotype' => 'FILE_HASH'),
         'tlsh' => array('brotype' => 'FILE_HASH'),
+        'cdhash' => array('brotype' => 'FILE_HASH'),
         'filename|authentihash' => array('brotype' => 'FILE_NAME', 'composite' => 'FILE_HASH'),
         'filename|ssdeep' => array('brotype' => 'FILE_NAME', 'composite' => 'FILE_HASH'),
         'filename|imphash' => array('brotype' => 'FILE_NAME', 'composite' => 'FILE_HASH'),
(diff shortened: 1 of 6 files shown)


iglocska:
To me it looks good, @cvandeplas - want to pitch in following the discussion on gitter?

cvandeplas:
It looks great!  Thank you very much !

iglocska:
Also, kudos for thinking about the export mappings too!

adulau:
It looks great.

We just need to update the attributes type descriptions in PyMISP. @Rafiot could you update it?

@droe  If you have any new specific objects, feel free to make a pull-request. Thanks a lot.

Rafiot:
done

cvandeplas:
Actually I forgot we've made a TODO list (and script) 
 https://github.com/MISP/MISP/wiki/Categories-&-Types-changes-CheckList

I'll take care of it.

cvandeplas:
@adulau Can you regenerate and upload the website pages? and the RFC if needed? Thanks

PR #3969, opened 2018-12-22
chg: [Objects] Sorts object references types in add reference form

cvandeplas:
#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

This change makes adding object references easier as the list of the reference types is now sorted.


adulau:
Good point! Indeed, it's better for usability.

PR #3973, opened 2018-12-28
Tag soft deletion patch

lucamemini:
see issue #3373 

For older installation need an database upgrade with new columns (es: mysql)

ALTER TABLE attribute_tags ADD deleted tinyint(1) NOT NULL DEFAULT 0;
ALTER TABLE event_tags ADD deleted tinyint(1) NOT NULL DEFAULT 0;

### DIFF
--- a/INSTALL/MYSQL.sql
+++ b/INSTALL/MYSQL.sql
@@ -57,6 +57,7 @@ CREATE TABLE IF NOT EXISTS `attribute_tags` (
   `attribute_id` int(11) NOT NULL,
   `event_id` int(11) NOT NULL,
   `tag_id` int(11) NOT NULL,
+  `deleted` tinyint(1) NOT NULL DEFAULT 0,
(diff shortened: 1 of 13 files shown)


adulau:
@mokaddem should we merge this after the new updater is included?

lucamemini:
hello! :-)
any news? :)

SteveClement:
> @mokaddem should we merge this after the new updater is included?

Was there a consensus? Should we fix this PR?

mokaddem:
Oh right...
The new updater is embedded in the monstrous branch so called ``zoidberg``.
Thus, I need to find the time to do the separation (not before start of June unfortunately).
So IMO either:
- We merge it and close it's 3 related issues at the cost of altering 1 big table (``attribute_tags``) without feedback regarding the update process
- We wait until the updater is merged into 2.4, then, we fix and merge this PR

PR #3974, opened 2018-12-30
Update Attribute.php

eCrimeLabs:
#### What does it do?

Added ja3-fingerprint-md5 as a dedicated data type. Also updating the object for ja3

#### Questions

#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch


adulau:
Thank you! I'll update the descType data files.

PR #3980, opened 2019-01-02
fix: [api] editing organisation attributes, other than name

JanSkalny:
#### What does it do?

Editing any organisation attribute other than name was not possible via API. This should fix it.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #4023, opened 2019-01-17
Allow multi-event exporting via misp-modules

FloatingGhost:
#### What does it do?

Allows throwing a list of comma-separated IDs to `/events/exportModule/<module>/<id_list>` to
export multiple events to the same file. 

It's not the best solution, a proper `event-collection` inputSource would likely replace this eventually.
However it works and is kinda useful for some stuff I guess

Additionally, composer.phar wasn't compatible with PHP7.3 so I updated that to the latest version.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


SteveClement:
@FloatingGhost ideally you would seperate the composer.phar update from your other changes.

iglocska:
The problem I see with this: It works for export formats where concatenation works, but it will break whenever an export contains any encapsulation (think headers in a CSV, encapsulation in something like S̻̲̞̦̞͈̻̈́̃̀͋̿T̵̫̺̼͚̼̺̠͋̂̂̆̈́͟͜Ỉ̵̛̤̤̰͓͎͕̝͚̀̇͆͋̓̓̽X̺̰͉̪̭̓͆́̕͘͠ͅ, etc.). The long term solution is to move every module that wants to be compose-able to a system similar to what we have in restsearch (responding to a request for header, footer, separator in addition to converting individual parts).

SteveClement (changes requested):
Please remove composer.phar from this PR, we will merge-in a new version in a seperate PR.

SteveClement:
@FloatingGhost any update on this?

PR #4049, opened 2019-01-24
Accessibility fix in the Logs view

obert01:
Accessibility: Fixed the aria-label properties of the filter buttons in the Logs view.

It is only a cosmetic change, only visible for screen reader users.


iglocska:
Thanks a lot, let us know if anything's missing labeling / if you see anything we could improve!

PR #4123, opened 2019-02-07
Add LDAP network timeout

czechnology:
#### What does it do?

Adds an option to timeout during an LDAP connection, if the server doesn't respond.
By default, it sets the network timeout to -1, meaning infinite timeout, same behavior as before this change.

Reference for the LDAP option: https://linux.die.net/man/3/ldap_set_option
> **LDAP_OPT_NETWORK_TIMEOUT**
> Sets/gets the network timeout value after which poll(2)/select(2) following a connect(2) returns in case of no activity. outvalue must be a struct timeval ** (the caller has to free *outvalue), and invalue must be a const struct timeval *. They cannot be NULL. Using a struct with seconds set to -1 results in an infinite timeout, which is the default. This option is OpenLDAP specific.

This option is supported since PHP 5.3.0

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Config/config.default.php
+++ b/app/Config/config.default.php
@@ -134,6 +134,7 @@ $config = array(
 			'apacheEnv'          => 'REMOTE_USER',           // If proxy variable = HTTP_REMOTE_USER
 			'ldapServer'         => 'ldap://example.com',   // FQDN or IP
 			'ldapProtocol'       => 3,
+			'ldapNetworkTimeout' => -1,  // use -1 for unlimited network timeout
 			'ldapReaderUser'     => 'cn=userWithReadAccess,ou=users,dc=example,dc=com', // DN ou RDN LDAP with reader user right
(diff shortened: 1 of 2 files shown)


adulau:
Sounds idea a good idea to have it as configuration. @iglocska what do you think?

iglocska:
Sounds good to me, if it works for you in practice - great!

czechnology:
Yes, in my case it helped overcome an issue when some entries returned by a DNS server were unreachable. If such IP was selected, LDAP kept hanging until user got a 504 Gateway Timeout. With the timeout option applied, the query is retried with different server.

Thanks!

adulau:
@czechnology thanks for the feedback and contribution.

iglocska:
Perfect, thanks a lot for sharing your fix!

PR #4125, opened 2019-02-08
RHEL7 installation fix implicitly SCL-enabled commands

iwitz:
#### What does it do?

The RHEL7 installation procedure currently assumes that the SCL-enabled shell opened in step 2.08 is used throughout the entire installation procedure, which can lead to problems when for example pausing the MISP installation and resuming it later. I changed commands that used SCL packages so that it is not assumed that these SCL packages were enabled prior to the command.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #4128, opened 2019-02-08
[RHEL7] Add systemd unit file in installation procedure

iwitz:
#### What does it do?

The installation procedure used the /etc/rc.local file to start MISP's background workers on startup.
This file contains the following by default : 
```
[...]
# THIS FILE IS ADDED FOR COMPATIBILITY PURPOSES
#
# It is highly advisable to create own systemd services or udev rules
# to run scripts during boot instead of using this file.
[...]
```
This PR removes the use of that file and instead uses a systemd unit to start the workers. This also has the following advantages :

- it makes sure the workers are started after their dependencies
- it lets systemd manage the workers' output
- it allows for an arguably easier management of the workers through the systemd commands

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/docs/INSTALL.rhel7.md
+++ b/docs/INSTALL.rhel7.md
@@ -260,7 +260,7 @@ echo bind-address=127.0.0.1 >> /etc/opt/rh/rh-mariadb102/my.cnf.d/bind-address.c
 systemctl restart rh-mariadb102-mariadb
 ```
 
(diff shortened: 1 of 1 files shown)


SteveClement:
@iwitz thanks for your efforts!

Issue #4134, opened 2019-02-11
Cannot update from web UI RHEL7 git not found


iwitz:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | RedHat 7.6
| OS version (client)       | Any
| PHP version               | 7.1
| MISP version / git hash   |  v2.4.101

### Expected behavior
An ability to update MISP from the web interface in the "diagnostics" page.

### Actual behavior
git is not found by MISP. As such, the following doesn't work : 
* displaying the current branch
* updating MISP through the web interface

However when I run `sh` from the command line then run `git` in the shell, the command is indeed found.
This might have to do with the fact that MISP is ran in an SCL-enabled environment ?

### Steps to reproduce the behavior
Go to the diagnostics page and click on "update MISP"

### Logs, screenshots, configuration dump, ...
After clicking on "update MISP" I have the following :
```
git checkout app/composer.json 2>&1

sh: git: command not found

git pull origin  2>&1

sh: git: command not found
sh: git: command not found

=================================

git submodule update --init --recursive 2>&1

sh: git: command not found
```

iwitz:
OK I just solved it, the bug was due to an environment modification I made to fix some other issues. I'll create a PR that solves that other issue without having the aforementioned undesired side effect. Sorry for the useless issue !

Issue #4084, opened 2019-02-04
[master] RHEL/CentOS 7.x/8.x Install Issues


SteveClement:
[Main install Guide for RHEL/CentOS](https://misp.github.io/MISP/INSTALL.rhel7/)

# Permissions

Revise the permission mess and consolidate in 1 function: 

https://misp.github.io/MISP/INSTALL.rhel7/#5-set-file-permissions
AND
https://misp.github.io/MISP/INSTALL.rhel7/#7-apache-configuration
Section: # Since SELinux is enabled, we need to allow httpd to write to certain directories

- [ ] Fixed

# Diagnostics Page
## php-cli
Cli
Issues reading PHP settings. This could be due to the test script not being readable.

- [x] Fixed

## python lief
lief:… Please install lief, documentation here: https://github.com/lief-project/LIEF

- [x] Fixed

Fix: Manual compilation of lief
Fix: Manual compilation of pyLIEF and add it to the venv path, BUT, it now will work on the CLI but NOT when called via PHP.
Fix: permissions: ```sudo chcon -t httpd_sys_script_exec_t $PATH_TO_MISP/app/files/scripts/lief/build/api/python/lief.so```

##  GnuPG
GnuPG installation and settings…FAIL: Failed to load GnuPG

~~This is most likely due to a bug in GnuPG 2.0.22~~
This is an unknown issue at this moment in time, more details eventually here: https://github.com/MISP/MISP/issues/4796

- [ ] Fixed

ToDo: check what CLI/Python scripts want to get launched withing PHP.

## ZeroMQ
ZeroMQ settings…ZeroMQ script not running.

If the permissions are wrong on ps, it will not work, restore them: ```sudo /sbin/restorecon -v /usr/bin/ps```

- [x] Fixed

ToDo: Works, some times, reproduce non-working states.

# Workers

Can: stop/start individual workers
Cannot: restart all workers after initial install

- [x] Fixed

Potential fix: sudo chcon -t httpd_sys_script_exec_t $PATH_TO_MISP/app/Console/worker/start.sh
ToDo: Check why some workers are ok with this fix, others are not...

# Emailing

Add and test:

```bash
setsebool -P httpd_can_sendmail on
```

- [ ] Fixed

# Redis

```
2019-06-24 17:22:38 Error: [RedisException] Permission denied
Request URL: /servers/serverSettings/diagnostics
Stack Trace:
[...]

PR #4136, opened 2019-02-11
Fix rhel7 diagnostics page

iwitz:
#### What does it do?

Fixes issues with the diagnotics page in RHEL7 and most likely CentOS 7 : 
* #4084 : with this commit the diagnostics page now detects, GPG, the PHP settings and the Python modules' versions properly without causing git not to be detected ( #4134 )
* allows the update through the web interface without permission issues by assigning ownership of /var/www/MISP to apache:apache instead of root:apache like what is done in other installation procedures ( Debian 9's, Ubuntu's... ).

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/docs/INSTALL.rhel7.md
+++ b/docs/INSTALL.rhel7.md
@@ -190,11 +190,7 @@ systemctl restart rh-php71-php-fpm.service
 ## 3.06/ Enable dependencies detection in the diagnostics page
 Add the following content to `/etc/opt/rh/rh-php71/php-fpm.d/www.conf` :
 ```
-env[PATH] = /usr/bin:/opt/rh/rh-php71/root/usr/bin:/opt/rh/rh-php71/root/usr/sbin:/opt/rh/rh-python36/root/usr/bin
-env[LD_LIBRARY_PATH] = /opt/rh/rh-php71/root/usr/lib64:/opt/rh/rh-python36/root/usr/lib64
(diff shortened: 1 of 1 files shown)


iglocska:
Ok, that's been a long time missing, thanks for providing a fix for this!

RichieB2B:
This PR changes the whole working directory to be owned by apache:apache instead of just the directories that need to be written to by MISP. Call me old fashioned, but setting the whole working directory including all the PHP source files writable for the httpd process looks like an unnecessary security risk to me. I know it is needed if you want to be able to do `git pull` directly from MISP but if you're running RHEL chances are high you are in a corporate environment where self-updating applications are not allowed anyway.

iglocska:
Hmm. I wonder what would be best in this case, perhaps describing both as alternatives with the risks described?

iwitz:
I understand your concern, and I think presenting both alternatives as well as the risks involved is a good solution. I'll create a PR with the changes tomorrow if that sounds right to you @RichieB2B

The issue that I had with the previous permissions isn't just that they prevent updates through the web interface from working, but that since f95e0b17c7f3f533b2ddc71753b76c6568a8d941 it is "very strongly recommended" to update MISP through the web interface. So basically the previous permissions were preventing the only supported way of updating.

PR #4137, opened 2019-02-11
fix: Typo in tag ID query

FloatingGhost:
In the REST `/tags/edit/:id` endpoint there was a typo causing the response to always be the tag with the lowest ID

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Woaaaah nice catch thanks

Issue #3528, opened 2018-08-03
MISP update error


ndhambi:
i am trying to update MISP using the webinterface but i get the following  error.

=================================================================================

git pull origin 2.4 2>&1

From https://github.com/MISP/MISP
 * branch            2.4        -> FETCH_HEAD
error: Your local changes to the following files would be overwritten by merge:
	app/composer.json
Please, commit your changes or stash them before you can merge.
Aborting
Updating aece239..ca6a40e

=================================

git submodule init && git submodule update 2>&

==========================================================================
Does anyone know how to solve this problem,

wkleinhenz:
I am seeing this exact same issue
`git pull origin 2.4 2>&1

From https://github.com/MISP/MISP
 * branch            2.4        -> FETCH_HEAD
error: Your local changes to the following files would be overwritten by merge:
	app/composer.json
Please, commit your changes or stash them before you can merge.
Aborting
Updating accf52b..b723395

=================================

git submodule update --init --recursive 2>&1`

wkleinhenz:
I was able to fix this issue by replacing composer.json with the composer.json file found here [https://github.com/MISP/MISP/blob/2.4.93/app/composer.json](url)

iglocska:
`git checkout /var/www/MISP/app/composer.json`

should work too

obert01:
Not exactly related, but since a few days, when updating from the GUI, I obtain the following non-fatal error:

```
git checkout app/composer.json 2>1

error: pathspec 'app/composer.json' did not match any file(s) known to git.

git pull origin 2.4 2>1

error: pathspec 'app/composer.json' did not match any file(s) known to git.
From https://github.com/MISP/MISP
 * branch                2.4        -> FETCH_HEAD
   7d0963bcb..575b29311  2.4        -> origin/2.4
Updating 7d0963bcb..575b29311
Fast-forward
 INSTALL/INSTALL.ubuntu1804.with.webmin.txt | 31 +++++++++++++++++++++---------
 app/View/Events/proposal_event_index.ctp   |  4 ++--
 app/View/ShadowAttributes/index.ctp        |  2 +-
 3 files changed, 25 insertions(+), 12 deletions(-)

```

How may I solve this issue?

SteveClement:
@obert01 this is not an issue.
Unfortunately this looks ugly but your MISP instance is being updated correctly.

Eventually we will fix this.

jimanderles:
I have the same problem, someone who solve the issue??

```
[...]

PR #4147, opened 2019-02-12
fix: working directory in update function

iwitz:
#### What does it do?

The commands in the update function were executed in the folder $MISP_ROOT/app/webroot which lead to the following bugs :
* `error: pathspec 'app/composer.json' did not match any file(s) known to git` When checking out app/composer.json, see #3528
* an error message along the lines of `this command must be executed from the root of the project` when executing `git submodule update`

This commit fixes these issues by making sure all the commands are executed in the project's root folder ( which is found using `git rev-parse --show-toplevel` )

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Model/Server.php
+++ b/app/Model/Server.php
@@ -4315,17 +4315,18 @@ class Server extends AppModel
     public function update($status)
     {
         $final = '';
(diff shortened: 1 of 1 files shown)


iglocska:
That's a neat trick, thanks!

PR #4149, opened 2019-02-13
Update software versions in RHEL7 installation procedure

iwitz:
#### What does it do?

Update PHP 7.1 to PHP 7.2 ; the repositories/package names do not change and I just tested the procedure with PHP 7.2, everything works as well as in 7.1. PHP 7.1 is EOL in December of this year.

Also, I tested the procedure with RHEL 7.6 so I updated the RHEL version too ( from 7.5 ).

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot for all the work!

PR #4155, opened 2019-02-14
RHEL7 installation security warning at step 5

iwitz:
#### What does it do?

Includes a security warning when setting the file permissions as discussed in PR #4136 and fixes a few minor mistakes ( non-sequential step indexes, wrong PHP version / user used to issue a command ).

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Lovely, also thanks for having it in an order where someone just blindly executing everything in the install script without checking what they're doing would end up with the more restrictive option.

PR #4157, opened 2019-02-14
new: Add CORS settings for external integration

### LINKED ISSUE #3996: Access API through browser Javascript (CORS)
If I try to access the MISP API using a bit of javascript served in a webpage, I get the predictable:
`Access to XMLHttpRequest at 'http://127.0.0.1/events/restSearch' from origin 'http://127.0.0.1:8000' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.`

Is there a possibility to enable cross-site access to the API (since access is protected by the `Authentication` header anyway)? Or do I have to make our own passthrough API proxy using PyMISP?

FloatingGhost:
Fixes #3996 

Implements a response for HTTP `OPTIONS` which will return a standard empty response, and if the user so desires, allow CORS requests from one or more specified origins

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/AppController.php
+++ b/app/Controller/AppController.php
@@ -103,6 +103,21 @@ class AppController extends Controller
 
     public function beforeFilter()
     {
+        if (Configure::read('Security.allow_cors')) {
+            // Add CORS headers
+            $this->response->cors($this->request,
+                    explode(',', Configure::read('Security.cors_origins')),
+                    ['*'],
+                    ['Origin', 'Content-Type', 'Authorization', 'Accept']);
+
+            if ($this->request->is('options')) {
+                // Stop here!
+                // CORS only needs the headers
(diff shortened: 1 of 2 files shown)


iglocska:
Thanks a lot!

FloatingGhost:
Hm, seems to not work for some endpoints

`/servers/getVersion` returns with the correct headers attached, however `/events/index` does not

the OPTIONS request comes back with headers

Do headers get changed somewhere else I wonder?

FloatingGhost:
They do indeed

`RestResponseComponent.__sendResponse()`, all previously set headers are wiped and a new response is created

I'll try to fix that tomorrow

PR #4158, opened 2019-02-14
fix: re-add CORS headers on REST Response

FloatingGhost:
Aaand another one really quickly

On REST response, all headers previously set were being purged, this re-adds CORS-related ones

There's likely a neater way of doing this, but it'll do for now

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #4163, opened 2019-02-15
fix: Correct X-Results-Count header

### LINKED ISSUE #4161: [restSearch] X-Result-Count is always limit
### Work environment

| OS version (server)       | Arch
| OS version (client)       | Arch
| PHP version               | 7.3
| MISP version / git hash   | 4b71eced00de

### Expected behavior

On restSearch, the `X-Result-Count` header should be the number of results *before* pagination to allow for clients to know if another page exists or not

### Actual behavior

It it always set to `limit`

```bash
curl \
  --header "Authorization: " \
  --header "Content-type: application/json" \
  --header "Accept: application/json" \
  --data-binary '{"returnFormat": "json", "page": 0, "limit": 10}' \
  -D- \
  -XPOST \
  http://localhost/events/restSearch 

...
HTTP/1.1 200 OK
X-result-count: 10
```

FloatingGhost:
Fixes #4161

#### What does it do?

Changes the `X-Result-Count` header to return the pre-pagination result count

This does add another database query, although will not return the data. Shouldn't add too much to the overall processing time

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/Component/RestResponseComponent.php
+++ b/app/Controller/Component/RestResponseComponent.php
@@ -428,6 +428,7 @@ class RestResponseComponent extends Component
             $headers["Access-Control-Allow-Headers"] =  "Origin, Content-Type, Authorization, Accept";
             $headers["Access-Control-Allow-Methods"] = "*";
             $headers["Access-Control-Allow-Origin"] = explode(',', Configure::read('Security.cors_origins'));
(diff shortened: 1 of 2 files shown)


FloatingGhost:
Scratch this for now it's inaccurate

PR #4165, opened 2019-02-15
[fix] Correct X-Result-Count

### LINKED ISSUE #4161: [restSearch] X-Result-Count is always limit

FloatingGhost:
#### What does it do?

Fixes #4161 

Also fixes my idiocy in #4163 (I forgot to remove the previously existing incrementing count :<)

Anyhow, it sets the `X-Result-Count` header to the pre-paginated result count, to allow for clients to display a "page count" 

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/Component/RestResponseComponent.php
+++ b/app/Controller/Component/RestResponseComponent.php
@@ -428,6 +428,7 @@ class RestResponseComponent extends Component
             $headers["Access-Control-Allow-Headers"] =  "Origin, Content-Type, Authorization, Accept";
             $headers["Access-Control-Allow-Methods"] = "*";
             $headers["Access-Control-Allow-Origin"] = explode(',', Configure::read('Security.cors_origins'));
+            $headers["Access-Control-Expose-Headers"] = ["X-Result-Count"];
         }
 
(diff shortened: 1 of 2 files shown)


iglocska:
Thank you!

FloatingGhost:
Just found out that the attributes controller has the same issue, but it's a bit nested and I can't find a clean way to solve that right now

PR #4168, opened 2019-02-18
[restSearch] Use pre-paginated result count for attributes x-result-count

FloatingGhost:
#### What does it do?

Very similar to #4165

This implements a pre-paginated result count for `/attributes/restSearch` 
#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Model/Attribute.php
+++ b/app/Model/Attribute.php
@@ -2812,7 +2812,7 @@ class Attribute extends AppModel
     //     conditions
     //     order
     //     group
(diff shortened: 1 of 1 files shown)


iglocska:
This looks super heavy for little gain (we basically execute the same query twice / round). The idea behind the result count was to show how many elements will be returned and processed by the export module. To paginate the full data set, just keep asking for a new page until no data is returned.

Issue #3780, opened 2018-10-19
Offer a more suitable filename for Download As... Rest Search


tomking2:
When using the "Download as" feature of an Event, it'd be nice to have a nicer named filename, such as:
`misp.event.<event_id>.<fileformat>`

Taking the "Download as" -> "MISP JSON", this would be formatted:
`misp.event.<event_id>.json`

A few users are noting that it's time consuming to keep renaming the `true.<fileformat>` for each event they download. Possibly could pass the filename down the chain?

iglocska:
Pretty sure this was fixed like a few days ago already, which version are you on?

tomking2:
Oh really @iglocska?
Let me pull the latest and take a look. Seems to have stem from the restSearch change

tomking2:
Doesn't look like it :(.

On the latest @ v2.4.96 (337de1600461c4d8d7df24784f0c88e0640c199f) 

From what I can see, issue is as a result of updating the EventsController.php to remove the filename generated from a restSearch (as MISP XML and MISP JSON "Download as..." use the restSearch now).
https://github.com/MISP/MISP/blob/337de1600461c4d8d7df24784f0c88e0640c199f/app/Controller/EventsController.php#L2999

Missing a 5th parameter which would be the filename, but the code no longer generates a suitable filename to be returned back to the user. Which is fine for a typical restSearch, but when using the "download as" feature on an event level, No filename is returned so it doesn't download the file, instead displays in browser (unless the browser can't handle rendering the content like IE).

Brings about another issue of that the original filename is:
`misp.event.<event_id>.<event_uuid>.<extension>`
EventID can be easily found (passed in the function call), but UUID is not obtained anywhere. To be done cleanly, would have to be returned from the Event model restSearch I guess?

PR #4176, opened 2019-02-19
chg: Patch back in the filename when returning from a REST search.

tomking2:
Work likely required!

#### What does it do?
This patches back in the filename as presented when producing a MISP export when using the "Download As" from the Event level view, e.g.:
https://mymispurl.com/events/view/17 > "Download As" > MISP JSON (as an example)

This should return a nice filename which will be auto-download by a browser with the following name as an example:
`misp.event.17.<uuid>.xml`

However as no filename is passed back to the `viewData` in the existing code, it is simply shown direct in the browser. And when you try to save the response, it always has the filename of `true.<extension>`

All of the background logic to add the filename is already present in the code which adds the `Content-Disposition` header. This PR is solely for identifying a suitable filename to present which adheres to how MISP previously generated the filename.

This will also fix the issue #3780 

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

#### Concerns
There is the following concern as it stands with this code:
- The filename will be given for _all_ restSearch queries. Thus this could affect existing user's integrations and may affect how data is returned, but typically, if you are conducting a restSearch you'd largely ignore (or most clients would ignore) any Content-Disposition header.

### DIFF
--- a/app/Lib/Export/NidsExport.php
+++ b/app/Lib/Export/NidsExport.php
@@ -647,7 +647,7 @@ class NidsExport
 
     public function checkWhitelist($value)
     {
-		if ($this->checkWhitelist) {
+		if ($this->checkWhitelist && is_array($this->whitelist)) {
 	        foreach ($this->whitelist as $wlitem) {
 	            if (preg_match($wlitem, $value)) {
 	                return true;


SteveClement:
Does not work as expected. See screenshot.

Screenshot from 2019-04-08 08-52-52

SteveClement:
@tomking2 any progress on this?

SteveClement:
@tomking2 any interest in fixing this? If not I shall close the relevant PR/Issues.

tomking2:
Hi Steve,

Yes there is still an interest, although feel free to close for the time being until I get a chance to review and improve the code. 

A bit strange however as the code appears to work on my instance, need to figure out why that is the case!

PR #4187, opened 2019-02-21
Fix errors in TagCollections import

RichieB2B:
There is no `TagCollection` controller without the `s`. The redirect to `/TagCollection` was confusing.


adulau:
Thank you!

PR #4194, opened 2019-02-21
Fix #4193 GlobalMenuSeparator not checking Roles

### LINKED ISSUE #4193: GlobalMenu separator role check
# This template is meant for bug reports, if you have a feature request, please be as descriptive as possible and delete the template

There are additional Seperators in the GlobalMenu than needed.
PR incoming

### Work environment
MISP vm

### Buggy behavior on the left, Expected behavior on the right
globalmenubug

Create  a user with org admin rights

gallypette:
#### What does it do?

If it fixes #<4193>`

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/View/Elements/genericElements/GlobalMenu/global_menu_separator.ctp
+++ b/app/View/Elements/genericElements/GlobalMenu/global_menu_separator.ctp
@@ -1,1 +1,8 @@
-<li class="divider"></li>
+<?php
+    /*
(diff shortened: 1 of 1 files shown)


iglocska:
Ha, nice catch. The separator is not a one liner after all :)

PR #4202, opened 2019-02-22
Fixed search for cluster value in /tags/search

obert01:
/tags/search was only looking for cluster synonyms, but not for cluster values. For example, /tags/search/Locky was not returning anything.


iglocska:
Woops, nice catch, thanks!

PR #4203, opened 2019-02-23
Update defaults.json (Phishtank - Exclude through regex)

eCrimeLabs:
Updated Phishtank feed with regex to exclude Phishtank url's
\/^http:\\\/\\\/www.phishtank.com\/i

#### What does it do?

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/files/feed-metadata/defaults.json
+++ b/app/files/feed-metadata/defaults.json
@@ -331,7 +331,7 @@
       "event_id": "5655",
       "publish": false,
       "override_ids": false,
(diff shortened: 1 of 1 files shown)


eCrimeLabs:
Added the following regex to remove attributes being created with phishtank URL
```
\/^http:\\\/\\\/www.phishtank.com\/i
```

adulau:
Is the JSON file parseable? It seems travis died on the json loading when verifying the default feed. Thanks a lot

eCrimeLabs:
Think it is an "\"-escape issue I'll dig into it

eCrimeLabs (commented):
Bugfix of the missing "\"

iglocska:
Thanks a lot!

adulau:
Yeah! Thank you.

PR #4229, opened 2019-02-27
chg: [tools] Various updates to CLI tools

SteveClement:
new: [tools] Experimental tool to upgrade MISP via GitHub
chg: [tools] updated MISP service script (systemd) to be more universal in terms of our Debian dependency


iglocska:
This could potentially cause some issues. Sometimes we can run into a situation when a worker is stuck, but is still alive - this would basically prevent us to restart stuck workers via the ui.

Perhaps something that would make sense is an interesting option - let the script behave as it did before, but have an optional parameter to only start what is currently missing.

iglocska:
Just read through the update script, that's cool AF

PR #4237, opened 2019-02-28
new: Add tests after install

Rafiot:
I'm currently testing that, and I'm unsure if it is the right place to run the tests, but in theory, the goal at the end of that is to discard every VM that doesn't pass the tests.

### DIFF
--- a/INSTALL/INSTALL.debian.sh
+++ b/INSTALL/INSTALL.debian.sh
@@ -827,17 +827,14 @@ checkSudoKeeper () {
 }
 
 runTests () {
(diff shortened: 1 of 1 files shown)


Rafiot:
Confirmed working on a fresh Ubuntu 18.04 install.

iglocska:
Awesome, thanks!

Rafiot:
weeeeel, not sure that should be merged... The script I modified is autogenerated, and I'm not sure how that works. @SteveClement: can you look at it before we forget about it?

iglocska:
oh woops. Sorry.

PR #4252, opened 2019-03-01
Sighting filtering

mokaddem:
- Added support of sighting filtering
- Clicking on sightings count redirect to the event with `sighting only` filter activated

I'm passing the ``sighting data`` to the function ``rearrangeEventForView()``. If there is a cleaner way to do it, please say so.


SteveClement:
I seem to not see the filter?

Or am I looking at the wrong place?

sightings

iglocska:
Wrong place indeed, it should be on the event view itself.

PR #4262, opened 2019-03-04
fix: Events index filter button i18n bug

4ekin:
Small fix to Events index template.
Button "Filter" and "Search field" were without correct i18n usage.
Also added new record in default.pot file

### DIFF
--- a/app/Locale/cze/LC_MESSAGES/default.po
+++ b/app/Locale/cze/LC_MESSAGES/default.po
@@ -6624,6 +6624,10 @@ msgstr ""
 msgid "Quickfilter"
 msgstr ""
 
+#: View/Events/index.ctp:95
+msgid "Enter value to search"
+msgstr ""
+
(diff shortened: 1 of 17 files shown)


iglocska:
Thanks a lot!

SteveClement:
For future reference, the pot files are generated.

Thus this will be overwritten by the next i18n export.
Also default.pot is generated dynamically.

I will do so now.

Thanks for the contribution.

4ekin:
> For future reference, the pot files are generated.
> 
> Thus this will be overwritten by the next i18n export.
> Also default.pot is generated dynamically.
> 
> I will do so now.
> 
> Thanks for the contribution.

Yes, sure, I saw a branch i18n, but it was too old to merge with (last commit in Feb)
So, Steve, I want to make translation to Russian language, what branch should I get?

Thank you for great work.

PR #4265, opened 2019-03-04
Event index: Aria label properties on view/edit/publish/delete links …

obert01:
Event index: Aria label properties on view/edit/publish/delete links for better consistancy accross screen readers and accessibility standards compliance.


iglocska:
Thanks a million!

PR #4266, opened 2019-03-05
Add correct PyMISP_dir to sys.path

RichieB2B:
The clever hack of loading pymisp from the MISP install directory didn't work. Now it does.


iglocska:
Thanks a lot!

PR #4275, opened 2019-03-05
Add workers stop script

RichieB2B:
Since the workers start script in _Advanced mode_ (when `jq` is installed) no longer stops workers, a separate stop script is handy when you absolutely want to restart the workers.


iglocska:
The default behaviour should still stop the workers first for start.sh (at least that was what was discussed, didn't actually test it yet). 

But anyway, this can't hurt, so merged. Thanks! B-)

RichieB2B:
No, in _Advanced mode_ the start script checks if the workers are still running. If they are it let's them run and does nothing.

PR #4299, opened 2019-03-13
Fix checking of incorrect flags when publishing to ZMQ

nikofil:
In some places, before publishing something to ZeroMQ, the wrong flag is checked.

* Check `Plugin.ZeroMQ_event_notifications_enable` instead of `Plugin.ZeroMQ_attribute_notifications_enable` in Event.php
* Check `Plugin.ZeroMQ_audit_notifications_enable` instead of `Plugin.ZeroMQ_user_notifications_enable` in Log.php
* Check `Plugin.ZeroMQ_object_notifications_enable` instead of `Plugin.ZeroMQ_attribute_notifications_enable` in MispObject.php

Signed-off-by: Nikos Filippakis <nikolaos.filippakis@cern.ch>

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Excellent, explains some weird behaviour I was seeing, thanks a lot! :)

PR #4337, opened 2019-03-18
Submodule diagnostic

mokaddem:
Dirty (usage of ``exec``) feature to get the submodule versions.
As usual, suggestions are more than welcome! :)

image

### DIFF
--- a/app/Model/Server.php
+++ b/app/Model/Server.php
@@ -4337,6 +4337,7 @@ class Server extends AppModel
         $status['commit'] = exec('git rev-parse HEAD');
         $status['branch'] = $this->getCurrentBranch();
         $status['latestCommit'] = $this->getLatestGitremote();
+        $status['submodules'] = $this->getSubmodulesGitStatus();
         return $status;
     }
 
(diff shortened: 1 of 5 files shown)


iglocska:
Looks good. Go for it!

app/Model/Server.php
```diff
@@ -4351,6 +4351,57 @@ public function checkoutMain()
         return exec('git checkout ' . $mainBranch);
     }
 
+    public function getSubmodulesGitStatus()
+    {
+        $submodulesNames = array('misp-galaxy', 'misp-taxonomies', 'misp-objects', 'misp-noticelist', 'misp-warninglists');
```
SteveClement, line 4356: What about PyMISP?

SteveClement: Also, as you already have an acceptable submodule list, I would suggest to do the following for module names: 

```bash
git submodule |cut -f 3 -d\ 
```

This will return all the current submodules. (Which you could then also catch if none found etc...)

SteveClement:
Tested working.

PR #4341, opened 2019-03-19
Addressing performance issues for fetching attributes when blocking attributes via proposals

liviuvalsan:
#### What does it do?

This patch is addressing performance issues for fetching attributes when blocking attributes via proposals.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #4342, opened 2019-03-19
Fixing bug when exporting to Bro MISP attributes from events that contain a percentage sign inside the event info

liviuvalsan:
#### What does it do?

This patch is fixing a bug when exporting to Bro MISP attributes from events that contain a percentage sign inside the event info. The event info used to be part of the export rule format, but of course % has a special meaning inside the sprintf rule format.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

Issue #4344, opened 2019-03-20
Att&ck matrix: display heatmap on a galaxy level for all tagged data


iglocska:
Would be great to be able to see: What sort of tactics does attacker A use?

iglocska:
It's just instance wide meta-data, simply fetch all event_tags / attribute_tags and compile the information out of that indiscriminately.

PR #4346, opened 2019-03-20
new: [cluster] Display heatmap on the Att&ck Matrix for all tagged data.

mokaddem:
As #4344  has been marked as ``prio-high``, The implementation was done quickly.
Even tough it does the job, queries can be be improved.


iglocska:
<3<3<3<3<3

PR #4370, opened 2019-03-26
Make the submit button focusable (tabindex) in the matrix view. 

obert01:
Make the submit button focusable (tabindex) in the matrix view. This is necessary  for use with screen readers.


iglocska:
Thanks a lot!

PR #4390, opened 2019-03-27
Add updateUser to default config

### LINKED ISSUE #4389: LDAP config option 'updateUser' is not listed in default config
While debugging some LDAP authentication code and figuring out why user roles were not updating, I ran across the code where there is a the updateUser config option which seems undocumented in the default config.

We should add this to the default configuration to make this more apparent to LDAP authentication users (despite little to no official documentation)

couchuser12345:
#### What does it do?

Add `updateUser` configuration to default config since this option is missing which fixes #4389 

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


SteveClement:
Thanks

PR #4399, opened 2019-03-29
Install python dependencies via Pipfile instead of manually cloning / installing them

sim0nx:

#### What does it do?

As discusses on gitter, it would be really nice if one could install python dependencies via a requirements.txt or Pipfile. I choose Pipfile as you are already using pipenv in travis and for PyMISP so I suppose this is the preferred way.
I used exactly the same sources and versions (any, specific refs, ...) as is defined in git submodules or documentation.

Ideally one would replace git refs by tags or pypi versions where possible.

Using pipenv a user simply does "pipenv sync" to put his virtualenv to the state the developers intended it to be. So this would be a win-win for users and developers :-).

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -6,12 +6,11 @@ verify_ssl = true
 [dev-packages]
 
 [packages]
-pymisp = {git = "https://github.com/MISP/PyMISP.git", ref = "64bcaad0e578129543cdffad532a232722615f6c"}
 mixbox = "*"
 cybox = "*"
 stix = "*"
 maec = "*"
-stix2 = {git = "https://github.com/MISP/cti-python-stix2", ref = "407f346eb8118d57b43035ef0da47e2ff77ed00e"}
-python-magic = "*"
-lief = {file = "https://github.com/lief-project/packages/raw/lief-master-latest/pylief-0.9.0.dev.zip"}
+
+pymisp = {git = "https://github.com/MISP/PyMISP.git", ref = "a68bd80ab9dceaee9674bd9a2b0bffc4f387fcdc", extras=["fileobjects"]}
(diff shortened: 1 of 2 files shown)


sim0nx:
Does this make sense to you ?
Please let me know if it does and what else you expect me to modify.

In case this would get merged, I would remove python related submodules.

Rafiot:
There are a few tricks in the pymisp package that will also automatically get the proper dependencies without having them in the pipfile directly: https://github.com/MISP/PyMISP/blob/master/Pipfile

But besides that, I don't see a problem. 

Removing the submodule all together may not be a good idea as some other users may be using the fact it's there in unexpected ways and they won't like.

sim0nx:
Thanks for the feedback. The Pipfile is not used when installing PyMISP as dependency here, but you have similar stuff in the setup.py.
With tricks you mean extras_require ?

Anything in particular I should include in the Pipfile here for PyMISP?

"fileobjects" would make sense.

BTW PyMIPS do you think it would be possible to tag releases when you make a new MISP release so that the submodule in MISP would point to a tag instead a commit-ref ?

sim0nx:
Please let me know if you require me to adapt anything else. If not, feel free to merge :-)

Rafiot:
Yes, I meant the `fileobjects` switch. Looks good to me, merging, thank you! 

You're making a good point regarding PyMISP and using the tagged releases instead of the commit ref. I'll see if that's possible with the submodules. 

Right now, the main reason we don't do that is because I often add new test cases and I want travis to run on them.

sim0nx:
Thank you for merging!

Regarding PyMISP and testing; maybe one could still lock a release tag in the Pipfile. As you keep the submodule, that won't change much for the testing part you mention.

PR #4410, opened 2019-04-02
Fix i18n in controllers, views, Attribute model

4ekin:
Fixed i18n in:
* BlacklistComponent controller
* PostsController
* UsersController
* WarninglistsController
* Plenty of Views
* Attribute model class

Updated default.pot file with new I18n string.
After merge need to update languages default.po files...

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/View/Jobs/index.ctp
+++ b/app/View/Jobs/index.ctp
@@ -1,8 +1,6 @@
 <div class="jobs index">
     <h2><?php echo __('Jobs');?></h2>
-    <h4>
-        Purge job entries:
-    </h4>
+    <h4><?php echo __('Purge job entries:');?></h4>
     <span>
         <?php
             echo $this->Form->postLink(
@@ -64,25 +62,25 @@
                     'children' => array(
                         array(
                             'url' => '/jobs/index',
-                            'text' => 'All',
+                            'text' => __('All'),
                             'title' => __('Show all queues'),
                             'active' => !$queue
                         ),
                         array(
                             'url' => '/jobs/index/default',
-                            'text' => 'Default',
+                            'text' => __('Default'),
                             'title' => __('Show default queue'),
                             'active' => $queue === 'default'
                         ),
                         array(
                             'url' => '/jobs/index/email',
-                            'text' => 'Email',
+                            'text' => __('Email'),
                             'titles' => __('Show email queue'),
                             'active' => $queue === 'email'
                         ),
                         array(
                             'url' => '/jobs/index/cache',
-                            'text' => 'Cache',
+                            'text' => __('Cache'),
                             'title' => __('Show cache queue'),
                             'active' => $queue === 'cache'
                         )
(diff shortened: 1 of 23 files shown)


app/View/Jobs/index.ctp
```diff
@@ -103,16 +101,16 @@
...
-                <th><?php echo $this->Paginator->sort('job_type');?></th>
+                <th><?php echo $this->Paginator->sort('date_created', __('Date created'));?></th>
+                <th><?php echo $this->Paginator->sort('date_modified', __('Date modified'));?></th>
+                <th><?php echo $this->Paginator->sort('process_id', __('Process ID'));?></th>
+                <th><?php echo $this->Paginator->sort('worker', __('Worker'));?></th>
+                <th><?php echo $this->Paginator->sort('job_type', __('Job type'));?></th>
                 <th><?php echo $this->Paginator->sort('job_input', __('Input'));?></th>
                 <th><?php echo $this->Paginator->sort('message');?></th>
-                <th><?php echo $this->Paginator->sort('Org.name');?></th>
+                <th><?php echo $this->Paginator->sort('Org.name', __('Organisation name'));?></th>
```
SteveClement, line 111: Not sure this is correct, @iglocska could you comment?

app/View/ObjectTemplates/index.ctp
```diff
@@ -51,16 +51,16 @@
...
             ?>
-                    <th><?php echo $this->Paginator->sort('active');?></th>
+                    <th><?php echo $this->Paginator->sort('active', __('Active'));?></th>
             <?php
                 endif;
             ?>
             <th><?php echo $this->Paginator->sort('id');?></th>
             <th><?php echo $this->Paginator->sort('name');?></th>
-            <th><?php echo $this->Paginator->sort('uuid');?></th>
+            <th><?php echo $this->Paginator->sort('uuid', __('UUID'));?></th>
```
SteveClement, line 60: I would refrain from translating "generic" technical terms, a UUID should always be a UUID, no matter the language, well, maybe some asian Languages might use some other names (Kanji) but they will also understand UUID

4ekin: Yes, I agree that UUID should always be a UUID.
This change I made because $this->Paginator->sort('uuid') produce column with name "Uuid" and I think that it will be better "UUID". What do you think? (Also UUID string already placed in default.pot file)

app/View/Jobs/index.ctp
iglocska: It's correct. By default it would "humanize" the name and just display it. Passing a second argument will simply force a display name, so it's cool.

app/View/ObjectTemplates/index.ctp
iglocska: Indeed, another option would be to simply use ('uuid', 'UUID'). Whether we want to translate it - I am not convinced either way and don't think I care enough. All decent translations would probably just translate it to UUID.

iglocska:
Looks great to me!

SteveClement:
Thanks for the contributions.

4ekin:
> Thanks for the contributions.

It's ok, thank you for great work!

PR #4411, opened 2019-04-02
Accessibility: added roles and aria labels for many icons/links

obert01:
This affected the index pages for: servers, tags, users, roles, attributes, warning lists, signatures, regexps, galaxies, templates  ...

There is still much work to be done. But we are on the road.

Ideally, and in order to avoid code dupplicates, we should have some utility functions to generate a proper and accessible yes/no icon, an accessible icon link, etc. This would prevent the code from being filled with "aria-label" tags, since the "title" and "aria-label" properties are identical 99% of the time.

### DIFF
--- a/app/View/Elements/Events/View/row_attribute.ctp
+++ b/app/View/Elements/Events/View/row_attribute.ctp
@@ -357,7 +357,7 @@
       <?php
             endif;
       ?>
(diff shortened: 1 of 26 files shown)


iglocska:
Thanks a lot, excellent stuff!

PR #4416, opened 2019-04-03
Fix blank $PATH_TO_MISP

tom564:
Move  PATH_TO_MISP=${PATH_TO_MISP:-$(locate MISP/app/webroot/index.php|sed 's/\/app\/webroot\/index\.php//')} outside of if statement checking if manual input is required otherwise is only executed if locate is unable to determine path.

#### What does it do?

`#<4415>`

#### Questions

- [ X] Does it require a DB change?
- [ Y] Are you using it in production?
- [X ] Does it require a change in the API (PyMISP for example)?

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


tom564:
Looks like the same changes may be needed for the restore script. can't seem to add any other changes to this PR so will leave it. have updated issue  https://github.com/MISP/MISP/blob/fcca264eb70e17ed3b280200f6f9c50df5700955/tools/misp-backup/misp-restore.sh#L146

SteveClement:
Hehe, indeed, the script worked for me because I kept hitting the if :)

Thanks.

PR #4419, opened 2019-04-04
Be consistent in quoting table names

RichieB2B:
Success isn't always about greatness. It's about consistency. Consistent hard work leads to success. Greatness will come. 

Dwayne Johnson


iglocska:
"Blood, Sweat and Respect. First Two You Give, Last One You Earn."

- Dwayne Johnson

Issue #3786, opened 2018-10-22
REST in case of SSO (shibboleth)


dsever:
Hi,

In case of using shibboleth as SP, I've configured by using instruction https (https://github.com/MISP/MISP/tree/2.4/app/Plugin/ShibbAuth):

       <Directory /var/www/MISP/app/webroot>
                Options -Indexes
                AllowOverride all
                #Require all granted
                <If "-T reqenv('HTTP_AUTHORIZATION')">
                  Require all granted
                  AuthType None
                </If>

                <Else>
                 Require valid-user
                 AuthType shibboleth
                 ShibRequestSetting requiresession On
                 ShibRequestSetting shibexportassertion On
                 ShibUseHeaders On
                </Else>
        </Directory>

Checked if, HTTP_AUTHORIZATION exists, I've tried event hardcode it as:

  SetEnvIf Authorization "(.*)" HTTP_AUTHORIZATION=xi9NdGJz9EmIshvQo7L2Be1o2TOWkZuiwfJ4EaqA

But it looks like it always fires shibboleth even there is Authorization variable in header. 
So does anyone has the same problem, and is there any kind of workaround?

Thanks
Dubravko

dsever:
After few days of debugging looks like rewrite in .htaccess comes to late, and HTTP_AUTHORIZATION is empty.
So workaround would me, check header variables like:
<If "-T req('Authorization')">

Dubravko

pettai:
Hi, could you elaborate on how you solved this? 
Did you patch ./app/webroot/.htaccess or the apache configuration?

dsever:
Sorry but don't remember anymore, it was on my previous job :(

SteveClement:
@dsever do you have the same issue?

pettai:
@SteveClement you mean me @pettai, right? 
Yes, we're also running MISP with Shibb-auth infront, but I published the API on another port as a workaround instead of debugging the issue, so, hence I was curious on howto solve that. (The docs for the Shibb plugin isn't working, probably incomplete)

marvin-sinister:
The fix was to do:
```
                #<If "-T reqenv('HTTP_AUTHORIZATION')">
                <If "-T req('Authorization')">
```

in ```/etc/httpd/conf.d/misp.conf```

The root of the problem is (and i just think so, i couldn't find a definitive answer) that rewrite rule in ```/var/www/MISP/app/webroot/.htaccess```:
```
    # Adds AUTH support to Rest Plugin:
[...]

PR #4470, opened 2019-04-13
Update README.md

pettai:
Makes API Authorization work

#### What does it do?

It fixes an existing issue, #3786


iglocska:
Thanks a lot!

PR #4478, opened 2019-04-15
Make xINSTALL.centos6.md work again

RichieB2B:
I just installed MISP on a clean CentOS 6 and needed to make some changes to make the installation instructions actually work.


SteveClement:
@RichieB2B thanks for that, I will try to balance CentOS 7&6 docs out so we are more or less on the same level.

PR #4484, opened 2019-04-16
Fixed case sensitivity in the tag search API (/tags/search).

obert01:
There was still a problem for matching the search on a cluster name. 

I have also slightly simplified the construction of the SQL request for better code readability.


iglocska:
Indeed, that is much more readable, thanks a lot!

PR #4492, opened 2019-04-17
[event:view] Major performance improvement for large event

mokaddem:
- Due to the introduction of ``tag`` and ``galaxy`` filtering in the event's querybuilder, ``Attributes`` were fetched 2 additional times so that all ``tags``/``galaxies`` could be retrieved, thus, adding a massive overhead when displaying an event (under ``events/view``).

:arrow_right: Now, ``tags``/``galaxies`` are extracted directly from the fetched event (before rearrangement/pagination)

- Improved UI (resizable widget)
- Improved Sanitation
- Fixed a bug with the ``toggleBoolFilter (Deleted flag)`` where it was resetting all the rules.


iglocska:
image

mokaddem:
> image

image

iglocska:
image

PR #4530, opened 2019-04-25
Several improvement on objects manipulation

mokaddem:
## Features
- Update object template to a newer version (core + UI)
- Merge object into a similar one including support in case of template inconsistencies (core + UI)

## Merge suggestion
image

## Merge helper
image


SteveClement:
this seems to work for me.

mokaddem:
Closed due to duplicated PR.

PR #4547, opened 2019-04-30
Proposition to fix bypassing required taxonomy

### LINKED ISSUE #4546: Required flag on taxonomies
Hello,

I added some taxonomies and force the usage of it.
When I create a new event, a message appeared when I'll try to **Publish (no email)**, which is good:
"Could not publish event - no tag for required taxonomies missing: <list taxo>"
But when I'll do the same using **Publish Event**, this alert message does not appeared and the event is published anyway. 

Is it a bug or is it normal?

If it is a bug, I'll have a look tomorrow.

FYI event I don't think it depends on it, I'll run the last version on misp on ubuntu 18.04

mokaddem:
`Publish event` does not bypass required taxonomy. Should fix #4546

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -2488,6 +2488,11 @@ class EventsController extends AppController
         $errors = array();
         // only allow form submit CSRF protection
         if ($this->request->is('post') || $this->request->is('put')) {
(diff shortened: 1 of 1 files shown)


iglocska:
Sorry, didn't see your PR...

PR #4581, opened 2019-05-07
RPZ - Add additional policy actions

pettai:
#### What does it do?
Add the last (missing) policy actions from the RPZ draft.
* PASSTHRU allows for testing without applying changes on the returned answer.
* TCP-only forces the client over to use TCP.

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Cool, thanks a lot!

PR #4584, opened 2019-05-07
Rework modules

chrisr3d:
Rework of the expansion and import module system, allowing them to return MISP attributes & objects as result.

- Including a new view to visualize the module results, using the same style as the event view
- Current system returning freetext not impacted and still running
- This code will not be raised until some [new format modules](https://github.com/MISP/misp-modules/tree/new_module) are merged in the misp-modules master branch
- New objects & attributes (from the module results) added into the event do not overwrite the already existing ones (even when the values are the same)

Some love for a full review is needed


adulau:
FYI, [misp-modules](https://github.com/MISP/misp-modules) is updated with the new sample modules using the new format and the previous modules operate as expected. I'm ready to start tagging misp-modules (as 2.4.110) when this branch is merged in MISP core.

Issue #4594, opened 2019-05-08
RPZ action policy updates


pettai:
RPZ has been around for some time and the draft has been updated a few times.
Current Draft: https://www.ietf.org/archive/id/draft-vixie-dnsop-dns-rpz-00.txt
MISP could be updated a bit to have a better default and to be even more aligned with the  terms in the draft.
  
1) The default policy "DROP" is not no in line with the main goal of RPZ then enforcing a policy: To make the client stubs fail hard so they don't keep on asking all nameservers in their configuration. Therefore, it would make sense to change the current default to NXDOMAIN ("The domain doesn't exist") instead.

2) The action policy "Local Data" is named "walled-garden" in MISP. I'd prefer to have the draft  terms (Local Data) so it matches other reference documentation, like http://www.zytrax.com/books/dns/ch7/rpz.html#policy-actions

adulau:
Thanks a lot for all the contributions! We will soon name you the official RPZ maintainer for MISP. If you see something else, don't hesitate.

PR #4595, opened 2019-05-08
rpz: make NXDOMAIN default

pettai:
Update default action policy from DROP --> NXDOMAIN
It fixes 1. part of an existing issue #4594 

#### Release Type:
- [X] Patch


iglocska:
Thanks a lot!

PR #4600, opened 2019-05-09
rpz: Local-Data

pettai:
Rename action policy "walled-garden" to --> Local-Data (as this is the term used in the draft and other RPZ documentation. (note that the data part input is still named "walled_garden") 
Fixes 2) in issue #4594

#### Release Type:
- [X] Patch


iglocska:
Thanks a lot, following the discussion on gitter - if anyone has a concern regarding this change, please let us know!

PR #4605, opened 2019-05-09
Worker for db update

mokaddem:
- Changed worker place in the update process
- When in cake console, perform the update inline (instead of creating a new worker)
- Deleted 1 useless function


adulau:
A small generic question, what happens if someone is stopping the httpd server where MISP is running while a job is still running and then they restart the worker when the other one doing the update is still running.

mokaddem:
PR no longer valid. See https://github.com/MISP/MISP/pull/5002

PR #4607, opened 2019-05-09
rpz: fix missing rpz policy actions

pettai:
Fix: Add two new policy actions added to RPZexport previously
complements: #4594

#### Release Type:
- [X] Patch


iglocska:
Thanks a lot.

PR #4610, opened 2019-05-09
Base URL fix

1earch:
#### What does it do?

Fixes issue #4535, providing support for MISP to work in a subdirectory:
`https://example.com/misp`

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

#### Some precisions
To perform these modifications, I considered the following:
1. Configure CakePHP to use `baseurl` information if given
2. Thus, don't use `$baseurl` with `redirect`, `Html->link` nor with `Html->postLink` (because managed by CakePHP)
3. Use `$baseurl` in `Form->create` only if the URL isn't defined by an array specifying an action of a controller (I think that when creating it specifying a controller, CakePHP will use its configured base URL)
4. Use `$baseurl` in all manually built HTML links

Maybe some of my hypothesis are wrong.


adulau:
This pull request **introduces 1 alert** when merging a22197929c4b9425c75e2405e58ce87231fd2e12 into e9bc2a40f4c2191322cce9917a34f7f2327f1f5a - [view on LGTM.com](https://lgtm.com/projects/g/MISP/MISP/rev/pr-640affaa20f66daeae1821c0dc2d9d89c8d9d099)

**new alerts:**

* 1 for Missing variable declaration

---

*Comment posted by [LGTM.com](https://lgtm.com)*

bschoenmaeckers:
I would love to see this merged.

SteveClement:
Ping reviewers: @iglocska 
But it is super busy so do not hold anyones breath.

app/Config/bootstrap.default.php
```diff
@@ -107,6 +107,19 @@
 	}
 }
 
+/**
+ * Configure base URL for CakePHP
```
SteveClement: Wouldn't you need to consider "MISP.external_baseurl" too?
/cc @iglocska

iglocska: External baseurl is simply for identifying the instance and has no bearing on the URLs.

SteveClement: Being the never-ending-n00b here, in which scope is this?

Who/where is the identifying done?

iglocska: When you create a sharing group, your instance is included in the list by default. This is implicitly part of the SG. When you synchronise, MISP attaches your own server's identifier - before we used the baseurl, which lead to your own server not being represented for empty baseurls or your server being misrepresented for internal baseurl settings (for example, 192.168.1.50 for your MISP is great to access it internally, sucks for everyone else). This is why you can now set the baseurl to whatever you want to use to reach your MISP but include the externally visible URL via the external_baseurl for sharing group creation.

SteveClement: Ok, I perhaps "understand" my confusion....

Because there is no clear separator (in the table of the MISP configuration explanation) I misread the thing.

UX ftw!

It was squeezed and I merged the tables... (in my brain, I guess)
Which make the explanation incomprehensible. Of course this might only attend to a minority, so no problem.

iglocska: No clear separator? o.O

image

SteveClement: Well if the value is NOT set, you have the Error Message "blending" into the left hand table.

1earch (commented):
**Concerning `$this->here`:**
I initially modified `$this->here` to `Router::url(null, true)` according to [a stack overflow answer](https://stackoverflow.com/questions/6836990/how-to-get-complete-current-url-for-cakephp#comment11184149_6875310). In this commit, I reverted back my changes, to hide Codacy warnings and after testing that `$this->here` effectually contains baseURL.

LetMeR00t:
Hi,
I've checked this PR and it's working great.
Hope it will be merged soon.

zmaktouf:
We are desperately waiting for this to be merged

PR #4629, opened 2019-05-14
Clean cached-exports

RichieB2B:
My training instance was getting quite large because someone enabled cached exports. We need to wipe all the tmp files after a training! :-)


iglocska:
Makes sense, thanks

PR #4638, opened 2019-05-15
fix: On-Demand Action missed side menu fixed

4ekin:
It fixes missed side menu on Admin On-demand action page

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


mokaddem:
Nice catch. Thanks for the fix!

PR #4639, opened 2019-05-15
GRANTs updated to DBUSER_MISP instead of DBNAME.

wasserman:
Two spots were GRANTing to the DBNAME instead of the DBUSER_MISP.  It still works if the user and DBNAME are the same.


iglocska:
Makes sense, thanks!

PR #4647, opened 2019-05-17
Match EDNS packets with snort rules

RichieB2B:
Since DNS Flag Day all DNS answers have the additional OPT RR set. As a result none of the snort rules generated by MISP match anything anymore. This patch fixes that.


adulau:
Thank you!

Issue #4625, opened 2019-05-13
misp-wipe ; starts at organizations id > 2


cudeso:
Why are the 'to delete' IDs of users and organizations in tools/misp-wipe/misp-wipe.sql set to > 3 and >2?

`
DELETE FROM `users` WHERE id > 3;
DELETE FROM `organisations` WHERE id > 2;
`

Seems to work perfectly well when both set to > 1

SteveClement:
Good question, maybe @iglocska knows.

iglocska:
1 will work fine. 0 should work too (it recreates the default org+user on the first login)

SteveClement:
@cudeso feel free to send a PR with a default value that makes sense for the larger community.

Thanks for the contribution.

PR #4649, opened 2019-05-17
misp-wipe ; delete all non-default orgs & users

cudeso:
#### What does it do?

misp-wipe : delete all non-default orgs & users
https://github.com/MISP/MISP/issues/4625

#### Questions

- [ Yes] Are you using it in production? 
- [ No] Does it require a change in the API (PyMISP for example)?

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


SteveClement:
Thanks.

PR #4666, opened 2019-05-22
Add multi-step login to the LinOTPAuth plugin

andir:
Following the merge of the initial LinOTP authentication support (https://github.com/MISP/MISP/commit/78d94870e4481141c11c8597811d6f08997c7592 / #4421 ) I started out implementing a more complete login experience featuring a custom login controller that allows (easier) multi-step logins.

I had to make a couple of changes to the project and separated them into different commits to make them individually simpler to review. Each of them should contain enough comments to grasp the idea of them.

If you see any kind of issues with this please let me know and I'll do my best to resolve them.

#### What does it do?

The initial version only supported logging in via the standard username and password style login. Logging in with an OTP was possible by submitting both the OTP and the password at the same time.

The user experience was not really great that way and a common pattern the user expects is to have multiple steps during the login procedure.

In order to implement a custom login view that supports multiple steps I had to add support of alternative login route. That basically means that the auth module will be re-configured on-the-fly depending on the activated plugins. For that to work I had to touch the AppComponent in a few places. Those changes mostly moved common code to helper functions and in one case added another match condition.

The LinOTPAUth Plugin itself received a few fixes and documentation updates that I initially wanted to get in when it was first merged.

The new controller and the auth class in the plugin are heavily commented and try to explain most of the things that I did run into and what decisions / assumptions were made.

To pass additional data between the Auth class and Controller I had to use custom request parameters that are being added to the request on-the-fly. I am not very fond of the solution but in order to stay with the mostly standard Auth interface it was unfortunately required. There are certainly other ways to do the same but that mostly sounded like bikeshedding to me. If you have some strong opinions and/or alternative idea to implement this please let me know.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


andir:
Not entirely sure how to deal with the `CodeFactor` errors since I reduced the complexity of the `AppComponent::beforeFilter` function and it still complains.. Basically why I started to not add more code there :)

andir:
Rebased because the tests should now pass as they have been fixed on `2.4`.

andir:
Fixed a minor typo that sneaked into this during one of the rebases.

@iglocska Any chance to get a review? :)

andir:
I did some more refinement on the implementation. Mostly removing some code branches and making a few statements a bit simpler.

iglocska:
Sorry for the crazy delays, started testing it, merged the latest changes into a branch called linotp on MISP/MISP. Thanks for all the work, the AppController looks a lot more sane!

andir:
I added 3e83f2eb0a6a0c03d137fd7f154c1b6baa016613 to this branch to silence the annoying (false) login error whenever you try to login. The error seems to originate from a few redirects and the information that you were not logged in is not being accounted for. If MISP would actually know that the user wasn't logged in while trying to access a "restricted" resource it could just avoid that message.

I believe the message should only be shown when a user is logged in and the resource they are trying to access isn't visible for them.

This would probably require a long overdue of much of the authentication / routing system in MISP.

challs:
I've updated the branch `andi:linotp` to reflect the current `MISP:linotp` branch plus the extra notifications fix in 3e83f2e.

challs:
Latest changes from 2.4 merged in and conflict resolved in 8a9a829

challs:
> Latest changes from 2.4 merged in and conflict resolved in [8a9a829](https://github.com/MISP/MISP/commit/8a9a8294a2a0202f2f90e51c58667c4755ac9cb6)

It turned out that the result of my previous merge was broken. I've replaced it with a merge from v2.4.114.

Also added further user feedback to help with getting started

* Warn the user if the plugin is enabled but the curl library is not available or configuration is invalid
* Modify the invalid user/password message to mention that a valid second factor must be available

challs:
@andir this PR is out of date now. I suggest you close it, and I'll open a fresh one when I am ready to submit a new version that works against the latest Misp version.

andir:
Roger.

Issue #4388, opened 2019-03-27
Function to merge same attributes types to a MISP object


Tux-Panik:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Feature Request
| OS version (server)       | CentOS
| OS version (client)       | 7
| MISP version / git hash   | 2.4.104

### Expected behavior
I suggest to have a WebUI function to merge existing attributes (same type) into a MISP Object.
For example, if you export a case from TheHive to MISP, all IOCs will be added in bulk.
To have a function that merges the one you selected to a new object could be useful.

### Actual behavior
Today, to transform several existing attributes to an object, you must:
1. Add a new object to your event,
2. Copy/paste each attribute you want to add to the object,
3. Finally, remove all duplicate attributes

Not sure to be able to add this feature from scratch, but I can contribute obviously.
Thank you.

Regards,
Julien

rommelfs:
Proposal:

Phase 1: Select attributes to be transformed 
Phase 2: MISP proposes matching objects
Phase 3: modifications 
Phase 4: accept (create object and delete attributes)

Tux-Panik:
Feature embedded in the 2.4.109 release.

PR #4672, opened 2019-05-24
Merge attributes into objects

mokaddem:
## TO DO
- Review the code of the valid template collection ``validObjectsFromAttributeTypes`` (Could be improved?)

## Allows to quickly merge multiple Attributes into an Object
- Pick a matching template
- Review before merge
- Perform the merge

mergeAttributes

Solve #4388

## Merging mechanism
- Duplicate the merged attributes
  - originals [O] get deleted flag set or gets hard-deleted if the event has never been published (end of procedure)
  - created [C] have their ``object_relation`` set
- Duplicate the tags attached to the old attribute [O] and attach them to the new ones [C]
- Duplicate sightings of [O] and attach to [C]
- Add a warning regarding references (As the attribute is converted into an object, the meaning of the references might change)


app/Controller/AttributesController.php
```diff
@@ -1497,11 +1426,11 @@ public function deleteSelected($id = false, $hard = false)
         $successes = array();
         foreach ($attributes as $a) {
             if ($hard) {
-                if ($this->__delete($a['Attribute']['id'], true)) {
+                if ($this->Attribute->__delete($a['Attribute']['id'], true)) {
```
iglocska: The parameters for __delete changed with the refactor, requires user to be passed along.

app/Model/Attribute.php
```diff
@@ -3909,6 +3908,74 @@ public function editAttribute($attribute, $eventId, $user, $objectId, $log = fal
         return true;
     }
 
+    public function __delete($id, $user, $hard = false)
```
iglocska: public function, should be called "delete()"
private would be __delete()
protected would be _delete()

app/Model/Attribute.php
```diff
@@ -3909,6 +3908,74 @@ public function editAttribute($attribute, $eventId, $user, $objectId, $log = fal
...
+    {
+        $this->id = $id;
+        if (!$this->exists()) {
+            return false;
+        }
+        $result = $this->find('first', array(
+            'conditions' => array('Attribute.id' => $id),
+            'fields' => array('Attribute.*'),
```
iglocska: 'recursive' => -1,
'contain' => array('Event')

would achieve the same and be cleaner.

app/Model/Attribute.php
iglocska: However, delete on the model level is already defined, so consider a different name.

app/Model/MispObject.php
```diff
@@ -725,4 +725,147 @@ public function removeOrphanedObjects()
...
+                'ObjectTemplate.name',
+                'COUNT(ObjectTemplateElement.type) as type_count'
+            ),
+            'conditions' => array(
+                'ObjectTemplate.active' => true,
+                'ObjectTemplateElement.type' => $attribute_types
+            ),
+            'joins' => array(
+                array(
+                    'table' => $db->fullTableName($this->ObjectTemplate->ObjectTemplateElement),
```
iglocska: Can just replace this with

'table' => 'object_template_elements'

app/Model/MispObject.php
```diff
@@ -725,4 +725,147 @@ public function removeOrphanedObjects()
...
+        if (empty($attributes)) {
+            return array('templates' => array(), 'types' => array());
+        }
+        $attribute_types = array();
+        foreach ($attributes as $i => $attribute) {
+            $attribute_types[$attribute['Attribute']['type']] = 1;
+            $attributes[$i]['Attribute']['object_relation'] = $attribute['Attribute']['type'];
+        }
+        $attribute_types = array_keys($attribute_types);
+        $db = $this->getDataSource();
```
iglocska: No need for this, see later.

app/Controller/AttributesController.php
mokaddem: Nice catch!

app/Model/MispObject.php
mokaddem: Though it was good practice.. ;)

app/Model/Attribute.php
mokaddem: Called it ``deleteAttribute``

iglocska:
Thanks a lot!

PR #4674, opened 2019-05-26
strict typing - snuffleupagus tests

juju4:
#### What does it do?

This add few casting to support php strict as enforced by snuffleupagus global strict mode (https://snuffleupagus.readthedocs.io/features.html) and included in my ansible role (https://github.com/juju4/ansible-MISP/)
Tested in Travis with PyMISP tests https://travis-ci.org/juju4/ansible-MISP/jobs/529771853.
There is also a larger patch for cakephp (https://github.com/juju4/ansible-MISP/blob/master/files/patch-cakephp-snuffleupagus-strict) to be reviewed separately.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/AppController.php
+++ b/app/Controller/AppController.php
@@ -520,7 +520,7 @@ class AppController extends Controller
 
     private function __convertEmailToName($email)
     {
-        $name = explode('@', $email);
+        $name = explode('@', (string)$email);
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks a lot!

PR #4694, opened 2019-05-30
fix: [authkey] Fixed The authkey variable (Viper should work again)

SteveClement:


adulau:
Thanks for the quick fix!

PR #4716, opened 2019-06-03
enable misp-backup where MySQL datastore isn't on localhost

certbe-trey:
The misp-backup script grabs the MYSQL host parameter from database.conf but it wasn't included in the call to mysqldump.


iglocska:
Makes sense, thanks a lot.

PR #4776, opened 2019-06-18
I am (g)root!

PaoloVecchi:
Apparently a freshly installed Ubuntu 18.04.2 is as 1337 as Kali so $EUID == 0 is a lie 

Before I was getting:
Next step: Checking if run as root and misp is present
This script cannot be run as a root

with $EUID -ne 0 I'm getting:
Next step: Checking if run as root and misp is present
id: ‘misp’: no such user
There is NO user called 'misp' create a user 'misp' (y) or continue as root (n)? (y/n) 
y
User misp added, password is: ....
.... and all the rest of the install.

#### What does it do?

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/INSTALL/INSTALL.sh
+++ b/INSTALL/INSTALL.sh
@@ -403,7 +403,7 @@ ask_o () {
 # Check if misp user is present and if run as root
 checkID () {
   debug "Checking if run as root and $MISP_USER is present"
-  if [[ $EUID == 0 ]]; then
+  if [[ $EUID -ne 0 ]]; then
     echo "This script cannot be run as a root"
     exit 1
   elif [[ $(id $MISP_USER >/dev/null; echo $?) -ne 0 ]]; then


SteveClement:
I will need to think about this and test it.

PaoloVecchi:
I'm not a bash expert so I realised, rechecking what I wrote, that I got the suggestion wrong.
Anyway, I've created a new Ubuntu 18.04.2 VM and ran the install script again to make sure there were no leftovers from previous test.

Running the script as it is I still get "This script cannot be run as a root" and it exits but shouldn't it check if the user misp is present and if not carry on and ask to create it?

Creating the user manually then I can launch the install as misp.

iglocska:
Based on @PaoloVecchi's comment, considering this resolved for now, Paolo let us know if you want us to merge it anyway ;)

PR #4807, opened 2019-06-27
Support empty attribute type in IOC import

RichieB2B:
Without this some of the IOC files we receive will not import (php will error out).


iglocska:
Thanks a lot

PR #4836, opened 2019-07-04
New option "internal tag" for tag sharing on internal instance

lucamemini:
Added support for "internally exportable tags" #2010 :)


lucamemini:
Cattura

iglocska:
Argh. What exactly is it doing? I have something that I'll push out in the release tomorrow that might be very similar :'(

iglocska:
image

lucamemini:
Doh!
hqdefault

adulau:
@lucamemini This has been merged in 2.4. 2a059f7b3171e3f3f8b26de6b8e7b5594e4ac3ee 

Let us know what you think about it and if this solves your issue. Thanks a lot for your contribution.

local-tags

lucamemini:
@adulau I will try in next days... thx :-)

lucamemini:
@adulau @iglocska Hi guys,
I tested your patch but this is different from my patch.

My patch works on tag table adding new state "internal" (currently states are only exportable and hidden).
The new tag state driving the synchronization process with remote server. if remote server configuration  had flag "internal" active the tag is pushed, without flag the tag is stripped. 

Thx! :)

iglocska:
A stupid question, how is that different from exportable? :$

iglocska:
OK, I got it I think. Would the new system with local tags work for you if they would not get stripped when pushing to internal instances? (I actually wanted to add that but forgot...)

lucamemini:
Yeppa... look this function:

```
    // since we fetch the event and filter on tags after / server, we need to cull all of the non exportable tags
    private function __removeNonExportableTags($data, $dataType, $server)
    {
        if (!empty($data[$dataType . 'Tag'])) {
            foreach ($data[$dataType . 'Tag'] as $k => $tag) {
                if (($server['Server']['internal'] == 1) && ($tag['Tag']['internal'] == 1)) { // Internal server, Internal Tag => Export
                    unset($tag['org_id']);
                    $data['Tag'][] = $tag['Tag'];
                } else if (($server['Server']['internal'] == 0) && ($tag['Tag']['internal'] != 1) && ($tag['Tag']['exportable'] == 1)) { // External Server, not internal tag,  exportable tag => Export
                    unset($tag['org_id']);
                    $data['Tag'][] = $tag['Tag'];
                } else if (($server['Server']['internal'] == 1) && ($tag['Tag']['exportable'] == 1)) { // Internal Server, Exportable tag => Export
                    unset($tag['org_id']);
                    $data['Tag'][] = $tag['Tag'];
                } else { // Drop tag (tag is not exportable) 
                    unset($data[$dataType . 'Tag'][$k]);
                }
            }
            unset($data[$dataType . 'Tag']);
        }
        return $data;
    }
```

> OK, I got it I think. Would the new system with local tags work for you if they would not get stripped when pushing to internal instances? (I actually wanted to add that but forgot...)

Lol :-)
BTW, yes I have 3 misp server linked with "internal instance" flag active... but and I need use internal tag to exchange additional info with this costellation :-)

RichieB2B:
Indeed: local tags are synced to local servers only. Non-exportable tags are never synced.

lucamemini:
Any update? :)

iglocska:
Sorry, was away on holidays for the past 2 weeks, will be back starting
Monday - expect delays ;)

tom564:
Being able to export local tags to internal MISP servers would be really useful for us as well

lucamemini:
up :*

github-germ:
+1 --> This would help if we move to local tags as per https://github.com/MISP/MISP/issues/5204#issuecomment-537043057

Thanks!

iglocska:
For our records: The plan is to refactor local tags to move from internal to internal instance.

PR #4861, opened 2019-07-12
fix: [utils] Added missing util "sponge" via moreutils

SteveClement:


github-germ:
Should `moreutils` install be included in other OS installs, e.g. RHEL?

PR #4877, opened 2019-07-16
Allow SSL verification to be disabled with config

tom564:
Allow SSL verification to be disabled with config. If I understand this right this will need to be scheduled with a cronjob if the expiration framework is wanted?

#### What does it do?
Allows for SSL verification for misp_retention.py to be configured.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


adulau:
Thank you for the fix.

Indeed, this is a support tool which can be triggered via cronjobs for the organisation willing to use it.

PR #4894, opened 2019-07-17
Make error clearer when canPush bit is missing

RichieB2B:
The canPush permission was introduced in a81894f14ca74a060f4d50713c6cf64f4243f954 a year ago. Today the missing bit actually means you are not allowed to push data.


iglocska:
Thanks!

PR #4902, opened 2019-07-18
fix: Load Galaxy mappings for misp2stix2 seperately from Objects

tomking2:
#### What does it do?

It fixes a newly discovered issue I've been facing where my 'Object' key from a MISP event is empty, but the 'Galaxy' key is not. As a result, the `load_objects_mapping` function was never called, and therefore none of the Galaxies used were being included in the generated STIX2 file.

To resolve, a new function has been defined that loads the Galaxy mapping upon confirming there is data within the 'Galaxy' key of a MISP event.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


chrisr3d (approved):

chrisr3d:
@tomking2 Thank you very much !

PR #4905, opened 2019-07-18
Allow for file download when using restSearch through front-end UI

tomking2:
#### What does it do?

This fixes the ongoing issue #3780 and is an improved update to the closed PR #4176 

Historically you could download files when exporting an event to a particular format. This brings this functionality back, setting the Content-Disposition header.

This places back in the filename as presented when producing a MISP export when using the `Download As...` modal from the Event level view, e.g.:
`https://mymispurl.com/events/view/17 > "Download As..." > MISP JSON (as an example)`

This should return a nice filename which will be auto-download by a browser with the following name as an example:
`misp.event.17.<uuid>.xml`

Currently, the response is shown direct in the browser and the "suggested" filename isn't very helpful.

Each of the downloads that leverage the EventsController (not AttributesController) through the front-end will auto-set the restSearch parameter `asDownload`, and attempt to generate a suitable filename to save the file as.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -3329,13 +3329,14 @@ class EventsController extends AppController
         $timestamp = false,
         $published = false,
         $enforceWarninglist = false,
-        $sgReferenceOnly = false
+        $sgReferenceOnly = false,
+        $asDownload = false
     ) {
         $paramArray = array(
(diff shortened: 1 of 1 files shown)


iglocska:
Ended up implementing it by default, there's no reason not to set the correct filenames in any case.

iglocska:
Thanks for the PR and sorry it took this long!

PR #4917, opened 2019-07-19
Allow pushing sightings with perm_sighting

RichieB2B:
Sightings are synced between MISP servers but only for new events. This patch allows sightings to be pushed  by a user with the perm_sighting permission.

This patch changes several things:

1. Allow using the sync API's with perm_sighting
1. Make filterEventIdsForPush() also return local (locked==0) events
1. Add sightings early when editing an event; updating the event data might fail when the user has no perm_sync
1. Update the timestamp of an event when a sighting is added; this notifies the server the event needs to be pushed

Todo: 
- audit logging for sightings


RichieB2B:
Comments from @iglocska 
> https://github.com/MISP/MISP/pull/4917/files#diff-d8c04cac8f5bb7d0517df5bab8464659R3849
> this might be an issue. Sightings should not update the timestamp of an event you will easily run into race condition issues with sightings being added on a remote where instances stop getting updates for certain events
> Basically, the timestamp should only ever be modified by the owner of the event
> I think we need to rethink how we store sightings and we need a new field for events (sighting_timestamp) or something like that
> Right now we have 2 separate processes running during the sync (in-line one after the other):
> 
>  1. we exchange event data based on the timestamp field (what is newer than what I / the other side has?)
> 2. Then we exchange the proposals based on a separate negotiation
> 
> we should ideally have a 3rd process that goes through all events that have an older sighting_timestamp and negotiates the updates based on that separately

RichieB2B:
This new version adds a new `sighting_timestamp` field for events. It syncs the sightings upon push if this timestamp is newer than the remote one. If a user only has `perm_sighting` and not `perm_sync` only the sightings are pushed, not any other event data.

RichieB2B:
This is a complete refactor of the sightings sync based on https://github.com/MISP/MISP/tree/wip/sighting_push

1. New pullSightings() cycle in server pull
2. New syncSightings() cycle in server push
3. "push on add" when a new sighting is added it is pushed directly
4. New servers property "push_sightings" so no surprises of sightings suddenly being pushed to foreign servers

I hope this is now acceptable to be merged.

PR #4924, opened 2019-07-22
slightly modified logroate config which rotates all logs in MISP/app/…

ag-michael:
…tmp/logs when they reach a 50MB limit, with maximum log size set to 500M. rotation is checked every hour

#### What does it do?

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/INSTALL/misp.logrotate
+++ b/INSTALL/misp.logrotate
@@ -4,7 +4,7 @@
     missingok
     notifempty
     compress
-    hourly
+    daily
     size 50M
     maxsize 500M
     copytruncate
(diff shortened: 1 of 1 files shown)


ag-michael:
slightly modified logroate config which rotates all logs in MISP/app/tmp/logs when they reach a 50MB limit, with maximum log size set to 500M. rotation is checked every hour.

SteveClement:
> slightly modified logroate config which rotates all logs in MISP/app/tmp/logs when they reach a 50MB limit, with maximum log size set to 500M. rotation is checked every hour.

I wold prefer a daily check.
For the rest it looks ok.
Perhaps @iglocska wants to give the final go.

ag-michael:
@SteveClement Daily is fine, I used hourly because if something is filling up logs fast you'd want it rolled before it had a chance to impact disk space usage significantly.

SteveClement:
Cool, I was more coming from the side of having the ease to debug any issues without decompressing random files ;)

PR #4939, opened 2019-07-30
Update MYSQL.sql

8ear:
## Fixed MYSQL.sql scheme file for table `organization`
If you install the MISP scheme and want to do it a second time. Then it hangs always at table `organization`. This PR fix this problem.


8ear:
The Travis CI stopped, but this is not a problem of this PR. :-)

iglocska:
Indeed, thanks a lot!

PR #4949, opened 2019-08-02
chg: More efficient Warninglist::__evalCIDR

JakubOnderka:
## What does it do?

Currently, the foreach loop in `__evalCIDR` function is going trough whole array even when the value is already found. With this change, after first match, the method returns `true` immediately .

We notice PHP timeout in this code area in events with a lot of attributes, so I hope this change can help.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Nice catch, that was stupid indeed, thanks!

PR #4951, opened 2019-08-02
chg: Use faster algorithm for Warninglist::__ipv6InCidr

JakubOnderka:
## What does it do?

Even after applying #4949, events with a lot of IP addresses still timeout during IP network comparison. This pull request change the algorithm for comparing IPv6 network with IPv6 address to algorithm used in Symfony framework, that is roughly two times faster.

## Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Excellent, thanks a lot!

PR #4955, opened 2019-08-04
fix: [UI] Row description in View Warninglists

JakubOnderka:
## What does it do?

Fixes row description in View Warninglists. Originally it looks like this: 

Snímek obrazovky 2019-08-04 v 9 58 41

This fix changes the second description from 'Accepted attribute types' to 'Enabled' and adds space between Yes and (disable).

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot, the copy-pasta monster took another victim.

PR #4965, opened 2019-08-08
chg: [warning-list] Filter CIDR warning list before eval

JakubOnderka:
Hello, 

even after patches from https://github.com/MISP/MISP/pull/4951 and https://github.com/MISP/MISP/pull/4949, CIDR warning lists are still slow.

One of the reason is, that CIDR warning lists are filtered for every IP in an event again and again. This patch changes the logic and list is filtered just once.

It also changes the comparison behavior, because before if CIDR warning list contains IP address without netmask part (for example just `1.1.1.1`), it was filtered out. Now this IP address is considered as `/32` for IPv4 or `/128` for IPv6.

After this change, loading times drop for one event with thousands of IP attributes from 261 seconds to 194 seconds and for another event with hundreds of IP attributes from 13 seconds to 10 seconds.

## Questions

## Release Type:

- [ ] Major
- [ ] Minor
- [x] Patch

### DIFF
--- a/app/Model/Warninglist.php
+++ b/app/Model/Warninglist.php
@@ -257,6 +257,33 @@ class Warninglist extends AppModel
         return $entries;
     }
 
+    private function filterCidrList($inputValues)
+    {
+        $outputValues = [];
+        foreach ($inputValues as $v) {
+            $parts = explode('/', $v, 2);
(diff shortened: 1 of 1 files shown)


iglocska:
Lovely! Thanks a lot!

PR #4969, opened 2019-08-08
Improved the accessibility of the galaxy matrix view.

obert01:
Improved the accessibility of the galaxy matrix view for screen readers. 
The table elements are now focusable, and only a short text is brailled/spoken by default.


iglocska:
Thanks a lot!

PR #4971, opened 2019-08-08
fix: [warning-list] Split value just if type is malware-sample or contains `|` char

JakubOnderka:
This pull request simplifies WarningList::__checkValue:
* It removes checking if value contains `|` chars, because the same do also `explode` method
* Also removes `$type` variable, because now it is not necessary and in previous implementation it was also buggy (look at comparison just with one `=`, so the condition was always true)
* Also removes empty `else` block

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Model/Warninglist.php
+++ b/app/Model/Warninglist.php
@@ -358,7 +358,7 @@ class Warninglist extends AppModel
 
     private function __checkValue($listValues, $value, $type, $listType, $returnVerboseValue = false)
     {
-        if (strpos($type, '|') || $type = 'malware-sample') {
+        if ($type === 'malware-sample' || strpos($type, '|') !== false) {
             $value = explode('|', $value);
         } else {
             $value = array($value);
@@ -382,9 +382,8 @@ class Warninglist extends AppModel
             if (!empty($result)) {
                 if ($returnVerboseValue) {
                     return $value[$component];
-                } else {
                 }
-                    return ($component + 1);
+                return ($component + 1);
             }
         }
         return false;


alecrabbit:
@JakubOnderka I'm sorry to comment here. Could you please look into PR's for your libs )
https://github.com/JakubOnderka/PHP-Console-Highlighter/pull/17
https://github.com/JakubOnderka/PHP-Console-Color/pull/16

iglocska:
The problem with this is that it would also explode text / comment etc attribute types. Not sure if that's a good idea.

JakubOnderka:
@iglocska: Thats true, but I just copied the previous behaviour. Original condition was 
```php
if (strpos($type, '|') || $type = 'malware-sample') {
  $value = explode('|', $value);
```
Thats means that string was split in all cases, because the comparison uses just one `=`, that means it is always true.

But if you think that it is incorrect behaviour, I can change the code to:

```php
if ($type === 'malware-sample' || strpos($type, '|')) {
  $value = explode('|', $value);
} else {
  $value = array($value);
}
```

That means a value will be exploded just for type `malware-sample` or when type contains `|`. Do you think it is better?

@alecrabbit I hope I will check them soon.

alecrabbit:
@JakubOnderka Thank You :)

iglocska:
That would be great and it would be inline with the original behaviour.

Basically:

```
$value = explode('|', $value);
```

This in the PR would trigger on any value that contains a '|', for example text, comment, etc

```
        if (strpos($type, '|') || $type = 'malware-sample') {
            $value = explode('|', $value);
        } else {
            $value = array($value);
        }
```
The above will trigger exclusively on composite type attributes (domain|ip, filename|md5, filename|sh256, malware-sample, etc)

JakubOnderka:
@iglocska The problem is that because of typo in original code (just one `=`, not `==` or `===` in `$type = 'malware-sample'`), the original behaviour is the same with this PR:

```
$ php -a 
Interactive shell

php > function test($type, $value) {
php { if (strpos($type, '|') || $type = 'malware-sample') {
php {         $value = explode('|', $value);
php {     } else {
php {         $value = array($value);
php {     }
php {     return $value;
php { }
php > var_dump(test("malware-sample", "aa|bb"));
array(2) {
  [0] =>
  string(2) "aa"
  [1] =>
  string(2) "bb"
}
php > var_dump(test("any-other-type", "aa|bb"));
array(2) {
  [0] =>
  string(2) "aa"
  [1] =>
  string(2) "bb"
}
```

So in the original code the value is always split to multiple parts when contains pipe (`|`) character.

But I will change my PR to match intended behavior.

iglocska:
Ah ok, completely missed the bug with the =, indeed. That should be == :)

JakubOnderka:
@iglocska Force pushed fixed version

iglocska:
Excellent, thanks a ton!

PR #4972, opened 2019-08-09
chg: modules can now pre-check a checkbox from userConfig

zaphodef:
This PR introduces the possibility for a module to pre-check a checkbox created through `userConfig`.


iglocska:
Sounds good, thanks!

PR #4981, opened 2019-08-12
Replace http with https

StefanKelm:
#### What does it do?

It replaces _http_ with _https_ in `config.default.php`

#### Questions

#### Release Type:
- [ ] Major
- [X] Minor
- [X] Patch


adulau:
Thank you!

PR #4986, opened 2019-08-13
chg: delete an object by its uuid, similar syntax to attribute's dele…

### LINKED ISSUE #4974: Unable to delete event or object by UUID
The same way we can get an event by UUID, it would be nice to delete an event or an object by UUID (it works for attributes).

zaphodef:
Should fix #4974 : add the possibility to delete an object based on its UUID. The syntax is inspired from what we already do for the attributes.


iglocska:
Thanks a lot!

PR #5000, opened 2019-08-14
fix: Fix error messages

### LINKED ISSUE #4913: Inconsistent response 'POST' attributes/edit/{attribute_id} and  attributes/delete/{attribute_id}
When the user try to edit an attribute owned by someone else, it gets a 403 back, with the error message "Invalid attribute." , when it should be "You do not have permission to do that." (as it is for the event edit).

On delete, the message is: "Attribute not found or not authorised."

zaphodef:
Should fix #4913: trying to edit/view/delete an attribute which you can't see will return "Invalid attribute" (avoid bruteforce attacks to guess valid attributes that you can't see), while trying to edit/view/delete an attribute you can see but you don't have permissions to edit will return a bad permission error.
Don't hesitate to double check my changes, especially the change from `$this->Attribute->read()` to directly setting `$this->Attribute->data`.

### DIFF
--- a/app/Controller/AttributesController.php
+++ b/app/Controller/AttributesController.php
@@ -810,7 +810,14 @@ class AttributesController extends AppController
         if (!$this->Attribute->exists()) {
             throw new NotFoundException(__('Invalid attribute'));
         }
-        $this->Attribute->read();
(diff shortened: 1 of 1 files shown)


iglocska:
Long overdue, thanks!

PR #5002, opened 2019-08-14
Revised update process

mokaddem:
## Fix
- Previously, ``database_number`` where incremented if the update process was locked
- Do not increment ``database_number`` if ``flag_stop`` is set and errors occurs
- Added sidebar
- Fixed FA classes

## New
- ``SQLSTATE[42S21]: Column already exists: 1060 Duplicate column name`` and ``SQLSTATE[42000]: Syntax error or access violation: 1061 Duplicate key name`` are not considered as an error anymore
- Changed worker place in the update process
- When in cake console, perform the update inline (instead of creating a new worker)
- Deleted 1 useless function
- Created a new worker named `update` which take care of all update processes
- Added database schema diagnostic in the `diagnostic` page 
- Added *TaskScheduler* library
- Improved ``updateProgress`` page
- Added another lock system that triggers if an update fails multiple time in a row. Therefore, preventing infinite update loop

## Note
Regarding question https://github.com/MISP/MISP/pull/4605#issuecomment-490865669
It should not be an issue as updates are executed only if the lock is not locked.

In the scenario from the question:
- You trigger the update
- The update is delegated to a worker
- You restart/kill the apache-server/worker
    - The ``database_number`` is not incremented
    - The lock is still locked
- No new updates can take place until the lock unlocked after a timer (default timer = 2h; can be changed)
- Eventually, updates are executed again; complete; ``database_number`` gets incremented


app/Model/AppModel.php
```diff
@@ -106,14 +106,15 @@ public function afterSave($created, $options = array())
     // this could become useful in the future
     public function updateMISP($command)
     {
+        $db_update_success = false;
         switch ($command) {
             case '2.4.20':
-                $this->updateDatabase($command);
+                $db_update_success = $this->updateDatabase($command);
```
JakubOnderka: I think variable names should be `camelCase` in MISP.

mokaddem: Hey @JakubOnderka thanks for the feedback (I just saw it..)!
Indeed, ``camelCase`` is widely used in the MISP codebase.
I must admit that I used the ``under_score`` notation because of personal taste but also because the rule is not applied everywhere (there are mixes between ``cameCase`` and ``under_score`` in a lot of places).

Even though (AFAIR) PSR-x does not explicitly specify the notation to be used on variable, I agree that having an homogeneous  code base would be better.

What do you think?

iglocska: Indeed, there's no recommendation besides being consistent. I am not too bothered with it either way, though it would look nicer if we were more consistent on it.

app/Console/Command/AdminShell.php
```diff
@@ -550,4 +525,25 @@ public function purgeFeedEvents()
             }
         }
     }
+
+    public function dumpCurrentDatabaseSchema()
```
iglocska, line 523: Nice idea

app/Model/AppModel.php
```diff
@@ -236,51 +238,41 @@ private function __addNewFeeds($feeds)
...
         if ($this->isUpdateLocked()) {
+            $this->Log->create();
+            $this->Log->save(array(
+                    'org' => 'SYSTEM',
+                    'model' => 'Server',
+                    'model_id' => 0,
+                    'email' => 'SYSTEM',
+                    'action' => 'update_database',
+                    'user_id' => 0,
+                    'title' => __('Issues executing the SQL query for ') . $command,
```
iglocska: really tiny improvement - change this to:
```
 'title' => __('Issues executing the SQL query for %s', $command),
```

That will allow translators to deal with languages with different grammar structures where you would have a different sentence order.

I realise that this has been there from before, but perhaps a good moment to fix it ;)

app/Model/AppModel.php
```diff
@@ -1340,43 +1332,54 @@ public function updateDatabase($command, $useWorker=false)
...
-                    $this->__setUpdateError($i);
-                    $error_count++;
-                    if ($exitOnError) {
-                        $flag_stop = true;
-                        break;
+                        'change' => __('The executed SQL query was: ') . $sql . PHP_EOL . __(' The returned error is: ') . $error_message
+                    );
+                    $this->__setUpdateResMessages($i, __('Issues executing the SQL query for ') . $command . __('. The returned error is: ') . PHP_EOL . $error_message);
+                    $error_duplicate_column = 'SQLSTATE[42S21]: Column already exists: 1060 Duplicate column name';
+                    $error_duplicate_index = 'SQLSTATE[42000]: Syntax error or access violation: 1061 Duplicate key name';
```
iglocska: The only issue I can see with this sort of error handling is that we're tying ourselves even closer to mysql. Perhaps something more generic would be handy? Simple substring search for duplicate key name / duplicate column name?

app/Model/Server.php
```diff
@@ -4255,6 +4255,140 @@ public function redisInfo()
         return $output;
     }
 
+    public function dbSchemaDiagnostic()
```
iglocska, line 4266: Cool stuff.

app/View/Elements/genericElements/IndexTable/headers.ctp
```diff
@@ -31,5 +31,8 @@
             __('Actions')
         );
     }
-    echo $headersHtml;
+    $thead = '<thead>';
```
iglocska, line 34: For next time if possible: Have generic changes like these in a separate commit, makes reviewing easier (don't change it for now though!)

app/View/Elements/genericElements/IndexTable/index_table.ctp
```diff
@@ -24,8 +24,8 @@
     if (!empty($data['html'])) {
         echo sprintf('<p>%s</p>', $data['html']);
     }
-    $skipPagination = isset($data['skip_pagination']) && $data['skip_pagination'];
-    if (!$skipPagination) {
+    $skip_pagination = isset($data['skip_pagination']) ? $data['skip_pagination'] : 0;
```
iglocska: this was fixed in a different PR since then, should also not really be part of this PR.

iglocska (commented):
Looks great, go for it!

app/Model/AppModel.php
mokaddem: Just had a look at [mysql-error-reference](https://dev.mysql.com/doc/refman/8.0/en/server-error-reference.html). Doing a substring search for ``Duplicate column name`` works but ``Duplicate key`` is also present in other *mysql* error messages.
Moreover for *posgresql*, the error message for duplicate column is ``ERROR: column "x" specified more than once``, so a substring might not work.
What I can do is also collect and whitelist errors from *posgresql*.
What do you think?

app/View/Elements/genericElements/IndexTable/headers.ctp
mokaddem: Sure! The problem was that I fixed the indexTable in this branch but in the meanwhile, it was also fixed in 2.4..
Sorry for that ;)

mokaddem:
Switched to camelCase as requested by @JakubOnderka ;)

PR #5006, opened 2019-08-14
fix: Replace not exists MethodNotFoundException with NotFoundException

JakubOnderka:
Hi, this pull request:
* replace `MethodNotFoundException` with `NotFoundException`, because in MISP codebase, the `MethodNotFoundException` is not defined.
* in `app/Model/Attribute.php` it remove format check, because the check is redundant https://github.com/MISP/MISP/blob/cf6bc6f2047284597aa3290f48235f91d2adef3e/app/Model/Attribute.php#L4177-L4181
* in `app/Model/Event.php` move the existence check, because in the code before check is accessed `$event` array

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/LogsController.php
+++ b/app/Controller/LogsController.php
@@ -173,7 +173,7 @@ class LogsController extends AppController
             )
         );
         if (empty($event)) {
-            throw new MethodNotFoundException('Invalid event.');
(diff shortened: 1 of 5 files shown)


iglocska:
Thanks a lot!

PR #5010, opened 2019-08-15
chg: Do not log ForbiddenException by default

JakubOnderka:
## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Good catch, thanks!

PR #5015, opened 2019-08-16
Redis optimisations

JakubOnderka:
This pull requests provides:
* small speed optimisations, because `class_exists('Redis')` is called just when Redis connection is created and not all the time.
* fix bug that prevent warning list cache deletion (when deleting cache `cacheWarninglistEntries` method was missing `$id` variable)
*  small speed optimisation, bacause we don't need to check if key in Redis exists and also if it is not empty – `sCard` returns zero for not exists keys.
* code cleanup, code with `?:` operator is shorter and less error prone 

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Model/AppModel.php
+++ b/app/Model/AppModel.php
@@ -1776,18 +1776,20 @@ class AppModel extends Model
 
     public function setupRedis()
     {
-        if (class_exists('Redis')) {
-            if ($this->__redisConnection) {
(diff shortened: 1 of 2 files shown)


iglocska:
Makes sense, thanks a lot!

PR #5017, opened 2019-08-16
Feed correlations optimisations

JakubOnderka:
## What does it do?

* Remove unused variables that are never and array `$objectKeys`, that is only append, but never read
* `getCompositeTypes` is call just once, not again and again for every attribute in event
* Sources from DB are fetched just when correlation is found
* Cache key prefix string is computed just once

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #5020, opened 2019-08-19
fix: Import modules using 'misp_standard' format can be called via RE…

### LINKED ISSUE #324: API and resolved_misp_format
When calling from the API an import module which uses `'format': 'misp_standard'` ([resolved to `resolved_misp_format` here](https://github.com/MISP/MISP/blob/2.4/app/Controller/EventsController.php#L5453)), we get an internal error. Reproducible with the csvimport module for instance. This is probably because there is no handler for REST requests, [as there is when using `resolved_attributes`](https://github.com/MISP/MISP/blob/2.4/app/Controller/EventsController.php#L5456).

zaphodef:
Should fix MISP/misp-modules/issues/324, though there could be a best RestResponse to return maybe?


iglocska:
Thanks a lot!

PR #5021, opened 2019-08-20
chg: Return the sighting when adding one through REST API

### LINKED ISSUE #4867: POST on sightings/add doesn't returns the sighting
it returns the following: `{'message': 'Sighting added'}`, which is inconsistent with every other call.

zaphodef:
Should fix #4867


iglocska:
Thanks a lot!

PR #5029, opened 2019-08-20
fix: Fix messages when we try to delete an attribute

zaphodef:
Better error messages when deleting an attribute, should fix [this issue in PyMISP](https://github.com/MISP/PyMISP/blob/e95948bcf6a5a14a67a00361a2d96065f143fd8b/pymisp/aping.py#L425). When the user can't see the attribute or the attribute doesn't exist, return "Invalid attribute". When the user can see the attribute but is not allowed to delete it, return "You do not have permission to do that.".

### DIFF
--- a/app/Model/Attribute.php
+++ b/app/Model/Attribute.php
@@ -4033,24 +4033,26 @@ class Attribute extends AppModel
         if (!$this->exists()) {
             return false;
         }
-        $result = $this->find('first', array(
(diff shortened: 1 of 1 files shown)


zaphodef:
(Travis will probably not be happy because of the corresponding workaround in PyMISP's tests)

iglocska:
Thanks a lot!

PR #5031, opened 2019-08-20
chg: Setting the 'Tag' key when editing a tag through API is not mand…

### LINKED ISSUE #4852: POST on tags/edit/{tag_id} inconsistent
Every other update calls accepts a json dict in that format:

```json

{"id": "<id>", ...}

```

Tag requires

```json

{"Tag": {
    "id": "<id>", ...
  }
}

```

It is also impossible to update a tag if the json doesn't contains the ID AND the name of the tag (ID only doesn't work)

zaphodef:
Should fix #4852


iglocska:
Nice, thanks! I have a feeling we're missing this in some other places too ;)

PR #5033, opened 2019-08-21
chg: Show descriptions in import modules

zaphodef:
The description already present in all the import modules was lost, while it could help the user to know what kind of file he's expected to upload.

### DIFF
--- a/app/View/Events/import_module.ctp
+++ b/app/View/Events/import_module.ctp
@@ -3,7 +3,7 @@
     <fieldset>
         <legend><?php echo h(Inflector::humanize($module['name']));?></legend>
         <?php if (isset($module['meta']['description'])) {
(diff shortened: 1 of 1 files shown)


app/View/Events/import_module.ctp
```diff
@@ -2,6 +2,9 @@
 <?php echo $this->Form->create('', array('type' => 'file'));?>
     <fieldset>
         <legend><?php echo h(Inflector::humanize($module['name']));?></legend>
+        <?php if (isset($module['meta']['description'])) {
+            echo '<p>'.$module['meta']['description'].'</p>';
```
iglocska: Please wrap it in a h() ;)

zaphodef: oh right thx!

iglocska:
Thanks a lot!

PR #5036, opened 2019-08-21
chg: /events/contact/{event_id} can now be properly called from API

zaphodef:
- create key 'Event' if not given in user's input
- check message is not empty (not sure if we should throw an Exception or return a RestResponse failure)
- return REST messages if needed


iglocska:
Cool, why not? :) Thanks!

PR #5043, opened 2019-08-22
fix: [ui] Show proper error message for ForbiddenException

### LINKED ISSUE #4973: Unable to toggle IDS flag on pulled event
### Work environment

| Type of issue             | Bug / Support
| OS version (server)       | Debian
| OS version (client)       | 9 (stretch)
| PHP version               | PHP 7.0.33-0+deb9u3
| MISP version / git hash   |  v2.4.112 (bf09668ab4fd80e5c0950954a4aaf43fe2da89d4) 
| Browser                   | n/a

### Expected behavior
Situation:

1. Be organisation A.
2. Have received event from organisation B, with several attributes with the IDS flag set.
3. We want to unset the IDS flag from one or more of the attributes.
4. We click the checkbox.
5. The flag is removed.

### Actual behavior
1. Be organisation A.
2. Have received event from organisation B, with several attributes with the IDS flag set.
3. We want to unset the IDS flag from one or more of the attributes.
4. We click the checkbox.
5. *Receive error message in web interface:* "Something went wrong - the queried function returned an exception. Contact your administrator for further details (the exception has been logged)."

See Actual behaviour.

error.log does not actually contain any errors around this timestamp. None of the files in `$MISP/app/tmp/logs` seem to contain any information relevant to this event's ID. 

Note that this was tried as Org Admin, and Publisher. I did not attempt this as admin user, as I was unsure if this would cause proposals to be sent to organisation B, which is something we do not necessarily want. The user that this was tried as, was able to toggle the IDS flag on the organisation's own events. 

So perhaps this is a user rights error, in which case this could be marked as user support. If I'm at the wrong address at this place, my apologies.

JakubOnderka:
## What does it do?

Fixes #4973

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/webroot/js/misp.js
+++ b/app/webroot/js/misp.js
@@ -1619,10 +1619,14 @@ function getPopup(id, context, target, admin, popupType) {
             $(popupType).html(data);
             openPopup(popupType, false);
         },
-        error:function() {
+        error:function(xhr) {
             $(".loading").hide();
             $("#gray_out").fadeOut();
-            showMessage('fail', 'Something went wrong - the queried function returned an exception. Contact your administrator for further details (the exception has been logged).');
+            if (xhr.status === 403) {
+                showMessage('fail', 'Not allowed.');
+            } else {
+                showMessage('fail', 'Something went wrong - the queried function returned an exception. Contact your administrator for further details (the exception has been logged).');
+            }
         },
         url: url
     });


iglocska:
Neat, thanks!

PR #5044, opened 2019-08-22
fix: [ui] Missing space and dot at export page

JakubOnderka:
## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #5048, opened 2019-08-23
fix: [ui] Link to REST client at Automation page

JakubOnderka:
## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


adulau:
Good catch! Thank you very much.

PR #5050, opened 2019-08-23
fix: [feed] Prevent MITM for feeds that support HTTPS

JakubOnderka:
## What does it do?

Use HTTPS connection to Feeds that support HTTPS to prevent MITM during downloading new info. 

All changed URL was tested that support HTTPS and that they use correct certificate.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


adulau:
Thank you!

Issue #5007, opened 2019-08-15
How to troubleshoot push-SYNC problems?


ITServ-DE:
Ladies and Gentleman,

I'm playing with MISP sync and two servers, and I have a working setup for **pulling** events from one host to the other. But I'd like to reverse the process: I want to **push**. For some reason, I don't get it work. The connection has been setup by creating a sync config on the target host and import of the settings on the source host.

The connectivity test succeeds, I can also explore the events on the remote server, but when I add and publish an event on the source start the push to the target, it does not show up on the target system. Still, pulling works.

During push, I see network traffic on the target host, so the communication is established. But I also see some HTTP 403 (forbidden) on the target host:
```
sourcehost - - [15/Aug/2019:08:45:07 +0200] "GET /servers/getVersion HTTP/1.1" 200 3943 "-" "CakePHP"
sourcehost - - [15/Aug/2019:08:45:07 +0200] "POST /events/filterEventIdsForPush HTTP/1.1" 200 4531 "-" "CakePHP"
sourcehost - - [15/Aug/2019:08:45:07 +0200] "GET /servers/getVersion HTTP/1.1" 200 4226 "-" "CakePHP"
sourcehost - - [15/Aug/2019:08:45:07 +0200] "GET /servers/getVersion HTTP/1.1" 200 4226 "-" "CakePHP"
sourcehost - - [15/Aug/2019:08:45:07 +0200] "GET /servers/getVersion HTTP/1.1" 200 4226 "-" "CakePHP"
sourcehost - - [15/Aug/2019:08:45:07 +0200] "GET /servers/getVersion HTTP/1.1" 200 4226 "-" "CakePHP"
sourcehost - - [15/Aug/2019:08:45:07 +0200] "GET /servers/getVersion HTTP/1.1" 200 4226 "-" "CakePHP"
sourcehost - - [15/Aug/2019:08:45:07 +0200] "GET /servers/getVersion HTTP/1.1" 200 4226 "-" "CakePHP"
sourcehost - - [15/Aug/2019:08:45:07 +0200] "GET /servers/getVersion HTTP/1.1" 200 4226 "-" "CakePHP"
sourcehost - - [15/Aug/2019:08:45:08 +0200] "GET /servers/getVersion HTTP/1.1" 200 4226 "-" "CakePHP"
sourcehost - - [15/Aug/2019:08:45:08 +0200] "POST /events HTTP/1.1" 403 4317 "-" "CakePHP"
[...]

PR #5052, opened 2019-08-25
Better error handling for fetching feeds

JakubOnderka:
## What does it do?

Currently, the error handling for fetching feeds is very bad and leads to error like in issue #5007. Sometimes methods return string as error message, but then return value of that methods are checked by `empty` method so they are considered as valid value. Some methods returns numbers as error code, but again, not properly checked. Some of error states are not checked at all (for example JSON parsing). 

This is the first step to fix that problem by using Exception and proper error handling and error logging when fetching feed manifest. Next steps (fetching feed events and free text feed) will follow in this branch. 

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Console/Command/ServerShell.php
+++ b/app/Console/Command/ServerShell.php
@@ -407,12 +407,26 @@ class ServerShell extends AppShell
         );
         $this->Job->save($data);
         $jobId = $this->Job->id;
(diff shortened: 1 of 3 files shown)


iglocska:
Looks great, thanks a lot!

PR #5053, opened 2019-08-25
fix: [feed] Preview feed event don't have id

JakubOnderka:
## What does it do?

Removes unnecessary notice from log. For preview event, ID is not defined.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #5054, opened 2019-08-25
fix: [UI] Correct class name in View/Elements/Feeds/View/row_attribute.ctp

JakubOnderka:
## What does it do?

HTML code fix.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #5059, opened 2019-08-26
attributeTagContainer wrapping

ghost:
#### What does it do?

This change ensures that the attribute table won't explode in width if
one uses more tags on an attribute.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

ghost:
Thanks for the fast merge.

PR #5062, opened 2019-08-26
fix: Default to 0 when no distribution is specified

zaphodef:
The current behavior conducted to set distribution to -1 in the returned json, and raise an 'Undefined index' notice


iglocska:
Thanks a lot!

PR #5066, opened 2019-08-27
chg: No need to encapsulate data in API request

zaphodef:
More coherent with other API calls (I quickly looked for similar cases, but that's the only "obvious"' one I found).


iglocska:
image

PR #5072, opened 2019-08-27
fix: SQL-error during obtaining dbSpaceUsage

4ekin:
Fixed SQL-error in PostgreSQL for viewing Diagnostics Page
Added check for datasource (MySQL or PG), added PostgreSQL handling (without
reclaimable memory).

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #5073, opened 2019-08-28
fix: [ui] Show proper error message for ForbiddenException again

JakubOnderka:
This was already fixed in https://github.com/MISP/MISP/pull/5043, but then removed in probably by mistake in https://github.com/MISP/MISP/commit/e4708c0b6c9f45578a29104bb9b7b4562b2fe84e :/


iglocska:
The famous "every now and again, atom doesn't update open files with changes on disk coming from a git pull" strikes again. Thanks!

PR #4927, opened 2019-07-23
fix: some PostgreSQL issues

4ekin:
Closes: #3066, #3067
Fixes issues:
- wrong boolean and smallint conversion;
- postgresql table and field naming (field 1_event_id is wrong name for
field for example, so lets escape this with doublequotes);
- postgresql grouping (you cannot select columns without grouping them,
so lets add columns to group by);
- wrong checkbox rendering without keyword "checkbox".

I've checked it on pre-installed MariaDB and PostgreSQL 10.9-0ubuntu0.18.04.1
Conversion from MariaDB to PostgreSQL by [your docs](https://github.com/MISP/MISP/blob/2.4/docs/xINSTALL.debian9-postgresql.md)

#### Questions

- [ ] Does it require a DB change?
- [x] Are you using it in production?
- [ ] Does it require a change in the API (PyMISP for example)?

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

PR #5083, opened 2019-08-29
fix: more issues with PostgreSQL

4ekin:
- fixed error during update Job date_modified field (SQLSTATE[22008]:
Datetime field overflow: 7 ERROR:  date/time field value out of range)
- fixed error during fetching events while updating from feeds (
SQLSTATE[42P01]: Undefined table: 7 ERROR: missing FROM-clause entry for
table events)
- fixed Feed edit view with wrong boolean forms (combobox instead
checkbox)

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Console/Command/AdminShell.php
+++ b/app/Console/Command/AdminShell.php
@@ -54,7 +54,7 @@ class AdminShell extends AppShell
         $this->Job->id = $jobId;
         $result = $this->Server->updateAfterPull($submodule_name, $userId);
         $this->Job->saveField('progress', 100);
-        $this->Job->saveField('date_modified', date("y-m-d H:i:s"));
+        $this->Job->saveField('date_modified', date("Y-m-d H:i:s"));
         if ($result) {
             $this->Job->saveField('message', __('Database updated: ' . $submodule_name));
         } else {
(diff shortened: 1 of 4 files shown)


iglocska:
Looks great and thanks for bringing us up to speed to be usable via PostgreSQL again. Are you using it in production?

4ekin:
> Looks great and thanks for bringing us up to speed to be usable via PostgreSQL again. Are you using it in production?

We have full-production instance on MariaDB, but soon we have to migrate to PG, so I'm trying now to find as much bugs as I can.
It is a big issue with boolean and smallinteger columns (for example I tried to fix it in commit [76d58e6e532598b89148e95c329bb1b04559360c](https://github.com/MISP/MISP/pull/4927/commits/76d58e6e532598b89148e95c329bb1b04559360c)), so I made little fix to cakephp

~app/Lib/cakephp/lib/Cake/Model/Datasource/Database/Postgres.php
```
@@ -957,6 +957,9 @@ class Postgres extends DboSource {
 		if ($column === 'uuid' && is_scalar($data) && $data === '') {
 			return 'NULL';
 		}
+		if ($column == 'smallinteger' && $data == '') {
+			$value = '0';
+		}
 		return $value;
 	}
```
What do you think, can I make PR to cakephp lib with this fix?

PR #5084, opened 2019-08-30
Fix Declaration of RestResponseComponent warning

RichieB2B:
FTFY:
```
Warning (2): Declaration of RestResponseComponent::initialize($controller) should be compatible with Component::initialize(Controller $controller) [APP/Controller/Component/RestResponseComponent.php, line 1645]
```


iglocska:
Woops, thanks

PR #5097, opened 2019-09-03
chg: [feed] Break loop when match is found

JakubOnderka:
## What does it do?

Small optimisation in Feed fetching algo.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Sounds good thanks!

PR #5101, opened 2019-09-04
chg: [user] Use machine readable format for fetching PGP keys

JakubOnderka:
## What does it do?

After this change, fetching PGP keys from key server uses machine readable format, so then it is not necessary to parse HTML file by regular expressions, so that should be more stable. 

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #5103, opened 2019-09-04
fix: [UI] GPG keys are fetched from CIRCL keyserver

JakubOnderka:
## What does it do?

Just small change in field description.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


adulau:
Indeed very good point. As a side note, we will have an alternative way soon to set pgp key server or other kind of server which can deliver keys too (HKP-like). Thank you for the contribution and attention to the details. This is really appreciated.

PR #5106, opened 2019-09-04
chg: Put require PHP version and extension into composer.json

JakubOnderka:
## What does it do?

With this change, `composer` will check during install required PHP version and installed extensions. So users will see, what is missing. 

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Why not, thanks!

Issue #5109, opened 2019-09-05
Error: [ArgumentCountError] Too few arguments to function Feed::__filterEventsIndex()


sp4wnxxx:
Hello everyone,

After upgrading to version "v2.4.114", i can no longer see "_Explore the events remotely_", generating the error  below:

_"2019-09-05 10:05:12 Error: [ArgumentCountError] Too few arguments to function Feed::__filterEventsIndex(), 2 passed in /var/www/MISP/app/Model/Feed.php on line 165 and exactly 3 expected
Request URL: /feeds/previewIndex/1
Stack Trace:
#0 /var/www/MISP/app/Model/Feed.php(165): Feed->__filterEventsIndex(Array, Array)
#1 /var/www/MISP/app/Controller/FeedsController.php(589): Feed->getManifest(Array, Object(HttpSocket))
#2 /var/www/MISP/app/Controller/FeedsController.php(568): FeedsController->__previewIndex(Array, Array)
#3 [internal function]: FeedsController->previewIndex('1')
#4 /var/www/MISP/app/Lib/cakephp/lib/Cake/Controller/Controller.php(499): ReflectionMethod->invokeArgs(Object(FeedsController), Array)
#5 /var/www/MISP/app/Lib/cakephp/lib/Cake/Routing/Dispatcher.php(193): Controller->invokeAction(Object(CakeRequest))
#6 /var/www/MISP/app/Lib/cakephp/lib/Cake/Routing/Dispatcher.php(167): Dispatcher->_invoke(Object(FeedsController), Object(CakeRequest))
#7 /var/www/MISP/app/webroot/index.php(92): Dispatcher->dispatch(Object(CakeRequest), Object(CakeResponse))
#8 {main}"_

Editing the file /var/www/MISP/app/Model/Feed.php, at line 165, displays "$ events = $ this -> __ filterEventsIndex ($ events, $ feed);"

and I think it should be "$ events = $ this -> __ filterEventsIndex ($ events, $ feed, $ HttpSocket);_"

Is this right??

Thanks in advance!

*If you would like to report a bug, please fill the template bellow*

### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | Ubuntu Server 18.04.3 LTS
| OS version (client)       | Windows 10
| PHP version               | 7.2
| MISP version / git hash   |  v2.4.114 (d783e0d03938186cba1e591a218b268cacc671b2)
[...]

PR #5112, opened 2019-09-05
fix: [feed] Fetching event manifest

JakubOnderka:
## What does it do?

Fix bug introduced in #5052, report from #5109

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot

PR #5116, opened 2019-09-06
chg: [user] Remove unused private method

JakubOnderka:
## What does it do?

Just small code cleanup.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Can't find any use for it any longer either, thanks!

PR #5117, opened 2019-09-06
fix: [UI] Fetching GPG keys

JakubOnderka:
## What does it do?

This error was introduced in 600e54051694ca4d479a9e2c82db45fe19a46a6c

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #5120, opened 2019-09-06
Install Crypt_GPG by composer

JakubOnderka:
## What does it do?

Install Crypt_GPG library by composer, so we can remove pear dependency and two Git submodules.

Preserve BC, so if user already have `Crypt_GPG` installed by pear, it will still works.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Indeed, makes a lot of sense, thanks!

PR #5122, opened 2019-09-07
chg: [user] LDAP login

JakubOnderka:
## What does it do?

I tried to implement LDAP user login trough ApacheSecureAuth (in pull request https://github.com/MISP/MISP/pull/5078 and manual in https://github.com/MISP/misp-book/pull/174), but I fail :/ 

There are a lot of caveats, but the biggest I realised after the work was done: Apache uses Basic auth protocol, so user password is send in HTTP header in all requests and also it uses the same HTTP header as MISP for auth keys – so when LDAP is enabled in Apache, auth keys doesn't work anymore.

So I reuse the code for communication with LDAP server and apply it for normal form authentication. That requires more changes, but then it works as normal authentication – users can log out, they see login screen etc.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/INSTALL/MYSQL.sql
+++ b/INSTALL/MYSQL.sql
@@ -1287,10 +1287,12 @@ CREATE TABLE IF NOT EXISTS `users` (
   `force_logout` tinyint(1) NOT NULL DEFAULT 0,
   `date_created` bigint(20),
   `date_modified` bigint(20),
+  `ldap_dn` varchar(255) COLLATE utf8_bin DEFAULT NULL,
   PRIMARY KEY (`id`),
   INDEX `email` (`email`),
   INDEX `org_id` (`org_id`),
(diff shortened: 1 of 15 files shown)


JakubOnderka:
Hi, this is almost done. We are currently using this code in production, but some problems can still happen (for example with different LDAP server or with different setting).

Manual how to configure LDAP authentication is in https://github.com/MISP/misp-book/pull/175. Is somebody interested in testing this change? I can provide my advices when deploying.

RichieB2B:
We are getting ready to use MISP with LDAP authentication. We have set up ApacheSecureAuth but haven't tested it with MISP authkeys yet.

Are you still using this patch yourself? Can it be applied to MISP 2.4.121?

PR #5125, opened 2019-09-09
fix: Require exact monolog version 1.24

### LINKED ISSUE #5085: Monolog breaking workers
### Work environment

| MISP version / git hash   | 2.4.0396cc5cb3725336c6f3d0ab0097fd238ef7a9f9

### Expected behavior
Misp workers should be running.

### Actual behavior
The workers are failing to start. It appears to be because the php code in `/var/www/MISP/app/Vendor/kamisama/php-resque-ex-scheduler/bin/resque-scheduler.php`.
It tries to call `$logger->getInstance()->addInfo` where the logger is `$logger = new MonologInit\MonologInit($LOG_HANDLER, $LOG_HANDLER_TARGET);`.

The new major version of Monolog doesn't have this function anymore. It should be in `/var/www/MISP/app/Vendor/monolog/monolog/src/Monolog/Logger.php`.

Blank install misp using provided installer `INSTALL.sh -c -M`.

/var/www/MISP/app/tmp/logs/resque-worker-error.log
```
Error: Call to undefined method Monolog\Logger::addInfo()
#0 /var/www/MISP/app/Vendor/kamisama/php-resque-ex/lib/Resque/Worker.php(360): Resque_Worker->log('Starting worker...', 200)
#1 /var/www/MISP/app/Vendor/kamisama/php-resque-ex/lib/Resque/Worker.php(194): Resque_Worker->startup()
#2 /var/www/MISP/app/Vendor/kamisama/php-resque-ex/bin/resque(109): Resque_Worker->work('5')
#3 /var/www/MISP/app/Vendor/kamisama/php-resque-ex/bin/resque(100): startWorker('email', 1, Object(MonologInit\MonologInit), '5')
#4 {main}
```

Monolog launched a new major version 3 days ago which appears to be the reason that the workers are failing to run.
https://github.com/Seldaek/monolog/releases

I will try to get more info and see if there is an easy way this could be updated to work with the latest version. I would like if the MISP install pinned to a particular version of these dependencies  to make the install more stable.

I know that this code is from the `app/Vendor` path which is in the gitignore so maybe I should report this elsewhere. If there is somewhere more relevant then I would appreciate being pointed in the right direction.

JakubOnderka:
## What does it do?

Fixes https://github.com/MISP/MISP/issues/5085 by requiring exact version of monolog package.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/composer.json
+++ b/app/composer.json
@@ -3,7 +3,8 @@
     "minimum-stability": "dev",
     "require": {
         "kamisama/cake-resque": "4.1.2",
-        "pear/crypt_gpg": "1.6.3"
+        "pear/crypt_gpg": "1.6.3",
+        "monolog/monolog": "1.24.0"
     },
     "suggest": {
         "elasticsearch/elasticsearch": "For logging to elasticsearch",


iglocska:
Thanks, looks like the upstream PR is being ignored :)

Kamforka:
@iglocska In which release will the hotfix be available?

iglocska:
This one is already merged, if you mean tagged release, the moment we tag again it will be included (should be later today).

Kamforka:
@iglocska Thank you for the update! So it should be there in version 2.4.116, right?

iglocska:
Yep. Just pushed it ;)

PR #5129, opened 2019-09-09
chg: [users] Remove unused method UsersController::arrayCopy

JakubOnderka:
## What does it do?

Just small code cleanup, that removes public method from controller that is never used and even doesn't look like it should be controller action.

Usage search before commit:
```
MISP$ rg arrayCopy
app/Controller/UsersController.php
1331:    public function arrayCopy(array $array)
1336:                $result[$key] = arrayCopy($val);

app/Controller/Component/ACLComponent.php
506:                    'arrayCopy' => array(),
```

Usage search after commit:
```
MISP$ rg -i arrayCopy
MISP$
```

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #5148, opened 2019-09-11
chg: [internal] Simplify installation with composer

JakubOnderka:
## What does it do?

Removed commands from installation script just duplicate the configuration from composer.json and can be removed. That will simplify MISP installation process and in future also updating process.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #5149, opened 2019-09-11
chg: [internal] Update composer.phar to version 1.9.0

JakubOnderka:
## What does it do?

Current version of composer.phar is very old (version 1.2.1 from 2016) and doesn't work with PHP 7.3. This pull request updates composer to version 1.9.0. This version still supports PHP 5.

Another solution is remove composer from repository and let user to install their own version of composer from distribution package.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot, missed this.

PR #5163, opened 2019-09-15
fix: [UI] Allow to skip pagination in IndexTable

JakubOnderka:
## What does it do?

`skip_pagination` parameter is used in 
https://github.com/MISP/MISP/blob/3e3318c4192513d3b1a08b64590654e321b6fe8c/app/View/Elements/healthElements/diagnostics.ctp#L202
but not used in View.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #5169, opened 2019-09-16
fix: [internal] Clear also cake core and model caches

JakubOnderka:
## What does it do?

By default, `Cache::clear` method clears just `default` cache type, but we need to clear also at least `_cake_model_` cache too.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


JakubOnderka:
Tested  and ready for merge. This is import when APCu cache is used, because then model cache is never deleted.

iglocska:
Thanks a lot!

PR #5178, opened 2019-09-17
fix: [internal] Allow to use GPG when it is installed by composer

JakubOnderka:
## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


JakubOnderka:
Replaced by https://github.com/MISP/MISP/pull/5240

PR #5187, opened 2019-09-18
Refactor app controller

challs:
#### What does it do?

These patches refactor some of the code in AppController's beforeFilter method into
smaller pieces, to aid readability and reusability.

This is part of a patch series originally submitted in pull request #4666. I am hoping that will be
easier to review by separating these changes out from the plugin.

#### Questions

- [-] Does it require a DB change? No
- [-] Are you using it in production? In regular testing
- [-] Does it require a change in the API (PyMISP for example)? No

#### Release Type:
- [X] Patch


iglocska:
It indeed makes it much easier to check in several steps, looks very good thanks!

PR #5202, opened 2019-09-23
fix: [UI] GnuPG diagnostic message

JakubOnderka:
## What does it do?

As you can see, error code 4 is returned when sign failed, not encrypt:

https://github.com/MISP/MISP/blob/1ebf403ee88a7c24551da42b4955b57a14910ff4/app/Model/Server.php#L4366-L4370

## Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #5205, opened 2019-09-23
fix: [UI] Error handling for submitPopoverForm function

JakubOnderka:
## What does it do?

This should show more information for user if there is an error when submitting form that uses `submitPopoverForm` JavaScript method.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


JakubOnderka:
Tested and ready for merge.

iglocska:
Makes sense, thanks a lot!

PR #5206, opened 2019-09-23
Optimize MYSQL.sql

erickatwork:
#### What does it do?

* Update `utf8_unicode_ci` -> `utf8_general_ci` as it's the default and it's faster
* Add primary key to `bruteforces` table
* Update `CHARSET=latin1 ` to `CHARSET=utf8`

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Sorry, saw several issues in the PR (auto incrementing IP, etc). We will move to utf8_mb4 in the next major release across the application. Thanks for the PR!

PR #5207, opened 2019-09-23
fix: [internal] Just site admin can force when saving freetext

JakubOnderka:
## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


SteveClement:
@iglocska looks sane, please Ack so we can merge.
And in case of feeling, comment and close ;)

mokaddem:
All good for me.
It doesn't really impact the import behavior as the `processFreeTextData` function already enforces the user to be `site_admin`
https://github.com/MISP/MISP/blob/cf9fcc45ce61cb92169349c1e66ee801fd23427f/app/Model/Event.php#L6017

SteveClement:
Thanks for the change!

PR #5210, opened 2019-09-23
chg: [user] GPG key fetching by server

JakubOnderka:
## What does it do?

* Removes schizophrenia, when key search is made by server, but keys fetch is made by user's browser.
* In future, it will be easier to change key server (is defined on one place).
* In future, it will allow us to check key fingerprint fetched from server.
* Better handlers error states.
* Adds warning about checking fingerprint.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


mokaddem:
Thanks a lot!

PR #5215, opened 2019-09-24
Add support to add local tags to an event using the API.

ancailliau:
#### What does it do?

The current API does not allow users to add local tags to an event. 

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/TagsController.php
+++ b/app/Controller/TagsController.php
@@ -894,7 +894,7 @@ class TagsController extends AppController
         return $object;
     }
 
-    public function attachTagToObject($uuid = false, $tag = false)
+    public function attachTagToObject($uuid = false, $tag = false, $local = false)
     {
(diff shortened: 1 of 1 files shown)


app/Controller/TagsController.php
```diff
@@ -894,7 +894,7 @@ private function __findObjectByUuid($object_uuid, &$type)
         return $object;
     }
 
-    public function attachTagToObject($uuid = false, $tag = false)
+    public function attachTagToObject($uuid = false, $tag = false, $local = 0)
```
JakubOnderka: Why not to use `$local` variable as boolean?

ancailliau: Indeed, I mapped what was in the database. I'll update tomorrow with Boolean values unless someone complain :-)

PR #5217, opened 2019-09-24
new: [internal] New AppModel::logException method

JakubOnderka:
## What does it do?

New `AppModel::logException` that can be used in all models for logging exception with backtraces. 

## Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Sounds good, thanks a lot.

PR #5222, opened 2019-09-25
fix: [UI] Notices margin

JakubOnderka:
## What does it do?

After fix:

Snímek obrazovky 2019-09-25 v 17 03 19

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #5225, opened 2019-09-25
fix: [UI] MISP logo is in center at login page

JakubOnderka:
## What does it do?

Put MISP logo in the middle of login page. Tested in Firefox, but I must also test in Chrome.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


JakubOnderka:
Tested and it looks good.

iglocska:
Thanks a lot!

PR #5226, opened 2019-09-26
chg: [feed] Compute md5 value just once

JakubOnderka:
## What does it do?

Just little speedup of feeds caching.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


adulau:
Thank you. It makes sense.

PR #5227, opened 2019-09-26
new: [internal] Allow to use custom CA

JakubOnderka:
## What does it do?

This pull request introduce new option `MISP.ca_path` that can be used for defining own path to CA file. 

By default, MISP uses CA file from Cake framework (`/var/www/MISP/app/Lib/cakephp/lib/Cake/Config/cacert.pem`), but it is 4 years old so it will not contains new CAs and also will contain CA that are already revoked. And also if you put your own CA into system keystore, you need also to put it also to Cake keystore.

## Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Lib/Tools/SyncTool.php
+++ b/app/Lib/Tools/SyncTool.php
@@ -6,7 +6,6 @@ class SyncTool
     public function setupHttpSocket($server = null)
     {
         $params = array();
-        App::uses('HttpSocket', 'Network/Http');
         if (!empty($server)) {
             if ($server['Server']['cert_file']) {
                 $params['ssl_cafile'] = APP . "files" . DS . "certs" . DS . $server['Server']['id'] . '.pem';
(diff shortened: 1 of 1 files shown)


iglocska:
Weird, my cacert.pem in the cake directory is definitely not that old (it should be less than a year old).

With that said, having an override for this makes sense, so why not. Added the server settings entries and a new validation script for it.

Thanks a lot!

JakubOnderka:
@iglocska: My `cacert.pem`: 

```
##
## Bundle of CA Root Certificates
##
## Certificate data from Mozilla as of: Wed Jan 20 04:12:04 2016
```

Thanks for merging!

iglocska:
Interesting, you are right. The file was updated less than a year ago but didn't check the contents. Hmph.

PR #5230, opened 2019-09-27
fix: [shell] Update updateWarningLists from CLI

JakubOnderka:
## What does it do?

Currently, the command `cake Admin updateWarningLists` updates Galaxy. After this pull, it will really update Warninglists :D

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot, copy pasta strikes again!

PR #5233, opened 2019-09-27
fix: [UI] Remove duplicate condition in footer.ctp

JakubOnderka:
## What does it do?

Just small code cleanup.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #5234, opened 2019-09-27
new: [UI] Download GPG public key from GPG homedir

JakubOnderka:
## What does it do?

Currently the GPG public key is stored in two locations: GPG homedir and `app/webroot/gpg.asc`. So for example if admin update key just in GPG homedir, users will still download old/expired key. 

This pull request introduce new config `MISP.download_gpg_from_homedir` and when this config is set to `true`, public key will be extracted and downloaded from GPG homedir.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/AppController.php
+++ b/app/Controller/AppController.php
@@ -373,7 +373,7 @@ class AppController extends Controller
                 }
             }
         } else {
-            $pre_auth_actions = array('login', 'register');
+            $pre_auth_actions = array('login', 'register', 'getGpgPublicKey');
             if (!empty(Configure::read('Security.email_otp_enabled'))) {
                 $pre_auth_actions[] = 'email_otp';
             }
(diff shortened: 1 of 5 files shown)


adulau (changes requested):
Thanks a lot. It's a good one but I see an issue if you fetch a key with a similar key id (email) in the same key ring. Maybe adding the ability to force a specific fingerprint?

JakubOnderka:
@adulau Thats true, but the logic is the same when signing e-mails during sending: https://github.com/MISP/MISP/blob/2eda47f21835037e9e2671ac8e4ba5124eee1c9d/app/Model/User.php#L909

So I can change the behaviour, but then it must be changed also in User model.

iglocska:
I think the problem is somewhat different - I can see separate keys being generated for the instance (info@myorg.org for example) in an automated fashion and landing on the same host's public keyring. Not sure how much of an issue this really is.

For users it's less of an issue as they're manually added keys by admins/users.

JakubOnderka:
@iglocska @adulau  Sorry, I still don't understand. 

The way is the same as for choosing key for signing e-mails – just by e-mail address. So if there is problem with key choosing for this change (public key download), the same problem is here already for signing e-mails.

PR #5240, opened 2019-09-30
chg: [internal] Refactor e-mail sending 

JakubOnderka:
## What does it do?

This is major refactoring how e-mails are send:
* GPG previously uses Inline PGP. First, at least from my experience, the mail clients didn't show signature properly. And also it is not that secure as MIME – see https://dkg.fifthhorseman.net/notes/inline-pgp-harmful/
* The similar changes is also made for S/MIME signing and encryption.
* E-mail headers (like Subject) are now also signed by GPG.
* With new option `GnuPG.obscure_subject` is possible to also encrypt e-mail subject.
* With new option `MISP.email_reply_to` is possible to specify Reply-To for all send e-mails (for example, when MIPS e-mail address is no-reply@...)
* Public PGP keys are added to e-mail header according to [Autocrypt specification](https://autocrypt.org/level1.html).
* It also adds many new debug messages to log when something bad happen.
* In future it will allow to send HTML e-mails.

But because this is big change and can have many corner cases (for example with different e-mail clients or e-mail format), I need your help. Can somebody that uses GPG or S/MIME encryption test this pull request?  Thanks!

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


Rafiot:
Soo I'm not sure what's going on, but the PR seem to work fine on PHP nightly (8.0), but fails on PHP 7.x, and I re-ran the texts multiple times in travis, they fail for real.

JakubOnderka:
I think the problem will be in this error log message:

```
2020-04-26 13:24:46 Error: Exception during sending e-mail
13292[SendEmailException] GPG signing is enabled, but GPG is not configured.
13293Stack Trace:
13294#0 /home/travis/build/MISP/MISP/app/Model/User.php(738): SendEmail->sendToUser(Array, '[ORGNAME MISP] ...', '===============...', 'A new or modifi...', Array)
13295#1 /home/travis/build/MISP/MISP/app/Model/Event.php(2975): User->sendEmail(Array, '===============...', 'A new or modifi...', '[ORGNAME MISP] ...')
13296#2 /home/travis/build/MISP/MISP/app/Console/Command/EventShell.php(138): Event->sendAlertEmail('71', Array, '0', '19')
13297#3 /home/travis/build/MISP/MISP/app/Console/Command/AppShell.php(35): EventShell->alertemail()
13298#4 /home/travis/build/MISP/MISP/app/Vendor/kamisama/php-resque-ex/lib/Resque/Job.php(199): AppShell->perform()
13299#5 /home/travis/build/MISP/MISP/app/Vendor/kamisama/php-resque-ex/lib/Resque/Worker.php(278): Resque_Job->perform()
13300#6 /home/travis/build/MISP/MISP/app/Vendor/kamisama/php-resque-ex/lib/Resque/Worker.php(241): Resque_Worker->perform(Object(Resque_Job))
13301#7 /home/travis/build/MISP/MISP/app/Vendor/kamisama/php-resque-ex/bin/resque(109): Resque_Worker->work('5')
13302#8 /home/travis/build/MISP/MISP/app/Vendor/kamisama/php-resque-ex/bin/resque(100): startWorker('email', 1, Object(MonologInit\MonologInit), '5')
13303#9 {main}
```

I need to figure out why is this happening.

JakubOnderka:
So this problem was fixed, the test still failing, but the problem is now just in different log message format (this PR removes double space in message).

PR #5247, opened 2019-10-01
Fixes #5244.

### LINKED ISSUE #5244: Unable to add tag to attributes within a MISP object
### Work environment

| OS version (server)       | ubuntu
| OS version (client)       | 18.04
| PHP version               | 7.2.19
| MISP version / git hash   | 2.4.116, a0ff8ba8e50227759c1fe0ca47be8a7cc2eb5fdc
| Browser                   | Chrome

### Expected behavior
When adding a global or local tag to an attribute contained within a MISP object, using the tag selector, the tag should be applied to the attribute.

### Actual behavior
When attempting to add a tag to an attribute contained within a MISP object, an error is received stating "Could not add tag".

It is still possible to remove tags already on the attributes (ability to add tags seems to have dropped off since v2.4.115). It is also possible to add tags on the attribute when using the REST API client. As such, appears to be a GUI specific issue.

1. Create (or use existing) MISP object with associated attributes
2. Try adding a tag using tag picker
3. Receive error "Could not add tag".

ancailliau:
AttributeController->addTag was searching for attribute id without flattening.

#### What does it do?

If it fixes an existing issue, see #5244.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/AttributesController.php
+++ b/app/Controller/AttributesController.php
@@ -3067,6 +3067,7 @@ class AttributesController extends AppController
                     $this->Auth->user(),
                     array(
                         'conditions' => array('Attribute.id' => $id, 'Attribute.deleted' => 0),
+                        'flatten' => 1,
                         'contain' => array('Event.orgc_id')
                     )
                 );


iglocska:
Thanks a lot!

PR #5249, opened 2019-10-01
fix some typo

garanews:
fix some typo


iglocska:
Ah, those grammar fixes are soothing. Could you however exclude the .po files from the PR? Those have to be edited in an external tool or else we'll incur the wrath of @SteveClement ;)

PR #5251, opened 2019-10-02
fix: undefined variable when no feed was enable

zaphodef:
Just set the variable in case no feed is enable and we don't enter the loop.


iglocska:
Thanks a lot!

PR #5254, opened 2019-10-02
chg: [UI] Collapse S/MIME or GPG key

JakubOnderka:
## What does it do?

Nicer view of GPG or S/MIME keys in UI.

After change:

Snímek obrazovky 2019-10-02 v 19 11 03

Note: The term 'Podrobnosti' is provided by browser according to browser language. 

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/View/Communities/view.ctp
+++ b/app/View/Communities/view.ctp
@@ -14,12 +14,19 @@
             )
         );
         $optional_fields = array(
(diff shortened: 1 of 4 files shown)


iglocska:
And using a new field template for it, lovely, thanks!

PR #5255, opened 2019-10-02
fix: [internal] Remove already removed git modules

JakubOnderka:
## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


JakubOnderka:
@iglocska Hi, this should be merged without any problem.

iglocska:
Thanks a lot!

PR #5262, opened 2019-10-03
fix: [UI] Do not show Advanced extraction when not possible

JakubOnderka:
## What does it do?

Show advanced extraction checkbox just when Malware checkbox is also checked for attachment upload.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Makes sense, thanks.

PR #5263, opened 2019-10-03
chg: [sync] Code cleanup

JakubOnderka:
## What does it do?

* Correct checking push technique value.
* Throw exception if technique value is invalid instead of calling not exists function.
* Simplified loop for checking server sharing group.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #5269, opened 2019-10-04
fix: [AttributesController] Fix issues related to undefined variables

zaphodef:
fixes point 1 of #5145 (`/attributes/addTag`)


iglocska:
Looks good, thanks!

PR #5272, opened 2019-10-05
chg: [internal] Refactoring malware handling

JakubOnderka:
## What does it do?

* Provides much better error messages when dealing with malware upload.
* Support creating ZIP archives with `zip` extension (with fallback to `zip` binary)
* Advanced extraction checkbox is disabled when is not available.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/View/Attributes/add_attachment.ctp
+++ b/app/View/Attributes/add_attachment.ctp
@@ -68,8 +68,10 @@
             echo $this->Form->input('advanced', array(
                     'type' => 'checkbox',
                     'checked' => false,
+                    'disabled' => !$advancedExtractionAvailable,
+                    'data-disabled-reason' => !$advancedExtractionAvailable ? __('Advanced extraction is not installed') : '',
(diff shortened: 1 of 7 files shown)


app/View/Attributes/add_attachment.ctp
```diff
@@ -68,8 +68,10 @@
             echo $this->Form->input('advanced', array(
                     'type' => 'checkbox',
                     'checked' => false,
```
cvandeplas, line 70: Please set `checked` to `true` and enable by default, most people want this feature all the time.
It's a comparable (default) feature to when all the hashes were taken automatically.

iglocska: added this

iglocska:
Sorry for waiting this long with the merge, absolutely fantastic PR, loving the move of all malware encryption related stuff to its own library. Kudos!

PR #5275, opened 2019-10-06
fix: [iternal] Remove unused method EventShell::__recursiveEcho

JakubOnderka:
## What does it do?

Code cleanup.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Makes sense, this should now be part of the json/xml exports. Thanks!

PR #5278, opened 2019-10-07
Value field template

JakubOnderka:
## What does it do?

* Remove two unused files
* Change `value_field.ctp` template, so adding new features will be easier in future 

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #5280, opened 2019-10-08
Server shell: use the right array key

vpiserchia:
#### What does it do?

It fixes the ServerShell module using the correct array keys

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


vpiserchia:
any update on this?
ty

iglocska:
Nice catch, thanks!

PR #5285, opened 2019-10-09
fix: check if variables are defined

zaphodef:
I wonder if it's actually worth it to create a function in ToolboxComponent.php for the repeated 3 lines.


iglocska:
Thanks, looks good to me. Don't think we really need to move it anywhere for this.

PR #5288, opened 2019-10-09
fix: [internal] Remove closing PHP tags

JakubOnderka:
## What does it do?

Removes closing PHP tag `?>` from end of files. It is bad practice and it can lead to empty lines in output.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Sounds good, thanks!

PR #5289, opened 2019-10-09
chg: [UI] Resizing images

JakubOnderka:
## What does it do?

This patch changes way how images are handled:
* When thumbnail is not requested, we just pass the image content without recoding.
* Thumbnails keep aspect ratio of original image.
* Thumbnails support transparent background.
* Thumbnails has higher quality.

The disadvantage is that thumbnails are in 32bit PNG format, so they can be little bit larger than in JPG format.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Model/Attribute.php
+++ b/app/Model/Attribute.php
@@ -1727,60 +1727,89 @@ class Attribute extends AppModel
         return $this->saveAttachment($attribute);
     }
 
-    public function getPictureData($attribute, $thumbnail=false, $width=200, $height=200)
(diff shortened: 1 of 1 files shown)


mokaddem:
Much better that way thanks!
I screw up the merge but the code is in 2.4 (https://github.com/MISP/MISP/commit/6f212dd97cfdd4943828f8133fa4b304aa74c305). Sorry :/

PR #5294, opened 2019-10-10
chg: [internal] New Attribute::getAttachmentFile method

JakubOnderka:
## What does it do?

This new method simplifies `AttributesController::__downloadAttachment` method and it is ready for next function that will come.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


JakubOnderka:
Tested and ready for merge.

PR #5295, opened 2019-10-10
chg: [internal] Much better error handling for server preview

JakubOnderka:
## What does it do?

Now implemented just for server preview to check if everything works as I expected, more will follow

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


mokaddem:
Thanks a lot!
I just changed error messages into translatable strings. :)

PR #5296, opened 2019-10-11
Fix RHEL 7 / CentOS 7 docs

RichieB2B:
I did a fresh install on RHEL 7.7 and noticed the documentation was lacking in various ways. This should bring it back to a complete working set of instructions.


adulau:
Great! Thanks a lot, many RHEL/CentOS 7 users will be happy.

PR #5302, opened 2019-10-12
Show thumbnails for list and search attributes view

JakubOnderka:
## What does it do?

* Shows thumbnails for list and search attributes view.
* Simplifies logic in `viewPicture`, so loading pictures can be little bit faster.
* Checks, if picture is real picture (according to extension).

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


mokaddem:
Thanks a lot!

PR #5304, opened 2019-10-13
fix: [internal] Load MISP version just once in AppController

JakubOnderka:
## What does it do?

Small code cleanup and small speedup.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/AppController.php
+++ b/app/Controller/AppController.php
@@ -164,10 +164,10 @@ class AppController extends Controller
         } else {
             $this->Auth->authenticate['Form']['userFields'] = $auth_user_fields;
         }
-        $versionArray = $this->{$this->modelClass}->checkMISPVersion();
         if (!empty($this->params['named']['disable_background_processing'])) {
             Configure::write('MISP.background_jobs', 0);
         }
+        $versionArray = $this->{$this->modelClass}->checkMISPVersion();
         $this->mispVersion = implode('.', array_values($versionArray));
         $this->Security->blackHoleCallback = 'blackHole';
         $this->_setupBaseurl();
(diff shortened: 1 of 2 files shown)


iglocska:
Thanks a lot!

Rafiot:
This PR broke freetext import, please make a new one is needed.

JakubOnderka:
@Rafiot Jaj, can you please provide more information? At least error message from log?

iglocska:
This one was a nasty one, totally not your fault. We had a long standing bug that was just revealed by your PR ;) 

Quick explanation:

We've had an erroneus command in the event model:

https://github.com/MISP/MISP/commit/1b13734483ec3e3044b47e01e65f682ce013acda

What happened: We have a setting in MISP that is super seldomly used, `disable_background_jobs` that can be passed via URL parameters to temporarily disable background processing for a single request. It's almost never a good idea to use it, but for the travis tests it allows us to have the execution of the tests and their assertions happen in a linear fashion with no breaks in between.

You moving the version check a few lines further down had a weird effect:

the version check itself is in AppModel, but in order to call the function, cakePHP needs to lazy-load the model. This all of the sudden happens after the on-demand disabling of the background processing via a Configure::write() command. Lazy loading the Event model triggered the renewed load of the configuration, leading to the purging of the on-demand setting. 

I was puzzled as hell when doing this during the testing:

```
debug(Configure::read('MISP.background_jobs'));
$this->Event;
debug(Configure::read('MISP.background_jobs'));
```

and receiving different results in the two debug calls. With that resolved, I've reverted the git revert and your PR is merged back again, sorry for the confusion :)

RichieB2B:
Hah! I remember introducing this "fix" with 886d36438ef7a0653123cb68f3f6bf79eff17d7e in 2015. Puzzled me as well why it was needed. Glad to see if finally resolved!

iglocska:
Yeah we should be good without it. Tried via CLI command too to see if we can run into a situation where the settings are not accessible but couldn't find anything.

iglocska:
2015 would explain it though, probably before we've had the current settings system ;)

PR #5305, opened 2019-10-13
chg: [internal] Speed up of loading attributes and events

JakubOnderka:
## What does it do?

This patch changes the way how Galaxy Cluster tags are remove from attributes and events. In previous behaviour, all GalaxyCluster tags was loaded from DB (thousands now) and then compared according to tag name. 

Now, because Galaxies are already part of attribute or event data, just tags that are part of that already loaded galaxies are removed.

### Speedup

(all values are in ms)

| Action.         | Before | After |
|-----------------|--------|-------|
| List attributes | 5319   | 4453  |
| View event      | 3128   | 2150  |
| View all attrs. | 2185   | 1380  |

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/AttributesController.php
+++ b/app/Controller/AttributesController.php
@@ -1775,25 +1775,21 @@ class AttributesController extends AppController
     private function __searchUI($attributes)
     {
         $sightingsData = array();
-        $sgids = $this->Attribute->Event->cacheSgids($this->Auth->user(), true);
         $this->Feed = ClassRegistry::init('Feed');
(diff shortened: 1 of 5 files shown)


JakubOnderka:
It looks like this change broke the test, I need to check whats wrong.

JakubOnderka:
> It looks like this change broke the test, I need to check whats wrong.

Aaa, the problem is that ACL checker consider private methods as controller action. So I move the method to Model and now test pass.

PR #5306, opened 2019-10-13
fix: [UI] Show image attachment for previewing event

JakubOnderka:
## What does it do?

Show image attachment for previewing event. Previously, it just shows "loading" image.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #5307, opened 2019-10-13
fix: [UI] Remove unused code from Communities view template

JakubOnderka:
## What does it do?

This code was probably copied from Organisation view template, but doesn't make sense for Communities.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Copy pasta, thanks!

PR #5308, opened 2019-10-13
fix: [UI] Removed unused JS from galaxy view

JakubOnderka:
## What does it do?

In the code, there is no:
* HTML element with class `delete-cluster` and
* JS method `removeTag`.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Nice catch of some copy-pasta, thanks!

PR #5309, opened 2019-10-13
fix: [UI] Input is not pair element

JakubOnderka:
## What does it do?

Code cleanup: HTML input element is not pair.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Woops, nice catch!

Issue #5172, opened 2019-09-17
Race condition in checkIfLoggedIn causes relogin


RichieB2B:
9f8e22e96ce27a472e8f8372b82fef377238ff44 introduced a redirect to `/users/login` when the jQuery get of `/users/checkIfLoggedIn.json` fails. This causes a race condition when a user clicks on a link in MISP at the same time the jQuery get call is made. What happens is:

1. jQuery call to `/users/checkIfLoggedIn.json` is made
1. User clicks on link in MISP at nearly the same time
1. `$.get()` is cancelled and `.fail()` is called
1. Browser gets redirected to `/users/login` instead of MISP link that was clicked on

This is driving our users nuts as it randomly cancels whatever they were doing. It is especially nasty when in the middle on an event edit. 

Is the .fail() redirect really needed to fix #4678 ? Setting `MISP.disable_auto_logout` also fixed that issue.

The logs below show the race condition. This user uses SSL client certificate authentication so a redir to `/users/login` simply redirects to `/users/routeafterlogin` and `/` in the end.

```
10.0.0.1 - - [17/Sep/2019:09:33:52 +0200] "GET /events/checkLocks/28003 HTTP/1.1" 200 - "https://misp.local/events/view/28003" "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:60.0) Gecko/20100101 Firefox/60.0"
10.0.0.1 - - [17/Sep/2019:09:33:53 +0200] "GET /users/checkIfLoggedIn.json HTTP/1.1" 200 2 "https://misp.local/events/view/28003" "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:60.0) Gecko/20100101 Firefox/60.0"
10.0.0.1 - - [17/Sep/2019:09:33:57 +0200] "GET /events/checkLocks/28003 HTTP/1.1" 200 - "https://misp.local/events/view/28003" "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:60.0) Gecko/20100101 Firefox/60.0"
10.0.0.1 - - [17/Sep/2019:09:33:58 +0200] "GET /events/edit/28003 HTTP/1.1" 200 20432 "https://misp.local/events/view/28003" "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:60.0) Gecko/20100101 Firefox/60.0"
[...]

PR #5311, opened 2019-10-14
fix: [UI] Automatic logout

JakubOnderka:
## What does it do?

Automatic logout was broken, because for not logged user the call `/users/checkIfLoggedIn.json` returns status code 403, that is considered as fail in jQuery and the code inside was not handled.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


RichieB2B (approved):
Seems that this is a proper fix for #5172

adulau:
It's indeed fixing the old automatic logout issue. Tested and it works as expected for me.

iglocska:
Thanks a lot!

PR #5312, opened 2019-10-14
Update misp_retention.py to new api, use local tags

RichieB2B:
This also fixes a bug where events that did not change never received the `expired` tag. So they were pocessed over and over.


iglocska:
Thanks a lot

PR #5313, opened 2019-10-14
fix: [internal] Load Regexp just when they are requested

JakubOnderka:
## What does it do?

Without this, Regexp are loaded twice for every request, even when they are never used. So this is small performance improvement. 

It also check if Regexp type is relevant for given type before regexp executing. 

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Yeah, good idea. Indeed doesn't make sense in like 90% of the cases.

PR #5317, opened 2019-10-15
fix: [internal] Remove unused ShadowAttributesController method

JakubOnderka:
## What does it do?

Code cleanup.

## Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Indeed, can't find any calls for this, thanks!

PR #5318, opened 2019-10-15
fix: [UI] Proposal attachment downloading

JakubOnderka:
## What does it do?

Before this patch, instead of proposal attachment, real attachment with the same ID was downloaded.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #5337, opened 2019-10-18
chg: [internal] Much faster GalaxyCluster::attachClustersToEventIndex

JakubOnderka:
## What does it do?

This change speedups loading events – instead of loading clusters in loop, all clusters are loaded in one SQL query.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


mokaddem:
Thanks!

PR #5341, opened 2019-10-20
chg: [internal] Faster loading list of attributes

JakubOnderka:
## What does it do?

Instead of loading all galaxy cluster tags and then check if we can remove them, they are removed by integrated functionality of `Event::massageTags` method.

Also remove unused `$overrideLimit` and `$sgids` variables. 

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/AttributesController.php
+++ b/app/Controller/AttributesController.php
@@ -1775,34 +1775,26 @@ class AttributesController extends AppController
     private function __searchUI($attributes)
     {
         $sightingsData = array();
-        $sgids = $this->Attribute->Event->cacheSgids($this->Auth->user(), true);
         $this->Feed = ClassRegistry::init('Feed');
-        if (!empty($options['overrideLimit'])) {
-            $overrideLimit = true;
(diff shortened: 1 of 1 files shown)


iglocska:
From what I can tell overrideLimit is in fact used. It culls the attached correlations for the view and simply displays an aggregate count over 10k objects.

JakubOnderka:
@iglocska Sorry for late reply.

But if you check the code:

https://github.com/MISP/MISP/blob/b4ff851b629d7f9cf94d75f839774fda8f409e8b/app/Controller/AttributesController.php#L1756-L1765

There is 
```php
if (!empty($options['overrideLimit'])) {
``` 
condition, but the problem is, that the `$options` variable is never defined and this condition is always considered as `false`.

iglocska:
Went through it, indeed, it doesn't do anything currently. We might fix the overridelimit functionality at one point if it's still used at all and causing issues.

PR #5342, opened 2019-10-20
chg: [feed] Use precomputed hashes to speedup attaching correlation

JakubOnderka:
## What does it do?

We don't need to compute `md5` hash again.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Makes sense, thanks!

PR #5343, opened 2019-10-20
fix: [internal] Remove dead code from AttributesController

JakubOnderka:
## What does it do?

Code cleaup.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks, indeed that was superceded by the soft delete a long time ago. Thanks!

PR #5349, opened 2019-10-22
Fixing BasicAuth using ldap or AD

rbnor:

#### What does it do?

The  $_SERVER[$envvar] was previously never set, this is effectively $_SERVER['REMOTE_USER']. 
Its the easiest to fix that here, as the  $_SERVER['PHP_AUTH_USER'] only is set if the user is authenticated by BasicAuth at this stage. It is also possible to fix this in the misp-ssl.config with the virtualhost, but it was simply easier to just fix it here, and at the same time a bit tricky to actually set the variable in the virtualhost section, its easier in PHP.

#### Questions

- [ YES] Are you using it in production?

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/AppController.php
+++ b/app/Controller/AppController.php
@@ -152,7 +152,8 @@ class AppController extends Controller
         }
         // check if Apache provides kerberos authentication data
         $envvar = Configure::read('ApacheSecureAuth.apacheEnv');
-        if (isset($_SERVER[$envvar])) {
+        if  (isset($_SERVER['PHP_AUTH_USER']))  {
+            $_SERVER[$envvar]=$_SERVER['PHP_AUTH_USER'];
             $this->Auth->className = 'ApacheSecureAuth';
             $this->Auth->authenticate = array(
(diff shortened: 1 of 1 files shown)


JakubOnderka:
Hi, this change is not correct. You can define own `$_SERVER` key in `ApacheSecureAuth.apacheEnv` config.

rbnor:
> Hi, this change is not correct. You can define own `$_SERVER` key in `ApacheSecureAuth.apacheEnv` config.

I agree, and thats smoother for sure to define the key there, and i considered doing that, but i fear $_SERVER['REMOTE_USER'] is used elsewhere as well, so just changing the key read in this example from 'REMOTE_USER' to 'PHP_AUTH_USER', will work here, yes, but will it work in all the places $envvar is read and used as key for an value in the  $_SERVER array? If any of those places still is 'REMOTE_USER' it wont work.

rbnor:
Adopted the suggested config change instead, works nicely.

PR #5366, opened 2019-10-29
Allow python scripts to write to exec-errors.log

RichieB2B:
We were missing a crucial SELinux rule to allow the httpd_sys_script_t role to write to httpd_log_t files. This caused audit entries like this:
```
type=AVC msg=audit(1571825472.085:2697584): avc:  denied  { write } for  pid=10159 comm="python3" path="/var/www/MISP/app/tmp/logs/exec-errors.log" dev="dm-2" ino=657769 scontext=system_u:system_r:httpd_sys_script_t:s0 tcontext=unconfined_u:object_r:httpd_log_t:s0 tclass=file permissive=0
```

### DIFF
--- a/INSTALL/misplogrotate.te
+++ b/INSTALL/misplogrotate.te
@@ -1,8 +1,9 @@
-module misplogrotate 1.1;
+module misplogrotate 1.2;
 require {
 	type httpd_t;
 	type logrotate_t;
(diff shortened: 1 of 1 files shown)


adulau:
Oops. Nice catch!

PR #5373, opened 2019-10-31
Force botvrij.eu feed to use HTTPS

cudeso:
#### What does it do?

Force botvrij.eu feed to use HTTPS

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot

PR #5375, opened 2019-10-31
fix: Remove unusued config option

JakubOnderka:
## What does it do?

Removes config option `ApacheShibbAuth. ssoAuth` from example config, because it is never used.

```
$ rg ssoAuth
app/Config/config.default.php
113:                    'ssoAuth'           => 'AUTH_TYPE',
```

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot, we don't use it internally so if it fails for anyone just open an issue please :)

PR #5400, opened 2019-11-15
fix: [REST] Python has no 'Null' type, it is called 'None'

SteveClement:
Wrong null type returned by generator.


Rafiot:
1. We can probably keep PyMISP in the template, as it will be the default in ~90 days when I purge all remaining python 2.7 things
2. Null vs. None, yes, definitely.

iglocska:
Thanks a lot.

iglocska:
We can already start preparing users for the switch by letting them generate expanded pymisp scripts by default. If they don't like it, they can ofc manually edit it to pymisp - there's nothing saying these scripts will be running on the same box as MISP itself so we can default to Python 3.6+

PR #5404, opened 2019-11-16
Added more Organisation statistics

mokaddem:
- Added number of Attributes per Organization
- Added activity sparkline (based on ``Event.timestamp``) for Organisation

### DIFF
--- a/app/Controller/UsersController.php
+++ b/app/Controller/UsersController.php
@@ -1756,10 +1756,11 @@ class UsersController extends AppController
             'group' => 'Event.orgc_id',
             'conditions' => array('Event.orgc_id' => array_keys($orgs)),
             'recursive' => -1,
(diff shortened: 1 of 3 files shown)


mokaddem:
One issue I see with that activity based on ``Event.timestamp`` is if an admin make any change, it will be marked as Organization activity.
Is this to be considered as an edge case?
If you have other concerns, please tell me. Otherwise I'll merge it

adulau:
> One issue I see with that activity based on `Event.timestamp` is if an admin make any change, it will be marked as Organization activity.
> Is this to be considered as an edge case?
> If you have other concerns, please tell me. Otherwise I'll merge it

Fine for me. The edge case is known and not sure how this can solved at the current stage. You can merge it.

iglocska:
Looks good, thanks!

PR #5412, opened 2019-11-19
Feature fix db inconsistencies

### LINKED ISSUE #5393: Critical warning! The MISP database seems to be in an inconsistent state. Immediate attention is required.
After an automatic upgrade from 117 to 118 this error is in my Diagnostics

Table name | Description | Expected schema | Actual schema
-- | -- | -- | --
attributes
Column `comment` is different | comment YES text 65535 utf8_bin | comment NO text 65535 utf8_bin
feeds
Column `settings` is different | settings YES text 65535 utf8_general_ci | settings NO text 65535 utf8_general_ci
galaxies
Column `icon` is different | icon NO varchar 255 utf8_bin | icon YES varchar 255 utf8_bin
logs
Column `action` is different | action NO varchar 20 utf8_bin | action NO varchar 80 utf8_bin
Column `change` is different | change YES text 65535 utf8_bin | change NO text 65535 utf8_bin
Column `ip` is different | ip NO varchar 45 utf8_bin | ip YES varchar 45 utf8_bin
object_references
Column `object_uuid` exists but should not
Column `source_uuid` does not exist but should | source_uuid YES varchar 40 utf8_bin |  
object_template_elements
Column `disable_correlations` exists but should not
Column `object_relation` is different | object_relation YES varchar 255 utf8_bin | object_relation YES varchar 255 utf8_unicode_ci
Column `type` is different | type YES varchar 255 utf8_bin | type YES varchar 255 utf8_unicode_ci
Column `disable_correlation` does not exist but should | disable_correlation NO tinyint 3 |  
org_blacklists
Column `comment` is different | comment NO text 65535 utf8_unicode_ci | comment YES text 65535 utf8_unicode_ci
shadow_attributes
Column `proposal_to_delete` is different | proposal_to_delete NO tinyint 3 | proposal_to_delete YES tinyint 3

image

mokaddem:
# What
This pull request introduces a prototype to quickly fix database schema inconsistencies detected in the diagnostic page.

# How does it work
- Based on the expected database schema, it creates queries that should fix the issues
- It also cleans the model cache. Not sure if it's needed though...

Peek 2019-11-19 15-12

Potentially fix #5393


app/Model/Server.php
```diff
@@ -4310,7 +4310,7 @@ public function dbSchemaDiagnostic()
      */
     private function __attachRecoveryQuery($field, $table)
     {
-        if ($field['is_critical']) {
+        if (isset($field['error_type'])) {
```
iglocska, line 4313: Do we want to provide the scripts even if we consider them unnecessary? I am not sure either way, just thinking of cases where the changes are actually beneficial (such as the action field length being set to 80 instead of 20 on the log table).

app/Model/Server.php
```diff
@@ -4327,30 +4327,57 @@ private function __attachRecoveryQuery($field, $table)
...
                             $field['expected']['data_type'],
                             $length !== null ? sprintf('(%d)', $length) : '',
-                            isset($field['expected']['default']) ? 'DEFAULT "' . $field['expected']['default'] . '"' : '',
+                            isset($field['expected']['column_default']) ? 'DEFAULT "' . $field['expected']['column_default'] . '"' : '',
                             $field['expected']['is_nullable'] === 'NO' ? 'NOT NULL' : 'NULL',
                             empty($field['expected']['collation_name']) ? '' : 'COLLATE ' . $field['expected']['collation_name']
                         );
                         break;
                 }
+            } elseif($field['error_type'] == 'missing_table') {
```
iglocska, line 4353: Not convinced by this. We currently basically only correctly configure types that we know in advanced - such as the type based length setting. If a table has anything we do not expect that script will create inaccurate settings. Perhaps what we should consider is loading new tables from MYSQL.sql directly and being stricter on keeping it up to date?

app/Controller/ServersController.php
```diff
@@ -2205,6 +2205,45 @@ public function dbSchemaDiagnostic()
...
+            $this->set('error', $dbSchemaDiagnostics['error']);
+            $this->set('remainingLockTime', $dbSchemaDiagnostics['remaining_lock_time']);
+            $this->set('updateFailNumberReached', $dbSchemaDiagnostics['update_fail_number_reached']);
+            $this->set('updateLocked', $dbSchemaDiagnostics['update_locked']);
+            $this->render('/Elements/healthElements/db_schema_diagnostic');
+        }
+
+    }
+
+    public function execSQLQuery() {
```
iglocska: Instead of executing the recovery scripts, I'd much rather just display it to the user and let them run it outside of MISP if they want. This is generally something that can be SLOW and them hammering away at their DB might have adverse effects. It shouldn't be something you trigger on a whim in some cases.

iglocska (commented):
I have some concerns as listed below, ping me when you have time to work on this and we can have a chat!

app/Model/Server.php
mokaddem: IMO, I would like my MISP database to be as close as possible to the excpected SQL schema. it cost absolutely nothing to provide a button with a quick fix.
That's why we have 2 alerts type:
- Error: The schema should be fixed (missing table/column)
- Warning: The schema can be fixed (NULL)

app/Model/Server.php
mokaddem: In that case, it means that we expect the schema to have a table which does not exists in the diagnosed database. As we know the complete configuration of the missing table thanks to the expected schema, we can also provide a quick fix.
I'm not sure if this issue will ever happen but if it does, users can quick fix it.

app/Controller/ServersController.php
mokaddem: I totally get your point. It's there to improve user experience.
I do not expected all site admins to be comfortable to interact with the database directly. The idea is to facilitate his task by providing a button to press.
Moreover, it can also be used by experienced admin that just want to quickly fix their DB issues.
In the end, the result will be the same. An SQL query will be executed.
If we keep this quick-fix-by-pressing-a-button feature, we definitely need to add a proper warning about the consequence.

mokaddem: Just in case, warning message added

iglocska: Keep in mind a click on this when it comes to an alter table on a large instance's log table might completely kill it until mysql is restarted. I'd reaaaaally not want this directly accessible in the UI.

iglocska: Providing a copy-paste script and letting admins sql-dump, shut their misp off and run them at their own risk is fine. But a warning is definitely not enough to prevent screw ups with this.

app/Model/Server.php
iglocska: "As we know the complete configuration of the missing table thanks to the expected schema"

We do, but the current builder doesn't know how to correctly interpret it. If we build something more complete - sure.

app/Model/Server.php
iglocska: A quick fix is not always a quick fix. Again, an alter table for fixing a NULL/NOT NULL flag that is ultimately of no consequence due to us handling it on a code level could lead to an alter table that runs for days on an older system with a large log table for example.

mokaddem: Totally get it.
I find it sad that we are not providing the query straight

mokaddem: Totally get it.
I find it sad that we are not providing the query despite the fact that we already have it.
For ``NULL/NOT NULL`` I don't see consequences but for other case like precision or collation. I think it would make sense.
Anyway. If you are against, I'll just remove that part of the code. No big deal

app/Model/Server.php
mokaddem: Indeed, I'm not considering indexes. However straight of my mind, I don't see how it's different from a column addition. The column configuration are basically the same.
In what sense could I improve it?
Although, I don't see a situation where we would end up missing a table so I can also scrap up this part of the code.

app/Controller/ServersController.php
mokaddem: Alright. Removing that part of the code ;)

app/Model/Server.php
iglocska: Indexes are one thing, I'm more thinking types. The part above will for example set the lengths based on type, but in some cases might set it incorrectly. In those cases we simply don't provide a recovery script for column level issues, but omitting a column in a table add is an issue. However, in those cases we could just grab the full table definition from mysql.sql

app/Model/Server.php
iglocska: We could provide it, especially if it's not executed through the application directly but simply provided as a command that admins can copy paste i don't mind either way

app/Model/Server.php
mokaddem: Still provide the SQL query but added a warning message

Issue #4365, opened 2019-03-25
Soft delete of tags & galaxies


chrisr3d:

mokaddem:
Related to #3973 and https://github.com/MISP/misp-galaxy/issues/308

PR #5428, opened 2019-11-26
new: Allow for attribute/event tag deletion via Event or Attribute edit

tomking2:
## What does it do?

This PR allows you to delete an event's or an attribute's tags via updating an event/attribute using the endpoints:

- `attributes/edit/{attribute_id}`
- `events/{event_id}`

This means you no longer have to use a separate endpoint to untag, you can simply untag in the same workflow. This is through setting a key of 'deleted' on the tag.

This has been built with the thought to integrate this into PyMISP, allowing you to simply do:
```python
from pymisp import ExpandedPyMISP

pm = ExpandedPyMISP("url", "key")
event = pm.get_event(1234)
selected_attribute = event.attributes[0]
selected_tag = selected_attribute.tags[0]
selected_tag.delete()
```

A separate PR is being placed in to enable this functionality in PyMISP

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


mokaddem:
Hey, sorry for the late response..
Just had a look at the PR and the feature totally makes sense!
I just have one worry about [this line](https://github.com/MISP/MISP/pull/5428/files#diff-4e170dc14483ed36d7c9658c82f61516R898) because ``editAttribute`` includes special edge cases for the synchronization.
I'll look deeper into it and do some tests later in week.
Thanks a lot for your contribution and your patience!

tomking2:
No problem!

OK thanks let me know how you get on, and if you need anything from my part!

SteveClement:
@mokaddem any news? conflicted by now, unfourtenatly.

mokaddem:
Wow. I'm really confused to see this issue pop up. I was convinced it was done. And it nearly is. I had few modification done in local but never actually pushed it. I opened a new PR based on this  one. Sorry for the mess #5837.

So @tomking2, I spotted one major issue and few minors.

- ``editAttribute`` is not meant to be used by the ``attributes/edit`` endpoint
- No support of Object's Attribute
- ``deleted: 0`` was actually deleting the tag
- We also want to have Tag soft-deletion https://github.com/MISP/MISP/issues/4365 in the future.

I had to modify a bit your Pull request but the introduced new functionalities are preserved.
I need to re-do my tests as things might have changed in 2 months and once it's review by one member of the team it will be merge.

Thanks a lot for the idea and the contribution!

tomking2:
Hey @mokaddem, thanks for picking this one back up!

Good spot on some of those issues, wasn't aware that function shouldn't be used by the `attributes/edit` endpoint.

Look forward to hopefully seeing it in the code base!

PR #5430, opened 2019-11-26
Allow pushing of sightings only for perm_sighting

RichieB2B:
Give users with perm_sighting (and not perm_sync) the ability to push sightings.

The diff of `app/Model/Server.php` looks like a mess, but it is mostly added indention, see `git diff -b`

This is a followup of #4917


iglocska:
Thanks a lot!

PR #5435, opened 2019-11-29
Select right servers for pushing sightings

RichieB2B:
Of course our first production case only pushes sightings and not events. This does not work until this patch is applied.


iglocska:
Thanks a lot!

PR #5443, opened 2019-12-03
Push sightings regardsless of distribution level

RichieB2B:
Sightings can (and should) be pushed upstream. So they should not be held back by the distribution level of the event. Pushing sightings will not create an event if it doesn't already exist on the remote server, instead those sightings will be ignored. So there is no risk of leaking information outside of the MISP sharing model. Also, the "push sightings" bit on servers provides fine-grained control over which servers get sightings.


iglocska:
Makes sense, we don't store the value, only the uuid in the sighting, so why not.

PR #5462, opened 2019-12-12
Zoidberg final

mokaddem:
# Features
This branch introduces changes on the database tables and the UI. An exhaustive list of features and changes is given below:

#### Major
- Support of ``first_seen`` and ``last_seen`` at Attribute, ShadowAttribute and Object level, up to micro-seconds resolution
- Event timeline: View and Modify data in an interactive fashion
    - Timezone support and  up to milli-seconds resolution
    - Support Attributes and Objects (With read-only overwrite from Object's Attribute)
    - Support different combination of ``first_seen`` and ``last_seen`` (single data point and data time span)
    - Searching capabilities
    - Edition is accessible only to users with correct permissions
- Quick Object's Attribute ``add`` in the event view: Allow users to quickly add a new attribute (if applicable) to an existing object in the event view.

#### Minor
- Quick edit on object's fields (Event view)
- Renamed ``rotation key`` to ``pivot key`` in the ``eventGraph``
- ``first_seen`` and ``last_seen`` fields are returned by the API (search not supported yet)

#### Questions

#### Release Type:
- [X] Major


app/Lib/Tools/EventTimelineTool.php
```diff
@@ -0,0 +1,138 @@
...
+                    'id' => $attr['id'],
+                    'uuid' => $attr['uuid'],
+                    'content' => $attr['value'],
+                    'event_id' => $attr['event_id'],
+                    'group' => 'attribute',
+                    'timestamp' => $attr['timestamp'],
+                    'first_seen' => $attr['first_seen'],
+                    'last_seen' => $attr['last_seen'],
+                );
+                array_push($this->__json['items'], $toPush);
```
JakubOnderka: Why not just `$this->__json['items'][] = $toPush;`? It is even faster and I think also cleaner.

app/Model/Attribute.php
```diff
@@ -604,6 +616,22 @@ public function afterFind($results, $primary = false)
             if (isset($v['Attribute']['object_relation']) && $v['Attribute']['object_relation'] === null) {
                 $results[$k]['Attribute']['object_relation'] = '';
             }
+            if (!empty($v['Attribute']['first_seen'])) {
+                $fs = $results[$k]['Attribute']['first_seen'];
```
JakubOnderka: It is possible to make function from this algo? Because the same implementation is copied four times.

app/Lib/Tools/EventTimelineTool.php
iglocska: Makes sense indeed.

app/Model/Attribute.php
mokaddem: Good point.

PR #5463, opened 2019-12-12
chg: new feature to selectively choose to import objects (or not)

zaphodef:
New feature if you decide not to import some objects, using an import module.


iglocska:
Makes total sense, thanks!

PR #5464, opened 2019-12-12
fix: Don't show an empty error when there is no error

zaphodef:
When there was no error, an empty red panel still was shown on the top of the page.


iglocska:
Oops, thanks ;)

PR #5470, opened 2019-12-16
chg: Selectively choose what you want to import

zaphodef:
Same purpose as #5463, but also works for attributes. In order not to make the table too big (it was getting out of my screen :p), I hid the UUID field, which is most certainly not used by the end user.


mokaddem:
Thanks a lot and sorry for the delay!
Cheers

PR #5488, opened 2020-01-06
Return STIX in JSON format when Accept header asks for it

RichieB2B:
According to https://www.circl.lu/doc/misp/automation/ MISP will return JSON when `Accept: application/json` is set. This was not working for /events/stix/download which always returned XML. This patch fixes that.


adulau:
Thank you! (and happy new year to you and all the team!)

Issue #5457, opened 2019-12-10
Upgrade from 2.4.118 to 2.4.119 database schema errors


jezkerwin:
I just updated to 2.4.119 from .118, ran the diagnotics and got a LOT of database schema errors.
Running on RHEL7.
Cleaned the model cache. no affect
Restarted the host after upgrade

Table name | Description | Expected schema | Actual schema |  
-- | -- | -- | -- | --
attributes
Column `object_relation` is different | object_relation YES varchar 255 utf8_bin | object_relation YES varchar 255 utf8_bin NULL |  
Column `comment` is different | comment YES text 65535 utf8_bin | comment YES text 65535 utf8_bin NULL |  
cake_sessions
Column `id` is different | id NO varchar 255 utf8_bin | id NO varchar 255 utf8_bin '' |  
decaying_models
Column `uuid` is different | uuid YES varchar 40 utf8_bin | uuid YES varchar 40 utf8_bin NULL |  
Column `parameters` is different | parameters YES text 65535 utf8_general_ci | parameters YES text 65535 utf8_general_ci NULL |  
Column `attribute_types` is different | attribute_types YES text 65535 utf8_general_ci | attribute_types YES text 65535 utf8_general_ci NULL |  
Column `description` is different | description YES text 65535 utf8_general_ci | description YES text 65535 utf8_general_ci NULL |  
Column `org_id` is different | org_id YES int 10 | org_id YES int 10 NULL |  
Column `ref` is different | ref YES text 65535 utf8_unicode_ci | ref YES text 65535 utf8_unicode_ci NULL |  
Column `version` is different | version NO varchar 255 utf8_bin | version NO varchar 255 utf8_bin '' |  
event_blacklists
Column `comment` is different | comment YES text 65535 utf8_unicode_ci | comment YES text 65535 utf8_unicode_ci NULL |  
event_delegations
Column `message` is different | message YES text 65535 utf8_general_ci | message YES text 65535 utf8_general_ci NULL |  
Column `sharing_group_id` is different | sharing_group_id YES int 10 | sharing_group_id YES int 10 NULL |  
event_graph
Column `network_name` is different | network_name YES varchar 255 utf8_unicode_ci | network_name YES varchar 255 utf8_unicode_ci NULL |  
Column `preview_img` is different | preview_img YES mediumtext 16777215 utf8_general_ci | preview_img YES mediumtext 16777215 utf8_general_ci NULL |  
events
[...]

PR #5490, opened 2020-01-07
Mark NULL vs "" as not critical

RichieB2B:
This is a simple fix for #5457 as it marks the differences of a default NULL value when there is none expected as a non-critical difference. 

It also uses @ as a regex divider instead of / so it does not collide with 'http://localhost' which is the default value for sightingdbs.host


iglocska:
Thanks a lot!

PR #5504, opened 2020-01-14
fix: Incorrect directionality for oldest_timestamp comparisons

tomking2:
#### What does it do?

Fixes issue where "First Recorded Change" on the event view is always set to the most recent attribute's date, not the earliest attribute's date. This was due to an incorrect comparison between the dates.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

Issue #5514, opened 2020-01-16
Suggestion to remove invalid default feed


Applenice:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Support
| OS version (server)       | Ubuntu
| OS version (client)       | Ubuntu
| PHP version               | 7.2
| MISP version / git hash   | MISP 2.4.117
| Browser                   | Chrome

### Support Questions
Some default feeds are no longer valid,  whether to consider deleting them?

status: `# ZeuS Tracker has been discontinued on Jul 8th, 2019`
https://zeustracker.abuse.ch/byebye.php?download=ipblocklist
https://zeustracker.abuse.ch/byebye.php?download=compromised

stauts `404 Not Found`
http://booterblacklist.com/data/booterlist_latest.txt
http://www.cert.at/static/downloads/data/conficker/all_domains.txt

stauts `# Ransomware Tracker has been discontinued on Dec 8th, 2019`
https://ransomwaretracker.abuse.ch/feeds/csv/ 

In addition, it seems that the ID numbers do not increase in order. What are the rules for defining ID numbers?
https://github.com/MISP/MISP/blob/4aa6490da7781be8835253d269803ab4e8d89e6b/app/files/feed-metadata/defaults.json#L1171
https://github.com/MISP/MISP/blob/4aa6490da7781be8835253d269803ab4e8d89e6b/app/files/feed-metadata/defaults.json#L1196

thanks.

PR #5515, opened 2020-01-16
remove invalid default feed

Applenice:
#### What does it do?
Some default feeds are no longer valid, I mentioned an issue before:
https://github.com/MISP/MISP/issues/5514

#### Questions

#### Release Type:
- [ ] Major
- [X] Minor
- [ ] Patch


iglocska:
Thanks a lot!

PR #5530, opened 2020-01-22
Add SAML (Shibboleth) login button

legoguy1000:

* Always make sure it is mergeable in the default branch (as of today 2020-01-21: branch 2.4)

#### What does it do?

Doesn't fix an issue, just adds additional options.

With Shibboleth and SAML you have 2 options, for SAML login and don't allow local login or allow both.  The example in the documentation forces (requires) SAML authentication and thus doesn't allow you to use local credentials if needed.  This adds a button below the login form to redirect to the Shibboleth login page if using passive Shibboleth auth.  To use passive auth set "ShibRequestSetting requireSession 0/false" instead of "ShibRequestSetting requireSession 1/true"

#### Questions

- [ ] Are you using it in production? Not yet, but might.

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/View/Users/login.ctp
+++ b/app/View/Users/login.ctp
@@ -43,6 +43,12 @@
         <?php
             echo $this->Form->button(__('Login'), array('class' => 'btn btn-primary'));
             echo $this->Form->end();
+            if (true == Configure::read('ApacheShibbAuth')):
+        ?>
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks a lot!

PR #5541, opened 2020-01-23
fix: Proper logout when `CustomAuth_custom_logout` is set

JakubOnderka:
#### What does it do?

Currently, when `Plugin.CustomAuth_disable_logout` config variable is set and user click on logout link, MISP cookie will still be preserved. 

After this change, MISP cookie will be deleted first and then redirected to external logout path.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Controller/AppController.php
+++ b/app/Controller/AppController.php
@@ -112,7 +112,14 @@ class AppController extends Controller
     public function beforeFilter()
     {
         $this->Auth->loginRedirect = Configure::read('MISP.baseurl') . '/users/routeafterlogin';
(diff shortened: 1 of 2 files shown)


legoguy1000:
THis should be merged.  This is great for SSO logout like SAML since you can logout locally and then be redirected to the Shibboleth Logout URL. I would love for this to be part of the next release!!

iglocska:
Indeed, makes sense, thanks a lot for all the work as always @JakubOnderka!

PR #5542, opened 2020-01-23
fix: [UI] Remove double escaping

JakubOnderka:
#### What does it do?

URL is already escaped in https://github.com/MISP/MISP/blob/7d9600c81d8e106e004c9278953e5227a5287e00/app/View/Elements/genericElements/SideMenu/side_menu_link.ctp#L16

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


mokaddem:
Thanks

PR #5553, opened 2020-01-26
Fix the CoinBlockerLists

ZeroDot1:
Delete the MiningServerIPList.txt feed because the feed is no longer available. 

All current downloads can be found via the CoinBlockerLists homepage.
https://zerodot1.gitlab.io/CoinBlockerListsWeb/downloads.html

Thanks to everyone for using the CoinBlockerLists, I appreciate it very much.

```json
{
    "Feed": {
      "id": "68",
      "name": "This list contains all IPs - A additional list for administrators to prevent mining in networks",
      "provider": "CoinBlockerLists",
      "url": "https://gitlab.com/ZeroDot1/CoinBlockerLists/raw/master/MiningServerIPList.txt?inline=false",
      "rules": "",
      "enabled": true,
      "distribution": "3",
      "sharing_group_id": "0",
      "tag_id": "0",
      "default": false,
      "source_format": "freetext",
      "fixed_event": false,
      "delta_merge": false,
      "event_id": "0",
      "publish": false,
      "override_ids": false,
      "settings": "{\"csv\":{\"value\":\"\",\"delimiter\":\",\"},\"common\":{\"excluderegex\":\"\"}}",
      "input_source": "network",
      "delete_local_file": false,
      "lookup_visible": true,
      "headers": ""
    }
  },
```

#### What does it do?
This fix corrects bugs and removes a feed that is no longer available.

#### Questions

I don't know, I hope the fix is helpful.

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/files/feed-metadata/defaults.json
+++ b/app/files/feed-metadata/defaults.json
@@ -1112,7 +1112,7 @@
     "Feed": {
       "id": "64",
       "name": "This list contains all domains - A list for administrators to prevent mining in networks",
-      "provider": "CoinBlockerLists",
+      "provider": "ZeroDot1 - CoinBlockerLists",
       "url": "https://gitlab.com/ZeroDot1/CoinBlockerLists/raw/master/list.txt?inline=false",
       "rules": "{\"tags\":{\"OR\":[],\"NOT\":[]},\"orgs\":{\"OR\":[],\"NOT\":[]}}",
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks a lot, both for the fix and of course for maintaining the feed :)

Issue #5554, opened 2020-01-27
Repeated Error during Install Invalid setting Notice Error: Array to string conversion in [/var/www/MISP/app/Model/Log.php, line 211]


utkonos:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | Ubuntu 18.04.3
| PHP version               | 7.2
| MISP version / git hash   | 2.4.120

### Expected behavior
Installed MISP

### Actual behavior
During install, an error occurs for many of the plugins with the same error type for each one.

### Steps to reproduce the behavior
```
apt install wget
adduser --disabled-password misp
usermod -a -G sudo misp
usermod -a -G staff misp
usermod -a -G www-data misp
su - misp


wget -O /tmp/INSTALL.sh https://raw.githubusercontent.com/MISP/MISP/v2.4.120/INSTALL/INSTALL.sh
bash /tmp/INSTALL.sh -c -M
```

### Logs, screenshots, configuration dump, ...
These errors repeat for multiple plugins, but not all.
```
Welcome to CakePHP v2.10.19 Console
---------------------------------------------------------------
App : app
Path: /var/www/MISP/app/
---------------------------------------------------------------
Invalid setting "Plugin.Enrichment_cve_enabled". Please make sure that the setting that you are attempting to change exists and if a module parameter, the modules are running.
Notice Error: Array to string conversion in [/var/www/MISP/app/Model/Log.php, line 211]

2020-01-27 00:10:08 Notice: Array to string conversion in [/var/www/MISP/app/Model/Log.php, line 211]
Setting "Plugin.Enrichment_cve_enabled" changed to 1
```

utkonos:
MISP install wasn't completing for a separate reason and that's fixed. This error doesn't appear to actually prevent the install.

iglocska:
Interesting. Those settings can only be set if the MISP modules are running and are dynamically created on the fly - in this case they're not so the setting doesn't exist. We should have a cleaner way of handling that.

JakubOnderka:
This change (https://github.com/MISP/MISP/pull/5555) will fix the notice message and exit with error code if setting doesn't exists.

utkonos:
Thanks!!

utkonos:
With the more recent versions, there are fewer errors. There is still the following message:

```
[...]

PR #5555, opened 2020-01-27
fix: [CLI] Die if setting name is not correct

JakubOnderka:
## What does it do?

Fixes notice in #5554.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Sweet, thanks!

PR #5559, opened 2020-01-27
fix: [log] Proper format log message for reset auth key

JakubOnderka:
In future, it will be also possible to filter auth keys in logs.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot

PR #5560, opened 2020-01-27
chg: [internal] Speed up of loading event page

JakubOnderka:
## What does it do?

Speedups loading event page.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Model/Event.php
+++ b/app/Model/Event.php
@@ -6948,4 +6948,27 @@ class Event extends AppModel
         }
         return $filters;
     }
+
+    /**
+     * @param array $event
+     */
+    public function removeGalaxyClusterTags(array &$event)
+    {
+        $galaxyTagIds = array();
+        foreach ($event['Galaxy'] as $galaxy) {
+            foreach ($galaxy['GalaxyCluster'] as $galaxyCluster) {
+                $galaxyTagIds[$galaxyCluster['tag_id']] = true;
+            }
+        }
+
+        if (empty($galaxyTagIds)) {
+            return;
+        }
+
+        foreach ($event['EventTag'] as $k => $eventTag) {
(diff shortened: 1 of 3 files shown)


SteveClement:
@iglocska speed is good, isn't it? ;)

app/Model/Event.php
```diff
@@ -6848,4 +6848,27 @@ public function addFiltersFromSubqueryElements($filters, $subqueryElements)
         }
         return $filters;
     }
+
+    /**
+     * @param array $event
+     */
+    public function removeGalaxyClusterTags(array &$event)
```
mokaddem, line 6955: Simple question/suggestion. Why not use `isset()` comparison on associative arrays instead of `in_array()`?
Something like this:
```
$galaxyTagIds = array();
foreach ($event['Galaxy'] as $galaxy) {
    foreach ($galaxy['GalaxyCluster'] as $galaxyCluster) {
        $galaxyTagIds[$galaxyCluster['tag_id']] = true;
    }
}

if (empty($galaxyTagIds)) {
    return;
}
foreach ($event['EventTag'] as $k => $eventTag) {
    if (isset($galaxyTagIds[$eventTag['tag_id']])) {
        unset($event['EventTag'][$k]);
    }
}
```

mokaddem (commented):
Hey, it looks way cleaner and I would like to merge it.
Could you resolve the conflict (should be easy - and sorry, you would not have to do it if it was merged months ago... :/ ) and have a look at my question?
Thanks a lot!

app/Model/Event.php
JakubOnderka: Changed, I was thinking that it should be not necessary, because it will contains just few tags,  but still make sense.

mokaddem:
Thanks!

PR #5574, opened 2020-01-30
fix: [internal] HTML code in view_event_distribution_graph

JakubOnderka:
## What does it do?

Just small HTML fix.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot

Issue #5498, opened 2020-01-10
Background workers need restart to notice new configuration settings


RichieB2B:
[app/Model/Sighting.php](https://github.com/MISP/MISP/blob/2.4/app/Model/Sighting.php) reads `Plugin.Sightings_policy` from the config in `attachToEvent()` which is used for the `publish_sightings` job. Unfortunately the background workers do not see the actual value unless I add
```
Configure::load('config');
``` 
at the top of `Sighting.php`
However, this might cause the issue discussed in #5304. How should this be properly solved?

iglocska:
That indeed makes sense and is a good point. The solution is to replicate the loading of the configuration on startup for the console scripts. Will have a look once I'm in a slightly more usable state

RichieB2B:
I see now. Because the workers are long lived PHP processes, changes in the config are not noticed until the workers are restarted. Makes sense indeed and would be nice to resolve but not crucial.

iglocska:
Indeed. Will have a look and will try to see if there's a hooking mechanism that could help. If not we'll add a new one.

RichieB2B:
This just got a lot more important to me. I needed to mass update some events so I set `MISP.disable_emailing=true` and started pulling in the updated events. Of course server pulls run in a background worker and they sent out notification E-mails anyway. :`(

RichieB2B:
Thanks! \o/

iglocska:
Forgot to reply to this. This should hopefully be fixed as of yesterday, let me know if it persists!

iglocska:
Oh just as I submit the message it refreshes and reveals your response. Argh.

PR #5590, opened 2020-02-04
Force refresh for proxy settings

thatreguy:
Observed that proxy settings were not getting read properly, unless issued Configure::load('config').
On a side note shouldn't every Configure::read without a load be treated as stale?

#### What does it do?

Fixes the issue of proxy address not getting utilized while fetching feeds even when set through UI.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Lib/Tools/SyncTool.php
+++ b/app/Lib/Tools/SyncTool.php
@@ -40,6 +40,8 @@ class SyncTool
      */
     public function createHttpSocket($params = array())
     {
+        // Refresh config
+        Configure::load('config');
         // Use own CA PEM file
         $caPath = Configure::read('MISP.ca_path');
         if (!isset($params['ssl_cafile']) && $caPath) {


iglocska:
Configure::load() is called at the start of every execution, so unless you directly modify the setting before executing it, it shouldn't cause issues. Can you describe what exactly is leading to the issue with the settings not being correctly reloaded?

thatreguy:
Can you please check again? As far as I can see, `Configure::load` is only called in bootstrap.php(via index.php) but for some reason, feed initiated by the workers is not taking the proxy settings. I faced this issue both in a corporate environment and while testing locally via tcpdump. Without `configure::load` proxy settings did not take effect.
Is there a test for this?

```#0 /var/www/MISP/app/Model/Feed.php(1548): Feed->getFollowRedirect(Object(HttpSocket), 'https://rules.e...', Array)
#1 /var/www/MISP/app/Model/Feed.php(199): Feed->feedGetUri(Array, 'https://rules.e...', Object(HttpSocket), true)
#2 /var/www/MISP/app/Model/Feed.php(792): Feed->getFreetextFeed(Array, Object(HttpSocket), 'csv', 'all')
#3 /var/www/MISP/app/Console/Command/ServerShell.php(199): Feed->downloadFromFeedInitiator('3', Array, '1357')
#4 /var/www/MISP/app/Console/Command/AppShell.php(32): ServerShell->fetchFeed()
#5 /var/www/MISP/app/Vendor/kamisama/php-resque-ex/lib/Resque/Job.php(199): AppShell->perform()
#6 /var/www/MISP/app/Vendor/kamisama/php-resque-ex/lib/Resque/Worker.php(278): Resque_Job->perform()```

RichieB2B:
This problem is a duplicate of #5498 and this PR is not the right solution

PR #5601, opened 2020-02-06
chg: [UI] Check if ssdeep PHP extension is installed

JakubOnderka:
## What does it do?

This extensions is necessary for ssdeep correlations, so this change provide info about status of this extension in Diagnostic page.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot

PR #5603, opened 2020-02-06
fix: Saving attribute with same value remove correlation

JakubOnderka:
## What does it do?

Before this patch, when user save attribute with correlation and keep the value of it, the correlations are missing.

The problem is, that correlations are not regenerated when value is not changed (makes sense), but all correlations are automatically deleted in `beforeSave` when editing existing attribute.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Model/Attribute.php
+++ b/app/Model/Attribute.php
@@ -653,10 +653,6 @@ class Attribute extends AppModel
 
         $this->data = $this->ISODatetimeToUTC($this->data, $this->alias);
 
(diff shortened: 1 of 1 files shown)


JakubOnderka:
This is fixed by 9913d194fafb78cb996d144c9d0a910c1ab57774, 20 days after I submitted the same fix :D

iglocska:
Ah shit missed that :( Sorry!

PR #5606, opened 2020-02-07
fix: [UI] Proper object table header when includeRelatedTags

JakubOnderka:
Also makes HTML code little bit smaller by removing unnecessary whitespaces.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


mokaddem:
Thanks!

Issue #5608, opened 2020-02-07
[Import] stix2misp.py assumes local installed pymisp 


coolacid:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | Debian
| OS version (client)       | Ubuntu
| PHP version               | 7.3.11-1~deb10u1 -- 20180731
| MISP version / git hash   | 2.4.120
| Browser                   | If applicable

### Expected behavior
Import of STIX2 event

### Actual behavior
Fails with:
```
Traceback (most recent call last):
  File "/var/www/MISP/app/files/scripts/stix2/stix2misp.py", line 39, in <module>
    with open('{_PyMISP_dir}/pymisp/data/describeTypes.json'.format(_PyMISP_dir=_PyMISP_dir), 'r') as f:
FileNotFoundError: [Errno 2] No such file or directory: '/var/www/MISP/PyMISP/pymisp/data/describeTypes.json'
```

### Steps to reproduce the behavior
In stall pyMISP as a system module, and not keep the /var/www/MISP/PyMISP/pymisp directory

### Logs, screenshots, configuration dump, ...
https://github.com/MISP/MISP/blob/2.4/app/files/scripts/stix2/stix2misp.py#L30

`_PyMISP Dir` assumes that pyMISP is installed in the MISP path, which when installing as a system level module, is not the case - thus the rest of the script fails. 

The correct way, which is used in misp2stix2.py: 
https://github.com/MISP/MISP/blob/2.4/app/files/scripts/stix2/misp2stix2.py#L146

jloehel:
I installed pymisp in a virtualenv. It's also not working for me. ~Please make this configurable.~

I would prefer:
~~~Python
_pymisp_dir = Path(os.path.dirname(pymisp.__file__))
with open(_pymisp_dir / 'data' / 'describeTypes.json', 'r') as f:
    _misp_types = json.loads(f.read())['result'].get('types')
~~~

chrisinmtown:
FWIW, the install scripts init/update the PyMISP submodule of MISP, which sets up the prerequisite for this little stix2misp script.  I would not say that PyMISP is *installed* in /var/www/MISP/PyMISP , we followed instructions and installed from source in that directory to our Python venv, but the PyMISP files remain available at the MISP/PyMISP path. I guess it's a case of "department of redundancy department"?

jloehel:
I am using now:
~~~Python
[...]

PR #5609, opened 2020-02-07
stix2misp.py: import pymisp and use it's __path__

coolacid:

#### What does it do?

If it fixes an existing issue, please use github syntax: `#5608`

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/files/scripts/stix2/stix2misp.py
+++ b/app/files/scripts/stix2/stix2misp.py
@@ -23,11 +23,12 @@ import time
 import io
 import re
 import stix2
+import pymisp
 from stix2misp_mapping import *
 from collections import defaultdict
 
 _MISP_dir = "/".join([p for p in os.path.dirname(os.path.realpath(__file__)).split('/')[:-4]])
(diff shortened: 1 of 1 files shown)


iglocska:
This is intentional. Maybe we should have your method in a try catch if the first one fails. The idea is this:
MISP servers often do not have PyMISP installed, hover, via the submodules the static files are always available for importing.

coolacid:
I'd disagree - All the install documents have pyMISP being installed in the virtual env `pip install .`  while in the pyMISP directory. 

Also, my changes makes it the same as misp2stix2.py : https://github.com/MISP/MISP/blob/2.4/app/files/scripts/stix2/misp2stix2.py#L146 

Finally, import is also how versions are checked: https://github.com/MISP/MISP/blob/aef4e51cdb610c77d8aa42ad106757594ad0a57e/app/files/scripts/stixtest.py

SteveClement:
Any thoughts @chrisr3d

Issue #5611, opened 2020-02-07
[Import] Warnings from stix2misp.py causes import to fail


coolacid:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | Debian
| OS version (client)       | Ubuntu
| PHP version               | 7.3.11-1~deb10u1 -- 20180731
| MISP version / git hash   | 2.4.120
| Browser                   | If applicable

### Expected behavior
Import of STIX2 event

### Actual behavior
Import does not work due to multiple warnings:
`ANTLR runtime and generated code versions disagree: 4.8!=4.7.1` 

### Steps to reproduce the behavior
Manulally run stix2misp.py and see the above warnings, with the `1` returned at the end. 

### Logs, screenshots, configuration dump, ...
Because the following checks the returned result against only a number `1` it fails if warnings are returned also from the stix2misp.py python script. 
https://github.com/MISP/MISP/blob/2.4/app/Model/Event.php#L5699

iglocska:
fixed via your PR #5612, thanks a lot!

PR #5612, opened 2020-02-07
Ensure we only have the last line from the shell command

coolacid:
#### What does it do?

If it fixes an existing issue, please use github syntax: `#5611`

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #5614, opened 2020-02-10
fix: [UI] Disable autocomplete for authkey

JakubOnderka:
#### What does it do?

To prevent saving it into browser cache.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Good idea, thanks!

PR #5615, opened 2020-02-10
chg: [internal] mispzmqtest.py

JakubOnderka:
#### What does it do?

- Also check if Redis Python library is installed.
- Do not print "OK" if libraries doesn't exists.
- Return error code 1 if library doesn't exists.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #5617, opened 2020-02-11
Adding instructions to build a Debian Package

stricaud:
It does not build a Debian package that can be pushed to the distribution yet, but it provides
an easy way to have a Debian package for MISP for minimal configuration efforts.

It is installed in /usr/share/misp and there are too many things happening in that directory,
such as logs, instead of being in /var/log/misp/.

However it can be useful to a lot of people, and I will gradually improve it over time.

-- STR

#### What does it do?

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


app/Model/Attribute.php
```diff
@@ -660,6 +660,11 @@ private function __alterAttributeCount($event_id, $increment = true)
 
     public function afterSave($created, $options = array())
     {
+       //STR
```
iglocska: image

stricaud: damn, this was a test I left :( let me remove that

iglocska: Thanks :)

iglocska:
Thanks a lot!

iglocska:
Awesome stuff :)

PR #5628, opened 2020-02-14
chg: [internal] Little bit faster ssdeep saving

JakubOnderka:
#### What does it do?

Just small speedup, with cleaner a more efficient code.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thank you!

PR #5630, opened 2020-02-14
chg: [mispzmq] Do not create array every 0.1 sec

JakubOnderka:
#### What does it do?

Should lead to little bit lower CPU usage.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

PR #5631, opened 2020-02-14
fix: [pubsub] Do not fetch setting for every push

JakubOnderka:
#### What does it do?

Before this change, for every push to Redis in PubSub tool, settings are loaded again and again and also Redis database is changed for every push, even when settings is already in `$this->__settings` variable and Redis database is selected when connecting.

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Wow, nice catch!

PR #5635, opened 2020-02-17
chg: [mispzmq] Save PID and settings to scripts tmp folder

JakubOnderka:
## What does it do?

Before this change, PID and setting file are stored at the same folder as python scripts. Then it is necessary to add write permission to the `mispzmq` folder, that can be dangerous. 

After this change, files are moved to scripts tmp folder.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Lib/Tools/PubSubTool.php
+++ b/app/Lib/Tools/PubSubTool.php
@@ -21,7 +21,7 @@ class PubSubTool
                 $settings[$key] = $temp;
             }
         }
-        $settingsFile = new File(APP . 'files' . DS . 'scripts' . DS . 'mispzmq' . DS . 'settings.json', true, 0644);
+        $settingsFile = new File(APP . 'files' . DS . 'scripts' . DS . 'tmp' . DS . 'mispzmq_settings.json', true, 0644);
         $settingsFile->write(json_encode($settings, true));
         $settingsFile->close();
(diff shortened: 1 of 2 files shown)


iglocska:
This looks indeed sane, we need to separate the config / data from the script files, something that has been muddied quite a bit when it comes to the files directory. However, merging this directly would disassociate the currently running zmq script from MISP when updating, so we need to add a script that would stop / start the script on update.

SteveClement:
> 
> 
> This looks indeed sane, we need to separate the config / data from the script files, something that has been muddied quite a bit when it comes to the files directory. However, merging this directly would disassociate the currently running zmq script from MISP when updating, so we need to add a script that would stop / start the script on update.

Any good pointers where we already did this?

mokaddem:
> Any good pointers where we already did this?

AFAIK we never did it. But what I would do is to add an entry in the `updateMISP` function that would restart the ZMQ (if it's enabled and currently running). Similar to what is done here
https://github.com/MISP/MISP/blob/ec634e426289b52c323fb04efe152af8bd25ec11/app/Model/AppModel.php#L241-L243

JakubOnderka:
I am closing this, I will create new pull with another solution.

Issue #5598, opened 2020-02-05
Unable to attach Galaxy if not previously used by (org)admin


coolacid:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | Debian
| OS version (client)       | Ubuntu
| PHP version               | 7.3.11-1~deb10u1 -- 20180731
| MISP version / git hash   | 2.4.120
| Browser                   | If applicable

### Expected behavior
Any user should be able to add a Galaxy to an event

### Actual behavior
A user without the Tag Editor Role option is unable to add a Galaxy if that Galaxy has not been used previously by someone that has the role. 

### Steps to reproduce the behavior
As a user without the Tag Editor Role, attempt to add a galaxy to an event that has not been used before. 

### Logs, screenshots, configuration dump, ...
It appears galaxies are just tags under the hood, and that tag is created dynamically when needed/used, not prior. Since there is no tag for a galaxy not yet used, the system attempts to create the tag. In 2.4.120, there is a check to ensure the user creating the Tag has the `perm_tag_editor` role. (https://github.com/MISP/MISP/blob/c574329e06a99de6ace60e359e8a99fcb912806b/app/Model/Tag.php#L308-L326) 

In the case of galaxies, it's the system creating the tag, not the user, thus this check should be bypassed.

coolacid:
Two possible fixes:

- Add to the $tag being sent an override to force the tag creation
- Pass a fake user to the `captureTag` function that has the `['Role']['perm_tag_editor']` set to true

iglocska:
Good point, this is indeed a defect. We should override the check for tags coming from galaxies.

iglocska:
```
 if ($user['Role']['perm_tag_editor']) { 
```

This check should have an OR for the tag coming from an enabled vocabulary /  galaxy

coolacid:
For Galaxies, I assume that by https://github.com/MISP/MISP/blob/c574329e06a99de6ace60e359e8a99fcb912806b/app/Model/Galaxy.php#L223 it's confirmed the galaxy is enabled - So passing a $force variable to the tag array and checking for that $force in the or would suffice?

iglocska:
Yep, that sounds right!

coolacid:
https://gitter.im/MISP/MISP?at=5e4d3ad13ca8a67fb8052b38 -- Vocabulary not effected, closing.

PR #5647, opened 2020-02-18
Allow forcing tag creation for galaxies

coolacid:
#### What does it do?

If it fixes an existing issue, please use github syntax: #5598 
Issue notes `vocabulary` this does NOT fix that section

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Very clean, thanks a lot!

Issue #5653, opened 2020-02-19
[Schemacheck] MySQL column case miss match


coolacid:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | Ubuntu (Docker)
| OS version (client)       | Ubuntu, ...
| PHP version               | 7.x
| MISP version / git hash   | 2.4.121
| Browser                   | If applicable

### Expected behavior
DB schema check shows all clear

### Actual behavior
Nope :)

### Steps to reproduce the behavior
`docker-compose up` on CoolAcid's docker images, login, check maint page

### Logs, screenshots, configuration dump, ...
Originally reported: https://github.com/coolacid/docker-misp/issues/17

MySQL has a setting that defines the case of column names returned. Normally, they are returned as lowercase, however, for some reason the instance used from docker hub returned uppercase. 

Given this could be changed by anyone at anytime, care should be taken to ensure the case matches expected values. In this case, the expected schema vs actual schema. 

MySQL setting reference: https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_lower_case_table_names

coolacid:
PR merged, closing

PR #5654, opened 2020-02-19
fix: Force schema columns lowercase to match expected

coolacid:
#### What does it do?

If it fixes an existing issue, please use github syntax: #5653 

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

NOTE: It might be possible for `array_change_key_case` to return false if `$column_schema['columns']` is not an array. However, I don't believe it could ever NOT be an array.


iglocska:
Thanks, excellent catch. Also, why the hell did they change that?...

coolacid:
I don't know if they changed that, or the docker hub MySQL version config changes it. I could test against a different image, but better to just handle the case in case someone has a weird config somewhere.

iglocska:
Yeah better safe than sorry, would make for a nice anecdote about mysql if it was a case, but nothing beyond that :)

PR #5672, opened 2020-02-27
Fixes STIX2 export failing with "ANTLR runtime and generated code versions disagree: 4.8!=4.7.1"

patriziotufarolo:
#### What does it do?

It fixes the error 
```
ANTLR runtime and generated code versions disagree: 4.8!=4.7.1
```
when exporting STIX 2.
The patch is the same as #5612 for issue #5611 (e3b1e8c74a0b40cdb54be938bcea4d9b28a7f0b9)

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot, indeed this seems to solve it!

patriziotufarolo:
You're welcome! the best, of course, would be to make the two versions number match, but I'm still investigating the real issue in the dependencies and in this way it works :D

PR #5677, opened 2020-03-05
Debian

stricaud:
This updates the MISP debian package to the new version.

There was also an mistake with the version number from the debian changelog, I fixed it to reflect the upstream release.

The biggest change it that it now installs workers via systemd.


stricaud:
I do not understand the error with Travis. Those commits only affect the debian/ directory to build the debian package.

PR #5705, opened 2020-03-19
new: usr: add support for RHEL in the install script

lfortemps:
#### What does it do?

It allows for the use of "INSTALL.sh" on RHEL and CentOS 7.x, essentially speeding up the installation process on these platforms. It reuses the script blocs in the documentation in the same way it is done for the other platforms.
Note: integrity check is currently non-blocking for RHEL, it needs to be enabled if/when PR is merged.

#### Questions

#### Release Type:
- [ ] Major
- [X ] Minor
- [] Patch


INSTALL/INSTALL.tpl.sh
```diff
@@ -676,9 +788,8 @@ fi
 # TODO: Move support map to top
 
 SUPPORT_MAP="
-x86_64-centos-8
+x86_64-centos-7
```
cvandeplas, line 791: Note: removing support for some centos en RHEL where the script doesn't (and didn't) work before.

SteveClement:
Looks good.

PR #5706, opened 2020-03-19
Fix venv and ssdeep for RHEL 7

RichieB2B:
Two fixed for the RHEL 7 / CentOS 7 installation documentation:

1. pip3 now installs virtualenv-20.0.12 which does not have a callable virtualenv.py anymore. /usr/local/bin/virtualenv should be used instead.
1. the ssdeep php modules was never installed. The braindead install script needs libfuzzy.so in /usr/lib/


SteveClement:
I changed it to: python3 -m venv
Seems more clean and perhaps future proof

PR #5707, opened 2020-03-20
Scoped css for widget

mokaddem:
- Behave transparently if the widget style is not marked as `scoped`
- If the widget view contain the HTML tag `style` contains the parameter `widget-scoped` like this:
```
<style widget-scoped>
[...]
</style>
```
It will be converted to normal `<style>` html tag and all subsequent CSS selector will have a random data selector prepend to them, so that scoped CSS behavior is achieved.


app/View/Helper/ScopedCSSHelper.php
```diff
@@ -0,0 +1,93 @@
...
+            if ($length == 0) {
+                return true;
+            }
+            return (substr($haystack, -$length) === $needle);
+        }
+
+        private function preppendScopedId($css, $seed)
+        {
+            $prependSelector = sprintf('[data-scoped="%s"]', $seed);
+            $cssLines = explode("\n", $css);
```
iglocska, line 19: PHP_EOL instead \n

PR #5709, opened 2020-03-20
Fixes missing MySQL ignore table statements.

kscheetz:
Installation of a new server with an existing database was failing.  It looks like three 'CREATE TABLE' statements were missing the 'IF NOT EXISTS' statement.  I've add the statement to where I believe it is missing, and this has fixed our issue.


SteveClement:
@kscheetz yet you also change stix python scripts, please seperate from SQL changes.

kscheetz:
> @kscheetz yet you also change stix python scripts, please seperate from SQL changes.

@SteveClement Ok, I have removed the commits and will split these into 2 requests.

mokaddem:
Thanks!

PR #5715, opened 2020-03-23
Added Malware Bazaar / Formatting in default.json

GlennHD:
Added Malware Bazaar / Formatting in default.json.


adulau:
Thank you for the pull-request. I fixed the timestamp field which was set with a boolean value instead of a string.

Issue #5688, opened 2020-03-09
Email based OTP


cvandeplas:
**Is your feature request related to a problem? Please describe.**
Increased security for authentication using email based one time password (OTP) system. 

**Describe the solution you'd like**
- An authentication system allowing the users to enter their username / password, and on a following page enter the OTP token that was sent by email. Only then the user can be considered as fully authenticated.
- The (nicely formatted) email should mention the IP address you've been connecting from, (keep in mind X-Forwarded-For) and also a unique token (a few numbers, easy to type) that is valid for a few minutes.
- The configuration of the EmailOTP solution should allow the administrator to exclude certain users from the OTP.  (allowing them to login even if the emailing is down)

**Describe alternatives you've considered**
- LinOTP offers similar functionality, and can integrate with MISP using the LinOTP auth mechanism. This option requires the installation and maintenance of another (large) software package, and makes things to configure much more complex than needed for a simple email-based OTP solution which could work directly with MISP.   Integrating it immediately with MISP would make this feature accessible to everyone without much effort from their side.

lfortemps:
I think this one can be closed as the email OTP PR was merged

iglocska:
Indeed, thanks again!

PR #5726, opened 2020-03-25
new: usr: Implementation of email-based OTP

lfortemps:
#### What does it do?

Implementation of #5688 
Using Redis instead of sessions due to volatility.
Introduce new server settings to configure OTP validity, length but also an exclusion list.

#### Questions

#### Release Type:
- [] Major
- [] Minor
- [X] Patch


app/View/Users/email_otp.ctp
```diff
@@ -0,0 +1,30 @@
+<?php echo $this->Flash->render(); ?>
+
+<div class="actions sideMenu">
+  <div style="padding: 10px;">
+    <p> Your administrator has turned on an additional authentication step which
+      requires you to enter a OTP (one time password) you have received via email.
+    </p>
+    <p> Make sure to check your SPAM folder. </p>
+    <a href='/users/email_otp'> <button class='btn'> Resend </button></a>
```
cvandeplas: I think this link needs to be generated dynamically. Otherwise the configurable base_url will not work.

app/Controller/AppController.php
```diff
@@ -406,6 +406,12 @@ public function beforeFilter()
                     if (!$this->_isRest()) {
                         $this->redirect(array('controller' => 'users', 'action' => 'change_pw', 'admin' => false));
                     }
+                } elseif (Configure::read('Security.email_otp_enabled') && !$this->_isRest() && !in_array($this->request->here, array($base_dir.'/users/terms', $base_dir.'/users/email_otp', $base_dir.'/users/change_pw', $base_dir.'/users/logout', $base_dir.'/users/login'))) {
+                    $redis = $this->{$this->modelClass}->setupRedis();
+                    $otp_authed = $redis->get('misp:otp_authed:'.$this->Auth->user('id'));
```
cvandeplas: token should not be on user_id , this would be a vulnerability in the use-case:
- legit user is logged in with OTP-is already validated.
- malicious user logs in with stolen credentials, (s)he does not get the request to the OTP.

app/Controller/AppController.php
```diff
@@ -406,6 +406,12 @@ public function beforeFilter()
                     if (!$this->_isRest()) {
                         $this->redirect(array('controller' => 'users', 'action' => 'change_pw', 'admin' => false));
                     }
+                } elseif (Configure::read('Security.email_otp_enabled') && !$this->_isRest() && !in_array($this->request->here, array($base_dir.'/users/terms', $base_dir.'/users/email_otp', $base_dir.'/users/change_pw', $base_dir.'/users/logout', $base_dir.'/users/login'))) {
```
cvandeplas: need to find another way than `$this->_isRest()` to distinguish REST API users. This can be easily circumvented by adding some request header accept response types.  
We might need to review the location of this code. Authentication is already succeeded, and afterwards we do extra authentication elements.

app/Model/Server.php
```diff
@@ -1243,6 +1243,53 @@ public function __construct($id = false, $table = null, $ds = null)
                             'type' => 'boolean',
                             'null' => true
                         ),
+                        'email_otp_enabled' => array(
+                                'level'=> 2,
+                                'description' => __('Enable two step authentication with a OTP sent by email.'),
```
cvandeplas: add a remark it's incompatible with ExternalAuth

app/Controller/UsersController.php
```diff
@@ -31,6 +31,9 @@ public function beforeFilter()
 
         // what pages are allowed for non-logged-in users
         $allowedActions = array('login', 'logout');
+        if(!empty(Configure::read('Security.email_otp_enabled'))) {
+          $allowedActions[] = 'email_otp';
```
cvandeplas, line 35: should this not be below the self-registration?
As self_registration should work independently of email_otp

app/Controller/UsersController.php
```diff
@@ -1224,6 +1209,35 @@ public function login()
...
+      $user['User']['last_login'] = $this->Auth->user('current_login');
+      $user['User']['current_login'] = time();
+      $this->User->save($user['User'], true, array('id', 'last_login', 'current_login'));
+      if (empty($this->Auth->authenticate['Form']['passwordHasher']) && !empty($passwordToSave)) {
+          $this->User->saveField('password', $passwordToSave);
+      }
+      $this->User->Behaviors->enable('SysLogLogable.SysLogLogable');
+      if ($lastUserLogin) {
+          $readableDatetime = (new DateTime())->setTimestamp($lastUserLogin)->format('D, d M y H:i:s O'); // RFC822
+          $this->Flash->info(sprintf('Welcome! Last login was on %s', $readableDatetime));
```
cvandeplas, line 1235: i18n? or is sprintf covered automatically ?

iglocska: Nice catch. Just replace sprintf with __ and all good :)

cvandeplas:
I have no remarks. Just to be sure we'll wait for the green light from Andras before merging

cvandeplas (approved):

SteveClement:
> 
> 
> I have no remarks. Just to be sure we'll wait for the green light from Andras before merging

@iglocska please ack, or not.

PR #5746, opened 2020-03-30
Fixed bugs with PostgreSQL in bruteforce and feed models

4ekin:

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot!

Issue #4796, opened 2019-06-25
GnuPG on CentOS/RHEL not working


SteveClement:
After some debugging the following issue becomes clear.

Tested with and without Passphrase.

[Similar issue](https://stackoverflow.com/questions/47785199/cannot-sign-document-with-crypt-gpg-fails-with-cannot-sign-data-no-passphrase)

Tested with GnuPG 2.0.2 and 2.2.9, permissions are correct.

I have no solution to this issue.

To debug, please feel free to use this [test.php](https://raw.githubusercontent.com/SteveClement/misp-test/master/test.php)

```
[...]

PR #5763, opened 2020-04-03
Set SELinux context for crypt-gpg-pinentry

RichieB2B:
As discussed in #4796 by @butefr crypt-gpg-pinentry needs to be set to httpd_sys_script_exec_t if you want to use a password for the gpg key.


SteveClement:
grazie

Issue #5765, opened 2020-04-04
Downloading events as CSV ignores the flag "Include non-IDS marked attributes" 


patriziotufarolo:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | RedHat
| OS version (client)       | Linux
| PHP version               | 7.4
| MISP version / git hash   | 2.4.124 (5ab0d937273d11967335d7a6dc0803dfe33c11f9) 
| Browser                   | -

### Expected behavior
Similar to #5704 

When downloading a published event as CSV, if the  "Include non-IDS marked attributes" flag is disabled the attributes without the IDS flag = false should be removed from the attributes.

### Actual behavior
All the attributes are actually included in the CSV output, despite of the value of the IDS flag and the "Include non-IDS marked attributes" flag.

### Steps to reproduce the behavior
1) Create an event
2) Add two attributes, one with the IDS flag enabled, one without.
3) Publish the event
4) Download the event as CSV through the left-side menu, without selecting "Include non-IDS marked attributes"
5) Look at the output: all the attributes are included in the output file

Note: this only happens with CSV export.

### Logs, screenshots, configuration dump, ...
![image](https://user-images.githubusercontent.com/4507407/78452115-af305980-7689-11ea-8ada-84277d9bbca3.png)
![image](https://user-images.githubusercontent.com/4507407/78452124-bce5df00-7689-11ea-9e7f-24931a16d0b3.png)
[misp.event.1.csv.txt](https://github.com/MISP/MISP/files/4431441/misp.event.1.csv.txt)

Issue #5704, opened 2020-03-18
Behaviour of 'ignore' flag changed in export attributes


cudeso:
# This template is meant for bug reports, if you have a feature request user the other template.

*If you would like to report a bug, please fill the template bellow*

### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | Ubuntu
| MISP version / git hash   | 2.4.121

### Expected behavior

If the '**ignore**' flag is not specified when exporting data from /events/csv/download/ then only the attributes with to_ids should be included. As per the documentation.
_Setting this flag to true will include attributes that are not marked "to_ids"._

### Actual behavior

If you do not specify the 'ignore' flag, all attributes are exported.
Including the flag, set to '0', only includes attributes with the to_ids flag.

### Steps to reproduce the behavior

```
 curl -X POST -k -H 'Accept: application/json' -H 'Authorization: ' -H 'Content-Type: application/json' -i 'https://misp/events/csv/download/' --data '{"last":"1d", "ignore": "0"}' 

 curl -X POST -k -H 'Accept: application/json' -H 'Authorization: ' -H 'Content-Type: application/json' -i 'https://misp/events/csv/download/' --data '{"last":"1d", "ignore": "0"}' 

```

cudeso:
/events/csv is deprecated, use /events/restSearch and /attributes/restSearch instead

Use for similar behavior **to_ids** and **published**, but would be good to keep the backwards compatibility 

```
"to_ids": [0,1],
"published": [0,1]
```

PR #5767, opened 2020-04-06
fix: [events:export-csv] Default to_ids to be 1

mokaddem:
Even though the recommended way is to use `restSearch`, it lets users use the web interface to download CSVs.
Should fix
- #5704
- #5765


iglocska:
Thanks a lot!

PR #5776, opened 2020-04-08
Removing mentioned stable support for Debian 9

srikwit:
As there is no file `INSTALL.debian9.txt` and we only have a file with the name `xINSTALL.debian9.txt`. The Debian 9 support seems to be experimental.

### DIFF
--- a/INSTALL/README.md
+++ b/INSTALL/README.md
@@ -6,7 +6,6 @@ The text files in this folder are symlink to ../docs - Which is the actual sourc
 
 Currently the following install guides are being tested on a regular basis:
 ```
-INSTALL.debian9.txt
 INSTALL.kali.txt
 INSTALL.ubuntu1804.txt
 ```


SteveClement:
ack

srikwit:
I would also like to mention that the script is very difficult to use in Debian buster when the user does not have `sudo` permissions which also seems to be the default installation difference between Ubuntu and Debian. 

I want to suggest a rewrite of the installer so that the relevant packages for MISP are installed in the newly created `misp` user which is configured with excessive permissions:
`%misp ALL=(ALL:ALL) NOPASSWD:ALL`

However, if the current script is to be maintained, The following binary paths needs to be given permissions from the installing user's perspective on a Debian 10 - buster - stable:
/usr/bin/apt-get, /usr/bin/apt, /usr/sbin/dmidecode, /usr/sbin/useradd, /usr/sbin/chpasswd, /usr/bin/mkdir, /usr/bin/chown, /usr/bin/git, /usr/bin/virtualenv, /usr/bin/php, /usr/sbin/phpenmod, /usr/bin/cp, /usr/bin/chgrp, /usr/bin/systemctl, /usr/bin/tee, /usr/bin/sh, /var/www/MISP/venv/bin/pip, /var/www/MISP/app/Console/cake, /usr/bin/su, /usr/bin/chmod, /usr/bin/wget, /usr/bin/tar, /usr/bin/make, /usr/bin/pecl, /usr/bin/cmake, /usr/sbin/ldconfig, /usr/bin/sed, /bin/pip

PR #5802, opened 2020-04-20
Various fixes on Sharinggroups, Events and correlations

mokaddem:
This PR fixes multiples related bugs on Sharing groups, Events and Correlations.

1. Prevent user not being part of a Sharing group to be able to view the Event, even if they are the owner. This can happen on pulled Events from improperly configured sync setup where the remote user can see the Events (if he/she is site_admin) but is not part of the sharing group.
   - Log entry is created, just in case
2. Similar to 1, Attributes and Objects are virtually removed from the Event if they are distributed with a sharing group where the user does not belong to it
3. Viewing correlation of related event is unaffected, meaning that it's still possible to view correlations (at Event level only) for Attributes that are not displayed. This is intrinsic to the design of the correlations and is meant to be changed upon correlation system revamp
4. Prevent user to save Sharing group if they are not part of it. Closely linked to 1, it is still possible to create the Sharing group if we are in a `local`/`internal` sync setup
5. Trigger re-correlation whenever an Event or Attribute get edited
6. Prevent save of Attribute if an invalid `sharing_group_id` is provided

:warning: Draft Pull Request! More testing need to be done

### DIFF
--- a/app/Controller/AttributesController.php
+++ b/app/Controller/AttributesController.php
@@ -858,6 +858,12 @@ class AttributesController extends AppController
             if (!isset($this->request->data['Attribute'])) {
                 $this->request->data = array('Attribute' => $this->request->data);
             }
+            if ($this->request->data['Attribute']['distribution'] == 4) {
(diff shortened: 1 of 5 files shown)


iglocska:
Thanks a lot!

PR #5809, opened 2020-04-23
Event timeline sightings

mokaddem:
The event timeline now supports sightings visualisation.
If sightings are not pooled regularly at a same rate, we need to extrapolate the range. The strategy used is prioritisation of false positive. Suggestion are of course welcome!

image

Extrapolation strategy:
- If only positive sightings: Will be from first to last sighting (including time.now())
- If both positive and false positive: False positive gets priority. The visual range will be marked as false positive between the two positive sightings.

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -4514,16 +4514,20 @@ class EventsController extends AppController
         if (!in_array($type, $validTools)) {
             throw new MethodNotAllowedException('Invalid type.');
         }
-
         App::uses('EventTimelineTool', 'Tools');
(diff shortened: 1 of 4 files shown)


iglocska:
Looks good to me, any reason to keep it unmerged @mokaddem? Anything in particular you want me to check?

iglocska:
The priority to false positive over positive sounds good to me.

mokaddem:
Reason why it's not merged it to discuss about the strategy to priorities FP. If it's cool, let's merge it ;)

iglocska:
It's cool, go for it!

PR #5811, opened 2020-04-23
Enable notice- and warninglists after misp-wipe

RichieB2B:
I found myself doing this a lot, so time to automate it.
This PR also wipes the tables for the newer MISP features and updates the decayingModel after the wipe.


SteveClement:
thanks, I flipped the defaults.

PR #5818, opened 2020-04-25
fix: Correct flash message when sending e-mail

JakubOnderka:
#### What does it do?

MISP should not lie to users. 

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot

Issue #5766, opened 2020-04-04
Suricata incorrect rules output?


Bryserker:
# This template is meant for bug reports, if you have a feature request user the other template.

*If you would like to report a bug, please fill the template bellow*

### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | Ubuntu
| OS version (client)       | Ubuntu
| PHP version               | 7.2
| MISP version / git hash   | 2.4.XX, 2670e2b443297cf6431bc2a9163ce2963a103694
| Browser                   | N/A

### Expected behavior

Suricata rules cannot have a FQDN as destination, only $EXTERNAL_NET or an IP address.

Example:

alert tcp $HOME_NET any -> $EXTERNAL_NET 443 <SNIP>

### Actual behavior
When downloading the Suricata rules from our MISP server I noticed we had various rules that looked like (example):

alert tcp $HOME_NET any -> github.com 443 <SNIP>

Suricata returned parse errors for these rules like this.

### Steps to reproduce the behavior
I find it hard to answer this as it might be specific to having certain events, I don't have an example I can show.

What I did:

curl --insecure --header "Authorization: <AUTHKEY>" --header "Accept: text/plain" --header "Content-Type: text/plain" -o misp.rules -X GET  https://<MISP_SERVER>/events/nids/suricata/download

And then doing a kill -USR2 <SURICATA_PID>

The suricata log then showed [ERRCODE: SC_ERR_ADDRESS_ENGINE_GENERIC(89)] - failed to parse address for the rules with a FQDN as destination.

### Logs, screenshots, configuration dump, ...

I believe the issue is in:

app/Lib/Export/NidsSuricataExport.php
Line 137: $suricata_dst_ip = $data['host'];

Looking at line 119: $suricata_dst_ip = NidsExport::getCustomIP($data['host']);

In our MISP instance I replaced line 137 with 119 and I no longer got faulty output. I assume this will be the same for line 148 & 165. 

Note: We started testing MISP just yesterday, I might be misunderstanding something. :)

stacsirt:
Just to add to this - seeing the same error here. Seems to have come in since 2.4.124.

I'll change the lines above to see if that works. Unable to do that for third party MISP instances that we get rules from, though.

stacsirt:
Looking at this (thanks for the hint for where to look!), I made a change on a test MISP instance, for the https case in;

app/Lib/Export/NidsSuricataExport.php

[...]

PR #5821, opened 2020-04-27
bugfix in Suricata export template

Cooper-Dale:
PR for reported bug https://github.com/MISP/MISP/issues/5766 based on suggestion @stacsirt, tested on my instance and it is working great

#### What does it do?

If it fixes an existing issue, described here: https://github.com/MISP/MISP/issues/5766

#### Questions

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


iglocska:
Thanks a lot

PR #5827, opened 2020-04-27
fix: [internal] syslog shouldn't end with new line

JakubOnderka:
Because then two lines are logged

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


mokaddem:
Thanks!

PR #5829, opened 2020-04-27
chg: [internal] Removed unused function

JakubOnderka:
## What does it do?

This function has typo in name `beforeValid*e*te`, so its never called. And because everything works, I think it is safe to remove it.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


mokaddem:
Well that's funky.
Moreover, it was also done in `beforeSave`.
Nice catch!

PR #5835, opened 2020-04-28
Event info from Stix2 report

kscheetz:
In the original Stix 2 to MISP converter, it would only set the event info if there was only one report object in the bundle.  Otherwise, a generic title is used.  This change updates the code to use the name of the first report encountered in the bundle.  The report attribute has be changed from a dict to an OrderedDict to preserve order.  If no report is encountered, the generic title is still used but it has been updated to include the file name.


iglocska:
Apologies, had to revert the merge for now, could you reopen the PR? The only issue I see is that it's mixed with an update of the submodules - could you separate that out into a separate PR / not include those? Thanks in advance and sorry for the inconvenience.

PR #5837, opened 2020-04-29
Feature/tags deletion

mokaddem:
Massively based on the work of #5428.
Allows the deletion of tags by passing the parameter
```
deleted: 1
```


tomking2:
Any update on this PR @mokaddem?

mokaddem:
@tomking2 Thanks a lot for your patience!
It took way longer than it should have been..
The deletion of tags on elements (events, attributes, objectAttributes) can be used by editing the element and adding the `deleted` key on the tag.
See #6143 for more information

PR #5847, opened 2020-05-01
Connection test

JakubOnderka:
## What does it do?

Server connection test returns info about client certificate. Usually, the PHP error messages for invalid certification doesn't make sense, so this should provide more info and help to debug problem with connection.

Snímek obrazovky 2020-05-01 v 11 41 34

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Lib/Tools/SyncTool.php
+++ b/app/Lib/Tools/SyncTool.php
@@ -89,17 +89,9 @@ class SyncTool
         }
 
         $params = stream_context_get_params($httpSocket->connection);
-        $certificateDetails = openssl_x509_parse($params['options']['ssl']['peer_certificate']);
-        if (!$certificateDetails) {
-            throw new Exception("Could't parse certificate: " . openssl_error_string());
-        }
-        return [
-            'name' => $certificateDetails["name"],
-            'serial_number' => $certificateDetails["serialNumber"],
-            'valid_from' => new DateTime("@{$certificateDetails["validFrom_time_t"]}"),
-            'valid_to' => new DateTime("@{$certificateDetails["validTo_time_t"]}"),
-            'signature_type' => $certificateDetails["signatureTypeSN"],
-        ];
+        $httpSocket->disconnect();
+
+        return $this->parseCertificate($params['options']['ssl']['peer_certificate']);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


adulau:
I love it. This is a great idea. Would it possible add also the type of TLS handshake negotiated?

JakubOnderka:
> I love it. This is a great idea. Would it possible add also the type of TLS handshake negotiated?

Problem is that PHP do not provide enough information about TLS connection (no info about used ciphers or TLS version) :/  So the only thing that I can add is information about remote certificate.

Snímek obrazovky 2020-05-04 v 10 21 27

SteveClement:
@adulau still ok?

app/Lib/Tools/SyncTool.php
```diff
@@ -60,4 +60,148 @@ public function createHttpSocket($params = array())
...
     }
+
+    /**
+     * @param array $server
+     * @return array|void
+     * @throws Exception
+     */
+    public function getServerRemoteCertificateInfo(array $server)
+    {
+        $httpSocket = $this->setupHttpSocket($server, 5);
```
mokaddem, line 71: Is the 5sec timeout on purpose or is there a reason not to lower it?
Because by putting the header `Connection: Keep-Alive`, `$httpSocket->get ...` only return after 5sec. So it adds an overhead of 5 sec everytime you do a test even though the connection might return directly.

app/webroot/js/misp.js
```diff
@@ -3276,93 +3276,143 @@ function getRemoteSyncUser(id) {
...
         },
         error: function(){
-            $("#connection_test_" + id).html('Internal error.');
-        },
-        success: function(response){
-            var result = response;
+            $("#connection_test_" + id).html('<span class="red bold">Internal error</span>');
+        },
+        success: function(result) {
+            function line(name, value, valid) {
```
mokaddem, line 3287: Could you use a safer way to generate the line like using jQuery's `.text()` function?
Because if one manages to craft a certificate with HTML inside it (which is easy), he/she could do an XSS.

mokaddem (commented):
Hey,
Very nice pull request!
I put some comments, could you have a look?
Thanks a lot!

mokaddem:
Hey @JakubOnderka what is the state of this PR?
Did you have time to look at my comment https://github.com/MISP/MISP/pull/5847#discussion_r425787414

PR #5850, opened 2020-05-02
Updates on debian package for 2.4.125

stricaud:
This updates the debian scripts to be used to generate a debian package. It fixes a few issues I have found over time using the package, updating the package etc. And at the same times, it bumps to the latest MISP version.


iglocska:
Thanks a lot!

PR #5857, opened 2020-05-04
fix: [stix export] Bump minimum CybOX version to 2.1.0.21

RichieB2B:
Because of 65048be032378a6a43bda2b8883f27aef1707545 the old minimal CybOX version 2.1.0.18 will throw an error. All is fine when 2.1.0.21 is used though.


chrisr3d:
Thank you!

PR #5861, opened 2020-05-05
fix: [UI] Always use UUID with capital letters

JakubOnderka:
## What does it do?

Just small UI fix to make things nice.

## Questions

## Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch


SteveClement:
This will probably fix previous PR, @iglocska any thoughts what this would break?

adulau:
:tada:

Issue #5801, opened 2020-04-20
STIX2 export failed


men-menouar:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       |  Ubuntu
| OS version (client)       | 18.04 LTS
| PHP version               |  7.2.24
| MISP version / git hash   |  v2.4.124 (aed771ef4a8f9bda969f0c4c1eaa13b1c4ce007d) 
| Browser                   | firefox

### Expected behavior

Export events on STIX2.

### Actual behavior

Exported files STIX2 contain errors:

- First export we got this error: Error while processing your query: module 'stix2.v20.base' has no attribute 'STIXJSONEncoder'
- Then this one : Error while processing your query: A(n) STIX Object with type 'x-misp-object-yara' already exists and cannot be registered again  

### Steps to reproduce the behavior

Install MISP on fresh ubuntu 18.04 LTS, export events on STIX2 format.

### Logs, screenshots, configuration dump, ...

 cat /var/www/MISP/app/tmp/logs/exec-errors.log --> no error

cudeso:
Similar error on Red Hat Linux 7.8.

`Error while processing your query: module 'stix2.v20.base' has no attribute 'STIXJSONEncoder'`

Export (**Download as...**) to STIX1 in XML and JSON works flawlessly.

In Diagnostics page, all green for STIX/STIX2 modules.

```
STIX: >1.2.0.6
CyBox: >2.1.0.18.dev0
mixbox: 1.0.3
maec: >4.1.0.14
STIX2: >1.2.0
PyMISP: >2.4.120
Other versions might work but are not tested / recommended.
Current libraries status…OK
STIX library version…OK
CYBOX library version…OK
MIXBOX library version…OK
MAEC library version…OK
STIX2 library version…OK
PYMISP library version…OK
```

cudeso:
Also after update to 2.4.125
(
```
- [stix2 export] Fixed STIX JSON Encoder import. [chrisr3d]

  - With the latest update it is no longer part of
    stix2.base but stix2.v20.base by default, so we
    need to import it from stix2.base manually
```
)

cudeso:
@men-menouar check that `MISP/app/files/scripts/stix2/misp2stix2.py` contains the changes as in https://github.com/MISP/MISP/commit/6b669cc60db9def177252f7db380db39447bbccb

[...]

PR #5871, opened 2020-05-06
fix: Fixes STIX2 export bugs when trying to use TLP Tags other than T…

tomking2:
…LP_WHITE, resolve attempted dual registration of custom STIX objects

#### What does it do?

Fixes a couple issues discovered when attempting to use the STIX2.0 export:
1) When using TLP tags on MISP events, anything other than TLP_WHITE would crash out the export as they weren't imported from the STIX library
2) Fixes an existing issue #5801 whereby STIX complains if you try to register the same custom object more than once when having two or more of the same attribute. E.g. A(n) STIX Object with type 'x-misp-object-yara' already exists and cannot be registered again

#### Questions


chrisr3d:
Makes sense, thank you!

PR #5876, opened 2020-05-07
chg: [pubsub] Refactored PubSub tool

JakubOnderka:
## What does it do?

* Allow to password protect ZMQ endpoint.
* ZMQ script using blocking call, so response is faster and uses less CPU.
* Settings and PID file moved to `tmp` folder, with backward compatibility.
* Setting file is readable just by owner (it contains secrets).
* Show user info about number of processed messages.
* ZMQ logs survive restart.
* Other small optimisations.

## Questions


iglocska:
Looks good to me, thanks for taking the old setting store into account!

PR #5882, opened 2020-05-09
Build fix

JakubOnderka:
## What does it do?

This fix broken Travis build.

## Questions


adulau:
Thank you very much!

PR #5885, opened 2020-05-10
chg: [UI] Replace famfamfam flags icons with unicode version

JakubOnderka:
## What does it do?

* Unicode flags are nicer and supports HiDPI displays.
* They don't require additional CSS and PNG file.

## Questions


iglocska:
Hmm, under Firefox it works well, but under Chrome it gets replaced with a unicode 2 character representation :(

JakubOnderka:
> Hmm, under Firefox it works well, but under Chrome it gets replaced with a unicode 2 character representation :(

I just tested it in Firefox :/ I will check also in Chrome, thanks for testing.

iglocska:
Standardisation going wrong again - this is why we can't have nice things.

JakubOnderka:
> Standardisation going wrong again - this is why we can't have nice things.

This screenshot is from latest Chrome on macOS. What kind of OS did you use?

iglocska:
I'm on kubuntu. Is it an OS font level issue? That's even more annoying...

PR #5889, opened 2020-05-11
IPv6 CIDR correlations

### LINKED ISSUE #5877: IPv6 CIDR correlations incorrect
### Work environment

| OS version (client)       | N/A
| MISP version / git hash   | 2.4.125 (be946bf09baf3d2543596bb1586d3e2d72e6c2c5)

### Expected behavior
MISP should correlate IPv6 /32 networks with the corresponding hosts

### Actual behavior
MISP correlates hosts with different networks

1. Enable advanced correlations
2. Create a event with a IPv6 host  (Ex: 2804:14d:1491:90b0:c4f4:b73e:f86c:a653)
3. Create a 2nd event with a CIDR /32 network (Ex: 2804:5330::/32)

see comment below

JakubOnderka:
## What does it do?

* Fixes #5877
* Removed support for `domain-ip` attribute (probably outdated?)
* Fixed removing dot from domain for `domain|ip` type.
* Use faster algorithm for IPv6 CIDR correlations.
* Faster correlation when inserting CIDR attribute (removed duplicated values).
* Some optimisations when creating Redis `misp:cidr_cache_list`.

## Questions

### DIFF
--- a/app/Model/Attribute.php
+++ b/app/Model/Attribute.php
@@ -760,7 +760,7 @@ class Attribute extends AppModel
                 }
             }
         }
-        if (Configure::read('MISP.enable_advanced_correlations') && in_array($this->data['Attribute']['type'], array('ip-src', 'ip-dst', 'domain-ip')) && strpos($this->data['Attribute']['value'], '/')) {
+        if (Configure::read('MISP.enable_advanced_correlations') && in_array($this->data['Attribute']['type'], array('ip-src', 'ip-dst')) && strpos($this->data['Attribute']['value'], '/')) {
             $this->setCIDRList();
         }
         if ($created && isset($this->data['Attribute']['event_id']) && empty($this->data['Attribute']['skip_auto_increment'])) {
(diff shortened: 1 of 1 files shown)


adulau (commented):
Thanks a lot for all the changes, very nice work.

Just wondering the logic. If I have two similar CIDR block 149.13.33.0/24 and 149.13.33.0/24 those won't correlate? Is there a reason to avoid the correlation of the block itself?

JakubOnderka:
> Just wondering the logic. If I have two similar CIDR block 149.13.33.0/24 and 149.13.33.0/24 those won't correlate? Is there a reason to avoid the correlation of the block itself?

If CIDRs are the same, they correlate. But for example 149.13.33.0/24 and 149.13.33.0/32, they are not correlated.

iglocska:
It looks good - just curious, why exclude domain-ip?

JakubOnderka:
> It looks good - just curious, why exclude domain-ip?

Because from the code, it looks like that `domain-ip` type doesn't exists and it is used just in correlation code.

iglocska:
Thanks a lot, awesome work!

PR #5892, opened 2020-05-11
Fixed event freetext feed speedup

JakubOnderka:
## What does it do?

Optimisations of freetext import with a lot of attributes.  Previously, fetching big feeds with a lot of attributes (for example `alienvault reputation generic feed`), could take tens of minutes even when attributes are the same. After this change, it take just tens of seconds. 

It also fixes progressbar for job when fetching just one event and speedups attribute save (when incrementing attributes count, instead of two database queries, just one atomic query is used).

## Questions

### DIFF
--- a/app/Model/Attribute.php
+++ b/app/Model/Attribute.php
@@ -659,20 +659,10 @@ class Attribute extends AppModel
 
     private function __alterAttributeCount($event_id, $increment = true)
     {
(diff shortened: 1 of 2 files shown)


iglocska:
Looks great, thanks, indeed the updateall is infinitely more clever than the old implementation. Thanks a lot!

PR #5895, opened 2020-05-12
chg: [feed] Use https when fetching DGAs feed

JakubOnderka:
## What does it do?

Make DGAs feed little bit safer. 

## Questions


iglocska:
Thanks a lot!

PR #5897, opened 2020-05-12
chg: [feed] Modify value when checking if value exists in current event

JakubOnderka:
## What does it do?

Because for example when fetching Tor exit nodes list and enabled 'delta_merge', for IPv6 addresses MISP delete old addresses (because they are converted to compressed format) and then inserted again.

This takes time and also generates a lot of audit log messages.

## Questions


iglocska:
Awesome idea, digging it a lot. Thanks a lot!

PR #5902, opened 2020-05-13
fix: [ui] Always show full logo for related events box

JakubOnderka:
## What does it do?

*Will add later*

## Questions


mokaddem:
Thanks a lot!

PR #5903, opened 2020-05-13
Correlation speedup

JakubOnderka:
## What does it do?

Speedup loading event page by two changes:
* do not correlate shadow attributes with cached servers if no shadow attribute for event exists
* do not check every attribute if no feed or server is cached

## Questions


iglocska:
Makes sense, thanks!

PR #5906, opened 2020-05-14
fix: [attribute] Do not allow for IPv4 CIDR masklen bigger than 32

JakubOnderka:
#### What does it do?

Do not allow for IPv4 CIDR masklen bigger than 32.

Snímek obrazovky 2020-05-15 v 15 12 07

#### Questions


iglocska:
Thanks a lot!

PR #5907, opened 2020-05-15
fix: allow_disabling_correlation not taken into account

### LINKED ISSUE #5898: Allow disabling correlation setting not taken into account
Running MISP 2.4.124

Despite having the setting MISP.allow_disabling_correlation set to True. When a user attempts to untick the correlation box, he gets an error. I can see in the logs :
[MethodNotAllowedException] Disabling the correlation is not allowed on this instance
/attributes/toggleCorrelation(...) 

I'm willing to work on the resolution of this bug and to submit a PR. Hence creating this issue for tracking purposes.

lfortemps:
#### What does it do?

Fixes #5898 

#### Questions

### DIFF
--- a/app/Controller/AttributesController.php
+++ b/app/Controller/AttributesController.php
@@ -3148,7 +3148,7 @@ class AttributesController extends AppController
 
     public function toggleCorrelation($id)
     {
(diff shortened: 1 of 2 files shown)


mokaddem:
Thanks a lot!

Issue #5379, opened 2019-11-04
HOWTO improve the speed of the correlations table rebuild


github-germ:
2.4.117

We are experimenting with enabling correlations in our dev lab, and decided to rebuild our `correlations` table via the WebUI. Granted we do have too many correlated indicators, and are working to make improvements as part of this effort.

**The rebuild is now in it's 11th day and still not complete!!**

* Yes, there are way too many rows, but why is this so slow?
* Is there something we can tune in the MySQL config to improve the process?
* Open to your suggestions.

Thanks.

```
MariaDB [misp]> select max(id) from correlations;
+------------+
| max(id)    |
+------------+
| 2460615880 |
+------------+
1 row in set (0.00 sec)

MariaDB [misp]> select date_created,progress,message from jobs where progress != 0 and job_type like '%enerate%';
+---------------------+----------+--------------------------+
| date_created        | progress | message                  |
+---------------------+----------+--------------------------+
| 2019-10-24 18:30:59 |       80 | Correlating Event 107938 |
+---------------------+----------+--------------------------+
1 row in set (0.00 sec)

MariaDB [misp]>
```

adulau:
I had to read the output 3 times (ok count() versus max() might a bit different). We are talking about 2.4 billions of correlation? That's really a lot. Have you played with caching? and split the data among instances, feeds and so on.  What's your ratio of correlation over the number of attributes?

Issue #481, opened 2015-05-08
Speed up correlations


elhoim:
Related to a comment in #454, it might be necessary to find a way to speed up correlations.

Just some food for thought: 
Divide the correlation for multiple attribute in small actions (ie: one per attribute being added) put in a work queue and have workers to pick it up to hammer the DB to check for correlations in parallel.

Potential downsides:
- If the server does not run with multiple cores, then it will be useless.
- Correlations are not available directly in the UI for the user.

elhoim:
Any other suggestions on what could be done to speed up correlation?

adulau:
Related to the cache engine as already discussed.

PR #5910, opened 2020-05-15
chg: [correlations] Refactored correlation saving

### LINKED ISSUE #5299: Why does correlations table include the event info?
Curiosity... We notice that there is an `info` column in the `correlations` table in addition to the `1_event_id`. We do see the size of the db influenced by the inclusion of the `info` as correlations grow.

Can the `info` be derived during processing via `1_event_id` or is it there perhaps for performance v. space, ...?

JakubOnderka:
## What does it do?

* Always show other correlating value (useful for CIDR and ssdeep correlations)
* Make correlation saving faster (move more work to database, do not fetch not necessary fields) 
* Fix a bug, that `ip-src|port` also correlate port number
* Remove `info` and `date` columns from `correlations` table and also index for `value` column to make table smaller (fixes #5299)
* Loading correlations is 2x-5x faster (depends on number of related events)
* Should help for #5379 and #481

## Questions

### DIFF
--- a/app/Model/AppModel.php
+++ b/app/Model/AppModel.php
@@ -81,7 +81,7 @@ class AppModel extends Model
         33 => false, 34 => false, 35 => false, 36 => false, 37 => false, 38 => false,
         39 => false, 40 => false, 41 => false, 42 => false, 43 => false, 44 => false,
         45 => false, 46 => false, 47 => false, 48 => false, 49 => false, 50 => false,
-        51 => false, 52 => false, 53 => false, 54 => false
+        51 => false, 52 => false, 53 => false, 54 => false, 55 => false,
     );
 
     public $advanced_updates_description = array(
@@ -1394,6 +1394,9 @@ class AppModel extends Model
                 break;
             case 54:
                 $sqlArray[] = "ALTER TABLE `sightingdbs` MODIFY `timestamp` int(11) NOT NULL DEFAULT 0;";
+                break;
+            case 55:
+                $sqlArray[] = "ALTER TABLE `correlations` DROP `info`, DROP `date`, DROP INDEX `value`;";
                 break;
             case 'fixNonEmptySharingGroupID':
                 $sqlArray[] = 'UPDATE `events` SET `sharing_group_id` = 0 WHERE `distribution` != 4;';
--- a/app/Model/Attribute.php
+++ b/app/Model/Attribute.php
@@ -2010,183 +2010,190 @@ class Attribute extends AppModel
         if (!empty($a['disable_correlation']) || Configure::read('MISP.completely_disable_correlation')) {
             return true;
         }
+
         // Don't do any correlation if the type is a non correlating type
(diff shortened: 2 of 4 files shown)


app/Model/AppModel.php
```diff
@@ -1395,6 +1395,9 @@ public function updateDatabase($command)
             case 54:
                 $sqlArray[] = "ALTER TABLE `sightingdbs` MODIFY `timestamp` int(11) NOT NULL DEFAULT 0;";
                 break;
+            case 55:
+                $sqlArray[] = "ALTER TABLE `correlations` DROP `info`, DROP `date`, DROP INDEX `value`;";
```
iglocska, line 1399: OK, with this we need to be careful. We have correlation tables with 100s of millions of entries so this will be extremely slow, perhaps if we want to go down this route the best would be to truncate the table, alter the table and initiate a recorrelation. Will think about this a bit more on Monday.

iglocska (commented):
Looks very good and should be a massive performance boost / cut down on the correlation table size with the removal of the info/date fields (one of those "sure seemed like a good idea" quirks).

However and the main reason we've delayed this so far, we need to purge the existing correlations for this to work reasonably well, an alter table could take a very long time on large tables. The described steps should make it somewhat less painful though, so if you could make those changes we can move forward with it.

app/Model/AppModel.php
JakubOnderka: Dropping index should be OK ([happens in place and without table rebuild](https://dev.mysql.com/doc/refman/8.0/en/innodb-online-ddl-operations.html)), but yes, dropping columns require table rebuild.

But because recorrelate all attributes can be very slow (much slower that altering table), we can create background job, that will drop columns. Or do not drop them, just don't use them anymore and provide script to drop them manually by MISP admins.

iglocska: It will be slower, but the system remains usable while it is running. We have a background processed system for slow upgrades like that, the only hurdle is that it will create a potentially hour+ long outage on instances for the alter table (we have some larger instances for example that are heavily used and are running on HDDs with massive correlation tables).

So our options are:

1. Truncate, alter table, recorrelate
2. Update system lock, drop indeces, alter table, reindex
3. Keep the fields just stop using them until an admin executes a removal task that will either run 1 or 2.

JakubOnderka: Now I am thinking that the best solution will be recorrelate everything. Because now  there are some fixes that affect correlations:
* port in `ip-src|port` attribute was improperly correlated with other attributes 
* IPv6 CIDR correlations was broken
* Value for CIDR and ssdeep correlation didn't show correlated value
* ssdeep correlation was also broken (#5917)

And because generating correlations should be now faster, I think it makes sense to truncate table, drop columns and run correlation generation.

JakubOnderka:
Closing, dropping columns will be solved in future MR.

PR #5911, opened 2020-05-15
Clean up errors when trying to update warning lists

coolacid:
#### What does it do?

Pre-setups the array so that in the case of a full failure you don't get a bunch of PHP errors. 

#### Questions


mokaddem:
Thanks!

PR #5913, opened 2020-05-16
chg: [kali] Kali linux fixes

SteveClement:


SteveClement:
kali is now working again.

PR #5916, opened 2020-05-17
chg: [correlations] Big speedup when correlating CIDR

JakubOnderka:
#### What does it do?

This change provides massive speed-up when saving CIDR.

For example, when creating correlations for event with 30 IPv4 CIDR IP addresses (and with 200 000 IPv4 addresses in DB), with current version it takes almost 25 to 30 minutes. After this change, it take just 30 seconds.

The main magic is to move comparison logic to database, that is much faster than PHP. The only problem, that used functions are available just in MySQL database, so old logic is still included for PostgreSQL. Also this speedup is not available for IPv6 addresses (it is probably possible to implement it, but IPv6 addresses are still not that usual).

#### Questions


iglocska:
Thanks, clever stuff!

PR #5917, opened 2020-05-18
fix: [correlations] Ssdeep check all chunks

JakubOnderka:
#### What does it do?

Previous conditions for query was defined like this:

```php
array(
    'AND' => array(
        'OR' => array(
            'FuzzyCorrelateSsdeep.chunk_size' => $chunks[0],
            'FuzzyCorrelateSsdeep.chunk_size' => $chunks[0] * 2,
        ),
        'OR' => array(
            'FuzzyCorrelateSsdeep.chunk' => $chunks[1],
            'FuzzyCorrelateSsdeep.chunk' => $chunks[2]
        )
    )
)
```

There was some problems: 
* `FuzzyCorrelateSsdeep.chunk_size` column doesn't exists.
* But that works, because `OR` was defined two times, so first was replaced with other.
* `FuzzyCorrelateSsdeep.chunk` was also defined two times, so just second part of ssdeep hash was used for correlation. 

This change fix all that problems.

#### Questions

### DIFF
--- a/app/Model/FuzzyCorrelateSsdeep.php
+++ b/app/Model/FuzzyCorrelateSsdeep.php
@@ -67,26 +67,19 @@ class FuzzyCorrelateSsdeep extends AppModel
 
     public function query_ssdeep_chunks($hash, $attribute_id)
     {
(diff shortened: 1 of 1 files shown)


mokaddem:
This how I feel.

image

Thanks a lot!

PR #5924, opened 2020-05-19
Fix notices in PHP 7.4

JakubOnderka:
## What does it do?

Bugfixes for PHP 7.4, because PHP 7.4 shows some new notices that fills debug log.

#### Questions


iglocska:
Thanks a lot!

PR #5925, opened 2020-05-19
fix: [UI] Galaxy cluster links should be clickable

JakubOnderka:
## What does it do?

No need to use clipboard anymore. 

## Questions


adulau:
Thank you!

PR #5926, opened 2020-05-19
new: [UI] Add event ID to page table

### LINKED ISSUE #5754: Page titles should be much more descriptive
Rather than simply using `Event` as the title of every event page and soforth, make page titles more descriptive, so pages are easy to identify and differentiate in browser tabs and history.

For example:

- Use that begging of the Info string  as the title for event pages
- Use the tag name for searches by tag

etc...

JakubOnderka:
#### What does it do?

With more tabs, navigation between tabs with different events can be pain, when all of them has the same title.

Partialy fixes #5754.

#### Questions


iglocska:
Nice thanks!

PR #5928, opened 2020-05-21
fix: [UI] Do not show Good-Bye when using custom logout

JakubOnderka:
#### What does it do?

Because without this patch, Good-Bye is show when user successfully log in.

#### Questions


iglocska:
Thanks a lot!

PR #5932, opened 2020-05-22
new: [internal] Use UUID V4 where possible

JakubOnderka:
#### What does it do?

UUID v4 is all random (so with lower probability of collision than v1), faster and do not leak MISP instance IP address.

#### Questions

### DIFF
--- a/app/Console/Command/UserInitShell.php
+++ b/app/Console/Command/UserInitShell.php
@@ -1,4 +1,6 @@
 <?php
+App::uses('UuidTool', 'Tools');
+
 class UserInitShell extends AppShell {
 	public $uses = array('User', 'Role', 'Organisation', 'Server', 'ConnectionManager');
 	public function main() {
@@ -46,7 +48,7 @@ class UserInitShell extends AppShell {
 					'description' => 'Automatically generated admin organisation',
 					'type' => 'ADMIN',
 					'date_created' => $date,
-					'uuid' => CakeText::uuid(),
+					'uuid' => UuidTool::v4(),
 					'local' => 1
 			));
 			$this->Organisation->save($org);
--- a/app/Controller/AppController.php
+++ b/app/Controller/AppController.php
@@ -26,6 +26,7 @@ App::uses('ConnectionManager', 'Model');
 App::uses('Controller', 'Controller');
 App::uses('File', 'Utility');
 App::uses('RequestRearrangeTool', 'Tools');
+App::uses('UuidTool', 'Tools');
 
 /**
  * Application Controller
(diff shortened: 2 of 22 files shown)


iglocska:
CakeText::uuid() uses uuid v4 already.

iglocska:
With that said, you MIGHT come across MISP data that is UUID v1 - simply because we often transpose identifiers from other sources when encoding data.

As an example, we have sensors that produce UUID v1 data, in which case we'll keep their UUID in place and reuse them for the attributes for further tracking / linking.

In the MISP standard we define all UUID fields as something that SHOULD be v4 (but we do not enforce it, however all UUIDs created by MISP are v4).

JakubOnderka:
Ha, thats weird. It is true that `CakeText::uuid()` says that it version UUID v4 (so all parts should be randomly generated), but in thats not true in reality:

* it uses current microtime
* as node ID, it uses IP address or crc32 of hostname and PID number
* it can even leak crc32 of `Security.salt`

```php
if (empty($node)) {
    $node = crc32(Configure::read('Security.salt'));
}
```

See https://github.com/cakephp/cakephp/blob/a4bd6c40ccafcf96b23d892c0f056b919bc8d1b5/lib/Cake/Utility/CakeText.php#L32

RichieB2B:
Auch. Looks like a PR for Cake is in order?

adulau:
Aie. This bug could be even classified as a vulnerability... and get a CVE ;-) @RichieB2B will you contact the guys at CakePHP?

iglocska:
Wtf

iglocska:
Even worse, it looks like they realised this by cake3:

https://github.com/cakephp/cakephp/blob/3.x/src/Utility/Text.php#L61

packet-rat:
Strongly suggest any constraints within MISP should just be an RFC 4122 compliant UUID.  For example:  Many of us have developed Deterministic IDs based on uuidv5 calculations that combine a Name Space, Keys, and immutable IOCs.  There is no value in constraining UUIDs beyond RFC compliance.

iglocska:
Definitely don't want to change the constraints. We have been pushing for
no constraints besides those imposed by RFC 4122 in other formats too for
the same reason as what you are describing. This discussion is purely about
the uuids that are auto-generated by MISP.

RichieB2B:
The uuid fix had landed upstream in [CakePHP](https://github.com/cakephp/cakephp/commit/59e12788fc406ee66180f41e8a2840b841c6051a)

iglocska:
Woah. Hats off!

iglocska:
cakephp bumped, thanks to both of you!

JakubOnderka:
Great, closing this.

PR #5933, opened 2020-05-23
chg: [UI] Nicer icon for form info

JakubOnderka:
#### What does it do?

Looks nicer on HiDPI displays.

#### Questions


iglocska:
Hehe, why not, thanks!

PR #5934, opened 2020-05-23
chg: [internal] Faster removing galaxy cluster tags from attributes

JakubOnderka:
#### What does it do?

Use the same approach as in #5560.

For example, clicking "view all" for event with one attribute take 500 ms instead of 2 seconds before this patch.

#### Questions


iglocska:
Thanks a lot!

PR #5935, opened 2020-05-23
fix: [UI] Bootstrap 2 doesn't support auto position for popover

JakubOnderka:
## What does it do?

Without this change, popover is position to upper top corner of page, thats annoying. 

For supported positions see https://getbootstrap.com/2.3.2/javascript.html#popovers

#### Questions


RichieB2B:
That link actually doesn't document the `placement = 'auto'` feature which confused me. This link does explain it https://getbootstrap.com/docs/4.0/components/popovers/#options

iglocska:
Thanks a lot!

PR #5936, opened 2020-05-24
fix: [internal] Notices in PHP 7.4 for login page

JakubOnderka:
#### What does it do?

I hope last one fix for PHP 7.4.

#### Questions


iglocska:
Thanks a lot!

PR #5942, opened 2020-05-26
Correlation saving

JakubOnderka:
#### What does it do?

This cherry picked two commits from #5910, these commits doesn't require database change. Database change (removing columns) can be done later.

* Always show other correlating value (useful for CIDR and ssdeep correlations)
* Make correlation saving faster (move more work to database, do not fetch not necessary fields) 
* Fix a bug, that `ip-src|port` also correlate port number
* Should help for #5379 and #481

#### Questions


iglocska:
Thanks a lot for making a stop-gap solution until we're ready to merge the column change. Lovely stuff!

PR #5955, opened 2020-05-29
fix: [UI] Module diagnostic colors

JakubOnderka:
#### What does it do?

Fixes super confusion module diagnostic, when „Connection refused” status is green.

Snímek obrazovky 2020-05-29 v 10 05 26

#### Questions


iglocska:
Woops. Quite the oversight, thanks!

PR #5958, opened 2020-05-30
[feed] Add phishstats.info

eschultze:
#### What does it do?

Add phishstats.info phishing URL list as default feed

#### Questions


adulau:
Thank you!

PR #5963, opened 2020-06-01
fix: [mail] Contact reporter body

JakubOnderka:
#### What does it do?

Do not send that GPG or Public key are sent as attachment, when user don't have them.

Also fix sending e-mails in Contact Reporter for when 'Submit only to the person that created the event' is checked.

#### Questions


iglocska:
Thanks a lot!

PR #5964, opened 2020-06-01
fix: [galaxy] Fetch all events for galaxy cluster

JakubOnderka:
#### What does it do?

Without this change, just tags with same case are loaded.

#### Questions


iglocska:
Thanks, makes sense!

PR #5965, opened 2020-06-01
fix: [UI] Show feed caching just for site admins

JakubOnderka:
#### What does it do?

Without this patch, when user is not site admin, for all feeds is showed 'Not cached', that is not true. And it also generates a lot of warnings to debug log.

#### Questions


iglocska:
Thanks a lot.

PR #5967, opened 2020-06-02
Decaying v2

mokaddem:
#### UserSetting
- New setting `default_restsearch_parameters` allowing users to supply restSearch parameters that will automatically be passed during API fetch. The main purpose of this new setting is to enable users to inject filters when integrating with third-party tools not offering the possibility to control the query performed against MISP.
image
- New setting `tag_numerical_value_override` allowing users to override the `numerical_value` of tags. The main purpose of this new setting is to let user have their own numerical value for tags. It does not constrain site-admin to diverge from the official misp-taxonomy repository and let them define new values for entries not possessing having one.
image
image

#### Attribute's `last_seen` will now takes precedence over the Attribute's `timestamp` if the former is set
In the decaying implementation prior to this pull request, if no sightings were recorded, the simulated last sighting was set on the `timestamp` value. However, once this pull request will be merged, the `last_seen` value will be used instead. Users will be able to alter Attributes (attach tags, modify `last_seen`, ...) without refreshing the decaying score to its maximum value.

### DIFF
--- a/app/Model/DecayingModel.php
+++ b/app/Model/DecayingModel.php
@@ -309,8 +309,9 @@ class DecayingModel extends AppModel
                                 unset($taxonomies[$namespace]['TaxonomyPredicate'][$p]['TaxonomyEntry'][$e]);
                             } else {
                                 $tag_name = sprintf('%s:%s="%s"', $taxonomy['namespace'], $predicate['value'], $entry['value']);
-                                if (isset($tags[strtoupper($tag_name)])) {
-                                    $taxonomies[$namespace]['TaxonomyPredicate'][$p]['TaxonomyEntry'][$e]['Tag'] = $tags[strtoupper($tag_name)]['Tag'];
+                                $upperTagName = strtoupper($tag_name);
+                                if (isset($tags[$upperTagName])) {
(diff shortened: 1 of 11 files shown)


adulau:
Great updates and it makes a lot of sense! Thanks a lot. Is there something fundamental which blocks the merge of the decay v2?

app/Model/DecayingModel.php
```diff
@@ -341,8 +349,17 @@ public function listTaxonomiesWithNumericalValue($create_non_existing_tags=1)
                                     'colour' => 'grey',
                                 );
                             }
-                            $taxonomies[$namespace]['TaxonomyPredicate'][$p]['Tag']['numerical_value'] = $predicate['numerical_value'];
                             $taxonomies[$namespace]['TaxonomyPredicate'][$p]['numerical_predicate'] = true;
+                            $taxonomies[$namespace]['TaxonomyPredicate'][$p]['Tag']['numerical_value'] = $predicate['numerical_value'];
```
iglocska, line 353: perhaps storing the tag name uppercased in a var once would make this cleaner

PR #5972, opened 2020-06-03
chg: [UI] Make Enrichment Results little bit nicer

JakubOnderka:
#### Questions


iglocska:
Thanks a lot!

PR #5973, opened 2020-06-03
Soft-delete Attributes when performing a feed delta-merge

### LINKED ISSUE #5949: Soft delete of attributes don't propagate across MISP instance syncing when they originate from a feed
*If you would like to report a bug, please fill the template bellow*

### Work environment

| OS version (server)       | ubuntu0.18.04.3 
| OS version (client)       | Win10
| PHP version               | 7.2.24-0
| MISP version / git hash   | 2.4.124

### Expected behavior
MISP instance A is ingesting external feeds such as Firehol level 1 using the delta-merge functionality to a static event. Attributes associated with that event are syncing as expected with new ones added and dead ones being soft deleted.

MISP instance A does a push/pull sync with MISP instance B.

It is expected that the event corresponding with the feed will be mirrored on MISP instance B.

### Actual behavior
New attributes associated with the feed event on instance A are being propagated on instance B as expected.
Dead attributes associated with the feed are being soft-deleted on instance A. Those soft deletes do not propagate on instance B.

Configure syncing between two MISP instances.
Configure feed event on the first instance.
Allow feed event to propagate on the second instance.
Update feed to remove some of the original attributes.
Allow first instance to perform delta-merge with feed.
Allow second instance ton sync with the first instance.

mokaddem:
When performing a feed delta-merge save, Attribute where hard-deleted. This PR soft-delete them.
It introduces an overhead that seems to be needed:
- Fetching all the Attributes (so that we have a trace in the logs)
- Iterating to change the `deleted` value
- Saving them (We need to execute the callback methods - `updateAll()` doesn't do it)
If you see a cleaner/faster way to do it. Please tell me :)

Fix https://github.com/MISP/MISP/issues/5949.

### DIFF
--- a/app/Model/Feed.php
+++ b/app/Model/Feed.php
@@ -996,8 +996,16 @@ class Feed extends AppModel
                 }
             }
             if ($feed['Feed']['delta_merge'] && !empty($existsAttributesValueToId)) {
-                $to_delete = array_values($existsAttributesValueToId);
-                $this->Event->Attribute->deleteAll(array('Attribute.id' => $to_delete, 'Attribute.deleted' => 0));
+                $attributesToDelete = $this->Event->Attribute->find('all', array(
+                    'conditions' => array(
+                        'Attribute.id' => array_values($existsAttributesValueToId)
+                    ),
+                    'recursive' => -1
+                ));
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks a lot!

PR #5974, opened 2020-06-03
Base URL fix #4610 - v2

vpiserchia:
#### What does it do?
This is an updated version of https://github.com/MISP/MISP/pull/4610

Fixes issue #4535, providing support for MISP to work in a subdirectory:
https://example.com/misp

#### Questions


iglocska:
Thanks for all the excellent work and for your patience with getting this merged!

I've found one tiny issue, rest looks good, I've applied a fix for that. 

One thing to note is that the two cake directives configured in bootstrap.default.php will not be used by existing installations by default - this needs to be modified for installations that already have their bootstrap.php configured from the installation.

PR #5985, opened 2020-06-05
fix: [mail] Fix body of passwordReset/newUser emails

vaclavbartos:
When MISP sends an email with new credentials, the body is generated from one of the configured templates - passwordResetText or newUserText. However, these two templates were swapped - the newUserText was used for password reset, while passwordResetText was used when new account is created.

This PR fixes it.

#### Release Type:

- [ ] Major
- [ ] Minor
- [x] Patch


iglocska:
Nice catch thanks!

PR #5988, opened 2020-06-05
fix: [internal] Removing attributes from empty event

JakubOnderka:
## What does it do?

In rare occasion, number attributes in `attribute_count` field can be different than real state. If `attribute_count` is already zero and we try to decrement that value, database returns type error (it is not possible to decrement unsigned value). This fix the behaviour.

#### Questions


iglocska:
Thanks a lot!

PR #5991, opened 2020-06-07
chg: [internal] Drop correlations indexes

JakubOnderka:
#### What does it do?

Removes `value` index from correlations table. This index is not used in any SQL query, it just takes database space (it can be around 20 % of table size) and computation time when inserting new correlation.

Also removes `org_id`, `sharing_group_id` and `a_sharing_group_id` indexes. These indexes are smaller, but still big and can be used just theoretically.

Removing index is in place operation, but to returning space to operation system, it is necessary to optimise correlations table, that must be run manually.

#### Questions


iglocska:
Thanks, nice catch!

PR #5992, opened 2020-06-07
chg: [internal] Log exception if exception is thrown during event dow…

JakubOnderka:
#### What does it do?

If something happen during remote event downloading, log exception.

#### Questions

### DIFF
--- a/app/Model/Event.php
+++ b/app/Model/Event.php
@@ -1424,6 +1424,15 @@ class Event extends AppModel
         return $attribute;
     }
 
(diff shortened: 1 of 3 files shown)


iglocska:
To me this looks mergable as is and makes a lot of sense - if you are ready with it just flip it an active PR and I'll merge it :) Thanks for the awesome work as always!

iglocska:
Lookd htrsy yo mr, yhsnkd@

PR #5993, opened 2020-06-07
fix: [pull] Correct progress for pull job

JakubOnderka:
#### What does it do?

Before this, progress can be logged to different job.

#### Questions


iglocska:
Thanks a lot!

PR #5999, opened 2020-06-09
new: [cli] Command for pulling from all remote servers

JakubOnderka:
#### What does it do?

Add ability to pull events by CLI command from all remote servers with enabled pulling. 

I am just not sure, if it is OK to enqueue jobs from CLI, because I think it is not used anywhere else. 

#### Questions


iglocska:
Thanks a lot!

Issue #6006, opened 2020-06-10
Event - Sharing Group null variable error


imidoriya:
Getting this error - looks like an unset variable `$sgsids` being null while defining the sharing group.   Line 6625 references `$sgsids`, but was probably intending to be `$sgids`, which is defined as a variable in the function.  Looks like it got missed when those variable names were changed on Sep 4, 2017 via https://github.com/MISP/MISP/commit/40ea22a272c536e68ad6feea10e0839a4d5a9579

### Error Message
```
2020-06-10 22:20:41 Warning: Warning (2): in_array() expects parameter 2 to be array, null given in [/var/www/MISP/app/Model/Event.php, line 6625]
Trace:
ErrorHandler::handleError() - APP/Lib/cakephp/lib/Cake/Error/ErrorHandler.php, line 230
in_array - [internal], line ??
Event::__attachReferences() - APP/Model/Event.php, line 6625
Event::fetchEvent() - APP/Model/Event.php, line 2177
EventsController::view() - APP/Controller/EventsController.php, line 1653
ReflectionMethod::invokeArgs() - [internal], line ??
Controller::invokeAction() - APP/Lib/cakephp/lib/Cake/Controller/Controller.php, line 499
Dispatcher::_invoke() - APP/Lib/cakephp/lib/Cake/Routing/Dispatcher.php, line 193
Dispatcher::dispatch() - APP/Lib/cakephp/lib/Cake/Routing/Dispatcher.php, line 167
[main] - APP/webroot/index.php, line 92
```

### Event.php - Line 6625
https://github.com/MISP/MISP/blob/ac6ed3235df25bc375d84dbb0100e763bf40b12a/app/Model/Event.php#L6611-L6636

### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | Ubuntu
| PHP version               | 7.4
| MISP version / git hash   | 2.4.126, ac6ed3235df25bc375d84dbb0100e763bf40b12a

PR #6007, opened 2020-06-11
Fix issue #6006 - sgsids is never set

imidoriya:
#### What does it do?
The value `$sgsids` is never set.  I expect it should be `$sgids` based on the function variable.  I believe this will fix the error described in #6006

#### Questions

- [ ] Does it require a DB change? **No**
- [ ] Are you using it in production?  **Not yet**
- [ ] Does it require a change in the API (PyMISP for example)?  **No**


iglocska:
Nice catch, thanks!

PR #6015, opened 2020-06-13
Removed hosts-file.net feeds from default feeds

GlennHD:
Malwarebytes has discontinued the feed:  https://forums.malwarebytes.com/topic/258056-hosts-filenet-domain-lists-are-broken-what-happened/


iglocska:
Thanks!

PR #6019, opened 2020-06-16
add [widget] Authentication failure widget

gallypette:
#### What does it do?

This widget make use of authentication-failure-report objects https://github.com/MISP/misp-objects/pull/262 contained in an "Authentication Failure Daily Event" to present bargraphs representations of its content.

Think Covid widget for authentication failures:
d4_ssh_widget

Just in time to miss .127

#### Questions


iglocska:
Thanks a lot!

PR #6022, opened 2020-06-17
STIX parsing updates

chrisr3d:
Updates on the stix 1 & 2 parsing, including:
- a reworked stix2 import
- updates on the stix1 & 2 exports
- updates on the stix1 import
- cleaner code in general
- a better way to import galaxies from external stix files, using a synonyms mapping dictionary
  - also galaxies are always imported with their tag name
  - if a known galaxy is imported, it will be displayed in the MISP event as a galaxy once the file is imported
  - if an unknown threat actor, malware, tool, etc. is imported but unknown as a galaxy in MISP, it will remain as tag (possible improvement with the future custom galaxies feature)
- updated test scripts & sample files


iglocska:
Holy crap that's a huge change. It indeed looks much cleaner, especially the mappings. Thanks a lot for the tremendous work and hope your sanity is still intact after this much stix!

PR #6026, opened 2020-06-18
Save sightings optimisation

JakubOnderka:
#### What does it do?

Small optimisations of sightings saving.

#### Questions


iglocska:
Looks good, should be a nice upgrade!

PR #6028, opened 2020-06-18
fix: [UI] Typo

JakubOnderka:
#### Questions


iglocska:
Thanks! Looks like the copy-pasta monster reared its ugly head again, nice catch ;)

PR #6029, opened 2020-06-18
MISP PUSH is not working as expected

srikanthprathi:
Created this PR by mistake


srikanthprathi:
> MISP Version: 2.4.126
> 
> I am having 45K events in my local MISP Instance and I tried to push them to the remote MISP instance but it is not happening.
> I am able to test the connection with the target remote MISP instance and used the AuthKey of the SyncUser role.
> When I push the events, the job is getting created and it is getting completed just in 5 seconds but in the target remote MISP instance, the events are not getting created.
> 
> Note: When there were 10k events, it worked fine but I am facing this issue when tried with 45k events.

srikanthprathi:
Closing this as I have created this by mistake

PR #6035, opened 2020-06-19
chg: [internal] Attribute REST search optimisations and error handling

JakubOnderka:
#### What does it do?

Small attribute REST search optimisations and better error handling.

#### Questions


iglocska:
Very cool idea. This should help loads with smaller queries, thanks!

PR #6036, opened 2020-06-20
chg: [feed] Faster freetext feed caching

JakubOnderka:
#### What does it do?

Faster freetext/csv feed caching by pipelining Redis communication.

#### Questions


iglocska:
Thanks a lot!

PR #6043, opened 2020-06-22
Update side_menu.ctp

StefanKelm:
Adjusts menu to be in line with "Global Actions"

#### What does it do?

Adjusts menu to be in line with "Global Actions"

#### Questions


iglocska:
Thanks a lot!

PR #6044, opened 2020-06-22
chg: [UI] Sort tags by name for server rules

JakubOnderka:
#### What does it do?

Make easier to find tags.

#### Questions


iglocska:
Thanks a lot!

PR #6045, opened 2020-06-22
chg: [internal] Faster loading sightings if the same attribute is req…

JakubOnderka:
#### What does it do?

In our case, it speed us loading attribute sighting graph from 1 minute to 478 ms.

#### Questions


iglocska:
Thanks a lot!

PR #6052, opened 2020-06-23
Using json parser to parse json configuration output from cake

stricaud:
This PR improves misp-config to use the json parser from Perl instead of reading line by line with a regex. This improves accuracy and avoid to have to decode json encoded string.


stricaud:
any love here?

iglocska:
Thanks a lot, sorry completely misunderstood what this was about :)

PR #6058, opened 2020-06-25
chg: [internal] Remove attributes.object_relation index

JakubOnderka:
#### What does it do?

To save disk size and CPU when inserting attributes.

#### Questions


iglocska:
I think it is a possible filter parameter on the attribute level rest search. If not it should be, object relation is basically a secondary type field when we're talking about object attributes.

JakubOnderka:
OK, closing then.

PR #6060, opened 2020-06-26
fix: [UI] Expanding attribute correlations on other pages

JakubOnderka:
#### What does it do?

Before, the link works just for first page, but after loading for example second page, the link didn't work anymore.

#### Questions


iglocska:
Thanks a lot!

PR #6061, opened 2020-06-26
fix: [internal] Do not create empty link for anonymized org sighting

JakubOnderka:
#### What does it do?

Just HTML code fix.

#### Questions


iglocska:
Thanks a lot!

PR #6062, opened 2020-06-26
fix: [feed] Incorrect call in Feed::__saveEvent

JakubOnderka:
#### What does it do?

Second parameter for `Event::_edit` method should be user array, but boolean value was given.

It also set more typehints to check incorrect function calls.

#### Questions


iglocska:
Nice catch, thanks!

PR #6063, opened 2020-06-26
fix: [UI] Fetching from not enabled feed should be error

JakubOnderka:
#### What does it do?

Info looks like it runs in background.

#### Questions


iglocska:
Good point, thanks!

PR #6064, opened 2020-06-26
Accessibility: ARIA properties for the "fix" button in the DB diagnostic tables

obert01:
Added proper ARIA properties for the "fix" button in the DB index and DB schema diagnostic tables.


iglocska:
Thanks a lot!

PR #6067, opened 2020-06-26
fix: [attributes] Possible duplicate attributes

JakubOnderka:
#### What does it do?

Before this patch, it was possible to create two attributes with same content if text contains `|` character.

It also contains small performance fixes.

#### Questions


iglocska:
Well spotted, thanks!

PR #6069, opened 2020-06-27
fix: [internal] Missing field for server model when editing event

JakubOnderka:
#### What does it do?

Field is provided for `_add` method, but missing for `_edit` method.

#### Questions


iglocska:
Woah, nice catch.

PR #6070, opened 2020-06-28
Cidr correlation optim

JakubOnderka:
#### What does it do?

Small optimisation when inserting IP address with advanced correlations enabled.

#### Questions


iglocska:
Thanks a lot!

PR #6071, opened 2020-06-28
fix: [stix] Store synonymsToTagNames.json file in tmp folder

JakubOnderka:
#### What does it do?

`script` folder is not required to be writeable, so storing in that location can silently fail.

#### Questions


iglocska:
Thanks a lot!

PR #6075, opened 2020-06-29
fix: [internal] Do not try to access bool as array

JakubOnderka:
#### What does it do?

This generates a lot of notices in PHP 7.4.

#### Questions


iglocska:
Thanks a lot!

PR #6078, opened 2020-06-30
chg: [ACL] Allow to access to fetchOrgsForSG and fetchServersForSG...

JakubOnderka:
…just with `perm_sharing_group`.
#### Questions


iglocska:
Thanks a lot!

PR #6079, opened 2020-07-01
Allow you to mannually set the API key for automation purposes

legoguy1000:
#### What does it do?

This PR adds the ability to set a user's API key to a user defined value using the CLI. This is useful, mainly in the situation where u are deploying MISP alongside Cortex and The Hive via automated/unattended deployment methods and want the integration setup without user intervention.

#### Questions

- [X] Are you using it in production? Sort of.  I just created this today and tested it and it works but haven't actually merged it into our Helm chart yet since i figured I would submit the PR here first.

#### Release Type:
- [ ] Major
- [ ] Minor
- [X] Patch

### DIFF
--- a/app/Console/Command/AuthkeyShell.php
+++ b/app/Console/Command/AuthkeyShell.php
@@ -14,7 +14,7 @@ class AuthkeyShell extends AppShell {
     public function main()
     {
         $this->ConfigLoad->execute();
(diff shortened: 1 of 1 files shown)


iglocska:
Sounds like a good idea, thanks!

PR #6085, opened 2020-07-02
fix: [audit] Show all attribute changes in event history

JakubOnderka:
#### Questions


iglocska:
It looks good, even if it doesn't fix the issue reported, including modifications / removal of the deleted attributes makes sense. I'd say go for it!

JakubOnderka:
> It looks good, even if it doesn't fix the issue reported, including modifications / removal of the deleted attributes makes sense. I'd say go for it!

True, but I think for tracking all changes we need to add new column to logs table with event ID.

PR #6086, opened 2020-07-03
chg: [internal] Reduce code complexity by using JsonResponse

JakubOnderka:
#### What does it do?

Reduce code complexity by using new `JsonResponse` class

#### Questions


iglocska:
Perhaps something we could do instead through https://github.com/MISP/MISP/blob/2.4/app/Controller/Component/RestResponseComponent.php

JakubOnderka:
> Perhaps something we could do instead through https://github.com/MISP/MISP/blob/2.4/app/Controller/Component/RestResponseComponent.php

So you mean something like this?
```php
return $this->RestResponse->viewData($data, 'json');
```

iglocska:
Yup!

PR #6087, opened 2020-07-03
new: [diag] Check if ZIP extension is installed

JakubOnderka:
#### What does it do?

Zip extension is required for fetching ZIP compressed feeds.

#### Questions


iglocska:
Thanks a lot!

PR #6088, opened 2020-07-04
fix: [UI] Attribute category select

JakubOnderka:
#### What does it do?

Before:
Snímek obrazovky 2020-07-04 v 10 36 16

After:
Snímek obrazovky 2020-07-04 v 10 35 59

#### Questions


iglocska:
Nice one thanks a lot

PR #6092, opened 2020-07-05
chg: [internal] Simplify and optimise eventUI method

JakubOnderka:
#### What does it do?

Code cleanup and faster UI for events with a lot of attributes (removed computing values that we don't need or we already have).

#### Questions


iglocska:
Looks good, go for it!

iglocska:
Thanks a lot!

PR #6093, opened 2020-07-06
fix: [security] Remove ShadowAttributesController::{fetchEditForm,editField}

JakubOnderka:
#### What does it do?

These methods are not used, but they allow to access attribute data without proper ACL checks.
#### Questions


iglocska:
Thanks a lot!

PR #6094, opened 2020-07-06
fix: [MispObject] Do not unpublish synced events, fixes #4838

### LINKED ISSUE #4838: Events keep being updated without any changes
We have a lot of events that keep getting updated on external servers without any of the metadata or attributes actually changing. One example is 5a6f379d-3854-4457-949e-41bb950d210f with the event history clearly showing the problem: https://misp.circl.lu/logs/event_index/10616

A result of this is that the timestamp of these events are always recent while the attributes (and threats) are actually quite old. Any thoughts on why this is happening?

RichieB2B:
This is a very safe fix for #4838 
See https://github.com/MISP/MISP/issues/4838#issuecomment-651739179 for further details.

### DIFF
--- a/app/Model/Event.php
+++ b/app/Model/Event.php
@@ -3691,7 +3691,7 @@ class Event extends AppModel
             $referencesToCapture = array();
             if (!empty($data['Event']['Object'])) {
                 foreach ($data['Event']['Object'] as $object) {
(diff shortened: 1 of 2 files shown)


iglocska:
Thanks a million, this tiny patch hides a crazy amount of digging, awesome stuff! ;)

PR #6095, opened 2020-07-06
fix: [security] Remove ShadowAttributesController::{getProposalsByUuid,getProposalsByUuidList}

JakubOnderka:
#### What does it do?

These methods are not used, but they let sync users to access proposals for any event.

#### Questions


iglocska:
Thanks a lot!

PR #6104, opened 2020-07-08
fix: Remove interoperability parameter from stix2.parse call

edcuba:
#### What does it do?

Remove unsupported parameter from stix2.parse API calls.

See: https://stix2.readthedocs.io/en/latest/api/stix2.parsing.html#stix2.parsing.parse

Interoperability parameter is not recognized by stix2 api and throws an error.

```
Traceback (most recent call last):
  File "/var/www/MISP/app/files/scripts/stix2/stix2misp.py", line 2037, in <module>
    main(sys.argv)
  File "/var/www/MISP/app/files/scripts/stix2/stix2misp.py", line 2029, in main
    event = stix2.parse(f.read(), allow_custom=True, interoperability=True)
TypeError: parse() got an unexpected keyword argument 'interoperability'
```

Stix2 JSON import starts to work after this fix.

#### Questions

### DIFF
--- a/app/files/scripts/stix2/stix2misp.py
+++ b/app/files/scripts/stix2/stix2misp.py
@@ -2033,7 +2033,7 @@ def from_misp(stix_objects):
 def main(args):
     filename = Path(os.path.dirname(args[0]), args[1])
     with open(filename, 'rt', encoding='utf-8') as f:
-        event = stix2.parse(f.read(), allow_custom=True, interoperability=True)
+        event = stix2.parse(f.read(), allow_custom=True)
     stix_parser = StixFromMISPParser() if from_misp(event.objects) else ExternalStixParser()
     stix_parser.handler(event, filename, args[2:])
     stix_parser.save_file()
--- a/app/files/scripts/stixtest/stix2_check.py
+++ b/app/files/scripts/stixtest/stix2_check.py
@@ -22,11 +22,11 @@ def externalise_event(event):
 
 def get_external(event):
     externalise_event(event)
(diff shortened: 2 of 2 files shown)


chrisr3d:
I can suggest you to use the stix2 python library you can find submoduled in MISP:
```
cd /var/www/MISP
sudo -H -u www-data git pull origin 2.4
sudo -H -u www-data git submodule update --init
cd cti-python-stix2
sudo -H -u www-data /var/www/MISP/venv/bin/pip install -U .
```
We use this version to avoid issues if ever you import STIX2 files containing non v4 uuids

mokaddem:
Can this PR be closed?

chrisr3d:
Indeed I guess it can be closed for now, if anything goes wrong following the previous instructions, please let me know

edcuba:
@chrisr3d thank you, building the submoduled version works.
Maybe it would be good to try to merge the custom changes (e.g. interoperability param support) to cti-python-stix2 master.

chrisr3d:
Well the story behind this additional parameter is simply we had a user facing issues to import their STIX2 files into MISP because of a uuid support issue. It turned out we did not want users to be blocked because of non v4 uuids and ended up having our own fork I keep up to date.

Now regarding uuids, the library is already a bit more lax and supports more uuid types, but it is still not what we expected so we are waiting for further changes or new ways to get around the issue.

I did submit a [pull request](https://github.com/oasis-open/cti-python-stix2/pull/235), but as it does not meet the requirements of the standard, it cannot be merged
If ever keeping our changes up to date is no longer useful and we can simply use the original library, for sure we will use so we avoid confusion for users

PR #6108, opened 2020-07-10
fix: [misp_retention] Support objects, use lists for build_complex_qu…

RichieB2B:
Pretty much as the title says. The script was broken because of https://github.com/MISP/PyMISP/commit/901afb32d914608e7f11d8e2bdf61440318c963e now it is working again. And it supports attributes in objects. :-)


iglocska:
Thanks a lot :)

PR #6109, opened 2020-07-10
fix: [stix export] log stack trace on error, support 'AMBER NATO ALLI…

RichieB2B:
This patch does 3 things:

1. Silently ignore missing temp files during delete phase
1. Log a python stack trace in exec-errors.log when misp2stix.py breaks
1. Map 'AMBER NATO ALLIANCE' to AMBER in Stix export

I needed 1 and 2 to fix 3 on our server..

### DIFF
--- a/app/Lib/Export/StixExport.php
+++ b/app/Lib/Export/StixExport.php
@@ -131,7 +131,11 @@ class StixExport
     {
         foreach ($this->__filenames as $f => $filename) {
             if ($index >= $f) {
-                unlink($this->__tmp_dir . $filename);
+                try {
+                    unlink($this->__tmp_dir . $filename);
+                } catch (Exception $e) {
+                    unset($e);
+                }
             }
         }
         $this->__stix_file->close();
--- a/app/files/scripts/misp2stix.py
+++ b/app/files/scripts/misp2stix.py
@@ -7,6 +7,7 @@ import datetime
 import re
 import ntpath
 import socket
+import traceback
 from misp2stix_mapping import *
 from collections import defaultdict
 from copy import deepcopy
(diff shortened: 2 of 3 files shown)


iglocska:
Thanks a lot!

iglocska:
Also, mumbl, grumbl, custom TLP tags. 

/cc @cvandeplas

app/Lib/Export/StixExport.php
```diff
@@ -131,7 +131,11 @@ private function __delete_temporary_files($index)
     {
         foreach ($this->__filenames as $f => $filename) {
             if ($index >= $f) {
-                unlink($this->__tmp_dir . $filename);
+                try {
+                    unlink($this->__tmp_dir . $filename);
+                } catch (Exception $e) {
```
JakubOnderka, line 136: @RichieB2B I think `unlink` never throw exception, so this catch will be never called. If you want to suppress error for non exists file, you should use [`@` operator](https://www.php.net/manual/en/language.operators.errorcontrol.php) or check file existence first.

RichieB2B: Good catch! It is indeed a warning which can be suppressed using the @ operator.
```
$ cat unlink.php 
<?php
unlink('/tmp/foobar');
try {
  unlink('/tmp/barfoo');
} catch (Exception $e) {
}
@unlink('/tmp/foofoo');
$ php unlink.php 
PHP Warning:  unlink(/tmp/foobar): No such file or directory in /tmp/unlink.php on line 2
PHP Warning:  unlink(/tmp/barfoo): No such file or directory in /tmp/unlink.php on line 4
```

PR #6110, opened 2020-07-11
fix: [StixExport] suppress unlink warnings

RichieB2B:
unlink throws a warning, not an error as mentioned by @JakubOnderka in #6109


iglocska:
Thanks a lot

PR #6118, opened 2020-07-14
Additionnal protection against XSS

lfortemps:
The response type defaults to html while it should be JSON. This avoids any HTML reflected input to be interpreted by the browser.


cvandeplas:
thank you very much !

PR #6120, opened 2020-07-14
[feature] Cluster relations and synchronization - aka Galaxy 2.0

mokaddem:
## Galaxy 2.0
Introduces various new features for *Galaxies* and their *Clusters* allowing the creation of **custom** *Clusters*, **connection** via *Relations* and **synchronization** to connected instances.

### Major features
- CRUD (Create, Read, Update and Delete) for *Clusters* and their *Relations*
    - New *Cluster* fields: `distribution`, `Org`, `Orgc`, `published`
    - *Cluster* fork capabilities, very similar to the *extended event* feature
    - **Tagged** *Relations* with their own distribution level between *Clusters*
    - Deletion
        - Similar to *Events*, *Clusters* have their own block-list
        - Similar to *Attributes*, *Clusters* can be soft or hard deleted
- Synchronization
    - Own synchronization mechanism which can be enabled with the `pull_galaxy_cluster` and `push_galaxy_cluster` flags
        - `Pull All`: Pull all remote *Clusters*
        - `Pull Update`: Update local *Clusters*
        - `Pull Relevant`: Fetches *Clusters* based on local *Tags*
        - `Push`: Triggered whenever a *Cluster* is published
    - Whenever an *Event* containing a custom *Cluster* is published, it will be pushed as well
- Visualization
    - Tree view of forked *Clusters*
    - Tree and network views for *Relations* between *Clusters*

# Screenshots

Screenshot from 2020-07-14 16-56-01
Screenshot from 2020-07-14 17-01-01
Screenshot from 2020-07-14 17-02-11


iglocska:
I'll go through some (often minor) issues while testing here:

iglocska:
image
Missing menu entry on the side and form elements not sized correctly.

iglocska:
When you are adding your own cluster and click toggle UI - clicking on add element will create an empty key/value pair which is correct. However, if left unfilled, the empty key/value pair is added to the dictionary.

iglocska:
Inconsistent naming in the UI: Fork vs extension - we should pick one and stick with it, it's a bit confusing currently (I fork a cluster then I see that my cluster is extending something / the original is extended by something).

iglocska:
Adding a relationship fails due to invalid group by statement:

image

iglocska:
When wanting to add a relationship, the target uuid picker barfs when trying to select a cluster from an att&ck like matrix

image

iglocska:
When trying to edit a default galaxy cluster, an error is thrown about it not being editable. Best would be to hide the edit button in the first place.

image

JakubOnderka:
- [x] Galaxy cluster blocklist add form do not provide validation, it is possible to save empty form
- [x] Snímek obrazovky 2020-07-27 v 16 19 48 – MISP images links lead to invalid org page.
- [x] Raw Galaxy Cluster export should return `application/json` MIME type.
- [x] Snímek obrazovky 2020-07-27 v 16 30 32 – it is not clear what user should put to "Relationship Type" field
- [x] After adding relationship, user should be redirected to relationship index page.
- [x] "Galaxy Cluster target" column on Relationship index page is empty.
- [x] It is not possible to edit galaxy cluster relationship. Editing ends with `Relationship could not be added. Reason: ["galaxy_cluster_id not provided"]`  error message.
- [x] When adding new Galaxy cluster relationship, selected tags are not saved.

mokaddem:
@JakubOnderka fixed all points you mentioned unless this one as I couldn't reproduce
>  When adding new Galaxy cluster relationship, selected tags are not saved.

Would you mind describing how you got the error?
Thanks!

JakubOnderka:
* [ ] <br> when saving invalid cluster relationship, no information what is wrong is provided. 
* [ ] When sending empty blocklist form, **green** flash message is shown with message „Done. Added 0 new entries to the blacklist. 1 entries could not be saved.”.
* [ ] Menu items are hiding without reason. For example, when clicking on „Add Galaxy Cluster“, active menu item is moved and name changed to „Add Cluster“.
* [ ] It is not possible to edit relationship.
* [ ] When adding galaxy cluster, it is not clear that Author field should be in JSON format. 

> Would you mind describing how you got the error?

Thanks, It works now. I am not sure what was the problem.

oaklet:
Going through the features in the galaxy-cluster2.0 build lately, I can add some user feedback. Feel free to address as needed. But first, let me thank you for all the work, looking forward to using new clusters in production.

Galaxies side panel
---
- 'Export Galaxy Clusters' also disappears when you choose a cluster within a galaxy.

- At least at first, jumping and ordering of elements in the left panel felt confusing.

For example, when you choose a galaxy and a cluster, it goes as follows:

```
Import Galaxy Clusters 
---
View Galaxy
View Cluster
Edit Cluster
Add Cluster
Fork Cluster
---
(Un)Publish cluster
---
View Correlation Graph
```


Perhaps it might help to simply rearrange items based on where they belong and how it gradually opens (without jumping), for example:

```
---
View Galaxy
Import Galaxy Clusters
Export Galaxy Clusters
Add Galaxy Cluster
---
View Cluster
(Un)Publish cluster
View Correlation Graph
Fork Cluster
---
```


View Galaxy
---
- At this moment, all intentions for columns, such as published, distribution and default are a bit unclear to me.

For example, updating galaxies will add default clusters which are unpublished and you can add them to events and attributes right away. However, when you add a custom cluster it will become available for use only after publishing, regardless of distribution. 

Suggestion: Perhaps it can be unified, similarly to events: a cluster to be usable by the creator/main org when created and its further visibility to behave according to distribution and as per 'MISP.unpublishedprivate' toggle in the settings?

- It leads to an error when trying to publish a default cluster. 

- An attempt to hard-delete a default cluster leads to an internal error.

- Suggestion: Remove right side icons for fork and delete cluster for non-admin users? (right now it leads to error messages)

List Relationships
---
- No edit available for a custom relationship where the source is a default cluster (seems to be available only for a custom source)
- Distribution of a relationship seems to be fully decoupled from distributions of clusters. Thus it can have a more benevolent distribution, and can lead to situation when you see clusters listed in the relationships (and their descriptions on mouseover) even when otherwise one shouldn't be able to see them. Would it make sense to put an upper limit on relationship distribution based on clusters'?
- Suggestion: remove right-side delete option for non-admin users.  
- Suggestion: to add Owner/Creator Org columns to the table.

Add Relationships
---
- At the relationship creation, unpublished custom clusters aren't available from the drop down menu. However, you can still create a relationship using cluster's UUID.
- Suggestion: change 'add/edit relationships' text to singular.
- Missing error messages for invalid/missing UUIDs.
- A minor usability note: when you just want to add a tag, using 'Pick tags' will overwrite any tags already present in the tag list. So you have to pick all of them again (or copy/paste as text). 
- Suggestion: perhaps might be useful to be able see/set relationship UUID itself too.
 
View Cluster
---
- Somewhat unintuitive that the header name on the 'View Cluster' is the name of its galaxy (initially thought it was the cluster).
- when viewing a default cluster the left side option 'Publish Cluster' does nothing.

Fork Cluster
---
- some default clusters cannot be forked, on submit it says "Cluster forks have to belong to the same galaxy as the parent" (not sure why, but it happened with a cluster that was listed in more than one galaxy).

mokaddem:
> Going through the features in the galaxy-cluster2.0 build lately, I can add some user feedback. Feel free to address as needed. But first, let me thank you for all the work, looking forward to using new clusters in production.

Thanks a lot for this very good feedback!
> ## Galaxies side panel
> 
>     * 'Export Galaxy Clusters' also disappears when you choose a cluster within a galaxy.
> 
>     * At least at first, jumping and ordering of elements in the left panel felt confusing.
> 
> 
> For example, when you choose a galaxy and a cluster, it goes as follows:
> 
> ```
> Import Galaxy Clusters 
> ---
> View Galaxy
> View Cluster
> Edit Cluster
> Add Cluster
> Fork Cluster
> ---
> (Un)Publish cluster
> ---
> View Correlation Graph
> ```
> 
> Perhaps it might help to simply rearrange items based on where they belong and how it gradually opens (without jumping), for example:
> 
> ```
> ---
> View Galaxy
> Import Galaxy Clusters
> Export Galaxy Clusters
> Add Galaxy Cluster
> ---
> View Cluster
> (Un)Publish cluster
> View Correlation Graph
> Fork Cluster
> ---
> ```
> 
- So the idea was to have all the different scopes at the top (galaxies/blocklist/relationships)
- Then import/export cluster
- Add/view/fork/...
- And action (publish/unpublish)

I thought that it made sense but it's easy to re-arrange if we want to change it
> ## View Galaxy
> 
>     * At this moment, all intentions for columns, such as published, distribution and default are a bit unclear to me.
> 
> 
> For example, updating galaxies will add default clusters which are unpublished and you can add them to events and attributes right away. However, when you add a custom cluster it will become available for use only after publishing, regardless of distribution.
> 
> Suggestion: Perhaps it can be unified, similarly to events: a cluster to be usable by the creator/main org when created and its further visibility to behave according to distribution and as per 'MISP.unpublishedprivate' toggle in the settings?
> 
>     * It leads to an error when trying to publish a default cluster.
> 
>     * An attempt to hard-delete a default cluster leads to an internal error.
> 
>     * Suggestion: Remove right side icons for fork and delete cluster for non-admin users? (right now it leads to error messages)
> 
All the bugs above have been fixed. When a viewing an unpublished custom cluster, an alert is displayed.
> 
> ## List Relationships
> 
>     * No edit available for a custom relationship where the source is a default cluster (seems to be available only for a custom source)
> 
Exactly, it is done on purpose. You can't edit default clusters so neither relations that have them as source.
>     * Distribution of a relationship seems to be fully decoupled from distributions of clusters. Thus it can have a more benevolent distribution, and can lead to situation when you see clusters listed in the relationships (and their descriptions on mouseover) even when otherwise one shouldn't be able to see them. Would it make sense to put an upper limit on relationship distribution based on clusters'?
> 
This was indeed a nasty bug. Forgot to take conditions of clusters into account. Nice catch!
>     * Suggestion: remove right-side delete option for non-admin users.
> 
Done
>     * Suggestion: to add Owner/Creator Org columns to the table.
> 
Done
> 
> ## Add Relationships
> 
>     * At the relationship creation, unpublished custom clusters aren't available from the drop down menu. However, you can still create a relationship using cluster's UUID.
> 
Indeed, fixed too. It makes sense to let user reference them.
>     * Suggestion: change 'add/edit relationships' text to singular.
> 
Done
>     * Missing error messages for invalid/missing UUIDs.
> 
Fixed
>     * A minor usability note: when you just want to add a tag, using 'Pick tags' will overwrite any tags already present in the tag list. So you have to pick all of them again (or copy/paste as text).
> 
This is a special case for the generic picker. Will keep that in mind though
>     * Suggestion: perhaps might be useful to be able see/set relationship UUID itself too.
> 
There are no relationship UUID, so not sure it's applicable here
> 
> ## View Cluster
> 
>     * Somewhat unintuitive that the header name on the 'View Cluster' is the name of its galaxy (initially thought it was the cluster).
Done
> 
>     * when viewing a default cluster the left side option 'Publish Cluster' does nothing.
> 
Fixed
> 
> ## Fork Cluster
> 
>     * some default clusters cannot be forked, on submit it says "Cluster forks have to belong to the same galaxy as the parent" (not sure why, but it happened with a cluster that was listed in more than one galaxy).

This is not good. I don't think you should have clusters belonging to more than one galaxy.
Could you tell me which cluster triggered it for you?

Thanks a lot again for this very good review!

oaklet:
Thanks for the kind words and kudos for the improvements!

>     * So the idea was to have all the different scopes at the top (galaxies/blocklist/relationships)
> 
>     * Then import/export cluster
> 
>     * Add/view/fork/...
> 
>     * And action (publish/unpublish)
> 
> 
> I thought that it made sense but it's easy to re-arrange if we want to change it

Sure, sounds good. Thoughts were more in the area if minimising the menu changes might improve the user experience. 

> > ```
> > * No edit available for a custom relationship where the source is a default cluster (seems to be available only for a custom source)
> > ```
> 
> Exactly, it is done on purpose. You can't edit default clusters so neither relations that have them as source.

However, 'edit relationship' leads only to the relationship editing. So it's confusing not to be allowed to edit the same parameters which were allowed to be adjusted at the relationship creation already. To be clear, this is only about custom relationships.

> > ```
> > * some default clusters cannot be forked, on submit it says "Cluster forks have to belong to the same galaxy as the parent" (not sure why, but it happened with a cluster that was listed in more than one galaxy).
> > ```
> 
> This is not good. I don't think you should have clusters belonging to more than one galaxy.
> Could you tell me which cluster triggered it for you?

Pre Attack - Intrusion Set galaxy  ->  APT1 - G0006  ->  Fork  ->  Submit

In the json view of APT1 - G0006, its 'galaxy_id' is of another galaxy (Enterprise Attack - Intrusion Set). And indeed I have this cluster listed in both of them (the same name and UUID). Btw. trying to fork from the other galaxy works.

Issue #5713, opened 2020-03-23
information sharing statistics in the (new) dashboards


cvandeplas:
**Is your feature request related to a problem? Please describe.**
The MISP-dashboards give interesting information sharing statistics that could stimulate information sharing; but does not respect data confidentiality based on the distribution and SharingGroups

**Describe the solution you'd like**
The new (currently developed) dashboards built in MISP web UI could provide similar information sharing statistics as the MISP-dashboard does. By design these automatically follow the data confidentiality as they use internal APIs.

This feature-request is therefore to create nicely looking dashboards with statistics of information sharing on this instance.  Stats should (at least) be organisation centric.

mokaddem:
What stats would you like to see in addition to Organisation activity (data creation/edition)?

cvandeplas:
Similar as what you tried to do with the MISP-dashboards:
- stimulate info sharing by gamification / trends of contributions
- better insight in data with trends of data (tags/galaxies, ...) 
- use of misp-objects
- use of references (linking objects/attributes)
- ...

cvandeplas:
As given by @mokaddem  in https://github.com/MISP/MISP/pull/6129#issuecomment-662298299 , which came from the [MISP dashboard](https://github.com/MISP/misp-dashboard) project that could also be interesting here:

- Specific badges and achievements for interacting with specific threat actors
- Rewards for discovering APT 41 C2, or new important TTPs
- Rewards for industry best champion (who produces the most/highest quality threat intel on actors targeting our industry)
    - Just be careful with the **most** as you want to avoid people submitting bad quality/junk content
- Better cross industry trends, filtering by industry
- Rewards or points for intra-industry collaboration, rather than analyst output

Not sure how to implement these point in a clean way, so food for thought

Credit @VVX7 ;)

cvandeplas:
Was fixed by PR: https://github.com/MISP/MISP/pull/6142

PR #6129, opened 2020-07-16
Feature/achievements widget

lfortemps:
### What does it do?

Initial contribution to #5713 . It brings gamification through an 'achievements' widget which can help new users discover the different MISP features. An example is shown below: 

achievements_widget

I'm planning to add more badges but would like to request an initial feedback, what do you think?

#### Questions

### DIFF
--- a/app/Lib/Dashboard/AchievementsWidget.php
+++ b/app/Lib/Dashboard/AchievementsWidget.php
@@ -53,7 +53,7 @@ class AchievementsWidget
          ),
          "taxonomies" => array(
              "icon" => "/img/custom/taxonomy.png",
-             "title" => __("Make sure to speak the same language as your conterparts by using existing taxonomies for your tags."),
+             "title" => __("Make sure to speak the same language as your counterparts by using taxonomies for your tags."),
              "help_page" => "https://www.circl.lu/doc/misp/taxonomy/"
          ),
          "galaxies" => array (
@@ -74,7 +74,7 @@ class AchievementsWidget
         "events" => __("Congratulations, you have shared your first event!"),
         "tags" => __("You have been using tags, good job!"),
         "taxonomies" => __("Taxonomies have been used in your events."),
-        "galaxies" => __("You have discovered how to use the galaxies."),
+        "galaxies" => __("Galaxies have no secrets for you in this Threat Sharing universe."),
         "attack" => __("MISP & MITRE ATT&CK is a great combo.")
     );
 
--- a/app/Lib/Dashboard/AchievementsWidget.php
+++ b/app/Lib/Dashboard/AchievementsWidget.php
@@ -9,7 +9,7 @@ class AchievementsWidget
   * - tlp_green.png --> https://raw.githubusercontent.com/MISP/intelligence-icons/master/square_png/64/tlp_green.png
   */
 
-	public $render = 'Achievements';
+  public $render = 'Achievements';
   public $title = 'Achievements of my organization';
   public $description = 'Earn badges and improve your usage of MISP.';
   public $width = 4;
--- a/app/Lib/Dashboard/AchievementsWidget.php
+++ b/app/Lib/Dashboard/AchievementsWidget.php
@@ -4,19 +4,29 @@ class AchievementsWidget
   /*
   * Note: for this widget to display as expected, you need all icons to be accessible in your webroot. (img/custom)
   * Icons used:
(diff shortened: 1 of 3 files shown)


cvandeplas:
cool  ! 

- A first visual impression is the text size which is imbalanced with the icons. (so increase text size or shrink images, or both?)
- For the icons you could use the MISP Object and MISP Event icons : MISPEvent96 and MISPObject96
- Taxonomies:  taxonomies or taxonomies 
- Galaxies: galaxy or  galaxy
- Relationships: relation or  relation

app/Lib/Dashboard/AchievementsWidget.php
```diff
@@ -0,0 +1,122 @@
...
+{
+  /*
+  * Note: for this widget to display as expected, you need all icons to be accessible in your webroot. (img/custom)
+  * Icons used:
+  * - relationship.png --> https://raw.githubusercontent.com/MISP/intelligence-icons/master/simple_png/64/relationship.png
+  * - report.png --> https://raw.githubusercontent.com/MISP/intelligence-icons/master/square_png/64/report.png
+  * - tlp_green.png --> https://raw.githubusercontent.com/MISP/intelligence-icons/master/square_png/64/tlp_green.png
+  */
+
+	public $render = 'Achievements';
```
SteveClement: I replaced the tabs with spaces.

lfortemps:
Slightly modified visually to take into account @cvandeplas comment. Also added 3 new badges.
achievements_widget2

cvandeplas:
To speed up the query, would using a `LIMIT 0,1` type of filter be a good idea? This way the DB can stop once it has found one matching item. 
If that speeds up things, we could also test removing the time-filter as logically it would slow down things again.

lfortemps:
As I'm using a "count" instead of a regular select, queries are really fast. At least on my small DB. I would have to test on a large DB. Why would you want to remove the time limit? If people go inactive for a long time, it's good to remotivate them :)

cvandeplas:
Excellent point on the time limit,
Could you do a benchmark on the `count` vs `limit 1` on prod data? Eager to see the effect on performance.

mokaddem:
I love the idea!
How do you plan to make it visible to the users?
Make their favorite page the 'achievement dashboard' page?

mokaddem:
He are some more ideas and remarks we received for the [MISP dashboard](https://github.com/MISP/misp-dashboard) project that could also be interesting here:

- Specific badges and achievements for interacting with specific threat actors
- Rewards for discovering APT 41 C2, or new important TTPs
- Rewards for industry best champion (who produces the most/highest quality threat intel on actors targeting our industry)
    - Just be careful with the **most** as you want to avoid people submitting bad quality/junk content
- Better cross industry trends, filtering by industry
- Rewards or points for intra-industry collaboration, rather than analyst output

Not sure how to implement these point in a clean way, so food for thought

Credit @VVX7 ;)

lfortemps:
@cvandeplas 
So as we could expect the larger the number of events, the better the "limit 1" will perform. For an organization having 20k events, this could go up to 0.4s via count and 0.01s for limit 1.
However for organizations having few events, the count completes faster (1ms) vs 10ms for the limit 1.  
So it's best if I switch the sql statements to select with limit instead of count.

@mokaddem Yes the plan is to put it in a dashboard that would be the default page after log in, next to potentially other widgets.
Thanks for the ideas, it might come in a subsequent PR but indeed it needs to be thought beforehand to make sure it doesn't introduce a huge computation cost and also define the appropriate weight/score on each contribution.

cvandeplas:
I' ve added @mokaddem 's feedback in the feature request: https://github.com/MISP/MISP/issues/5713#issuecomment-662441777

cvandeplas:
thank you very much ! 
Is there something else needed (SQL) to add it as part of the default dashboard?

lfortemps:
I have checked and there are a couple of problems with making this the default for all MISP install:

- It is indeed stored in SQL, so we would need to modify the INSTALL/MySQL.sql but there would be a broken reference because no user_id (template creator) can be populated at that time. Or we can make it part of a DB migration step? but again how do we decide on the user_id
- The pictures would need to be added in the repository by default
- Some people might not like to have it accessible for all orgs in their instance
- The layout can be great for our screen and instance data but it might not suit someone's else instance

So my vote would be to let MISP admin chose to create a template with the widgets available instead of forcing a default template. What do you guys think?

cvandeplas:
thank you very much !

PR #6132, opened 2020-07-17
fix: check Misp time fields exist before using them

denny-lclin:
#### What does it do?
According to [MISP RFC](https://github.com/MISP/misp-rfc/blob/master/misp-core-format/raw.md), first_seen & last_seen fields' requirement levels are **MAY**, while `handle_time_fields(attribute, timestamp, stix_type)` function in app/files/scripts/stix2/misp2stix2.py does not check if first_seen, last_seen exist and will cause exception if those keys don't exist.

#### Questions


chrisr3d:
Thank you!

PR #6136, opened 2020-07-20
fix: [security] Respect ACL when event edit

JakubOnderka:
This PR changes almost all event, attributes and objects controller methods to provide one code to checks ACLs. Before, different methods use different approach and for example just few respects "User can modify just own events" permission.

It also reduce a lot of code duplication. 

#### Questions

### DIFF
--- a/app/Controller/AttributesController.php
+++ b/app/Controller/AttributesController.php
@@ -2586,7 +2586,7 @@ class AttributesController extends AppController
 
     public function addTag($id = false, $tag_id = false)
     {
-        $this->Taxonomy = $log = ClassRegistry::init('Taxonomy');
+        $this->Taxonomy = ClassRegistry::init('Taxonomy');
         $rearrangeRules = array(
             'request' => false,
             'Attribute' => false,
@@ -2689,15 +2689,9 @@ class AttributesController extends AppController
                 } else {
                     $attribute = $attributes[0];
                 }
-                if (!$this->userRole['perm_tagger']) {
+                if (!$this->__canModifyTag($attribute, $local)) {
                     $fails++;
                     continue;
-                }
-                if ((!$this->userRole['perm_sync'] && !$this->_isSiteAdmin()) && $attribute['Event']['orgc_id'] !== $this->Auth->user('org_id')) {
-                    if (Configure::read('MISP.host_org_id') != $this->Auth->user('org_id') || !$local) {
-                        $fails++;
-                        continue;
-                    }
                 }
                 $eventId = $attribute['Attribute']['event_id'];
                 $event = $this->Attribute->Event->find('first', array(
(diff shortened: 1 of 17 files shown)


app/Controller/AttributesController.php
```diff
@@ -3069,30 +2888,16 @@ public function removeTag($id = false, $tag_id = false)
 
     public function toggleCorrelation($id)
     {
-        if (!$this->_isSiteAdmin() && !Configure::read('MISP.allow_disabling_correlation')) {
```
mokaddem, line 3086: Restore this conditions?

app/Controller/EventsController.php
```diff
@@ -1575,12 +1576,16 @@ private function __viewUI($event, $continue, $fromEvent)
...
     }
 
-    public function view($id = null, $continue=false, $fromEvent=null)
+    public function view($id = null, $continue = false, $fromEvent = null)
     {
-        // find the id of the event, change $id to it and proceed to read the event as if the ID was entered.
-        $id = $this->Toolbox->findIdByUuid($this->Event, $id);
-        $this->Event->id = $id;
-        $conditions = array('eventid' => $id);
+        if (is_numeric($id)) {
```
mokaddem, line 1568: Why not use `Event->fetchSimpleEvent`?

app/Controller/EventsController.php
```diff
@@ -1668,11 +1674,12 @@ public function view($id = null, $continue=false, $fromEvent=null)
         if (empty($results)) {
             throw new NotFoundException(__('Invalid event'));
         }
+        $event = $results[0];
+        $this->Event->id = $event['Event']['id'];
         //if the current user is an org admin AND event belongs to his/her org, fetch also the event creator info
-        if ($this->userRole['perm_admin'] && !$this->_isSiteAdmin() && ($results[0]['Org']['id'] == $this->Auth->user('org_id'))) {
-            $results[0]['User']['email'] = $this->User->field('email', array('id' => $results[0]['Event']['user_id']));
+        if ($this->userRole['perm_admin'] && !$this->_isSiteAdmin() && ($event['Org']['id'] == $this->Auth->user('org_id'))) {
+            $event['User']['email'] = $this->User->field('email', array('id' => $event['Event']['user_id']));
```
mokaddem, line 1668: See line 1700.
Replace `$results[0]['Event']['orgc_id']` by `$event['Event']['orgc_id']`

app/Controller/AttributesController.php
JakubOnderka: Yes, my mistake.

app/Controller/EventsController.php
JakubOnderka: Sorry, I don't understand. This method uses `Event->fetchEvent` for fetching event from database, because we need whole event with attributes.

app/Controller/EventsController.php
JakubOnderka: Thanks, fixed.

app/Controller/EventsController.php
mokaddem: Brain glitched. You are right ;)

PR #6141, opened 2020-07-23
chg: [UI] Add proposal form refactor

JakubOnderka:
##  What does it do?

Changes the form to match Add Attribute style. It also fix some bugs (for example adding proposal from event page by clicking `+` icon).

### Before:

Snímek obrazovky 2020-07-24 v 14 11 51

### After:

Snímek obrazovky 2020-07-24 v 14 12 08

#### Questions

### DIFF
--- a/app/Controller/ShadowAttributesController.php
+++ b/app/Controller/ShadowAttributesController.php
@@ -290,11 +290,6 @@ class ShadowAttributesController extends AppController
         } else {
             $this->set('ajax', false);
         }
-        if (empty($eventId)) {
-            if (empty($event)) {
-                throw new NotFoundException(__('Invalid Event'));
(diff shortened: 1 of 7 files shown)


mokaddem:
Hey it looks great!
Now that it looks very similar to the attribute's `add.ctp`, would it make sense to use it instead or would it make the view to cluttered?
What do you think?

JakubOnderka:
> Hey it looks great!
> Now that it looks very similar to the attribute's `add.ctp`, would it make sense to use it instead or would it make the view to cluttered?
> What do you think?

Definitely, but first I wanna join `ShadowAttributes/add.ctp` and  `ShadowAttributes/edit.ctp` templates.

PR #6142, opened 2020-07-24
new: [widgets] Additional widgets for sharing statistics and layouts

lfortemps:
#### What does it do?

Another contribution to #5713 . It adds a couple of additional widgets to stimulate sharing. Note that the ACL is applied so the dashboard will look different for each organization depending on what they can see.
Here is an example with some test data:

dashboard_example

#### Questions


iglocska:
Brilliant, thanks a million!

PR #6154, opened 2020-07-28
chg: [internal] Initialize UserSetting just when needed

JakubOnderka:
#### What does it do?

Should speed up loading a lot of tags.

#### Questions


mokaddem:
Thanks!

PR #6187, opened 2020-08-12
chg: [doc] Various changes to CentOS/RHEL docs

SteveClement:


JKD32332:
Thank you for this, but is there something else in the instructions that are not working? I was able to get past this part thanks to you adding the missing workerstartsh.te file but now when I'm in the misp-modules section (9.07), I'm trying to do the pip install, $SUDO_WWW $PATH_TO_MISP/venv/bin/pip install -U -I -r REQUIREMENTS command, and I'm getting the error that there is no file or directory with the name REQUIREMENTS...

Is this section not working as well? If I have a choice to choose between a centos 7 and centos 8 for deploying MISP, should I stick with centos 7?

PR #6203, opened 2020-08-17
Show proper number of attributes and events for tags

JakubOnderka:
#### What does it do?

Before that change, no ACL was applied, so MISP shows incorrect numbers.

#### Questions

### DIFF
--- a/app/Controller/GalaxyClustersController.php
+++ b/app/Controller/GalaxyClustersController.php
@@ -1,6 +1,9 @@
 <?php
 App::uses('AppController', 'Controller');
 
+/**
+ * @property GalaxyCluster $GalaxyCluster
+ */
 class GalaxyClustersController extends AppController
 {
     public $components = array('Session', 'RequestHandler');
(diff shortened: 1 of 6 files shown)


iglocska:
I think this was one of those cases where we said we'd be fine ignoring the ACL due to the potential performance hurdles this would lead to. Are you seeing a substantial increase in query times with this applied?

JakubOnderka:
> I think this was one of those cases where we said we'd be fine ignoring the ACL due to the potential performance hurdles this would lead to. Are you seeing a substantial increase in query times with this applied?

Good point. I just test it on my testing instance and it looks good (but I didn't measure it), but I will check it in larger instance and measure the difference.

JakubOnderka:
> > I think this was one of those cases where we said we'd be fine ignoring the ACL due to the potential performance hurdles this would lead to. Are you seeing a substantial increase in query times with this applied?
> 
> Good point. I just test it on my testing instance and it looks good (but I didn't measure it), but I will check it in larger instance and measure the difference.

I just tested it and fix one problem (SQL query was send even if tag doesn't contains any attribute or event). Showing page with 50 tags takes on my big instance 670 ms instead of 650 ms before.

iglocska:
That sounds like an acceptable loss, let's give it a whirl! Thanks a lot!

iglocska:
Sorry reverted it for now. It is causing ~140GB of memory usage for some users... :)

PR #6208, opened 2020-08-18
Experimental: Faster tag fetching for events

JakubOnderka:
#### What does it do?

This change should provide even faster tag fetching for events. 

#### Questions

### DIFF
--- a/app/Model/Event.php
+++ b/app/Model/Event.php
@@ -7,6 +7,7 @@ App::uses('TmpFileTool', 'Tools');
 
 /**
  * @property Attribute $Attribute
+ * @property EventTag $EventTag
  */
 class Event extends AppModel
 {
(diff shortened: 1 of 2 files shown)


iglocska:
I think prefetching all might be overkill in many cases. For a full exports I can see it being beneficial, but for searches this could really bog things down for instances with a large number of tags.

app/Model/Event.php
```diff
@@ -2346,6 +2346,8 @@ public function fetchEvent($user, $options = array(), $useCache = false)
                 $results[$k] = $this->__mergeExtensions($user, $result['Event']['uuid'], $result);
             }
         }
+        // Clear tags cache after fetching all events, because another fetch can use different tag conditions.
+        $this->assetCache['tags'] = [];
```
iglocska: Is that something we really want? The idea was to persist the assetCache between the execution of individual events for restSearch for example (which uses a lot of chunked fetchEvent() calls).

JakubOnderka: It was just temporary fix, changed now. But I still need test that.

PR #6212, opened 2020-08-20
Event ID translation feature

### LINKED ISSUE #6128: Event id to/from uuid translator with sync servers
An event id is easy to work with in ticketing tool, reports etc. because it is short and can be easily typed manually if copy-paste is not an option. When working with multiple MISP instances, this event id will be different for the same event (same uuid).

Example: a user from a public MISP instance would like to notify us about an interesting event id, our staff only having access to the private MISP instance won't easily find it as the event ids are not the same and they don't have direct access to the public instance.

In the Sync Actions, a new submenu "Event id translator" which can do the translation.
One of the following input:
- Local event id
- UUID
- Remote event id (From servers with pull and/or push relationship) - to select in a dropdown

Output:
- local event id (+link)
- UUID
- remote event ids for all sync servers

The translation will be done on the fly (during the request processing). Timeout in place for unresponsive servers. If it takes too much time to complete in practice, a background job will be investigated.

Security controls:
- In case of local id or uuid --> access to the local event is checked
- In case of remote id --> uuid is fetched, checked locally for presence and access.

** Additional Context **

I'm willing to work on this feature in the coming weeks

lfortemps:
#### Overview

Implement a new feature to translate an event ID into the equivalent ID on the sync servers. The feature is only accessible to users belonging to the host org. 
A small modification has been done to the downloadEventFromServer to speed it up when the use case doesn't care about attributes or attachments linked to the event.

See related Issue for more details.

Fixes #6128 

#### Questions


app/Controller/ServersController.php
```diff
@@ -1266,6 +1266,88 @@ private function __checkVersion()
...
+
+        // The id translation feature is limited to people from the host org
+        if (!$this->_isSiteAdmin() && $this->Auth->user('org_id') != Configure::read('MISP.host_org_id')) {
+            throw new MethodNotAllowedException(__('You don\'t have the required privileges to do that.'));
+        }
+
+        //We retrieve the list of remote servers that we can query
+        $options = array();
+        $options['conditions'] = array("pull" => true);
+        $servers = $this->Server->find('all', $options);
```
JakubOnderka, line 1279: You can use 'list' instead of 'all': see https://book.cakephp.org/2/en/models/retrieving-your-data.html#find-list

app/Controller/ServersController.php
```diff
@@ -1266,6 +1266,88 @@ private function __checkVersion()
...
+            } else if (!empty($this->request->data['Event']['uuid']) && $this->request->data['Event']['local'] == "remote" && !empty($this->request->data['Server']['id'])) {
+                //We check on the remote server for any event with this id and try to find a match locally
+                $conditions = array('AND' => array('Server.id' => $this->request->data['Server']['id'], 'Server.pull' => true));
+                $remote_server = $this->Server->find('first', array('conditions' => $conditions));
+                if(!empty($remote_server)) {
+                    try {
+                        $remote_event = $this->Event->downloadEventFromServer($this->request->data['Event']['uuid'], $remote_server, null, true);
+                    } catch (Exception $e) {
+                        $this->Flash->error(__("Issue while contacting the remote server to retrieve event information"));
+                        $this->log($e->getMessage(), LOG_NOTICE);
```
JakubOnderka: You can use `$this->logException`, so the whole exception with backtrace will be logged.

app/Controller/ServersController.php
```diff
@@ -1266,6 +1266,88 @@ private function __checkVersion()
...
+                // We check if the server was not already contacted in phase 1
+                if(count($remote_events) > 0 && $remote_events[0]['server_id'] == $s['Server']['id']) {
+                    continue;
+                }
+
+                try {
+                    $remote_event = $this->Event->downloadEventFromServer($local_event['Event']['uuid'], $s, null, true);
+                    $remote_event_id = $remote_event[0]['id'];
+                } catch (Exception $e) {
+                    $this->log($e->getMessage(), LOG_NOTICE);
```
JakubOnderka: After logging exception, you should use `continue` to skip unresponsive server.

app/Model/Event.php
```diff
@@ -1386,19 +1386,28 @@ private function __updateAttributeForSync($attribute, $server)
...
     {
         $url = $server['Server']['url'];
         $HttpSocket = $this->setupHttpSocket($server, $HttpSocket);
         $request = $this->setupSyncRequest($server);
-        $uri = $url . '/events/view/' . $eventId . '/deleted[]:0/deleted[]:1/excludeGalaxy:1';
-        if (!empty($server['Server']['internal'])) {
-            $uri = $uri . '/excludeLocalTags:1';
+        if ($metadataOnly) {
+            $uri = $url . '/events/index';
+            $data['eventid'] = $eventId;
```
JakubOnderka: Cleaner is to use
```php
$data = ['eventid' => $eventId];
```

app/Model/Event.php
```diff
@@ -1386,19 +1386,28 @@ private function __updateAttributeForSync($attribute, $server)
...
+    public function downloadEventFromServer($eventId, $server, $HttpSocket=null, $metadataOnly=false)
     {
         $url = $server['Server']['url'];
         $HttpSocket = $this->setupHttpSocket($server, $HttpSocket);
         $request = $this->setupSyncRequest($server);
-        $uri = $url . '/events/view/' . $eventId . '/deleted[]:0/deleted[]:1/excludeGalaxy:1';
-        if (!empty($server['Server']['internal'])) {
-            $uri = $uri . '/excludeLocalTags:1';
+        if ($metadataOnly) {
+            $uri = $url . '/events/index';
```
iglocska, line 1400: This is actually going to return a lot of data and will be rather heavy for what you need.

`$uri = $url . '/events/index/minimal:1';` is what you want.

iglocska: Ignore this, there's no need, missed the next line in that you restrict to an event id. The added benefit of minimal:1 is not worth it.

app/Controller/ServersController.php
lfortemps: Thanks for the suggestion, but I need to know which server is unresponsive, so that I can give a feedback to the user. So the next step where the server is added to the list needs to be executed.

app/Controller/ServersController.php
lfortemps: I use the same array later in the code and I need to use the URL at that point which I couldn't get via list I think

JakubOnderka: True, my mistake.

app/Controller/ServersController.php
JakubOnderka: Makes sense. But then there is problem with `$remote_event_id` variable. Imagine situation that first server will respond and `$remote_event_id` will contains for example value `3`. But then second server will be not responsive, but variable `$remote_event_id` will still contains value `3`. So when exception is thrown, you should set the variable to `null`.

lfortemps: Good point, I will modify that.

cvandeplas:
Thank you very much Loïc !

PR #6220, opened 2020-08-21
A few accessibility fixes for users of screen readers

obert01:
A few accessibility fixes for users of screen readers:
- Added aria label and role for the representation of booleans in generic index tables,
- Fixed Aria label for actions in generic index tables,
- Set titles for actions in the admin user index table,
- Added a few missing aria labels in the global menu.


iglocska:
Thanks a lot, as always your fixes are much appreciated!

PR #6221, opened 2020-08-21
MISP-SNMP Monitor script

cudeso:
#### What does it do?

Script to return statistics which can then be picked up via SNMP.

Post for monitoring with Cacti (inspired by OpenNSM) will follow shortly.

#### Questions

- [X ] Are you using it in production? Yes


adulau:
Thank you very much. The blog post has been published https://www.misp-project.org/2020/08/22/MISP-Monitoring-with-Cacti.html

PR #6237, opened 2020-08-25
new: Add the ability to customize the IP header field when logging

### LINKED ISSUE #5661: Improve IP logging
Add x-forwarded-for IPs additionally to capture users through proxies.

- as pointed out by @coolacid

jtdroste:
#### What does it do?

Fixes #5661

This change adds a new configuration value (`MISP.log_client_ip_header`) which lets you specify what header is used for "whose IP is connecting to me" when logging IPs. If this configuration value is not set, it'll fall back on the standard `REMOTE_ADDR`.

NB: Still working on getting a development environment setup, but the change looks relatively straight forward based on my code reading.

#### Questions

### DIFF
--- a/app/Model/Log.php
+++ b/app/Model/Log.php
@@ -104,8 +104,15 @@ class Log extends AppModel
         if (!empty(Configure::read('MISP.log_skip_db_logs_completely'))) {
             return false;
         }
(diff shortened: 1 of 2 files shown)


iglocska:
Looks sane to me, thanks!

marcnil815:
This does not seem to work. Will create bug report.

PR #6168, opened 2020-08-03
chg: [db_schema] Add constraint on UUID for Attribute, Object and Event

mokaddem:
Add constraints on UUID for Attribute, Object and Event as well as clean up methods taking care of duplicates

- Fixes https://github.com/MISP/MISP/issues/6166
- Probably mitigate https://github.com/MISP/MISP/issues/6034

PR #6254, opened 2020-08-30
new: [diagnostic] Check if database index is unique

JakubOnderka:
#### What does it do?

New ability for database diagnostic to check, if columns has unique index. Because it looks like some MISP instances missing unique indexes, that are important (see https://github.com/MISP/MISP/pull/6168#issuecomment-672761255). 

#### Questions

### DIFF
--- a/app/Model/Server.php
+++ b/app/Model/Server.php
@@ -4752,9 +4752,9 @@ class Server extends AppModel
         $dbActualIndexes = array();
         $dataSource = $this->getDataSource()->config['datasource'];
         if ($dataSource == 'Database/Mysql' || $dataSource == 'Database/MysqlObserver') {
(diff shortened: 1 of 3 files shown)


iglocska:
Yep, the issue is quite tricky. We’ve identified the update that caused this and I’m working on a strategy to resolve the duplicates but it’s not all that straight forward. The diagnostic is a great first step!

JakubOnderka:
> Yep, the issue is quite tricky. We’ve identified the update that caused this and I’m working on a strategy to resolve the duplicates but it’s not all that straight forward. The diagnostic is a great first step!

Good point. So I add check if column contains just unique values – if not, SQL hint how to create index will be not provided.

iglocska:
Sounds good, cheers!

PR #6262, opened 2020-09-01
chg: [event] Deduplicate attribute related tags

JakubOnderka:
#### What does it do?

Before:

Snímek obrazovky 2020-09-01 v 9 20 52

After:

Snímek obrazovky 2020-09-01 v 9 21 17

#### Questions


iglocska:
Good point, thanks for the fixing it!

PR #6267, opened 2020-09-01
Rename blacklist and whitelist to alternatives

lfortemps:
#### What does it do?

It renames the terms blacklist and whitelist to alternatives which do not carry any unintended traces of racism.

Because the endpoints are renamed, a change at PyMISP level will also be required.

#### Questions


iglocska:
Big thumbs-up from me, thanks a lot!

PR #6272, opened 2020-09-01
Experimental: UUID validation

JakubOnderka:
#### What does it do?

UUID is validated that is is real UUID and not just fake one and also UUID is normalized to lowercase form.

@iglocska Is there any reason why „fake” UUID (not RFC 4122) are accepted? Because I think it is dangerous and it can break other tools and it is against MISP format RFC, that require RFC 4122 compliant UUID.

#### Questions


adulau:
The RFC is pretty clear, we just recommend version 4 but we don't enforce version 4. The proposed change is just validating an UUID at large (whatever version), so the change makes sense to me. @iglocska any objection for this change ?

iglocska:
Agreed

PR #6282, opened 2020-09-03
fix: Support IE with no template literal support

tomking2:
#### What does it do?

Fixes support for IE 10/11 by removing template literal in misp.js

#### Questions


iglocska:
Thanks a lot and my condolences for having to use IE still ;)

Issue #6239, opened 2020-08-26
Event date query - date_from and date_to issues


imidoriya:
Would it be possible to add an option for sort order.  From what I can tell, the query only comes back `ASC`.  It would be great if we could specify `DESC`.  With close to 150,000 events, often I want the latest event first and I haven't found a way to do that with a search query.  

One option I've tried is to specify `date_from` and `date_to`, but that works on publication date.  I'm often updating old stuff with new tags, info, etc.  That is not what I'm after.  I'd like a DESC query, which should be easy to do from an SQL perspective `Order By DESC`.  It looks like you already have this ability in the GUI as you can sort on many fields.

iglocska:
For which API in particular? Also, for the date_from/date_to issue - use timestamp instead, that one looks at the creation/modification of the data.

imidoriya:
I'm using ExpandedPyMisp search.

imidoriya:
I see 4 timestamps in the UI.
* Date
* Published
* First recorded change
* Last recorded change

I expect either `Date` or `First recorded change` would give me something created after a certain date without regard to modification - essentially events created from x day forward.  However, from your description and the PyMisp description, I can't figure out if it does that or not.  Perhaps it's just the wording, but the options are:
* date_from - Events with the date set to a date after the one specified. This filter will use the **date of the event**.
* publish_timestamp - Restrict the results by the last **publish** timestamp (newer than).
* timestamp - Restrict the results by the timestamp (**last edit**)
* event_timestamp - Only return attributes from events that have received a **modification** after the given timestamp.

So it looks like `date_from` should be the GUI `Date`, but it doesn't appear to be the case.  `publish_timestamp` appears to be `published`, `timestamp` appears to be `Last recorded change` (which is usually the same as `published`), and then I'm not sure what `event_timestamp` really is, looks like timestamp but excludes older attributes.  Don't really see anything for `First recorded change`, which is essentially the same as `Date`.

imidoriya:
Here is an example where it appears to ignore `date_from`.  All the other timestamps don't apply to `date` or `First recorded change`.  
```json
{
"headerless": true,
"published": true,
"controller": "events",
"limit": 10,
"date_from": 1598995358
}
```
<img width="319" alt="Screen Shot 2020-09-08 at 5 24 21 PM" src="https://user-images.githubusercontent.com/45568592/92529454-8a23b200-f1f8-11ea-88a7-cd2f6ff85af7.png">

[...]

PR #6304, opened 2020-09-11
Fix date filter on to / from #6239

imidoriya:
Fixes the to / from search filter issue described in #6239.  Looks like it should be converting the input to the date format for compare.

#### Questions

- [ ] Does it require a DB change?   No, but an index on the Event.date (and other filter fields) could be considered
- [X] Are you using it in production?  Yes
- [ ] Does it require a change in the API (PyMISP for example)?  It's a fix for the API (PyMISP works correctly after the change)


imidoriya:
While this fixes the issue, I don't think it's the root cause.  Will discuss and resubmit.

PR #6307, opened 2020-09-11
Fix handling deleted attributes when viewing event

JakubOnderka:
#### What does it do?

In current version, the behaviour for fetching deleted attributes is broken and is different for view event and AJAX fetch. This patch unite the behaviour:

* `deleted:2` - exclude deleted (default)
* `deleted:1` - show just deleted 
* `deleted:0` - show both

This patch also cleans some controller code.

#### Questions


mokaddem:
Much cleaner indeed!

PR #6308, opened 2020-09-11
Anonymise pushed sightings

RichieB2B:
Introducing a new Sightings_anonymise_as so all pushed sightings can be hidden behind a single organisation. This is relevant in the scenario where a CERT wants to report sightings upstream but does not want to disclose which constituent had the sightings.

Pulled sightings are unaffected. Sightings visibility (for push and pull) still follow the Sightings_policy.


iglocska:
Good idea, thanks a lot!

PR #6310, opened 2020-09-11
Only push events/sightings when selected for server

RichieB2B:
When pressing the "Push all" button on the Servers page, events were always pushed even when only "Push sightings" was selected for a server. Sightings were always synced even when "Push sightings" was not selected.

This patch fixes both issues.


iglocska:
Thanks a lot!

PR #6316, opened 2020-09-15
Better Docker compatibility in INSTALL script

stevengoossensB:
#### What does it do?

It changes the database setup so there's no need for the 'expect' anymore, but uses mysqladmin to configure the basic database.

#### Questions


SteveClement:
This is a good idea. I implemented it a little differently, but removed the expect magic.

PR #6323, opened 2020-09-18
chg: [search] Allow to search in attributes or objects UUID in quicksearch

JakubOnderka:
#### What does it do?

Currently, just event uuid is checked when searching. This can be useful if you know exact attribute or object UUID and you need to quickly go to event.

#### Questions

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -162,7 +162,7 @@ class EventsController extends AppController
 
     /**
      * @param string|array $value
-     * @return array Event ID that match filter
+     * @return array Event IDs that match filter
      */
     private function __quickFilter($value)
     {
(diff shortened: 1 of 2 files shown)


iglocska:
Perhaps something to make this more feasible would be to pass the raw query string to the condition for the uuids instead of doing a like query on %searchterm% - that way we wouldn't incur the cost of an extra substring search per uuid field.

JakubOnderka:
> Perhaps something to make this more feasible would be to pass the raw query string to the condition for the uuids instead of doing a like query on %searchterm% - that way we wouldn't incur the cost of an extra substring search per uuid field.

Yes, thats true and i was thinking about the same. But there are two reasons I choose substring search:
1) Sometimes it can makes sense to search just by part of UUID (something like short git commit ID) if you for example need to copy UUID from physically disconnected networks or when somebody tell you UUID by phone.
2) The same behaviour already exists for Event UUID. 

https://github.com/MISP/MISP/blob/ae5434fbc66cf61376986424841df976e9c70fef/app/Controller/EventsController.php#L240

And in reality, it will be not that match slower, because the slowest part of search is going thought all attributes (it is not possible to use index).

iglocska:
The event uuid is targeting a massively smaller table though. Also, we could run a check on the string via Validation::uuid() - if it returns true do a string match, otherwise a LIKE. That would preserve the use-case for substring matches.

JakubOnderka:
> The event uuid is targeting a massively smaller table though. Also, we could run a check on the string via Validation::uuid() - if it returns true do a string match, otherwise a LIKE. That would preserve the use-case for substring matches.

I changed the behaviour, so now if search term is valid UUID, we try to find exact match or if search term can be part of UUID, we try to find anywhere in string. Do you agree?

PR #6330, opened 2020-09-22
chg: [UI] Show event creator for users from event org

JakubOnderka:
#### What does it do?

Event creator e-mail should be shown for everybody that has audit permission and is from the same org or for site admins. 

And we should prefer putting that info into `event_creator_email` field that is defined by MISP core format RFC (`User.email` is not defined for Event).

#### Questions

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -1612,10 +1612,6 @@ class EventsController extends AppController
         }
         $event = $results[0];
         $this->Event->id = $event['Event']['id'];
-        //if the current user is an org admin AND event belongs to his/her org, fetch also the event creator info
-        if ($this->userRole['perm_admin'] && !$this->_isSiteAdmin() && ($event['Org']['id'] == $this->Auth->user('org_id'))) {
-            $event['User']['email'] = $this->User->field('email', array('id' => $event['Event']['user_id']));
-        }
         if (isset($this->params['named']['searchFor']) && $this->params['named']['searchFor'] !== '') {
             $this->__applyQueryString($event, $this->params['named']['searchFor']);
         }
--- a/app/Model/Event.php
+++ b/app/Model/Event.php
@@ -2144,9 +2144,6 @@ class Event extends AppModel
             unset($params['contain']['ShadowAttribute']);
             unset($params['contain']['Object']);
         }
-        if ($user['Role']['perm_site_admin']) {
-            $params['contain']['User'] = array('fields' => 'email');
-        }
         $results = $this->find('all', $params);
         if (empty($results)) {
             return array();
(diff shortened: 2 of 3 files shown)


iglocska:
Be careful, this is actually dangerous:

Until now, normal user accounts exporting data and sharing it outside of MISP with their peers did not have the user information included in the export - now it will be included. Make sure that the change only affects the UI and not API calls.

JakubOnderka:
> Be careful, this is actually dangerous:
> 
> Until now, normal user accounts exporting data and sharing it outside of MISP with their peers did not have the user information included in the export - now it will be included. Make sure that the change only affects the UI and not API calls.

Understand. So whats the reason for keeping `event_creator_email`? When it is useful? Because now it is included in export if user has auditor role. In RFC, it is described just in JSON Schema without any other information, not in text.

https://github.com/MISP/misp-rfc/blob/92b133f9298d45bd90d6c03e6fc66103069c09fa/misp-core-format/raw.md.txt#L2489

iglocska:
Hmm that's weird, I thought it was only for site-admins. I might be wrong.

iglocska:
If there's no change in the behaviour there then go for it, the risk is if we disclose it when we didn't before and the output suddenly doesn't match the expectation based on prior behaviour.

JakubOnderka:
Look here (current version):

https://github.com/MISP/MISP/blob/650f32144c211d14fd5361bec9e924d0c0504884/app/Model/Event.php#L2196-L2201

Before my change, `event_creator_email` is included in export for:
* user with audit permission if event is owned by same org as user belongs

After my change:
* user with audit permission if event is owned by same org as user belongs
* site admins

So maybe keep current behaviour and for site admins include user creator e-mail just in UI?

iglocska:
Ah in that case it's great, go for the changes! Making changes on the UI level is fine, if you want to include it there for auditors, if for non site admins there's no change in the API behaviour we're all good.

JakubOnderka:
> Ah in that case it's great, go for the changes! Making changes on the UI level is fine, if you want to include it there for auditors, if for non site admins there's no change in the API behaviour we're all good.

OK, so I kept the current behaviour. 

```php
// Include information about event user email to event. This information is included just for:
// - org admins from event creator org
// - auditors from event creator org
// - site admins
// In export, this information will be included in `event_creator_email` field just for auditors of event creator org.
```

JakubOnderka:
Sorry, another change. I think that the user creator email should be shown for all user from the same org. It makes sense for our org (usually, you don't need to hide creator if they are from the same org) and the same information is also available from event history, that can see every user from same org as event org. This doesn't change any export output.

What do you think about that?

iglocska:
For the UI, makes sense, go for it!

PR #6331, opened 2020-09-22
Remove extraneous spaces from E-mail subject

RichieB2B:
77833be9609f025c5375abb608eda94297c72db introduced two additional spaces in the E-mail subject that should not be there. This is especially apparent when `MISP.threatlevel_in_email_subject` and `MISP.extended_alert_subject` are set to false.


iglocska:
Thanks a lot!

PR #6349, opened 2020-09-26
chg: [UI] Nicer warning box with link to show just warnings

JakubOnderka:
#### What does it do?

Before:

Snímek obrazovky 2020-09-26 v 20 57 56

After:

Snímek obrazovky 2020-09-26 v 20 59 26

#### Questions


iglocska:
Cool idea!

PR #6356, opened 2020-09-28
new: [settings] New security setting to allow id translation for all users (disabled by default)

lfortemps:
#### What does it do?

The id translation feature might be useful for other orgs than just the host org. However it exposes a way to query remote servers. A new security setting is introduced to let administrators chose to open it to all users if they have enough trust in the local orgs (for instance on an internal instance).

Also fixing a small display bug for the "list servers" menu which was not following the ACL config.

#### Questions


lfortemps:
Should I port it into the new code base? This is something we use for an internal instance so I thought it would be good to include it in the project.

PR #6359, opened 2020-09-28
Acl menu item

JakubOnderka:
#### What does it do?

Use ACLComponent for menu item permission. ACLComponent provides nice way how to define access rules for different controllers and actions and it is sad that it cannot be reused for defining menu access permission. This is proof of concept how it can works also for menu items. What do you think?

#### Questions


mokaddem:
Good idea from my side

iglocska:
Looks very good indeed!

PR #6360, opened 2020-09-28
chg: [internal] Simplified menu code

JakubOnderka:
#### What does it do?

Make code cleaner.

#### Questions

### DIFF
--- a/app/View/Elements/genericElements/GlobalMenu/global_menu_separator.ctp
+++ b/app/View/Elements/genericElements/GlobalMenu/global_menu_separator.ctp
@@ -1,8 +1,7 @@
 <?php
-    /*
-     *  This template creates a separator and check for the conditions under which it hsould be displayed.
-     */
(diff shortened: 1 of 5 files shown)


mokaddem:
Cleaner indeed. We should refactor the sidebar into a sidebar factory at some point.

iglocska:
> Cleaner indeed. We should refactor the sidebar into a sidebar factory at some point.

Yeah planning on backporting it from Cerebrate at some point:

https://github.com/cerebrate-project/cerebrate/blob/main/templates/element/genericElements/side_menu_scaffold.php
and
https://github.com/cerebrate-project/cerebrate/blob/main/templates/element/genericElements/header_scaffold.php

Both the top and the side menu are using the same data source there (https://github.com/cerebrate-project/cerebrate/blob/main/src/Model/Table/AppTable.php#L16)

JakubOnderka:
So for now I will merge this and I will look forward for new implementation!

PR #6362, opened 2020-09-29
Fix date filter on to / from #6239

imidoriya:
Fixes the to / from search filter issue described in #6239.  Event.date needs to be compared to a `Y-m-d` format.  PyMISP sends the date filter `date_from` and `date_to` as a timestamp, which was not converted in MISP.  This allows the MISP API to accept a timestamp for those filter dates.

#### Questions

- [ ] Does it require a DB change?   **No, but as a filter, an index on Event.date should be considered** 
- [X] Are you using it in production?  **Yes**
- [ ] Does it require a change in the API (PyMISP for example)? **No**

### DIFF
--- a/app/Model/Event.php
+++ b/app/Model/Event.php
@@ -2877,9 +2877,17 @@ class Event extends AppModel
     public function set_filter_timestamp(&$params, $conditions, $options)
     {
         if ($options['filter'] == 'from') {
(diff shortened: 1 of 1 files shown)


iglocska:
Looks good, thanks a lot!

PR #6370, opened 2020-10-01
Fix update no prio

mokaddem:
- Only use the `update` worker to perform updates or do it inline
- Fixed wrongly diagnosed queue status


iglocska:
Thanks a lot!

PR #6371, opened 2020-10-01
fix: [tag] Show correct count of tag attributes and events

JakubOnderka:
#### What does it do?

Fixed version of pull #6203.

#### Questions

### DIFF
--- a/app/Controller/TagsController.php
+++ b/app/Controller/TagsController.php
@@ -2,6 +2,9 @@
 
 App::uses('AppController', 'Controller');
 
+/**
+ * @property Tag $Tag
+ */
 class TagsController extends AppController
 {
     public $components = array('Security' ,'RequestHandler');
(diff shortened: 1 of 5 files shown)


app/Controller/TagsController.php
```diff
@@ -71,31 +74,31 @@ public function index($favouritesOnly = false)
         }
         if ($this->_isRest()) {
             unset($this->paginate['limit']);
-            unset($this->paginate['contain']['EventTag']);
```
iglocska: This will still mean that we load every event tag and every attribute tag. On instances with high tag volumes (think attributes in the 100M range, 7-8 tags per attribute) this will return a massive amount of data.

JakubOnderka: Yes, I know. Now I am working how to fix it without loading all `EventTag` and `AttributeTag`.

iglocska: An option would be to loop through the results and run a count() on those two objects for each value. It would be slow in terms of querying the data, but at least we wouldn't run into crazy memory usage issues. I am still not convinced that having the accurate number here is really something worth the massive performance hit.

Issue #6368, opened 2020-10-01
Configurable user for AdminShell.php:runUpdates()


jloehel:
**Is your feature request related to a problem? Please describe.**
I'd like to use a different user (misp) for the workers. The user has group access to the necessary files but the hard-coded user restrictions preventing the usage:

https://github.com/MISP/MISP/blob/bab68641aa4fc5a8b209a07c2a98293de4ad0567/app/Console/Command/AdminShell.php#L354

It would be nice to get this working without patching AdminShell.php.

**Describe the solution you'd like**
Make the user configurable in `Config/config.php` with a default to `www-data`.

chrisinmtown:
+1 from me for making the update-user configurable.  Checking the person with admin rights to perform updates should be done by some kind of role/priv mapping in the database, not hardcoded strings in a php file.

PR #6373, opened 2020-10-01
Allow OS user to be set for upgrades

RichieB2B:
As requested in #6368


iglocska:
That's a clean solution without being too permissive. Lovely!

chrisinmtown:
I voted -1 because the PR includes no documentation change for this new configuration setting.  Need to help users get this right, to understand it, etc.

Issue #6342, opened 2020-09-25
MISP incorrectly states base URL is incorrect


todonaki:
MISP appears to be functioning normally, but gives critical error.

Error received: **Critical**
MISP.baseurl: https://misp.XXXXXXXX.global
**The  base url of the application (in the format  https://www.mymispinstance.com or https://myserver.com/misp). Several  features depend on this setting being correctly set to function. | Invalid baseurl, it has to be in the "https://FQDN" format.**

Type of issue: Bug
OS: Ubuntu 18.04 LTS
PHP version: 7.4
MISP version / git hash   | 2.4.132

**### Expected behavior**
No error from properly formatted base URL

**### Actual behavior**
Properly formatted base URL shows as critical error

**### Logs, screenshots, configuration dump, ...**
<img width="1465" alt="misp_error_25_09_2020" src="https://user-images.githubusercontent.com/57256761/94321013-70a29a00-ff5c-11ea-912f-53de1ec6f4df.png">

MISP appears to be working correctly despite this error.

chrisinmtown:
Can you show the full hostname in that URL? File app/Model/server.php, line 3624 checks the URL using this regex:

    $regex = "%^(?<proto>https?)://(?<host>(?:(?:\w|-)+\.)+[a-z]{2,5})(?::(?<port>[0-9]+))?(?<base>/[a-z0-9_\-\.]+)?$%i";

I'm not entirely familiar with all the features used this regex, so it will help to know the input string exactly.

There's also the chance that the host portion of that URL doesn't match the result of the getHost() function, see line 3580 of that same file.

todonaki:
I think you may be on to something here.  Line 3624 regex allows for **2-5** characters for TLD, but the **.GLOBAL extension is 6 characters** in length.  The regex doesnt allow for many of the new TLD's.  I'm going to try editing the regex in the file to allow for 6 characters and see if the problem is resolved.

todonaki:
That was it.  Regex does not allow for enough characters for all TLD types.  I would consider this a bug.  I did not experience this behavior until the latest (132) release. @chrisinmtown, appreciate the assistance and helping point me to the right location where I could make the change.

chrisinmtown:
[...]

PR #6375, opened 2020-10-01
Do not limit TLD to 5 characters

RichieB2B:
As requested in #6342


iglocska:
Sure seemed like a good idea at the time... :)

BillClarksonAntill (commented):
did you guys consider .co.nz domains? for New Zealand based organizations?

PR #6381, opened 2020-10-03
UI small fixes

JakubOnderka:
#### What does it do?

Small user interfaces changes and fixes that will make MISP interface little bit nicer and more usable.

* Use HiDPI icons for sightings 
* Disable IDS checkbox if user don't have permission to modify event
* Support for quick select for UUIDs
* Do not send request twice when pressing ENTER for quick search fields
* Threat level icons
Snímek obrazovky 2020-10-05 v 18 58 55

* Return back not working sightings popover
Snímek obrazovky 2020-10-04 v 17 10 04
 

#### Questions


mokaddem:
Interface looks nice!

JakubOnderka:
> Interface looks nice!

Merged, if you find any problem, I will be happy to fix them!

PR #6400, opened 2020-10-07
chg: [cookie] Set session cookie SameSite to Lax to avoid browser warnings

JakubOnderka:
#### What does it do?

This change will affect just new installations.

#### Questions

### DIFF
--- a/app/Config/core.default.php
+++ b/app/Config/core.default.php
@@ -181,8 +181,15 @@ Configure::write('Session', array(
 	'cookie_timeout' => 10080 ,  // Cookie timeout, default is 1 week
 	'defaults'       => 'php',
 	'autoRegenerate' => false,
-	'checkAgent' => false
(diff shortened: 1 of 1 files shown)


adulau:
Reading "https://www.php.net/manual/en/session.security.ini.php" - `cookies using Lax will be accessible in a GET request originated from another registrable domain` what's the exact meaning of another registrable domain?

JakubOnderka:
> Reading "https://www.php.net/manual/en/session.security.ini.php" - `cookies using Lax will be accessible in a GET request originated from another registrable domain` what's the exact meaning of another registrable domain?

Thats means for example if you click on link from e-mail account on Gmail or other provider or from other website, session cookie will be send. In new browsers is this default value.

adulau:
Just wondering about the security implication of this but I see why it's useful for this specific use-case.

PR #6405, opened 2020-10-07
Change hover enrichment user interface

JakubOnderka:
#### What does it do?

What do you think?

Before:

Snímek obrazovky 2020-10-07 v 19 21 27

After:

Snímek obrazovky 2020-10-07 v 19 21 46

#### Questions

### DIFF
--- a/app/Controller/AttributesController.php
+++ b/app/Controller/AttributesController.php
@@ -2403,15 +2403,12 @@ class AttributesController extends AppController
         if (empty($attribute)) {
             throw new NotFoundException(__('Invalid Attribute'));
         }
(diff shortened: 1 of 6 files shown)


adulau:
It's indeed a great improvement. Thanks a lot!

iglocska:
Awesome, prefer it greatly to what we have now. Much not in line with the look and feel of the rest of the ui.

JakubOnderka:
> Awesome, prefer it greatly to what we have now. Much not in line with the look and feel of the rest of the ui.

Great! So I will merge this and if you don't like something, feel free to create issue or write here.

PR #6411, opened 2020-10-09
Attachment malware protection

JakubOnderka:
#### What does it do?

This pull request adds a functionality to scan all uploaded attachment (from web interface, MISP feed or server or by API) by ClamAV antivirus and show warning for web interface users that the file contains malware (API responses are not changed):

Snímek obrazovky 2020-10-09 v 11 15 08

I am not sure if this will make sense for all MISP instances, but I think it makes sense for large instances shared by a lot of org. The most probably use case will be if user upload a malware sample without checking "Is malware sample" checkbox – then other user can download it and run malware on computer. Also implementing malware protection can be required by some cybersecurity law (for example ours) or by ISMS (MISP [current status](https://www.misp-project.org/compliance/ISO-IEC-27010/) for *12.2.1 Controls against malware* is N/A).

And I think processing files directly on server would be tricky, because for a while malware samples are saved unencrypted on disk during uploading process.

What do you think? Makes sense for you?

#### Questions


adulau:
This is a cool idea. Could we rely on misp modules for the functionality and let the admin select scanning modules from available misp modules? Doing a clamav module would be super simple.

JakubOnderka:
> This is a cool idea. Could we rely on misp modules for the functionality and let the admin select scanning modules from available misp modules? Doing a clamav module would be super simple.

I was thinking about the same, for example combine more AV or just send hash to VT. But I miss something like standardized interface, every module returns different attributes or objects.

adulau:
Maybe we could agree on a common tag for such use case and the tag is set by the modules output by default « misp:malicious-sample-detection=« true » » we would need to review the module doing file scanning and return in case of positive result.

JakubOnderka:
> Maybe we could agree on a common tag for such use case and the tag is set by the modules output by default « misp:malicious-sample-detection=« true » » we would need to review the module doing file scanning and return in case of positive result.

To use tag for that is interesting idea. 

We are actually working on implementing some AV scanners to MISP modules (no ETA yet), but it will require some changes in MISP modules – we need to send file to external tool (for example sandbox) and then pool results and put them back to MISP after some time (maybe minutes?). And I think that is not possible with current implementation. 

So what about to merge PR and after we will implement this mechanism to MISP and MISP modules, we will need just to change `AttachmentScan` class.

JakubOnderka:
> Maybe we could agree on a common tag for such use case and the tag is set by the modules output by default « misp:malicious-sample-detection=« true » » we would need to review the module doing file scanning and return in case of positive result.

So it was easier then I thought. I didn't know that MISP contains `av-signature` and `sb-signature`. So I just checks if module returns that object and if yes, attachment is considered as malware.

PR #6414, opened 2020-10-09
Flag icons and country list

JakubOnderka:
#### What does it do?

* Uses flags for countries from [Twemoji](https://github.com/twitter/twemoj) project (because unicode flags are problematic, see #5885).
* Organization country options are fetch from country galaxy – but because some countries has different name in Galaxy (for example Czech Republic –> Czechia), we keep backward compatibility and user can still save organisation with „old” country name.

Snímek obrazovky 2020-10-11 v 10 53 19

#### Questions


adulau:
Everything sound good to me. I also checked the licensing of the icons project and it's compatible with the AGPL. Maybe we need to keep track on how to update the flag file on a regular basis. For me it's fine to merge. @mokaddem any specific issue before merge?

app/View/Helper/IconHelper.php
```diff
@@ -0,0 +1,31 @@
...
+     * @param string $countryName Full country name for title
+     * @return string
+     */
+    public function countryFlag($countryCode, $countryName = null)
+    {
+        if (strlen($countryCode) !== 2) {
+            return '';
+        }
+
+        //return '<span class="famfamfam-flag-' . strtolower(h($countryCode)) . '" ></span>';
```
mokaddem: Leftover comment

app/Model/Organisation.php
```diff
@@ -456,4 +454,41 @@ public function checkDesiredOrg($suggestedOrg, $registration)
         }
         return $suggestedOrg;
     }
+
+    private function getCountryGalaxyCluster()
+    {
+        static $list;
+        if (!$list) {
+            $file = new File(APP . '/files/misp-galaxy/clusters/country.json');
```
mokaddem, line 462: Add a check if file doesn't exists? It can happens if submodules are not updated

mokaddem (changes requested):
Whenever you had a look at the comments, feel free to merge! :)

JakubOnderka:
> Everything sound good to me. I also checked the licensing of the icons project and it's compatible with the AGPL. Maybe we need to keep track on how to update the flag file on a regular basis. For me it's fine to merge. @mokaddem any specific issue before merge?

There is a script for that: https://github.com/MISP/MISP/blob/d95f8acda12d1d228af9dd7147baf23cc008046a/tools/get-flags.py, but I don't know how to run it on a regular basis :/

adulau:
> > Everything sound good to me. I also checked the licensing of the icons project and it's compatible with the AGPL. Maybe we need to keep track on how to update the flag file on a regular basis. For me it's fine to merge. @mokaddem any specific issue before merge?
> 
> There is a script for that: https://github.com/MISP/MISP/blob/d95f8acda12d1d228af9dd7147baf23cc008046a/tools/get-flags.py, but I don't know how to run it on a regular basis :/

Thanks! All good. Maybe I'll have a look at the action. Maybe I can add a generic GitHub action to update the various external references.

adulau:
@JakubOnderka Feel free to merge after the sami comments are fixed. Thanks again for the contribution!

app/Model/Organisation.php
JakubOnderka: Fixed, thanks!

PR #6415, opened 2020-10-09
Event report optimisations

JakubOnderka:
#### What does it do?

Just optimisation for `getProxyMISPElements` method. For my test event with just few attributes, loading is twice faster than before. For big events, it must be much more with a lot of tags or sightings it must be much more. What do you think @mokaddem? It this OK for you?

#### Questions

### DIFF
--- a/app/Model/EventReport.php
+++ b/app/Model/EventReport.php
@@ -1,6 +1,9 @@
 <?php
 App::uses('AppModel', 'Model');
 
+/**
+ * @property Event $Event
+ */
 class EventReport extends AppModel
 {
     public $actsAs = array(
@@ -416,7 +419,7 @@ class EventReport extends AppModel
      */
     public function getProxyMISPElements(array $user, $eventid)
     {
-        $event = $this->Event->fetchEvent($user, ['eventid' => $eventid]);
+        $event = $this->Event->fetchEvent($user, ['eventid' => $eventid, 'noSightings' => true]);
         if (empty($event)) {
             throw new NotFoundException(__('Invalid Event'));
         }
@@ -426,7 +429,7 @@ class EventReport extends AppModel
         ]]);
         $completeEvent = $event;
         if (!empty($parentEventId)) {
-            $parentEvent = $this->Event->fetchEvent($user, ['eventid' => $parentEventId, 'extended' => true]);
+            $parentEvent = $this->Event->fetchEvent($user, ['eventid' => $parentEventId, 'extended' => true, 'noSightings' => true]);
             if (!empty($parentEvent)) {
                 $parentEvent = $parentEvent[0];
                 $completeEvent = $parentEvent;
--- a/app/Model/EventReport.php
+++ b/app/Model/EventReport.php
@@ -448,12 +448,10 @@ class EventReport extends AppModel
         $templateConditions = [];
         foreach ($completeEvent['Object'] as $k => $object) {
             $objects[$object['uuid']] = $object;
-            $objectAttributes = [];
-            foreach ($object['Attribute'] as $i => $objectAttribute) {
-                $objectAttributes[$objectAttribute['uuid']] = $object['Attribute'][$i];
-                $objectAttributes[$objectAttribute['uuid']]['object_uuid'] = $object['uuid'];
+            foreach ($object['Attribute'] as $objectAttribute) {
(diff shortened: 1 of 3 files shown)


app/Model/EventReport.php
```diff
@@ -410,71 +406,86 @@ public function reArrangeReport(array $report)
...
+            $parentEvent = $this->Event->fetchEvent($user, array_merge(['event_uuid' => $event['Event']['extends_uuid']], $options));
             if (!empty($parentEvent)) {
-                $parentEvent = $parentEvent[0];
-                $completeEvent = $parentEvent;
+                $event = $parentEvent[0];
             }
         }
-        $attributes = Hash::combine($completeEvent, 'Attribute.{n}.uuid', 'Attribute.{n}');
-        $this->AttributeTag = ClassRegistry::init('AttributeTag');
+
```
mokaddem, line 438: We should record event tags from both the parent and extending event

app/Model/EventReport.php
```diff
@@ -410,71 +406,86 @@ public function reArrangeReport(array $report)
...
-        $allTagNames = array_merge($allTagNames, $attributeTags, $parentEventTags);
+
+        $attributes = [];
+        foreach ($event['Attribute'] as $attribute) {
+            unset($attribute['ShadowAttribute']);
+            $attributes[$attribute['uuid']] = $attribute;
+            foreach ($attribute['AttributeTag'] as $at) {
+                $allTagNames[$at['Tag']['name']] = $at['Tag'];
+            }
+            $this->Event->Attribute->removeGalaxyClusterTags($attribute);
```
mokaddem, line 453: Should `removeGalaxyClusterTags` be moved up? It seems to do nothing useful.

app/Model/EventReport.php
```diff
@@ -76,8 +81,8 @@ public function beforeValidate($options = array())
         // These fields all have sane defaults either based on another field, or due to server settings
         if (!isset($this->data['EventReport']['distribution'])) {
             $this->data['EventReport']['distribution'] = Configure::read('MISP.default_attribute_distribution');
-            if ($report['EventReport']['distribution'] == 'event') {
-                $report['EventReport']['distribution'] = 5;
+            if ($this->data['EventReport']['distribution'] == 'event') {
```
mokaddem, line 84: Nice catch..

app/Model/EventReport.php
```diff
@@ -271,27 +276,18 @@ public function buildACLConditions(array $user)
...
-            ));
-            if (empty($temp)) {
-                if ($throwErrors) {
-                    throw new NotFoundException(__('Invalid report'));
-                }
-                return array();
-            }
-            $reportId = $temp['EventReport']['id'];
-        } elseif (!is_numeric($reportId)) {
+        if (is_numeric($reportId)) {
```
mokaddem, line 279: Much cleaner

mokaddem:
This is absolutely fantastic! Thanks a lot for the very quick improvements!
Could you have a look at the comments? :)

app/Model/EventReport.php
JakubOnderka: Can you please describe, how it should works? Image that I have three events:
```
Parent event A
|- Child event B
|- Child event C
```
When I am editing report for *Child event B*, should I use attributes, objects and tags from all three events or just from *Child event B* and *Parent event A*?

app/Model/EventReport.php
JakubOnderka: Nice catch, my mistake.

app/Model/EventReport.php
mokaddem: - My idea is to use Attributes of `event A` (if the extended flag is provided, it will includes attributes from its childs)
- Same goes for objects
- For tags, I thought it would make sense to use *event tags* from `A` and `B` and not `C` - but it's debatable

JakubOnderka: OK, so I change the implementation to follow your rules. I can change it any time.

JakubOnderka:
> This is absolutely fantastic! Thanks a lot for the very quick improvements!
> Could you have a look at the comments? :)

Thanks for review, it is ready to merge.

adulau:
@mokaddem I let you decide if you merge it before the release 2.4.133 ;-)

mokaddem:
Thanks for the work!

PR #6416, opened 2020-10-10
Remove object checkbox

JakubOnderka:
#### What does it do?

Object checkboxes just do nothing, just makes confusion. So this patch removes them.

Snímek obrazovky 2020-10-10 v 12 46 50

#### Questions


adulau:
Good point indeed!

Issue #2362, opened 2017-07-27
Enable default tag per MISP server connection or default tag per MISP instance


maraboutingermany:
Just like the MISP feeds allow default tagging, I think it would be good to be able to default tag events that come per MISP server connection, the CIRCL MISP instance for example has events that have no tagging at all and a default TLP tag capability to be set per server would be a great time win

bocyb:
It will be also very useful to have a default tagging for User's  - ex. internal sandbox API user publishing events. Looking forward for the future like this in MISP

cudeso:
Default tagging for events coming from a specific server would certainly be a big win.

cudeso:
One possible solution is by sending the event to ZMQ

In **/app/Model/Server.php** change the function **__pullEvent** by adding the lines after "//Publish to ZMQ"

    private function __pullEvent($eventId, &$successes, &$fails, $eventModel, $server, $user, $jobId)
    {
        $event = $eventModel->downloadEventFromServer(
                $eventId,
                $server
        );
        ;
        if (!empty($event)) {
            if ($this->__checkIfEventIsBlockedBeforePull($event)) {
                return false;
            }
            $event = $this->__updatePulledEventBeforeInsert($event, $server, $user);
            if (!$this->__checkIfEventSaveAble($event)) {
                $fails[$eventId] = __('Empty event detected.');
            } else {
                $this->__checkIfPulledEventExistsAndAddOrUpdate($event, $eventId, $successes, $fails, $eventModel, $server, $user, $jobId);
            }
        } else {
            // error
            $fails[$eventId] = __('failed downloading the event');
        }

        // Publish to ZMQ
        $pubSubTool = $this->getPubSubTool();
        $pubSubTool->event_save(array('Event' => $eventId, 'Server' => $server['Server']['id']), 'add_from_connected_server');

PR #6429, opened 2020-10-13
Send a message to ZMQ when an event from a connected server is added or edited

cudeso:
#### What does it do?

It sends a message to ZMQ when an event from a connected server is added or edited.
Suggestion for https://github.com/MISP/MISP/issues/2362

#### Questions

- [x ] It does add two new events in ZMQ (when subscribed to the events stream): **add_from_connected_server** and **edit_from_connected_server**


adulau:
It makes sense. Thanks a lot for the contribution.

PR #6434, opened 2020-10-13
Add warninglist wildlist feature

kirzaks:
#### What does it do?

Adding new type "wildmask" in warninglist. This list will **exclude** all subdomains what is from warninglist, as at this moment all domains is working as wildmask.
For example. by default IOC **test.example.com** will be marked as warning because I have domain **example.com**in warninglist (type hostname). Adding domain **example.com* in warninglist by type **wildmask** will not mark as warning for example.com subdomains

Also domains included in **wildmask** warninglist wil not generate warn by self.
##### POC:
* type *wildmask* warninglist
```json
{
  "description": "List of wildmask domains",
  "list": [
    "armins.lv",
    "example.com"
  ],
  "matching_attributes": [
    "domain",
    "hostname"
  ],
  "name": "Wildmask domains",
  "type": "wildmask",
  "version": 19
}
```

* type *hostname* warninglist
```json
{
  "description": "My own list",
  "list": [
    "armins.lv",
    "palms.lv"
    
  ],
  "matching_attributes": [
    "domain",
    "hostname"
  ],
  "name": "My domains",
  "type": "hostname",
  "version": 19
}
```
##### Results:
* IOC `[test.armins.lv, armins,lv, test.palms.lv, palms.lv. test.example.com, example.com]`

Screenshot from 2020-10-13 17-46-18
In this case warn IOC should be only armins.lv, palms.lv and test.palms.lv as domains armins.lv and palms.lv is added in "My domains" list
test.armins.lv is not in warn as domain armins.lv is added in "Wildmask domains" list

#### Questions

NO
NO, but tested in Lab env.
NO

### DIFF
--- a/app/Controller/WarninglistsController.php
+++ b/app/Controller/WarninglistsController.php
@@ -282,10 +282,11 @@ class WarninglistsController extends AppController
 
             $hits = array();
             $warninglists = $this->Warninglist->getEnabled();
+            $wildlist = $this->Warninglist->getWildList($warninglists);
             foreach ($data as $dataPoint) {
                 foreach ($warninglists as $warninglist) {
                     $values = $this->Warninglist->getFilteredEntries($warninglist);
-                    $result = $this->Warninglist->quickCheckValue($values, $dataPoint, $warninglist['Warninglist']['type']);
+                    $result = $this->Warninglist->quickCheckValue($values, $dataPoint, $warninglist['Warninglist']['type'], $wildlist);
                     if ($result !== false) {
                         $hits[$dataPoint][] = array('id' => $warninglist['Warninglist']['id'], 'name' => $warninglist['Warninglist']['name']);
                     }
--- a/app/Model/Warninglist.php
+++ b/app/Model/Warninglist.php
@@ -509,7 +509,11 @@ class Warninglist extends AppModel
 		foreach ($warninglists as $warninglist) {
 			if ($warninglist['Warninglist']['type'] == 'wildmask') {
 				$mvalues = $this->getFilteredEntries($warninglist);
-				$wildlist = array_merge($wildlist, $mvalues);
+				foreach ($mvalues as $mvalue) {
+					if (substr($mvalue, 0,2) === "!.") {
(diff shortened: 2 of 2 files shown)


JakubOnderka:
Can you please describe more, why is this useful? For me it is contra intuitive that when enabling this warninglist, some warnigns will be excluded.

It is really necessary to have another warninglist? What about to use similar solution as gitignore format, so in "domain" warninglist has something like:

```
example.com
!.example.com
```

Then `example.com` will be marked as warning, but not any subdomain of example.com.

kirzaks:
Currently there are many warninglist from various sources, Microsoft, Mozilla, Alexa etc. Another warninglist for subdomains is a good solution  as there is no need to change every existing and upcoming lists and thus parser/downloader. 

Even more, with this new warninglist I can choose, which domain subdomains will not be marked as as warning because all subdomains are marked as warning by default. 

Just example: domain `example.com`
Imagine that domains are in two big warninglists with various domains: **list-1**, **list-2**
* list-1 contain records: `example.com`, `!.example.com` (to exclude all subdomains) + 1k various domains
* list 2 contain records: `example.com` + 1k various domains
If I use both lists, I will get subdomains without warning as list-1 contents `!.example.com`. But what if I want to mark all subdomains as warning? Disable list-1and I will lose 1k domain from this list? It is an issue and separated wildlist list will resolve it.

Yes, maybe just domain in wildlist type is confusing and adding  prefix `!.` will more show that this record will ignore subdomains.

adulau:
I tend to agree the prefix `!.` is much more intuitive. Indeed there are some use cases, where we have exceptions in the warning-lists and having a completing warning for some exceptions could useful.

@kirzaks Could you update the pr to change the prefix? and maybe add an example in the warning-list as a pr and I’ll merge it when this pr is merge to show how to use the new type. Thanks a lot!

kirzaks:
Updated and also warning-list PR is created

adulau:
Could you rebase to the latest version? Thanks a lot

PR #6447, opened 2020-10-15
chg: [internal] Normalize AS type to asplain notation

JakubOnderka:
#### What does it do?

This patch normalize AS number from asdot notation to asplain and removes AS prefix if provided. Proper normalization is important for computing correlations.

#### Questions


adulau:
Indeed, very good point.

PR #6451, opened 2020-10-16
Extending documentation of ShibbAuth plugin

Wachizungu:
#### What does it do?

Extends the documentation for the ShibbAuth plugin

#### Questions


adulau:
Thank you very much!

PR #6461, opened 2020-10-19
Event UI changes and fixes for event reports

JakubOnderka:
#### What does it do?

This changes a lot but small in event view user interface. Most important changes:
* 'Add event report' button will not be shown when user don't have permission for event edit
* Do not debounce tag loading for event reports for first load 
* Move 'Add event report' button to one line with other buttons to match MISP style.
Snímek obrazovky 2020-10-19 v 12 25 37

#### Questions


mokaddem:
Awesome!

PR #6467, opened 2020-10-19
chg: [internal] Slightly optimise attribute REST fetch

JakubOnderka:
#### What does it do?

Attribute REST fetch should be slightly faster when nothing was found.

#### Questions


app/Model/Attribute.php
```diff
@@ -4655,21 +4655,25 @@ private function __iteratedFetch($user, &$params, &$loop, TmpFileTool $tmpfile,
         $continue = true;
         while ($continue) {
             $results = $this->fetchAttributes($user, $params, $continue);
+            if (empty($resultsCount)) {
```
iglocska: This is only set later on in the function

PR #6471, opened 2020-10-20
new: [UI] Allow to disable hover enrichment

JakubOnderka:
#### What does it do?

With new option `Plugin.Enrichment_hover_popover_only` you can disable automatic hover enrichment and keep just intentional enrichment by click on magnifier glass icon. This can be useful to have more control, what data user sends to external services.

#### Questions


adulau:
Good idea!

PR #6492, opened 2020-10-23
Fix Bro IDS export docs

pettai:
#### What does it do?

As per https://github.com/MISP/MISP/pull/1726 the "allowNonIDS" option was explicitly removed from Bro IDS export, update the docs accordingly
(some hairpulling was done prior to this finding...)

#### Questions


adulau:
Good catch!

PR #6518, opened 2020-10-29
 chg: [UI] Remove Source Format from related feed popover

JakubOnderka:
#### What does it do?

I think that information is mostly irrelevant for user, if feed is MISP, CSV or text format.

Snímek obrazovky 2020-10-29 v 16 25 03

#### Questions


iglocska:
Indeed, makes sense, thanks!

PR #6519, opened 2020-10-29
fix: [internal] Properly set login time for custom auth

### LINKED ISSUE #5797: Update users.last_login with external auth user login
2.4.121

We are now using `CustomAuth` external authenticated user logins into MISP. Works real nicely: thanks!

Would it be possible to have the db `users.last_login` updated? Also, perhaps a login row in the `logs` table?

I do see that the db `logs` table gets a new row for every WebUI click where
```
logs.title = "Successful authentication using MyLogins External authentication key"
```
However, it is difficult to deduce which one is an initial login. I believe a non-auth user login does cause a distinct row insert into `logs`, but not for ext auth.

Open to suggestions on how we may collect external auth user login stats. Thanks...

JakubOnderka:
#### What does it do?

Will fix #5797

#### Questions

### DIFF
--- a/app/Controller/AppController.php
+++ b/app/Controller/AppController.php
@@ -1101,6 +1101,7 @@ class AppController extends Controller
                 if ($user['User']) {
                     unset($user['User']['gpgkey']);
                     unset($user['User']['certif_public']);
+                    $this->User->updateLoginTimes($user['User']);
                     $this->Session->renew();
                     $this->Session->write(AuthComponent::$sessionKey, $user['User']);
                     if (Configure::read('MISP.log_auth')) {
@@ -1189,6 +1190,7 @@ class AppController extends Controller
         if ($this->Auth->startup($this)) {
             $user = $this->Auth->user();
             if ($user) {
+                $this->User->updateLoginTimes($user);
                 // User found in the db, add the user info to the session
                 $this->Session->renew();
                 $this->Session->write(AuthComponent::$sessionKey, $user);
--- a/app/Controller/UsersController.php
+++ b/app/Controller/UsersController.php
@@ -1231,13 +1231,7 @@ class UsersController extends AppController
       ));
       $lastUserLogin = $user['User']['last_login'];
       unset($user['User']['password']);
-      $user['User']['action'] = 'login';
-      $user['User']['last_login'] = $this->Auth->user('current_login');
-      $user['User']['current_login'] = time();
-      $this->User->save($user['User'], true, array('id', 'last_login', 'current_login'));
-      if (empty($this->Auth->authenticate['Form']['passwordHasher']) && !empty($passwordToSave)) {
-          $this->User->saveField('password', $passwordToSave);
-      }
+      $this->User->updateLoginTimes($user['User']);
       $this->User->Behaviors->enable('SysLogLogable.SysLogLogable');
       if ($lastUserLogin) {
           $readableDatetime = (new DateTime())->setTimestamp($lastUserLogin)->format('D, d M y H:i:s O'); // RFC822
--- a/app/Model/User.php
+++ b/app/Model/User.php
@@ -1325,6 +1325,24 @@ class User extends AppModel
     }
 
     /**
+     * Updates `current_login` and `last_login` time in database.
+     *
+     * @param array $user
+     * @return array|bool
+     * @throws Exception
+     */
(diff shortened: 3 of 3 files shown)


app/Model/User.php
```diff
@@ -1324,6 +1324,22 @@ public function registerUser($added_by, $registration, $org_id, $role_id) {
         }
     }
 
+    /**
+     * @param array $user
```
chrisinmtown, line 1330: Please state what the function does (or should do)

app/Controller/UsersController.php
```diff
@@ -1231,13 +1231,7 @@ private function _postlogin()
       ));
       $lastUserLogin = $user['User']['last_login'];
       unset($user['User']['password']);
-      $user['User']['action'] = 'login';
-      $user['User']['last_login'] = $this->Auth->user('current_login');
-      $user['User']['current_login'] = time();
-      $this->User->save($user['User'], true, array('id', 'last_login', 'current_login'));
-      if (empty($this->Auth->authenticate['Form']['passwordHasher']) && !empty($passwordToSave)) {
```
chrisinmtown, line 1238: what about this special treatment of passwords?  The updateLoginTimes() function does not cover this

app/Controller/AppController.php
```diff
@@ -1101,6 +1101,7 @@ private function __customAuthentication($server)
                 if ($user['User']) {
                     unset($user['User']['gpgkey']);
                     unset($user['User']['certif_public']);
+                    $this->User->updateLoginTimes($user['User']);
```
chrisinmtown, line 1104: I believe, but am not 100% certain, this will update the login times for every page request. Ideally the server should do the update only on a transition from unauthenticated to authenticated.  But detecting that transition in a stateless server is difficult!

app/Controller/UsersController.php
JakubOnderka: Yes, because this functionality is broken. It depends on `$passwordToSave` variable, that is not defined in that function.

app/Controller/AppController.php
JakubOnderka: This should be OK, because after login, user has session stored in cookie. So this login process is repeated just after session expires.

app/Model/User.php
JakubOnderka: Thanks, done.

app/Controller/AppController.php
chrisinmtown: Ok so that's how you detect the transition - if there's no MISP cookie, call login which sets the cookie AND updates the current & last login times, do I have that correct?

chrisinmtown (approved):

chrisinmtown:
If you don't mind a distracting question, please point me to the code/function where the last_login is updated today in the normal non-custom-auth path.  I really struggled to find it.

iglocska:
> If you don't mind a distracting question, please point me to the code/function where the last_login is updated today in the normal non-custom-auth path. I really struggled to find it.

function _postlogin() in the UsersController. It is called on the login function as well as email_otp() function.

Also, this is not the field you want to use for auditing most likely, but rather the audit logs (action == login)

app/Controller/UsersController.php
iglocska: Yep, looks like this was broken when the login function was split in two with _postlogin(). We can clean quite a bit of code up, it's from a few years ago when we transitioned to bcrypt for the password hashing.

chrisinmtown:
> function _postlogin() in the UsersController. It is called on the login function as well as email_otp() function.

Thanks for that.  I need to learn more about PHP before I can understand your answer.  I naively tried grep to find invocations of the login() function in UsersController but found none.

app/Controller/AppController.php
JakubOnderka: Yes, thats correct.

iglocska:
> > function _postlogin() in the UsersController. It is called on the login function as well as email_otp() function.
> 
> Thanks for that. I need to learn more about PHP before I can understand your answer. I naively tried grep to find invocations of the login() function in UsersController but found none.

Basically every public function in a controller is user exposed and invoked by querying the given endpoint: baseurl/[controller_alias]/[controller_function]

However, _postlogin() is a private function and only invoked by other functions in the controller (such as login())

Hope that makes sense!

PR #6549, opened 2020-11-03
chg: [UI] Better handling missing variable in ajaxTags.ctp

JakubOnderka:
#### Questions

### DIFF
--- a/app/View/Elements/ajaxTags.ctp
+++ b/app/View/Elements/ajaxTags.ctp
@@ -2,10 +2,16 @@
     if (empty($scope)) {
         $scope = 'event';
     }
-    $searchUrl = '/events/index/searchtag:';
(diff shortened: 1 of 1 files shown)


iglocska:
This causes several issues. First of all the scope can be something not in the switch statement, in which case we don't rely on the variables being overriden and go with the defaults, the introduced exception breaks the tag index for example:

image

Furthermore, for all other scopes, we just default to the event level search (so if we want to show a tag somewhere, the default behaviour for a user when they click on it is to search the event index for the tag). When removing the exception call, this produces issues with the search URL not being set:

image

iglocska:
This was my bad, I didn't push the new index causing the issues.

https://github.com/MISP/MISP/commit/8fdcd3d3bf5bc9884cdbc51a07b78fc15a1c92c9

PR #6580, opened 2020-11-13
Fix: Fix spelling of sightings_anonymise_as description

Maddosaurus:
#### What does it do?

There is no existing issue as far as I can tell. 
I have found a spelling mistake in the description text for the Sightings Plugin (Plugin.Sightings_anonymise_as) and fixed it.

#### Questions


mokaddem:
Thanks a lot!

PR #6584, opened 2020-11-14
fix: [test] Update database before generating new user

JakubOnderka:
#### What does it do?

This should fix Travis after new autkeys implementation.

#### Questions


iglocska:
Good idea. I've also pushed the update of MYSQL.sql so we sohuldn't need to run the updates anymore on a fresh setup that caused the issues - but it's indeed a good idea to reverse the order of the execution for future changes, thanks!

PR #6585, opened 2020-11-14
Autkey fixes

JakubOnderka:
#### What does it do?

Fixes a security bug in new AuthKey implementation, authkey was actually validated just by first and last 4 chars. 

#### Questions


iglocska:
Woah indeed, nice catch. Merged, thanks!

PR #6590, opened 2020-11-16
Feature report from event

mokaddem:
Generates a report from an event based on filtering criterias


app/View/EventReports/ajax/indexForEvent.ctp
```diff
@@ -31,6 +31,15 @@
...
                                     'requirement' => $canModify && $importModuleEnabled,
                                 ),
+                                array(
+                                    'onClick' => 'openGenericModal',
+                                    'onClickParams' => [$baseurl . '/eventReports/reportFromEvent/' . h($event_id)],
+                                    'active' => true,
+                                    'text' => __('Generate report from Event'),
+                                    'title' => __('Based on filters, create a report summarizing the event'),
+                                    'fa-icon' => 'list-alt',
+                                    'requirement' => $canModify,
```
JakubOnderka, line 41: Maybe not necessary? It can be available for everybody with read access, no?

mokaddem: Not really, it actually creates an event report from the event.
So you have to be able to modify the event to run this function.

PR #6592, opened 2020-11-16
chg: Add action workflow

Rafiot:
Quick and dirty port of the travis file


JakubOnderka:
@Rafiot Try to include https://github.com/MISP/MISP/pull/6599/files, this should help with broken update script

PR #6612, opened 2020-11-18
chg: [galaxy] Do not fetch full galaxy info for event view UI

JakubOnderka:
#### What does it do?

Loading all cluster info is so slow and not necessary.

#### Questions


mokaddem:
Awesome!

PR #6615, opened 2020-11-18
fix: [internal] Tags are not required for galaxy relation

JakubOnderka:
#### What does it do?

Should fix notices from test:

```
Notice Error: Undefined index: tags in [/home/runner/work/MISP/MISP/app/Model/Galaxy.php, line 215]
```

#### Questions


JakubOnderka:
@mokaddem Can you please confirm that `tags` field should be optional?

JakubOnderka:
Already fixed in https://github.com/MISP/MISP/commit/91f6479660037883591f1af47f911de6b29bb025

PR #6665, opened 2020-11-29
fix: [security] Remove hashed advanced keys from response

JakubOnderka:
#### What does it do?

Hashed keys should not be accessible.

#### Questions


iglocska:
Nice catch, thanks!

PR #6670, opened 2020-11-30
new: [objectTemplate] Allow fetching the raw template stored on disk by UUID or name

mokaddem:
Fetching the raw object's template stored on disk can be done by using the following URLs
- `/objectTemplates/getRaw/dc6a8fa2-0a43-4a0c-a5aa-b1a5336ca80e`
- `/objectTemplates/getRaw/ail-leak`

:warning: The `perm_object_template` permission is required


Rafiot:
Just to be sure: if there are two templates with the same name, do we get a list?

app/Model/ObjectTemplate.php
```diff
@@ -316,4 +316,41 @@ public function setActive($id)
...
     }
+
+    public function getRawFromDisk($uuidOrName)
+    {
+        $templates = $this->readTemplatesFromDisk();
+        if (Validation::uuid($uuidOrName)) {
+            return $templates[$uuidOrName];
+        } else {
+            foreach ($templates as $uuid => $template) {
+                if ($template['name'] == $uuidOrName) {
```
chrisinmtown: @Rafiot  suggests that multiple templates might have the same name.  This would randomly bring back only one of them.  But given this reads directories below files/misp-objects/objects, and those names are forced to be unique by the filesystem, this actually looks ok :/

app/Model/ObjectTemplate.php
```diff
@@ -316,4 +316,41 @@ public function setActive($id)
         }
         return 1;
     }
+
+    public function getRawFromDisk($uuidOrName)
```
chrisinmtown, line 321: does this function need to be public, or could it be private like readTemplatesFromDisk ?

app/Model/ObjectTemplate.php
```diff
@@ -316,4 +316,41 @@ public function setActive($id)
...
+    {
+        $templates = $this->readTemplatesFromDisk();
```
chrisinmtown: It seems pretty wasteful to read every single JSON object to find one of them.  Would you consider reading them and stopping when a match on UUID is found? Maybe you don't want to cache them to ensure the cache does not go stale?

@Rafiot  what is the relationship of an object template directory name and the object template name?  Today the directory with the definitions.json file seems to set the name, and there's no way to have two directories of the same name below files/misp-objects/objects/

app/Model/ObjectTemplate.php
```diff
@@ -316,4 +316,41 @@ public function setActive($id)
...
+                if ($template['name'] == $uuidOrName) {
+                    return $template;
+                }
+            }
+        }
+    }
+
+    private function readTemplatesFromDisk()
+    {
+        $objectsDir = APP . 'files/misp-objects/objects';
```
chrisinmtown: Please move this to a constant to avoid storing this path here? See line 52 in function update where this *exact* same path string also appears, that could use the constant also

app/Model/ObjectTemplate.php
```diff
@@ -316,4 +316,41 @@ public function setActive($id)
...
+        $directories = glob($objectsDir . '/*', GLOB_ONLYDIR);
+        foreach ($directories as $k => $dir) {
+            $dir = str_replace($objectsDir, '', $dir);
+            $directories[$k] = $dir;
+        }
+        $templates = [];
+        foreach ($directories as $dir) {
+            if (!file_exists($objectsDir . DS . $dir . DS . 'definition.json')) {
```
chrisinmtown: is the string 'definition.json' available as a constant somewhere?  Similar usage on lines 63 and 66 in this file.

app/Model/ObjectTemplate.php
mokaddem: Exactly! You can't have duplicated name on the filesystem, so there is no need to return more than once. Same goes for UUIDs, you should not have 2 objects with the same UUID.

app/Model/ObjectTemplate.php
mokaddem: This function is used by the controller and could be used for other purpose.

app/Model/ObjectTemplate.php
mokaddem: There is a lot of duplication in the Model. This and the update process should be refactored.
I thought the feature was needed as fast as possible and that optimizations and refactoring for the whole model could wait.

chrisinmtown: In my experience, duplication like this is rarely or never cleaned. It's far better to start clean instead of adding technical debt like this. This is only one string constant that can be used in two places within this file, should be easy, right?

app/Model/ObjectTemplate.php
chrisinmtown: The method for checking the name uses the string from inside the file, not the directory name.  Where is the code that reads these in to the database and/or UI, and what does it do?  I recommend extending this method with a comparison of the directory name with the name inside the file, and log an error/skip the entry if they do not match.

Rafiot:
Passing an object template to MISPObject is now possible: https://github.com/MISP/PyMISP/commit/0d86a4339f13f9d6634ec1523a25ecae226f8340

app/Model/ObjectTemplate.php
Rafiot: Alright, that should work fine then, as long as we're sure to never endup with two templates with the same name at any time in the future.

chrisinmtown: Well the names *in* the files could overlap, but the names *of* the files never can.  So if the code checks for that case, it should be pretty solid.

chrisinmtown:
Hi @mokaddem would you please consider my comments here and move this along?

app/Model/ObjectTemplate.php
chrisinmtown: Maybe this is unnecessary optimization, but if the parameter is a name, the operation can be done in O(1) aka constant time - look for the file, read the file, return the contents (or error if no file).  If the parameter is a UUID, then the operation requires O(N) bcos you have to scan all the files.  I expect that the most frequent use case from the PyMISP side is to fetch an object template by name, because that's convenient to use in code, it's meaningful to people, etc.

chrisinmtown:
good day CIRCL, is this stalled? please advise if there's some way I can help

chrisinmtown:
happy new year to the CIRCL team, can you please find a minute to review this PR and consider my comments?

mokaddem:
Implemented comments and change request by @chrisinmtown (thanks!)
PR will be merged on `develop` branch. Closing.

chrisinmtown:
good day @mokaddem thanks for pushing this along.  I guess this has not yet been merged to branch 2.4? Do you plan to release this as part of version 2.4.138?

iglocska:
Yeah, anything going to develop is part of the next release (so 2.4.138)

PR #6677, opened 2020-12-01
new: [internal] HTTP header with username in the reponse

lfortemps:
#### What does it do?

It sets the X-Username header in the HTTP response which allows Apache to log the username in the access logs

#### Questions

### DIFF
--- a/app/Controller/AppController.php
+++ b/app/Controller/AppController.php
@@ -333,6 +333,7 @@ class AppController extends Controller
             if ($this->Auth->user('Role')['perm_site_admin'] || (Configure::read('MISP.live') && !$this->_isRest())) {
                 $this->{$this->modelClass}->runUpdates();
             }
+            header('X-Username: '.$this->Auth->user('email'));
             $user = $this->Auth->user();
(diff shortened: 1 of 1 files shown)


cvandeplas:
Thank you very much!
As discussed it would be great to be able to distinguish the user from API access with each API key.
So for example:
- user/pass => "$username"
- api key (old) => "$username/API-default"
- api key (new) => "$username/API-$keydescription"

JakubOnderka:
Great idea! Because I heavily refactored that part of code, I included that in https://github.com/MISP/MISP/pull/6581/commits/9898c163e49d1ae73d025acc3119f1e374b8f679. I want to merge that this week.

The format will be:
* `example@example.com` for user/password authentication
* `example@example.com/API/default` for old API keys
* `example@example.com/API/1` for new API keys where 1 is API key ID

What do you think?

cvandeplas:
Looks great! And it includes tests, excellent!

In just wondering if we really need to add yet another configuration variable for this non breaking feature, that has no negative impact on anything.
MISP already has a gazillion features to enable, and I'm not sure of the added value of one more for this minor header.

PR #6699, opened 2020-12-05
fix: Don't fail writable attachment dir test for S3

folbricht-stripe:
#### What does it do?

Prevents a failure when using S3 attachment storage. There's a test in the code confirming if the directory is writable and it uses `is_dir()` to do so. Currently the following warning is written to the log on startup

```
2020-11-21 02:46:22 Warning: is_dir(): Unable to find the wrapper "s3" - did you forget to enable it when you configured PHP? in [/var/www/MISP/app/Model/Server.php, line 3589]
```

The same happens when running `cake Admin setSetting "MISP.attachments_dir" "s3://"`

#### Questions

### DIFF
--- a/app/Lib/Tools/CryptGpgExtended.php
+++ b/app/Lib/Tools/CryptGpgExtended.php
@@ -73,7 +73,8 @@ class CryptGpgExtended extends Crypt_GPG
     }
 
     /**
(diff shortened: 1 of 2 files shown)


JakubOnderka:
Thanks, merged into `develop` branch.

PR #6737, opened 2020-12-10
fix: [internal] Remove compact method call that do nothing

JakubOnderka:
#### What does it do?

Just generates noise inside logs.

#### Questions


iglocska:
Indeed, no ordered params on those functions anyway, nice catch!

PR #6778, opened 2020-12-18
Tag info popover

JakubOnderka:
#### What does it do?

Before I was sad, because useful info from taxonomy about tag is hard to find. For example if user wants to know more about `tlp:white` tag, he need to go to taxnomy and manually find correct tag.

Now it is easy with dynamic popover.

Snímek obrazovky 2020-12-18 v 20 54 21

#### Questions

### DIFF
--- a/app/Controller/TagsController.php
+++ b/app/Controller/TagsController.php
@@ -1108,8 +1108,12 @@ class TagsController extends AppController
                 $conditions['OR'][] = array('LOWER(Tag.name) LIKE' => $t);
             }
         } else {
-            foreach ($tag as $k => $t) {
-                $conditions['OR'][] = array('Tag.name' => $t);
+            foreach ($tag as $t) {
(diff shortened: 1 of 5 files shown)


adulau:
Love it. Maybe displaying the color and the numerical value if present would be also super cool. Thanks a lot!

JakubOnderka:
> Love it. Maybe displaying the color and the numerical value if present would be also super cool. Thanks a lot!

I added numerical value, but I think that tag color is not that useful information to show in this box. You mean color as hex representation? Can you describe your use case?

adulau:
> > Love it. Maybe displaying the color and the numerical value if present would be also super cool. Thanks a lot!
> 
> I added numerical value, but I think that tag color is not that useful information to show in this box. You mean color as hex representation? Can you describe your use case?

Just a small coloured box within the tag info with the colour. As a reminder for the user.

PR #6782, opened 2020-12-20
chg: [UI] Convert taxonomies to default view

JakubOnderka:
#### What does it do?

Use generic view for taxonomies, that allows to search and filter just enabled/disabled taxonomies.

#### Questions


iglocska:
Lovely, thanks!

PR #6783, opened 2020-12-21
Add the possibility to have a '-' in the baseurl

FafnerKeyZee:
With the actual regex in testBaseURL, we can not have a '-' inside the BaseURL, I did a quick fix

With the actual the domain must be: https://sub.mydomain.ext
Witht this PR we can have: https://sub.my-domain.ext

Best regards,


adulau:
Good catch! Thanks a lot!

PR #6799, opened 2020-12-23
INSTALL.rhel8.md - update EoL for CentOS 8

simonflood:
Maintenance for CentOS 8 will now end on 31 December 2021


adulau:
Thanks for the update.

PR #6860, opened 2021-01-14
fix: [log] Correctly handle limit and page params

JakubOnderka:
#### What does it do?

Will fix https://github.com/MISP/PyMISP/issues/680

#### Questions


app/Controller/LogsController.php
```diff
@@ -117,16 +110,14 @@ public function admin_index()
     // Shows a minimalistic history for the currently selected event
     public function event_index($id)
     {
-        // check if the user has access to this event...
```
chrisinmtown, line 120: is this change related to the limit & page params fix?

chrisinmtown:
Thanks for finding time for this fix!

PR #6890, opened 2021-01-20
fix: [UI] Show proper unit for diagnostics

### LINKED ISSUE #6886: MISP serverSettings, diagnostics section mis-reports PHP memory_limit showing M not G
Using MISP version 2.4.133.  The server settings, diagnostics section controller seems to mis-parse a gigabyte value.  The server's PHP ini file has this entry:
```
memory_limit = 128G
```
But the web UI at https://misp.my.company.com/servers/serverSettings/diagnostics shows this:
```
 memory_limit (128M)… Low (recommended: 2048M)
```

I think the problem is near here:
https://github.com/MISP/MISP/blob/0859f0d0601fb975f3f5cff24204474e88e04aae/app/Controller/ServersController.php#L1087

### Work environment

| OS version (server)       |  RedHat
| OS version (client)       | any
| PHP version               | 7.4
| MISP version / git hash   | 2.4.133
| Browser                   | any

### Expected behavior

I expect the serverSettings, diagnostics section to report accurately the configured PHP memory_limit value.

### Actual behavior

The serverSettings, diagnostics section shows the wrong memory_limit units, claims limit of 128M when the real limit is 128G

Configure php.ini with a G value, e.g., 64GB then view the serverSettings/diagnostics page.

I think all the details are above but I'll be glad to add more, just let me know.

JakubOnderka:
#### What does it do?

Should fix #6886.

#### Questions

### DIFF
--- a/app/Controller/ServersController.php
+++ b/app/Controller/ServersController.php
@@ -1042,35 +1042,35 @@ class ServersController extends AppController
                 $this->set('branch', $gitStatus['branch']);
                 $this->set('commit', $gitStatus['commit']);
                 $this->set('latestCommit', $gitStatus['latestCommit']);
+
                 $phpSettings = array(
                     'max_execution_time' => array(
                         'explanation' => 'The maximum duration that a script can run (does not affect the background workers). A too low number will break long running scripts like comprehensive API exports',
                         'recommended' => 300,
-                        'unit' => false
+                        'unit' => 'seconds',
                     ),
                     'memory_limit' => array(
                         'explanation' => 'The maximum memory that PHP can consume. It is recommended to raise this number since certain exports can generate a fair bit of memory usage',
                         'recommended' => 2048,
-                        'unit' => 'M'
+                        'unit' => 'MB'
                     ),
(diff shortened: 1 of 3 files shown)


app/Controller/ServersController.php
```diff
@@ -1042,35 +1042,35 @@ public function serverSettings($tab=false)
...
                 $phpSettings = array(
                     'max_execution_time' => array(
                         'explanation' => 'The maximum duration that a script can run (does not affect the background workers). A too low number will break long running scripts like comprehensive API exports',
                         'recommended' => 300,
-                        'unit' => false
+                        'unit' => 'seconds',
                     ),
                     'memory_limit' => array(
                         'explanation' => 'The maximum memory that PHP can consume. It is recommended to raise this number since certain exports can generate a fair bit of memory usage',
                         'recommended' => 2048,
```
chrisinmtown, line 1054: you could make a good case for changing the number to just 2 and using unit 'GB', right?

chrisinmtown:
thanks @JakubOnderka for taking this one so quickly!

PR #6895, opened 2021-01-20
fix: [login] Convert old password hash to blowfish

JakubOnderka:
#### Questions


cvandeplas:
thanks, it indeed fixes the issue. I've merged it to 2.4 branch

PR #6916, opened 2021-01-27
fix: [UI] Allow to sort feeds by name

### LINKED ISSUE #6915: Enable name column on Feeds page to sort the table rows
Cannot sort feeds table by name in the web UI Feeds page https://misp.mycompany.comfeeds/index/scope:all because the column head "Name" is not a link / not clickable.

I'm using version 2.4.133 on the server, viewing the page in Firefox version 84.  Is the Name column enabled for sorting in more recent versions or different browsers?  In other words, am I missing something obvious? 

Add an appropriate link/action to the Name column header so the table rows can be sorted on name, just as is possible for almost every other column: id, enabled, caching, format, provider, ..

If it's easy, enable *every* column head. The ones not enabled in v2.4.133: name, headers, target, distribution, tag.

I can sort on provider which gets me close to sorting on name.

Thanks for listening!

JakubOnderka:
#### What does it do?

Fixes #6915

#### Questions

### DIFF
--- a/app/View/Feeds/index.ctp
+++ b/app/View/Feeds/index.ctp
@@ -82,7 +82,7 @@
                     )
                 ),
                 array(
-                    'name' => __('Id'),
+                    'name' => __('ID'),
                     'sort' => 'Feed.id',
                     'class' => 'short',
                     'data_path' => 'Feed.id',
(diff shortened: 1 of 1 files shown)


chrisinmtown:
That was fast, thanks.

PR #6932, opened 2021-01-30
fix: [API] Allow 'hard' param in POSTed body for deleting a cluster, send b…

tomking2:
#### What does it do?
This allows you to post the 'hard' param in a POSTed body for deleting a MISP Galaxy Cluster. Once deleted, it sends back the generated message instead of the default delete REST response.

Also spotted some crashes when deleting clusters, whereby the `distribution` was not set, causing the validate function to crash.

Also fixes a bug where if you attempt to update a cluster and pass Authors as a valid array, it crashes (tries to decode it as JSON).

Will shortly be followed by a PR in PyMISP to add in galaxy cluster manipulation capability.

#### Questions


mokaddem:
Hey, thanks a lot for the PR!
Sorry it took so long to be reviewed.
It's now merged in the [develop branch](https://github.com/MISP/MISP/tree/develop).
Thanks again!

Issue #6912, opened 2021-01-27
Unselectable (hidden) Tags usable by anyone


marjatech:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | Ubuntu
| OS version (client)       | Win10, Ubuntu
| PHP version               | 7.4
| MISP version / git hash   | v2.4.137 (99b405b73e34ff3fc84af642d14d4e8c99eff7ba) 

### Expected behavior
If a Tag is edited to be unselectable, it would be shown as Hidden in the Tag List, and not selectable by editing an event in the GUI.

### Actual behavior
1. Hidden Tags can be selected by any user at least under 'Custom Tags'. Is this intended?
2. In the Tag-List these Tags also are not marked as Hidden.

### Steps to reproduce the behavior
- Edit any Tag
- Enable `Unselectable Tag`
- Try to add this tag to any event

marjatech:
PR fixes 2. (tag-list view)
How about 1. Is this intended?

PR #6937, opened 2021-02-01
fix: correctly show hidden tags in tag-list

marjatech:
#### What does it do?

Fixes one part of #6912 
hidden tags not beeing shown as hidden in tag-list.

#### Questions


JakubOnderka:
Thanks for contribution

PR #6938, opened 2021-02-01
[API] Update attribute sightings from REST POST

tomking2:
#### What does it do?

This allows you to update attribute-level sightings when POSTing a MISP Event to "updateEvent". This is especially useful in PyMISP when adding sightings to attributes, and updating the entire event to include these sightings. Otherwise, you currently have to manually add in sightings in separate API requests which is long winded.

#### Questions


app/Model/Attribute.php
```diff
@@ -4077,6 +4077,12 @@ public function editAttribute($attribute, $eventId, $user, $objectId, $log = fal
                 $attribute['distribution'] = 5;
             }
         }
+        if (isset($attribute['Sighting']) && !empty($attribute['Sighting'])) {
+            $this->Sighting = ClassRegistry::init('Sighting');
+            foreach ($attribute['Sighting'] as $s) {
```
JakubOnderka: Thanks for contributing. Better to use `Sightings::captureSightings` method.

Example usage:
https://github.com/MISP/MISP/blob/bd98ea937de43b8a311b1f71912ba71d3e674746/app/Model/Attribute.php#L3977-L3979

tomking2: Thanks @JakubOnderka, didn't spot that was an available function, will update the code :)

tomking2:
Thanks @JakubOnderka - Updated to use the recommend function. Had to add a further check into that function to pull the Org ID from the user object otherwise the sighting was added without an organisation.

JakubOnderka:
Thanks for contribution!

PR #6984, opened 2021-02-09
new: [oidc] OpenID Connect authentication

JakubOnderka:
#### What does it do?

Another authentication method. 

#### Questions

### DIFF
--- a/app/Model/User.php
+++ b/app/Model/User.php
@@ -7,6 +7,8 @@ App::uses('SendEmail', 'Tools');
 
 /**
  * @property Log $Log
+ * @property Organisation $Organisation
+ * @property Role $Role
  */
 class User extends AppModel
 {
--- a/app/Plugin/OidcAuth/Controller/Component/Auth/OidcAuthenticate.php
+++ b/app/Plugin/OidcAuth/Controller/Component/Auth/OidcAuthenticate.php
@@ -1,0 +1,244 @@
+<?php
+use Jumbojett\OpenIDConnectClient;
+App::uses('BaseAuthenticate', 'Controller/Component/Auth');
+
+/**
+ * Config options:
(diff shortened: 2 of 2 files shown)


garrit-schroeder:
@JakubOnderka Could you please help me out to setup oidc with your code. I am having struggle to get this working. 
I get directed to my IDP (Keycloak) that works fine but then I end up in a redirect loop until my browser stops because of too many redirects. 

I debug the code and within the library of Jumbojett 
The following if clause is never run through (even though i can see in my browser that it has an attribute in the url ?....&code=XXX)

OpenIDConnectClient.php Line 297

```
// If we have an authorization code then proceed to request a token
if (isset($_REQUEST['code'])) {

    $code = $_REQUEST['code'];
    $token_json = $this->requestTokens($code);
```
Your help is much appreciated

garrit-schroeder:
Hello @JakubOnderka 

I was able to setup authentication with open id connect. But I am having trouble persisting the config file in my MISP installment.

MISP somehow reverts my config.php file and removes the following configuration block if a change a setting in Server Settings & Maintenance via the Web interface:

```
  'OidcAuth' => [
        'provider_url' => 'https://KEYCLOAK/auth/realms/REALM/.well-known/openid-configuration',
        'client_id' => 'mymisp',
        'client_secret' => 'XXX',
        'role_mapper' => [ // if user has multiple roles, first role that match will be assigned to user
            'user' => 3, // User
            'admin' => 1, // Admin
        ],
        'default_org' => 'default',
    ],
    
  The Security configuration in config.php changes itself to the following:
```
```
'Security' => 
  array (
    ...
    array (
      0 => 'OidcAuth.Oidc',
    ),
  ),

garrit-schroeder:
https://github.com/MISP/MISP/issues/7208

PR #6988, opened 2021-02-10
fix: Elasticsearch complains when an IP is an empty string

tomking2:
#### What does it do?

Since the IP field was being populated with an empty string as part of commit https://github.com/MISP/MISP/commit/14ca7ddf49556e12a58a91929770a6219b3786c1, when this data is sent to ElasticSearch it crashes the log, due to the "IP not being a valid string literal" (this is when the IP field is set as type ip). Setting it to null if it is in fact empty allows this to continue.

#### Questions


adulau:
Indeed, good catch! Thanks a lot for pr.

PR #7033, opened 2021-02-18
fix: [tools] misp-wipe updated list of table to truncate

mokaddem:
Updated list of table to truncate in `misp-wipe`


iglocska:
Thanks a lot!

PR #7037, opened 2021-02-18
Adds 'List Auth Keys' button to Administration in global menu

Wachizungu:
#### What does it do?

Adds 'List Auth Keys' button to Administration in global menu.
We find this view useful with the new auth keys functionality and it seems there is no link to it in the menu yet.

#### Questions

- [ ] Are you using it in production? Not yet but we will


adulau:
Thank you!

PR #7052, opened 2021-02-21
fix: [server] Add application/x-pie-executable to the list of accepted mimetypes in testForBinExec

### LINKED ISSUE #7051: MISP.python_bin refuses python executable when its mimetype is application/x-pie-executable
### Work environment

| OS version (server)       | OS Independent
| OS version (client)       | Client OS independent
| PHP version               | PHP independent
| MISP version / git hash   |  HEAD, feefa591e9e864701c487c66ba6d7f065fd30b5

### Expected behavior

MISP.python_bin can point to an executable even if its mimetype is "application/x-pie-executable"

### Actual behavior

If mimetype of the python executable == "application/x-pie-executable", validator fails for MISP.python_bin

- Create a path independent virtualenv
- Set MISP.python_bin accordingly

N/A

Reason is that
https://github.com/MISP/MISP/blob/feefa591e9e864701c487c66ba6d7f065fd30b50/app/Model/Server.php#L1539
strictly checks mimetypes, but the provided list  is not complete

### Proposed solution
include application/x-pie-executable as well in the list

patriziotufarolo:
#### What does it do?

Fixes #7051 

#### Questions

### DIFF
--- a/app/Model/Server.php
+++ b/app/Model/Server.php
@@ -1536,7 +1536,7 @@ class Server extends AppModel
             return true;
         }
         if (is_executable($value)) {
-            if (finfo_file($finfo, $value) == "application/x-executable" || finfo_file($finfo, $value) == "application/x-sharedlib") {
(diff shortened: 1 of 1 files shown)


adulau:
Good catch! We completely forgot the position-independent executables. Thank you for the PR.

Issue #7040, opened 2021-02-18
[Enhancement] - Can you please define a ZeroMQ parameter for the "host"


joydragon:
**Is your feature request related to a problem? Please describe.**
Whenever you activate the ZeroMQ plugin it'll always start with 0.0.0.0

**Describe the solution you'd like**
Can you please define a ZeroMQ parameter for the "host" that's going to be listen to on the plugin settings page?
It wouldn't be a huge change, and it'll prevent the server to be open to everyone everytime.
The change would be needed at:
/MISP/app/files/scripts/mispzmq/mispzmq.py line 105-106

```
self.socket.bind("tcp://*:{}".format(self.settings["port"]))
self._logger.debug("ZMQ listening on tcp://*:{}".format(self.settings["port"]))
```
for something like:
```
self.socket.bind("tcp://{}:{}".format(self.settings["host"], self.settings["port"]))
self._logger.debug("ZMQ listening on tcp://{}:{}".format(self.settings["host"],self.settings["port"]))
```

**Describe alternatives you've considered**
Making the change on my own on the line considered above.
I've already setup the server to start as 127.0.0.1, but it'd be better if you can change it on the MISP interface.

adulau:
This is a very good point.

adulau:
Related to #7039

Issue #7039, opened 2021-02-18
ZMQ plugin listens on 0.0.0.0 and can't be changed


eCrimeLabs:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | Debian
| OS version (client)       | Windows 10
| PHP version               | 7.4
| MISP version / git hash   |  v2.4.139 (8233f0f6143472be376ab31ea7ca6d18d670b5fd)

### Expected behavior
It should be possible under the plugin settings around ZeroMQ to define what IP the ZeroMQ should listen on, similar to that is is possible to set the port.

### Actual behavior
In the plugin settings around ZeroMQ it is as default listening on 0.0.0.0 meaning if not firewall on the machine this would be exposed.

https://github.com/MISP/MISP/blob/8d9100b4a41c658d14956c7dd94ba3dd9c18ecbf/app/files/scripts/mispzmq/mispzmq.py 
on line 105 and 106
```
       self.socket.bind("tcp://*:{}".format(self.settings["port"]))
        self._logger.debug("ZMQ listening on tcp://*:{}".format(self.settings["port"]))
```

To mitigate this the above lines should change the * to 127.0.0.1

### Logs, screenshots, configuration dump, ...
![image](https://user-images.githubusercontent.com/14878613/108407714-62fbf280-7224-11eb-930a-0f225d78b807.png)

PR #7053, opened 2021-02-21
Fix for ZeroMQ - #7040 and #7039

eCrimeLabs:
#### What does it do?
The following commit contains the fix for ZeroMQ only listening on 0.0.0.0 also mentioned in #7040 and #7039 there is now a feature under plugins/ZeroMQ allowing to configure what IP the ZeroMQ should be listening on. 

image

This is directly related to: #7039 and #7040 

#### Questions

- [Maybe] Does it require a change in the API (PyMISP for example)?  I'm unaware if there are any other places where the ZeroMQ can be configured from, so uncertain.


adulau:
Thank you very much!

PR #7056, opened 2021-02-21
chg: [UI] Add can access check for correlation exclusions menu entry

Wachizungu:
#### What does it do?

Adds can access check to the global menu entry of correlation exclusions (to go to the index).

The following commit added correlation exclusions functionality:
https://github.com/MISP/MISP/commit/b8823b86e286ef0b821547f131f9e350a613a149

However, a menu entry shows up for read only users even though ACL blocks them from seeing that page. This change should fix that. Also changed the array format just to align with the rest of the file.

#### Questions


JakubOnderka:
Thanks for contribution.

PR #7059, opened 2021-02-22
Inclusion of full-name under person

adammchugh:
Proposing the inclusion of full-name under person to allow for better capture and correlation of full names of identified persons in events. Particularly where there are multiple identities within an event which may create confusion with multiple first-name and last-name entries.

#### What does it do?

#### Questions


iglocska:
As for the issue described in the PR - to avoid intermixing a bunch of first/last names in an event, please use the person object, that should resolve that already.

As for the correlations being muddied by common first / last names this is indeed a good idea, mainly merged it for this reason.

Thanks a lot!

PR #7062, opened 2021-02-22
chg: [permissions] User with Tag editor can edit and delete tags

JakubOnderka:
#### What does it do?

Currently, tag can modify or delete just site admin. But according to description of `perm_tag_editor` *This permission gives users the ability to create, modify or remove tags.* This patch changes the behavior according to permission description.

#### Questions

### DIFF
--- a/app/Controller/Component/ACLComponent.php
+++ b/app/Controller/Component/ACLComponent.php
@@ -594,8 +594,8 @@ class ACLComponent extends Component
             'tags' => array(
                     'add' => array('perm_tag_editor'),
                     'attachTagToObject' => array('perm_tagger'),
-                    'delete' => array(),
-                    'edit' => array(),
+                    'delete' => array('perm_tag_editor'),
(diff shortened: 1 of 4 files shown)


iglocska:
I like the changes to the view file (removing the ability of non site admins to restrict to an org - but for the deletions and modifications of the tag, it should be restricted to site admins.

Basically tags are used for flow control (imagine setting a sync filter for tlp:amber) - if a malicious user would edit the tag to tlp:idontcare then the flow filter would be ignored and the instance would start leaking information. 

If we really wanted to implement editing / deletion, we would first need to check if ANY data is tagged, any collection has a tag, any default instance filters are set to any of the tags, any feeds have a tag attached and if yes, block the deletion/edit by anyone but site admins.

So tl;dr - only site admins should edit/delete, fixing the description is the better way forward.

JakubOnderka:
@iglocska Thanks, makes sense. I will change perm description.

PR #7070, opened 2021-02-23
fix: syntax error in constructTaxonomyInfo()

fsoderblom:
## What does it do?

Fixes a syntax error in misp.js that prevents login on a freshly installed system since when chaining expressions, trailing commas are not allowed which leads to error and stops further JS processing on firefox and chrome. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Unexpected_token

## Questions


iglocska:
Thanks a lot!

PR #7071, opened 2021-02-23
Required predicate

JakubOnderka:
#### What does it do?

Some taxonomies combines different types of tags. But currently, required can be just whole taxonomy so that means for publishing it is OK to add any taxonomy tag. This patch changes that and allow for site admins set just predicates as required, so site admin can require specific taxonomy predicates or even multiple predicates from taxonomy.

Makes sense for you?

Snímek obrazovky 2021-02-23 v 17 52 25

Snímek obrazovky 2021-02-23 v 17 50 34

#### Questions


JoePJisc:
We were discussing how useful this would be yesterday; we have our own taxonomy and would like to mandate certain predicates form it for all our events before they can be published, in the same way we mandate the TLP namespace.

Hope this can be added to MISP soon.

PR #7074, opened 2021-02-23
fix: ui  _ function does not exist, l10n function is __

patriziotufarolo:
_ function is unknown, maybe a typo

#### What does it do?

I didn't open an issue, but this prevents submodule's table to work properly if error reporting is enabled in php

#### Questions


JakubOnderka:
Merged, thanks for contribution.

PR #7091, opened 2021-02-25
fix: [email_otp] skip OTP for disabled users

lfortemps:
#### What does it do?

With the implementation of the preauh_actions, a disabled user would still go to the OTP step after entering username/password correctly.
This small fix prevents this behavior, so that the user directly receives the feedback.

#### Questions


adulau:
Thank you!

PR #7092, opened 2021-02-25
fix: [email_otp] Trim value for increased UX

lfortemps:
#### What does it do?

When double-clicking on the email containing the OTP, some spaces might be copied at the same time. This small fix will discard the spaces automatically in the OTP field to improve user experience.

#### Questions


iglocska:
Thanks a lot!

PR #7100, opened 2021-02-26
Prevent pulling local tags

### LINKED ISSUE #6836: Local tags synced with non-internal instance?
### Work environment

| MISP version / git hash   | 2.4.136
| Browser                   | If applicable

### Expected behavior

When I attach local tags (exportable) to events in instance A, my assumption would be that they stay in instance A and optionally propagate to connected internal sync servers. Those tags usually are for internal use such as workflow etc. they do not need to be propagated outside.

### Actual behavior

From instance B, I trigger a pull from instance A and the local tags are also collected and they become "global tags" in instance B. This seems like a bug.

lfortemps:
#### What does it do?

Fix #6836 in my tests. It is not a security control because implemented on the client side. It is just to prevent "pollution" by importing local tags from a remote instance during a pull.

Let me know if you prefer having this logic in another function.

#### Questions

### DIFF
--- a/app/Model/Server.php
+++ b/app/Model/Server.php
@@ -286,6 +286,14 @@ class Server extends AppModel
                     $event['Event']['distribution'] = '1';
                     break;
             }
+            // We remove local tags obtained via pull
+            if (isset($event['Event']['Tag'])) {
+                foreach ($event['Event']['Tag'] as $key => $a) {
+                    if ($a['local']) {
+                        unset($event['Event']['Tag'][$key]);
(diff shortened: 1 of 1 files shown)


iglocska:
This has been fixed recently on the producing side (all synchronisations pass a flag along to exclude local tags to the event view - though this flag was broken).

With that said, let's include your fix for the time being - until everyone updates to 2.4.140 we'll definitely be seeing this issue otherwise.

Thanks for the fix!

PR #7101, opened 2021-02-26
fix: [sync] SG stay inactive after pull

### LINKED ISSUE #6503: Sharing groups become active automagically
I have noticed some sharing groups became active while they were previously put as passive. This is coming from synchronisation of events using those SGs. I will try to find out why but if someone has some ideas feel free to shout.

*If you would like to report a bug, please fill the template bellow*

### Work environment

| OS version (server)       | RHEL
| OS version (client)       | Win10
| MISP version / git hash   | 2.4.132

### Expected behavior

Once sharing groups are put as passive by an administrator, their definition should remain as such until changed again manually.

### Actual behavior

Sharing groups get the active flag automatically without any confirmation.

1. Have 2 MISP instance with a pull relationship
2. Define a sharing group on one instance, create an event, pull from the other instance
3. After the pull, the SG definition is accessible in the second MISP, mark it as passive
4. Create a new event on the first MISP with the same SG, pull again from the second MISP
5. The SG is now active again on the second MISP

lfortemps:
#### What does it do?

Fix #6503 for existing SG. If existing SG are marked as inactive, they will no longer become active after a pull containing events distributed to that SG.

#### Questions

### DIFF
--- a/app/Model/SharingGroup.php
+++ b/app/Model/SharingGroup.php
@@ -701,7 +701,7 @@ class SharingGroup extends AppModel
                 if ($force) {
                     $sgids = $existingSG['SharingGroup']['id'];
                     $editedSG = $existingSG['SharingGroup'];
-                    $attributes = ['name', 'releasability', 'description', 'created', 'modified', 'active', 'roaming'];
+                    $attributes = ['name', 'releasability', 'description', 'created', 'modified', 'roaming'];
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks for the fix, I've added it manually in the end since the SharingGroup model has diverged. Nice find!

PR #7107, opened 2021-02-27
chg: [UI] correct edit org blocklist entries view

Wachizungu:
#### What does it do?

Corrects copypasta error ('Edit Event Blocklist Entries' legend for edit org blocklist entry view) and removes paragraph that doesn't really add value.

#### Questions


JakubOnderka:
Thanks!

PR #7114, opened 2021-03-01
chg: [UI] Add small description of what event block rules do

Wachizungu:
#### What does it do?

Adds small description of what event block rules do to the set event block rules page.

#### Questions


iglocska:
Thanks a lot!

PR #7126, opened 2021-03-02
fix: [test] Sync user should be able to edit just non local groups

JakubOnderka:
#### What does it do?

The condition in old code was different: 

https://github.com/MISP/MISP/blob/19ef63f521e64d33b666c9e38ef853726445b0cc/app/Model/SharingGroup.php#L696

#### Questions

### DIFF
--- a/app/Model/SharingGroup.php
+++ b/app/Model/SharingGroup.php
@@ -641,7 +641,7 @@ class SharingGroup extends AppModel
      * @param array $existingSG
      * @param array $sg
      * @param boolean syncLocal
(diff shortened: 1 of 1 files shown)


iglocska:
Crap. Just saw this and ended up implementing something similar after investigating the issue. Sorry, should have checked the PRs first!

JakubOnderka:
> Crap. Just saw this and ended up implementing something similar after investigating the issue. Sorry, should have checked the PRs first!

OK, closing then.

Issue #7128, opened 2021-03-02
API calls not working correctly when called from web UI if using custom authentication


patriziotufarolo:
*If you would like to report a bug, please fill the template bellow*

### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | N/A
| OS version (client)       | N/A
| PHP version               | 7.4
| MISP version / git hash   | 2.4.139, 19ef63f521e64d33b666c9e38ef853726445b0cc
| Browser                   | N/A

With custom authentication with header forward is enabled (e.g. LDAP authentication through HTTP Basic with header forward, Kerberos authentication) API calls made through the WEB UI (i.e. without API key but using the current session)  fail for missing API key.

### Expected behavior
API calls can use already authenticated session if called without API key

### Actual behavior
API calls fail

### Steps to reproduce the behavior
1) Install  latest MISP
2) Configure LDAP or Kerberos authentication with Header forward
3) Login into the Web UI and try to download the CSV file for an event

### Reason and proposed solution
Bug introduced by e5e855b3c290af23eeba12bf7fb3b5cd57b1183c 
Line https://github.com/MISP/MISP/blob/19ef63f521e64d33b666c9e38ef853726445b0cc/app/Controller/AppController.php#L231 should be changed
to
```
if ($this->__loginByAuthKey() === false && $this->Auth->user() === null) {
```

JakubOnderka:
Hello, what do you mean by *API calls made through the WEB UI*? What URL do you try to access?

patriziotufarolo:
Hi,
restSearch is enough to trigger the bug, even if called by the Web UI (e.g. exporting a CSV).

If you configure custom authentication (ldap through http basic, kerberos and so on), and try to export a CSV, MISP gives error.
Relying on classic authentication (DB + Misp Login) solves the issue.
This is due to restSearch expects for an API key to be present in the Authorization header, while with a custom authentication schema the authorization header is sent with different content. If my patch is not good, then the issue might exist n the __loginByAuthKey function

patriziotufarolo:
Hi,
I looked a bit deeper into the code, and indeed my patch is wrong.
This is due to a mixture of different approaches when handling multiple authentication schemas.
[...]

PR #7129, opened 2021-03-02
fix: security: authentication logic via API key or current session

patriziotufarolo:
Logic was wrong in this point. With my change:
1) check first for API key
2) if not api key checks then for current session

solves #7128

#### What does it do?

#### Questions


JakubOnderka:
Hello, thanks for contribution, but your fix looks incorrect, because test doesn't pass.

PR #7130, opened 2021-03-02
chg: [UI] fix keyboard shortcut manager popup triangle

Wachizungu:
#### What does it do?

fixes / brings back the shortcuts popup triangle above the footer (bottom right)
See https://github.com/MISP/misp-book/tree/main/shortcuts for the related documentation

Looks like https://github.com/MISP/MISP/commit/2c20b305330b3c85386ae5852428f0c59ec6e548 might have broken it / made it invisible. Even after fixing that the menu was still not popping out, hence the change to app/webroot/js/keyboard-shortcuts.js as well.

#### Questions


iglocska:
Thanks a lot!

PR #7168, opened 2021-03-07
chg: [attributes] fix copypasta error leading to internal server erro…

Wachizungu:
…r on addtag with tag name

#### What does it do?

Fixes following issue:
Adding tag to attribute using /attributes/addTag gives internal server when trying to add with tag name.
Looks like line was not correctly modified after copying logic from /events/addTag to add tag collection

#### Questions


JakubOnderka:
Thanks!

PR #7171, opened 2021-03-07
chg: [attributes] fix attribute addtag by name conditions for find no…

Wachizungu:
#### What does it do?

Fixes issue with attribute addtag find conditions not being set for when adding a tag by name, leading to just adding the first tag found

#### Questions


Wachizungu:
Please double check it works for you as well before merging here. This should go into 2.4 later as well as fixing the 500 caused by another bug now leaves the door open for this issue to occur.

My bad for not double checking the correct tag was being added.

PR #7172, opened 2021-03-07
chg: [restClient:querybuilder] add events and attributes addTag and r…

Wachizungu:
…emoveTag actions.

#### What does it do?

Adds the addTag and removeTag actions of events and attributes endpoint to the restclient query builder.

Also changes relevant help text values.

#### Questions


adulau:
Thank you!

PR #7188, opened 2021-03-10
added newest DataPlane.org feeds

jtkristoff:
I added the remaining DataPlane.org feeds that were not listed.  I used the same default as the other DataPlane.org feeds.  All the feeds are essentially of the same format except proto41 has two time stamp fields, a firstseen and lastseen.

John


adulau:
Thanks John, it’s great to see new feeds from dataplane.org. I’ll do a tweet about it too!

PR #7205, opened 2021-03-15
fix: [UI] Fixes bug that stops country flag being displayed alongside country

tomking2:
Fixed a minor bug introduced when the galaxy cluster element view was changed to a generic factory and the path passed to the country flag was not updated. Caused no country flag to be shown and error to be thrown

#### Questions


JakubOnderka:
Thanks!

PR #7206, opened 2021-03-15
fix: [api] Fixes crash when a new indicator in existing event has a sighting

tomking2:
#### What does it do?

When an existing event has a new indicator added, it goes to the 'editAttribute' function, which before expected an ID when creating a sighting. As a result, new indicators where there was no ID, this was not available until post-save. Moved the code further down following the save. Same code, but works better now :)

#### Questions


iglocska:
Nice catch! Thanks!

PR #7217, opened 2021-03-17
Allow saving of OIDC plugin settings

jozuatec:

#### What does it do?

It allows for saving settings for the OIDCAuth plugin (https://github.com/MISP/MISP/tree/2.4/app/Plugin/OidcAuth) back into config.php so those settings do not get wiped on a restart.

#### Questions

No
Yes
No


jozuatec:
In light of pull request #7213, this one can be closed.

PR #7230, opened 2021-03-22
Update OidcAuthenticate.php

jozuatec:

#### What does it do?

With our IDP the user roles do not get delivered through claims. With this edit (get roles through "requestUserInfo" when claims fails to do so), our IDP can deliver the roles through an "Extra Attributes" field.
I am already using this code in our production, it works fine for us.

#### Questions

- [ ] Does it require a DB change? no
- [ ] Are you using it in production? yes
- [ ] Does it require a change in the API (PyMISP for example)? no


JakubOnderka:
Thanks!

PR #7231, opened 2021-03-22
chg: [ShibbAuth] Add login entry on logging in for audit

Wachizungu:
#### What does it do?

Adds a login entry that is visible in the audit logs when someone logs in with ShibbAuth.

Would still like to test this in production before it is merged but feedback is appreciated if this is not the best way to log login.

#### Questions


adulau:
Thank you.

Issue #7208, opened 2021-03-15
Changing a setting in Server Settings & Maintenance removed openid connect configuration block in config.php


garrit-schroeder:
# This template is meant for bug reports, if you have a feature request user the other template.

*If you would like to report a bug, please fill the template bellow*

### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | Debian
| OS version (client)       | 
| PHP version               | 7.3
| MISP version / git hash   | 2.4.140 2c20da190f71d0d2e99ed7dac475ccf050e678c9
| Browser                   | If applicable

### Expected behavior

Changing a setting in Server Settings & Maintenance will not remove the configuration block for the open id connect configuration in config.php.

### Actual behavior

After changing for the setting misp.live or hosting organisation for example the configurations regarding open id connect is removed entirely from config.php file.
```
'Security' => 
  array (
    ...
    array (
      0 => 'OidcAuth.Oidc',
    ),
  ),
// below this gets removed
  'OidcAuth' => [
        'provider_url' => 'https://KEYCLOAK/auth/realms/REALM/.well-known/openid-configuration',
        'client_id' => 'mymisp',
        'client_secret' => 'XXX',
        'role_mapper' => [ // if user has multiple roles, first role that match will be assigned to user
            'user' => 3, // User
            'admin' => 1, // Admin
        ],
        'default_org' => 'default',
    ],

garrit-schroeder:
It is fixed now. Thanks

ArnoldSpyros:
In version 2.4.145 the configurations regarding open id connect are still removed.

PR #7233, opened 2021-03-23
fix: [internal] Keep AadAuth setting in config.php when modify setting value from UI 

EvaYiYang:
#### What does it do?

Fixes a similar issue like #7208 when using AadAuth
Similar fix to #7213 

We run a docker image with a config file that enables `AadAuth` plugin. Every time the container runs, the config file re-generates and flushes all of our `AadAuth`settings. 
By adding `AadAuth` to `settingsToSave`, it will save `AadAuth` settings in config file by default and when `AadAuth` is not enabled, it will shown as below, just like other plugins `ApacheSecureAuth `and `ApacheShibbAuth`
```
'AadAuth' => NULL
```

#### Questions


iglocska:
Thanks a lot, nice catch!

PR #7234, opened 2021-03-23
fix: [Event:set_filter_value] Allows searching for composite attributes

mokaddem:
Allows searching for composite attributes.
Searching for `color|blue` will result in the following condition
```php
'OR' => array(
    'Attribute.value1' => array(
        (int) 0 => 'color|blue',
        (int) 1 => 'color'
    ),
    'Attribute.value2' => array(
        (int) 0 => 'color|blue',
        (int) 1 => 'blue'
    )
)
```

Let me know if you think something may go wrong


marjatech:
Hey guys,
Thank you @mokaddem for the fix!  
I'm really awaiting this to be in a release for some time... Doesn't seem like I could help on this right? 
Just waiting for a code-review?

PR #7235, opened 2021-03-23
chg: [tag] Use detailed message in tag return

imidoriya:
#### What does it do?
Gives a more detailed message response when 1 tag is added (similar to when a tag is deleted).

A detailed message is created on lines 870 (local) and 877 (global) when a tag is attached, however, these messages are never used in the response as it is overwritten by the generic message on line 888. The generic message, that includes count, would be better served only when more than one tag is attached.

#### Questions

- [x] Does it require a DB change? **No**
- [x] Are you using it in production? **Yes**
- [x] Does it require a change in the API (PyMISP for example)? **No**


iglocska:
That makes sense, thanks a lot!

PR #7236, opened 2021-03-23
chg: [statistics] fix typo in statistics_data view - monthly attribut…

Wachizungu:
…es styling check

#### What does it do?

Fixes a copy paste error / typo. The check should be for the correct parameter ('attribute_count_month').

#### Questions


adulau:
Good catch! Thank you.

PR #7238, opened 2021-03-24
fix: [doc] rhel 7 install doc initial fixes

SteveClement:
Still broken, but better.


garanews:
Can you check also
Downloading packages:

https://cdn.redhat.com/content/dist/rhel/server/7/7Server/x86_64/os/Packages/p/python-rpm-macros-3-34.el7.noarch.rpm: [Errno 14] curl#35 - "SSL peer was not expecting a handshake message it received."
Trying other mirror.
Error downloading packages:
  python-rpm-macros-3-34.el7.noarch: [Errno 256] No more mirrors to try.

PR #7255, opened 2021-03-25
chg: [UI] fix debugon for debug = 1. fix #7131

### LINKED ISSUE #7131: logic to set debug styling of the GUI is incorrect
### Work environment

| OS version (server)       | Ubuntu
| OS version (client)       | N/A
| MISP version / git hash   | 2.4.139, 7431beefa12bff1807dee5351dae9e6586e2c5a9

### Expected behavior
setting debug to 'Debug on' leads to setting debugOn class for certain GUI elements.

### Actual behavior
setting debug to 'Debug on' leads to setting debugOff class for certain GUI elements.

Set setting debug to Debug on. Inspect element on footer, check the classes assigned to the footer div. It will be "footer debugOff".

### Notes

I don't really understand why this difference in classes is there so didn't want to push a potential fix myself without asking input. Many different view pages have conditional formatting based on $debugMode (footer is one of them).

Three options for debug setting are:
Debug off -> 0
Debug on -> 1
Debug + SQL dump -> 2
image

Culprit line / function seems to be:
https://github.com/MISP/MISP/blob/2.4/app/Controller/AppController.php#L789
Should be `(Configure::read('debug') >= 1)` instead of `(Configure::read('debug') > 1)` I think.

If for some reason the conditional formatting should only be done for Debug + SQL dump then this issue can just be closed and I'll document it somewhere.

Wachizungu:
#### What does it do?

fixes #7131

#### Questions

### DIFF
--- a/app/Controller/AppController.php
+++ b/app/Controller/AppController.php
@@ -808,7 +808,7 @@ class AppController extends Controller
      * Configure the debugMode view parameter
      */
     protected function _setupDebugMode() {
-        $this->set('debugMode', (Configure::read('debug') > 1) ? 'debugOn' : 'debugOff');
+        $this->set('debugMode', (Configure::read('debug') >= 1) ? 'debugOn' : 'debugOff');
     }
 
     /*


adulau:
Thank you!

PR #7291, opened 2021-04-02
Added Threatfox to default feeds

stevengoossensB:
#### What does it do?

It adds Threatfox (Abuse.ch) to the list of default feeds so it's easier for new users to onboard it in MISP

#### Questions


adulau:
Thank you!

PR #7334, opened 2021-04-13
chg: [routes] fix allowedlists routes. Renamed from whitelists.

Wachizungu:
#### What does it do?

fixes allowedlists routes. Renamed from whitelists.

#### Questions


iglocska:
Thanks a lot!

PR #6661, opened 2020-11-27
Azure Active Directory Authentication

cudeso:
#### What does it do?

Fixes #6641 
Adds Azure Active Directory authentication (first 'rough', but working version)

#### Questions

- [ ] Does it require a DB change? No
- [ ] Are you using it in production? Yes
- [ ] Does it require a change in the API (PyMISP for example)? No

PR #7336, opened 2021-04-14
Change config.default.php to have everything needed for Azure AD auth

stevengoossensB:
#### What does it do?

Small documentation/comments update related to https://github.com/MISP/MISP/pull/6661 . The config.default.php now contains the necessary information by default, which causes less confusion when trying to implement AAD authentication.

#### Questions


adulau:
Thank you.

PR #7343, opened 2021-04-15
chg: [MispObject] fix copy paste error in editObject

Wachizungu:
#### What does it do?

Fixes some likely copy paste errors in editObject.  Considering the surrounding code and fact that IDE was complaining about $attribute not existing I assume this should be the correct version.

#### Questions


iglocska:
The copy pasta monster strikes again, thanks!

PR #7345, opened 2021-04-16
new: [Dashboard] Add user count evolution widget

Wachizungu:
#### What does it do?

Adds a graph widget showing the historic total user count up to now.
Allows setting a number of days, weeks or months for the graph to plot. Interval is 1 day, 1 week or 1 month respectively. The smallest option is chosen if multiple options were provided (e.g. if someone sets days to 10 and months to 20 the graph will plot 10 days).

Original idea was to have an overview of current status at the end with enabled users, users with pgp keys etc for the current date but MultiLineChart seems to expect you to provide data points for all labels for all dates so it would have led to a confusing graph.

Any feedback is greatly appreciated. If there are modifications to do I'll do my best.

image

#### Questions

### DIFF
--- a/app/Lib/Dashboard/UsersEvolutionWidget.php
+++ b/app/Lib/Dashboard/UsersEvolutionWidget.php
@@ -1,0 +1,82 @@
+<?php
+
+
+class UsersEvolutionWidget
+{
+    public $title = 'Evolution of user count';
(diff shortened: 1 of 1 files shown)


iglocska:
Lovely stuff, thanks!

PR #7349, opened 2021-04-17
chg: [Dashboard:MultiLineChart] make enabling 'total' line on initial…

Wachizungu:
… render configurable

#### What does it do?

Makes it possible to enable the 'total' line by default for widgets using MultiLineChart.
Passing something like below as config:

{
    "widget_config": {
        "enable_total": "1"
    }
}

Working on a widget that uses this:

image

#### Questions


adulau:
Great idea! Thanks a lot.

PR #7350, opened 2021-04-17
new: [Dashboard] Add org count evolution widget

Wachizungu:
#### What does it do?
Similar to https://github.com/MISP/MISP/pull/7345

Adds a graph widget showing the historic organisation (local org, remote org) count up to now.
Allows setting a number of days, weeks or months for the graph to plot. Interval is 1 day, 1 week or 1 month respectively. The smallest option is chosen if multiple options were provided (e.g. if someone sets days to 10 and months to 20 the graph will plot 10 days).

With following config you also get the total displayed by default (this relies on another commit that was recently merged into develop).
{
    "widget_config": {
        "enable_total": "1"
    }
}

Any feedback is greatly appreciated. If there are modifications to do I'll do my best.

Example below:
image

#### Questions

### DIFF
--- a/app/Lib/Dashboard/OrgsEvolutionWidget.php
+++ b/app/Lib/Dashboard/OrgsEvolutionWidget.php
@@ -1,0 +1,96 @@
+<?php
+
+class OrgsEvolutionWidget
+{
+    public $title = 'Evolution of orgs count';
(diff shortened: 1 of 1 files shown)


iglocska:
That's really cool! Thanks a lot!

PR #7354, opened 2021-04-20
new: [servers:edit] Fetches available orgs and tags from remote server

mokaddem:
- Componentized views and made them responsive
- Usage of picker for orgs and tags
- For server pull rules, fetches available choices from the remote server

image


app/View/Elements/serverRuleElements/pull.ctp
```diff
@@ -1,80 +1,186 @@
...
+            )
+        } else {
+            $('div.notice-pull-rule-fetched.alert-warning').show().find('.reason').text('<?= __('The server must first be saved in order to fetch remote synchronisation rules.') ?>')
+            $pickerTags.parent().remove()
+            $pickerOrgs.parent().remove()
+            $('div.notice-pull-rule-fetched.alert-primary').hide()
+        }
+    }
+
+    function getPullFilteringRules(cb, fcb, acb) {
```
iglocska, line 139: For maintainability, probably better to give the anonymous functions names that are a bit more clear as to what their function is.

app/View/Elements/serverRuleElements/pull.ctp
```diff
@@ -1,80 +1,186 @@
...
+        $.getJSON('/servers/queryAvailableSyncFilteringRules/' + serverID, function(availableRules) {
+            cb(availableRules)
+        })
+        .fail(function(jqxhr, textStatus, error) {
```
iglocska, line 143: This will always be the case for older instances.

app/Model/Server.php
```diff
@@ -4450,6 +4450,44 @@ public function removeOrphanedCorrelations()
         return $success;
     }
 
+    public function queryAvailableSyncFilteringRules($server)
+    {
+        $HttpSocket = $this->setupHttpSocket($server, null);
+        $uri = $server['Server']['url'] . '/servers/getAvailableSyncFilteringRules';
```
iglocska, line 4456: We have some options to mitigate the lack of this function on the remote side (if we want that, the alternative is to just not care and display that this feature only works for MISP > v2.4.142):

Using /organisations/index in junction with /tags/index. Though both the query and the response will be heavier this way.

PR #7374, opened 2021-04-27
fix: [UI] Correctly display last login time

lfortemps:
#### What does it do?

Until now, we were showing the "one before last" login time, this fixes the issue by reading the last_login after it has been updated.

#### Questions


adulau:
Thank you!

PR #7384, opened 2021-04-29
fix: [db_schema] Update to version 68

JakubOnderka:
#### Questions


iglocska:
Thanks a lot!

PR #7392, opened 2021-05-04
chg: [statistics:UsageData] fix active proposal count, exclude deleted entries

Wachizungu:
#### What does it do?

Realized statistics usage data counts all proposals, even deleted ones. This shouldn't be the case I think as it's clearly marked as 'active proposals'. Since I copied this code for the related dashboard widget it needs to be fixed in two places.

#### Questions

### DIFF
--- a/app/Controller/UsersController.php
+++ b/app/Controller/UsersController.php
@@ -1848,7 +1848,7 @@ class UsersController extends AppController
         $stats['correlation_count'] = $this->Correlation->find('count', array('recursive' => -1));
         $stats['correlation_count'] = $stats['correlation_count'] / 2;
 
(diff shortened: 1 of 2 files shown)


iglocska:
Indeed, that's a tricky one. In reality counting both is best for statistics, as it shows the interactions between organisations and data they do not own - when a proposal is accepted, what happens technically is that an attribute gets created/modified and the proposal is deleted. So if you are diligent with accepting proposals it will look as if there was no proposals being made for your shared data, which isn't the case.

With that said, the name of the statistic is indeed active proposals so let's merge this now and then consider adding a new field in the future for total proposals made.

Wachizungu:
Hmm, indeed. Thinking about that it might also be good to be able to tell whether a proposal was accepted or rejected, I don't see a way of doing that except by looking at logs though.

PR #7456, opened 2021-05-31
Fix mysql error index sightings by event

righel:

#### What does it do?
Fixes mysql error when calling /sightings/index/[event_id]. Error:
`17:38:36	SELECT `Sighting`.`id`, `Sighting`.`event_id` FROM `misp`.`sightings` AS `Sighting` WHERE `Sighting`.`event_id` = 1  GROUP BY `Sighting`.`event_id` LIMIT 0, 1000	Error Code: 1055. Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'misp.Sighting.id' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by	0,0012 sec`

#### Questions

### DIFF
--- a/app/Controller/SightingsController.php
+++ b/app/Controller/SightingsController.php
@@ -15,9 +15,9 @@ class SightingsController extends AppController
     }
 
     public $paginate = array(
-            'limit' => 60,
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks!

PR #7459, opened 2021-06-01
fix: [UI] Restore notice list warnings when adding or editing attribute

Kagee:
Restore the notice_message div that vanished in commit 0d4df7c98b0fc67618b1c3c298e64efb668fc4fe.

#### Questions


iglocska:
Thanks a lot!

Issue #7210, opened 2021-03-15
Custom warninglists


iglocska:
The idea is that admins should be able to create and modify warninglists directly through the interface. This would allow us to transition the old allowlist system into the warninglist system and get rid of the legacy baggage, whilst also making warninglists more usable.

Requirements:

[] Warninglists loaded from JSON files are protected and immutable by this system
[] The old Allowlist, if not empty, is transitioned to a new warninglist automatically on upgrade, called "Legacy allowlist"
[] Warninglists are exportable/dumpable to make the sharing with the OSS community easier
[] Bonus: enforceWarninglist, in addition to accepting true/1 to enforce all enabled list can take a list of warninglist IDs/names as a parameter to only enforce certain lists on demand

chrisinmtown:
👍

JakubOnderka:
Idea: make possible to add addition information to warninglist value, like comment, that will be also visible for user when attribute value match warninglist entry.

PR #7464, opened 2021-06-02
Custom warninglists

### LINKED ISSUE #7434: Feature request: Warninglist categories
Warninglist currently mix different categories of list. 
* First category is 100 % sure invalid IoC, like internal IP range or hash of empty file
* Second category is most probably false positive, like communication to public DNS server or Windows Update IP.
* Third category is know indicator, like IP address that belongs to public cloud.

Every warninglist will belongs to one of category. When viewing event, *Warning: Potential false positives* box will change color and description according to highest warninglist category. Also this information will be available thought API.

JakubOnderka:
#### What does it do?

Fixes #7434 and part of #7210.

#### Questions


iglocska:
Just played with it a bit, looks good, but there are a few things that we'd need to change before merging: 

1. The DB version needs to be bumped before we merge
2. The DB update script needs to kick the user out (=> true) otherwise we get notice errors until logout
3. The categories are confusing, I myself have no idea what the other two categories are meant to do. We've had some discussions on the proposed categories in a prior discussion (block, warn and something else that I can't recall) where the idea would be that we separate out warninglists that are meant to be used in the automation and ones that are purely meant to warn the users. I'll try to find the discussion back.
4. No edit button on the index

JakubOnderka:
> The DB version needs to be bumped before we merge

Fixed.

> The DB update script needs to kick the user out (=> true) otherwise we get notice errors until logout

Fixed.

> No edit button on the index

There should be edit icon for custom warninglist.

> The categories are confusing

Yes, that possible. My idea was:
* Invalid identifier - IoC that is 100 % sure that is invalid, like private IP address, hash of empty or EICAR file
* False positive - known good, like known DNS servers, Windows Update or Wikipedia IP addresses or domains
* Known identifier - something that is just on the list, but it doesn't mean that it is false positive, like public cloud providers or Top 1000 domains etc.

I will be happy if you propose better names or maybe even new category, or we can merge first two categories into *False positive*.

iglocska:
Ah I see, so you would differentiate between the certainty of those two. The problem is that invalid identifier is in the eye of the beholder - on a shared community my own infrastructure is ground truth to me in being 100% not something I want to detect / block on, but if I have a compromise others in the community might disagree. So what I'd do: Merge those two simply have a false positive + an informational/purely warning list (false positive and known identifier are fine for the naming).

Later on we should change the behaviour of enforceWarninglist to accept, in addition to the boolean values, a list of warninglist IDs, that way we can selectively enforce lists that are of our own infrastructure if we wish not to use the false positive lists.

JakubOnderka:
Changed, not there are just two categories:
* False positive
* Known identifier

PR #7468, opened 2021-06-03
Add openapi spec

righel:

#### What does it do?
Add initial OpenAPI spec for MISP Automation API, add ReDoc js files to render the spec in the UI. 

#### Questions


iglocska (commented):
Could you remove the Rest client from event actions? I dig the change of a separate API root, but now we have it double :) 

(Sami)ACL: should be open to all, not just site admins. `array()` means site admin only, should be `array('*')`

app/Controller/ServersController.php
```diff
@@ -2527,4 +2527,7 @@ public function getAvailableSyncFilteringRules()
         $syncFilteringRules = $this->Server->getAvailableSyncFilteringRules($this->Auth->user());
         return $this->RestResponse->viewData($syncFilteringRules);
     }
+
+    public function openapi(): void {
```
iglocska: Perhaps remove the return promise for now - we'll definitely make use of those once we move to cake 4 but until then we still know of some happy PHP 5.6 users (cough) ;)

iglocska:
Thanks a lot, killer feature

PR #7469, opened 2021-06-04
Fix SG Sync issues

mokaddem:
- Make sure to re-use the correct sharing group ID when capture ORGs and Servers
- Allow pushing SG even if the remote **internal** server is not in the list of of the SG servers


mokaddem:
Merged in develop

PR #7470, opened 2021-06-04
Improvements for cycat integration

mokaddem:
- Fixes some bugs of the original implementation
- Moved CyCat data from the meta-table to its own child component, this using ajax to fetch it
image


iglocska:
Thanks a lot!

PR #7476, opened 2021-06-07
fix: [db] rename org_blacklists to org_blocklists everywhere

RichieB2B:
The old name `org_blacklists` was still present in several install files.

### DIFF
--- a/INSTALL/MYSQL.sql
+++ b/INSTALL/MYSQL.sql
@@ -637,21 +637,6 @@ KEY `type` (`type`)
 -- -------------------------------------------------------
 
 --
--- Table structure for `org_blacklists`
---
-
-CREATE TABLE IF NOT EXISTS `org_blocklists` (
-  `id` int(11) NOT NULL AUTO_INCREMENT,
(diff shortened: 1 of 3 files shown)


adulau:
Thank you very much!

adulau:
Sorry I revert it. Because the SQL file is used an initial state for applying the updates via the updater. So it should be fine at the first run of the updater.

RichieB2B:
Not according to the db_version at https://github.com/MISP/MISP/blob/2.4/INSTALL/MYSQL.sql#L1481
MYSQL.sql is at db_version 61 but the rename was done in db_version 56. So there is no need for org_blacklists to exist in MYSQL.sql anymore. The current MYSQL.sql  creates both org_blacklists and org_blocklists which never makes sense.

See https://github.com/MISP/MISP/blob/2.4/app/Model/AppModel.php#L1415

mokaddem:
Hey @RichieB2B you were absolutely right!
That `org_blacklists` modification was overlooked when updating the sql file.
However for the Postgresql one, the `db_version` is way older than 56 so these changes should not be done there.
Thanks for your feedback!

PR #7479, opened 2021-06-07
chg: log remote IP for authkey use attempt if remote IP not allowed b…

Wachizungu:
…y key

#### What does it do?

Fixes below:
When a user tries to use Authkey from a remote IP that is not allowed by the key, the used IP is not logged.

-> change just adds the IP to the log entry

See gitter: https://gitter.im/MISP/MISP?at=60b622acfec22b4786f21a38

#### Questions


iglocska:
Looks good, but didn't the IP logging capture it before? With that said, no harm in merging this either way, thanks!

Wachizungu:
To be honest I also expected I'd see something already.
I have these settings and we had nothing in the GUI logs (/admin/logs/index):

image


This was the only type of entry we got (didn't have the IP before the change):
image

This particular instance is not hosted by us so for debugging we can't easily check the db or whatever.

If I missed something please let me know.

PR #7483, opened 2021-06-08
fix: typo

bartblaze:
😅 

#### What does it do?

Fixes a typo.

#### Questions


adulau:
Nice catch! Thank you.

Issue #7481, opened 2021-06-08
Propagate attribute tag deletion across syncs


3c7:
**Is your feature request related to a problem? Please describe.**
If I understood it correctly, the current MISP behaviour is to add new tags to an updated (synced) attribute. If attribute tags are used for information which likely changes, e.g., priority or confidence, it can be misleading to have multiple tags describing different levels of priority or confidence.

**Describe the solution you'd like**
It would be great, if events and attributes sync tag deletions, so remote instances have exactly the events and attributes that the publisher intended.

**Describe alternatives you've considered**
Maybe there are use cases, where this behaviour does not fit. In that case it might be a good idea to make it configurable per server in the server list.

PR #7493, opened 2021-06-11
new: [sync] Allow option to delete tags on event sync

tomking2:
#### What does it do?

Since the commit 96c1f41ffd3020fb9bdb98e35f2be7434fe8f2b4 that was made back in Jan 2020, any deletions of tags on an event are not propagated to another instance when pulling. As per #7481 raised by @3c7, some MISP instances use frequently changing tags such as confidence and priority. As a result events can end up with multiple tags applied when only one should be present at any one time. This means MISP events no longer match to what is specified by the original source and causes inconsistencies.

This PR re-enables tag deletion through a sync setting, and suggest this is an option until soft-delete tags are in place. Local tags are not touched, only global tags are amended.

#### Questions


tomking2:
Bit of a conflict with upgrade 71 for the DB, will need to be merged together.

tomking2:
Merge conflicts should now be resolved :)

app/View/Servers/edit.ctp
```diff
@@ -236,7 +240,8 @@ var formInfoValues = {
         'ServerPublishWithoutEmail' : '<?php echo __('Publish new event without email (working with Pull event).');?>',
         'ServerSubmittedCert' : "<?php echo __('You can also upload a certificate file if the instance you are trying to connect to has its own signing authority.');?>",
         'ServerSubmittedClientCert' : "<?php echo __('You can also upload a client certificate file if the instance you are trying to connect requires this.');?>",
-        'ServerSelfSigned' : "<?php echo __('Click this, if you would like to allow a connection despite the other instance using a self-signed certificate (not recommended).');?>"
+        'ServerSelfSigned' : "<?php echo __('Click this, if you would like to allow a connection despite the other instance using a self-signed certificate (not recommended).');?>",
+        'ServerRemoveMissingTags': "<?php echo __('Remove any global tags from events on local instance that are not present on an updated event being received from the server. This will remove any custom global tags you have set (working with Pull event).');?>"
```
mokaddem: The description is great but it would be nice if it includes the fact that local tags are unaffected and that it only work for pulling and that pushing is not affected. (This is nitpicking but for a feature that only works in one way, I would prefer to be as transparent as possible)

mokaddem:
@tomking2 thanks a lot for the PR. Please find some comments that need to be addressed before merging:
1. Could you please align it with the current state of `develop`?
2. The installation script has been modified
    - These changed should be moved in another PR
    - Also, these files should not be modified only the installation files in `docs` should be (https://github.com/MISP/MISP/tree/2.4/INSTALL#installsh-hacking)
3. Is it intentional that only attributes are affected? What about events and galaxyClusterRelations?

Thanks again!

tomking2:
Hey @mokaddem,

1. Should now be aligned with develop.
2. Not a clue why the installation script was modified, must be some weird thing as I was basing off of 2.4. Merged with develop and those changes are gone. This PR is only focused on tag sync.
3. It is as this is what we have found changes frequently in our use case. Event tags have stayed consistent, the same with galaxyClusterRelations. But happy to follow up with making these sync if you think it's the right move.

tomking2:
Hey @mokaddem, just updated to ensure it's not conflicting with the `develop` branch. Are you able to review my comments and determine if it's OK to now be pushed out?

syloktools:
Any updates on this one?

tomking2:
Hi @mokaddem, @iglocska ,

I've just updated to align with the latest develop branch, are we able to see if we can progress this before I have to update the DB schema versions again!

Thanks

tomking2:
Updated again to fix conflicts - OK to review @mokaddem, @iglocska, @righel?

mokaddem:
@tomking2 your contribution has been merged! Please come to [CTIS-2022](https://cti-summit.org/) to collect your reward! ;)
The CFP is not totally closed.

Without kidding, this was long overdue.
Many apologies for the waiting time. And many, many thanks for your patience and your good work.
Cheers!

PR #7500, opened 2021-06-15
Stix to misp types path

JakubOnderka:
#### What does it do?

If PyMISP is not installed from Git modules, but from Python package, file `describeTypes.json` is not in expected location. This fix that and loads `describeTypes.json` directly from package.

#### Questions


adulau:
@chrisr3d could you have a look and see if it's also the issue is applicable for the recent repo too?

PR #7511, opened 2021-06-22
new: update to handle network connection objects

emmekappa86:
#### What does it do?

This update concerns the possibility of creating snort/suricata rules from network-connection objects using multiple attributes (instead of splitting those attributes into a series of unrelated rules).
This change was mentioned during the CONCORDIA MISP Training (May 27th 2021) and authored by (me) Marco Caselli from Siemens AG.

#### Questions


app/Lib/Export/NidsExport.php
```diff
@@ -16,11 +16,14 @@ class NidsExport
 				'fields' => array('threat_level_id')
 			)
 		),
-		'flatten' => 1
+		#'flatten' => 1
```
mokaddem: Any debugging / test commented lines should be removed

app/Lib/Export/NidsExport.php
```diff
@@ -16,11 +16,14 @@ class NidsExport
...
 	);
 
 	public function handler($data, $options = array())
 	{
+	
+		//NOTES: Here the scope "Object" should be probably checked
```
mokaddem: Not sure I understand the meaning of this comment. However, it should be noted that the `network-connection` object is only supported when the scope `event` is used. Would it be worth investing more time trying to have the same feature while using the `attribute` scope?

app/Lib/Export/NidsExport.php
```diff
@@ -38,15 +41,16 @@ public function handler($data, $options = array())
 				$this->__convertFromEventFormat($data['Attribute'], $data, $options, $continue);
 			}
 			if (!empty($data['Object'])) {
-				foreach ($data['Object'] as $object) {
-					$this->__convertFromEventFormat($object['Attribute'], $data, $options, $continue);
-				}
+				#foreach ($data['Object'] as $object) {
```
mokaddem: Delete line?

app/Lib/Export/NidsExport.php
```diff
@@ -69,6 +73,44 @@ private function __convertFromEventFormat($attributes, $event, $options = array(
 		return true;
 
 	}
+	
+	private function __convertFromEventFormatObject($objects, $event, $options = array(), $continue = false) {
+		
+		#CakeLog::debug("ConvertFromEventFormatObject");
```
mokaddem: Delete line?

mokaddem:
Hello!
Thanks a lot for the improvements.
In addition to the few comments placed in the PR itself, please find some general comments below:
1. The biggest issue I see is that when using the `event` scope (`/events/restSearch`), any `attributes` belonging to an object will be ignored if the object is not of the type `network-connection`. This is an issue as we have lots of different objects (`domain-ip`, ...) that should also be included in the export. This is not an issue for the `attribute` `scope` (see point 2).
2. Not sure if it's done on purpose or not but `network-connection` objects are only handled as intended when the `scope` is `event`. Not while using `attribute` (See related comment inline of this PR).
3. Also here is a mix between tabs and spaces in the file. Could it be possible to uniform it? (Preferably with spaces?).

emmekappa86:
Hi mokaddem,

Thanks for the feedback and sorry for my late response.
I will go through your comments and suggestions and try to fix everything.

Best regards

             Marco

app/Lib/Export/NidsExport.php
emmekappa86: Done.

app/Lib/Export/NidsExport.php
emmekappa86: The comment should have been deleted as it was written at the very beginning while studying the code. However, I am not sure I get exactly the situation in which an "object" signature will be created while using the "attribute" scope. Could you make an example?

app/Lib/Export/NidsExport.php
emmekappa86: Done.

app/Lib/Export/NidsExport.php
emmekappa86: Done.

emmekappa86:
Hi mokaddem,

Here is my update:

1.  I have changed the code to ensure that any object not included in the "supportedObject" list is treated as it used to be (thus, considering the attributes one by one
2. As written in the comments above, I might not have fully understood the way scopes are used. What I thought was that the feature was needed only when the scope is "Event". Therefore, I have implemented it this way. Furthermore, I tried to follow the previous code (the objects were only checked within the "Event" scope)
3. I replaced all tabs with spaces

app/Lib/Export/NidsExport.php
```diff
@@ -7,84 +7,125 @@ class NidsExport
...
+                }
+
+                $rearranged[] = array(
+                    'Attribute' => $object, // Using 'Attribute' instead of 'Object' to comply with function export
+                    'AttributeTag' => $objectTag, // Using 'AttributeTag' instead of 'ObjectTag' to comply with function export
+                    'Event' => $event['Event']
+                );
+
+            } else { // In case no custom export exists for the object, the approach falls back to the attribute case
+                $this->__convertFromEventFormat($object['Attribute'], $data, $options, $continue);
```
mokaddem: Replace `$data` by `$event`

app/Lib/Export/NidsExport.php
```diff
@@ -142,6 +183,9 @@ public function export($items, $startSid, $format="suricata", $continue = false)
 
             $sid = $startSid + ($item['Attribute']['id'] * 10); // leave 9 possible rules per attribute type
             $sid++;
+            
+            if(!empty($item['Attribute']['type'])) { // item is an 'Attribute'
```
mokaddem, line 187: Fix indentation

app/Lib/Export/NidsExport.php
```diff
@@ -201,10 +245,116 @@ public function export($items, $startSid, $format="suricata", $continue = false)
                 default:
                     break;
             }
+
+          } else if(!empty($item['Attribute']['name'])) { // Item is an 'Object'
```
mokaddem: Fix indentation

mokaddem:
Hello @emmekappa86,
Add a look at your changes and I still have some nitpicking.
The rest looks good.
Looking forward for these fixes to merge the PR!

app/Lib/Export/NidsExport.php
emmekappa86: Done.

app/Lib/Export/NidsExport.php
emmekappa86: Done.

app/Lib/Export/NidsExport.php
emmekappa86: Done.

emmekappa86:
Thanks for the review. I fixed the reported mistakes.

emmekappa86:
Hi @mokaddem, I wanted to ask if this feature can be merged. Is there anything else I should do/fix in the code?

mokaddem:
Hey, finally merged it into the [`develop` branch](https://github.com/MISP/MISP/commit/8f4c850cde2b5296b05efa2ded21f0e4aee7466a)
Thanks a lot for the work!

PR #7551, opened 2021-07-06
fix: wrong attribute value hash computed

Phat3:
#### What does it do?

The problem that this proposed PR solves is the following. When trying to add a duplicated object that contains in one of its attributes a text that will be modified by the method `Attribute->runRegexp`, the duplication is not detected. This happens because inside the method `MispObject->checkForDuplicateObjects` the hash of the attribute value that has to be inserted gets computed **before** performing the regular expression substitution, while the hash of the pre-existing attribute gets computed **after** this substitution. Therefore, the 2 hashes would never match. Specifically, this substitution happens later on during validation inside the file `app/Model/Attribute.php#591`.

As an example, if the attribute value contains the string `Os: Windows`, one of the default MISP regex will modify it as `O%WINDIR%`. When trying to insert this attribute twice the method `MispObject->checkForDuplicateObjects` will erroneously compare the `sha256("Os: Windows")` with the `sha256("O%WINDIR%")` and the duplicate won't be found. During the save process the attribute will then be modified in `O%WINDIR%` resulting in having 2 attribute with the same value.

The proposed solution performs the substitution on the attribute before computing the hash.

#### Questions

Signed-off-by: Sebastiano Mariani <smariani@vmware.com>

### DIFF
--- a/app/Model/MispObject.php
+++ b/app/Model/MispObject.php
@@ -365,9 +365,13 @@ class MispObject extends AppModel
                     $attribute['value'] = $attribute['value'] . '|' . md5(base64_decode($attribute['data']));
                 }
             }
+
+            $attributeValueAfterModification = $this->Attribute->modifyBeforeValidation($attribute['type'], $attribute['value']);
+            $attributeValueAfterModification = $this->Attribute->runRegexp($attribute['type'], $attributeValueAfterModification);
+
             $newObjectAttributes[] = hash(
                 'sha256',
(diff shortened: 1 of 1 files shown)


mokaddem:
Hello!
Thanks a lot for the contribution!
I merged it manually in the `develop` branch (https://github.com/MISP/MISP/commit/c51ef9da8ed187347adde16a805c90752cd57007).
(It's not showing up as merged because you created the PR on the `2.4` branch)
I also provided a small feedback showing which object triggered the duplication check to fire.
Thanks again!

PR #7571, opened 2021-07-16
Migrate org blocklists index view

righel:
#### What does it do?

Migrate to new view factories:
* /org_blocklists/index
* /org_blocklists/add
* /org_blocklists/edit

#### Questions


iglocska:
This fails for me. The uuid field should be called "uuids" not "org_uuids" - otherwise the blocklist component will fail with undefined uuids.

iglocska:
Thanks a lot!

PR #7574, opened 2021-07-19
Migrate threads index view

righel:
#### What does it do?

Migrated /threads/index view to new views factory.

#### Questions


iglocska:
Thanks a lot!

PR #7580, opened 2021-07-20
Migrate event blocklists views

righel:
#### What does it do?

Migrate event_blocklists view to views factory:
* /event_blocklists/index
* /event_blocklists/add
* /event_blocklists/edit

#### Questions


iglocska:
Thanks a lot!

PR #7593, opened 2021-07-22
Migrate templates views

righel:
#### What does it do?

Migrate Template views to view factory.
* `/templates/index`
* `/templates/add`
* `/templates/edit/:id`
* `/templates/view/:id`

#### Questions


iglocska:
Thanks, looks great!

PR #7602, opened 2021-07-26
chg: [shibbauth] added option to block organisation changes at login

liviuvalsan:
#### What does it do?

New ApacheShibbAuth.BlockOrgModifications setting added, defaults to false, boolean. If set to true, will block updates to the organisation of existing users on authentication. This preserves any modifications made by a site admin in MISP and is similar to ApacheShibbauth.BlockRoleModifications (same logic applied to role modifications).

#### Questions


adulau:
Thank you for the extension. It makes sense.

PR #7613, opened 2021-07-29
quick fix sticky buffers

lk-dll:

#### What does it do?

According to documentation (https://suricata.readthedocs.io/en/suricata-6.0.3/rules/http-keywords.html#http-keywords) sticky buffers should be before content, http.header and http.uri isn't marked as sticky buffers, but rules are wrongly generated and reported to logs. Tested on stable Suricata v6.0.1+

#### Questions

- [ x ] Are you using it in production?


Cooper-Dale:
agree

adulau:
Thank you very much!

PR #7633, opened 2021-08-09
chg: migrate news views to factory.

righel:
#### What does it do?

Migrate news views to factory:
* `/news/index`
* `/news/ad/[id]`
* `/news/edit/[id]`

#### Questions


iglocska:
Tested and looks good, thanks a lot!

PR #7641, opened 2021-08-11
chg: migrate sharing_views/view/[id] to factory

righel:
#### What does it do?

Migrate `sharing_groups/view/[id]` view to factory.

#### Questions


iglocska:
Thanks a lot!

PR #7648, opened 2021-08-11
chg: [sync] Remove commit and MISP-version from HTTP header

JakubOnderka:
#### What does it do?

Commit and MISP version is also part of User-Agent, to it doesn't make sense to have that information in three separate headers. Also `commit` and `MISP-version` headers is not allowed in HTTP protocol.

#### Questions


iglocska:
Thanks a lot, indeed the user agent has the information anyway.

PR #7656, opened 2021-08-12
Migrate tags views

righel:
#### What does it do?

migrate tags view to factory
* `/tags/add`
* `/tags/edit/[id]`

#### Questions


iglocska:
Thanks a lot!

PR #7701, opened 2021-08-27
chg: [UI] Make the Event Info field resizable

noat:

#### What does it do?

#### Questions

### DIFF
--- a/app/View/Events/add.ctp
+++ b/app/View/Events/add.ctp
@@ -47,7 +47,7 @@
                     'field' => 'info',
                     'label' => __('Event Info'),
                     'class' => 'input span6',
-                    'type' => 'text',
+                    'type' => 'textarea',
                     'placeholder' => __('Quick Event Description or Tracking Info')
                 ),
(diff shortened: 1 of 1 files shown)


mokaddem:
Dear @noat,
Thanks a lot for your contribution.
The size of the `Event info` field input is indeed small. However, this is on purpose.
The aim of this field is to provide a quick and concise description of the event.
I'm not very in favor of allowing user to enlarge it as it might give them the false impression that this field could contain a longer description.
However, these are my 2 cents. Maybe @iglocska or @adulau have other thoughts?

adulau:
Should we make it optional as a configuration? It might be a specific requirements for a specific sharing communities?

iglocska:
I agree with Sami, definitely wouldn't change the default behaviour. It used to be resizable and creative users just crammed all technical information in the info field.

With that said, we could make it indeed an option, however that might result in info fields being shared that get truncated whilst viewing them on instances without the setting set.

PR #7720, opened 2021-09-02
chg: [misp-wipe] wipe auth_keys tables

RichieB2B:
This PR makes misp-wipe also wipe the `auth_keys` table.


adulau:
Thank you! Good catch.

PR #7734, opened 2021-09-07
chg: skip dev dependencies when installing via INSTALL.sh script.

righel:
#### What does it do?
do not install composer dev dependencies when installing via INSTALL.sh automated install script.
reason: dev dependencies should be only installed on development/test instances.

#### Questions


SteveClement:
Looks good.

PR #7744, opened 2021-09-10
fix: [misp-retention] use update_tag

RichieB2B:
edit_tag no longer exists in PyMISP so we need to use update_tag instead.


mokaddem:
Looks good!

mokaddem:
Out of curiosity, have you tried using the built-in decaying feature (it also support the `retention` taxonomy)?
If yes and you decided not to use it, would you mind sharing the reasons behind that choice?
Thanks!

RichieB2B:
The retention script is working well for us. We have not looked into replacing it with the MISP decaying feature yet.

PR #7817, opened 2021-10-08
new: Supports MinIO as alternative to AWS S3

fandigunawan:
Needs an alternative S3 compatible file storage as it is against their policy. The widely used S3 compatible system includes [MinIO](https://min.io/). MinIO uses PHP AWS S3 SDK as pointed in MISP Composer requirements and MinIO's documentation.

#### What does it do?

Supports MinIO as an alternative to AWS S3 for file storage #3560
Check file whether it is exists, if it is not exists it will throw exception that will notify upper layer code.

#### Questions

### DIFF
--- a/app/Lib/Tools/AWSS3Client.php
+++ b/app/Lib/Tools/AWSS3Client.php
@@ -1,6 +1,7 @@
 <?php
 
 use Aws\S3\S3Client;
+use Aws\Exception\AwsException;
 
 class AWSS3Client
 {
@@ -14,7 +15,9 @@ class AWSS3Client
                 'bucket_name' => 'my-malware-bucket',
                 'region' => 'eu-west-1',
                 'aws_access_key' => '',
-                'aws_secret_key' => ''
+                'aws_secret_key' => '',
+                'aws_endpoint' => '',
+                'aws_compatible' => false
         );
 
         // We have 2 situations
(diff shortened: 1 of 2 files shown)


app/Lib/Tools/AWSS3Client.php
```diff
@@ -39,10 +42,39 @@ private function __getSetSettings()
...
+                 //   - https://docs.guzzlephp.org/en/5.3/clients.html#verify
+                 // Example: 
+                 // -- Verify certificate
+                 //    'http'    => ['verify' => '/usr/lib/ssl/certs/minio.pem'],
+                 // -- Do not verify certificate, securitywise, this option is not recommended, however due to 
+                 //    internal deployment scheme it is acceptable risk to set this to false
+                 //    'http'    => ['verify' => false],
+                 // -- Verify againts  built in CA certificates
+                 //    'http'    => ['verify' => true],
+                 'http'    => ['verify' => false],
```
JakubOnderka: Thats dangerous and maybe makes sense in your env, but it should not be default.

fandigunawan: I will make it default to trust internal trust store

adulau: @fandigunawan Can you update the pull-request? Like that we can merge it. Thanks a lot.

fandigunawan: Sorry, I was very busy lately.
I have pushed the patch to default the TLS validation to true and adds custom trusted CA file.

adulau: Thank you

adulau:
Thank you for the contribution and the updates!

PR #7818, opened 2021-10-08
fix: [internal] withCredentials property was added into $.ajaxSetup()…

MrBoba:
… to get rid of 403 and 302 responses

#### What does it do?

It fixes missing cookies in AJAX requests like:
 - /events/checkLocks/<ID>/<TS>
 - /threads/view/<ID>/true
 - /eventReports/index/event_id:<ID>/index_for_event:1
 - ...
I've a new MISP installation and when go into event details - issues appear.
In some cases on the backend the ajax request doesn't determine as ajax - 'X-Requested-With' header fixes this.

#### Questions

### DIFF
--- a/app/webroot/js/misp.js
+++ b/app/webroot/js/misp.js
@@ -5379,3 +5379,13 @@ $('td.rotate').hover(function() {
     var t = parseInt($(this).index()) + 1;
     $table.find('td:nth-child(' + t + ')').css('background-color', '');
 });
+
+$.ajaxSetup({
+    xhrFields: {
+        withCredentials: true
+    },
+    headers: {
+        'X-Requested-With': 'XMLHttpRequest',
+    },
+});
+


iglocska:
Nice, that should solve some annoying issues, thanks!

iglocska:
Reverted for now, as suggested by @JakubOnderka this is a bit too lax when it comes to requests to other domains. However a new PR that only affects queries such as checkLocks specifically would be nice!

JakubOnderka:
What browser are you using? Can you also check if you use correct domain? Because it looks like AJAX tries to send requests to another domain (cross domain requests) and thats the reason why cookies are not sent.

MrBoba:
I am using Firefox 93.0 (64-bit). I used the machine's IP as the host and this IP is the same for regular and AJAX requests. Am I doing something wrong?

MrBoba:
I've tried with domain and it works perfect.

PR #7831, opened 2021-10-12
fix: keep tag local state when importing from json or sync from internal

marjatech:
#### What does it do?

Fixes MISP#7810
When importing an Event via JSON, local tags inside the json should stay local after import too, and not be attached as global ones.
Same applies for Sync-Operations from internal instances (for any other instance local tags get stripped anyway)

#### Questions


iglocska:
Thanks a lot!

PR #7878, opened 2021-10-22
Exclude the /venv/* as it causes confusion

imidoriya:
.gitignore update


adulau:
Good point! Thank you for the contribution.

PR #7887, opened 2021-10-25
Clarify some aspects of the Shibboleth config

thijskh:


adulau:
Thanks for the PR. We are checking with some existing MISP Shibboleth installation how the changes proposed could impact those.

PR #7892, opened 2021-10-27
chg: [tags] possibility to define local_only tags and galaxies

### LINKED ISSUE #7876: Possibility do define "local only" tags and galaxies
Some tags, taxonomies and galaxies are made to only be used internally and we would like to avoid spreading them to external MISP. One way to do it is to add local tags, taxonomies and galaxies to an event or an attribute. Unfortunately it is not possible to enforce a specifc tag, taxonomy or galaxy to only be used as local. (the user can always make a mistake)

Ideally a new checkbox next to tags, taxonomies and galaxies in the UI "Local only" that would be enforced on the UI and on the backend to make sure that these taxonomies and galaxies cannot be added as global.

Making the tags not exportable does not help because we still want to access the tags via the API on our own instance.

lfortemps:
#### What does it do?

Fix #7876 .
Add local only options for tags and galaxies that shouldn't be put as global by the user.

Open for review

#### Questions

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -3496,7 +3496,7 @@ class EventsController extends AppController
                 $tag = $this->Event->EventTag->Tag->find('first', array(
                     'conditions' => $conditions,
                     'recursive' => -1,
-                    'fields' => array('Tag.name')
+                    'fields' => array('Tag.name', 'Tag.local_only')
                 ));
                 if (!$tag) {
                     $fails[$tag_id] = __('Tag not found.');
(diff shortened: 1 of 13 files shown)


iglocska:
Thanks a lot, very nice idea for how to implement it!

github-germ:
We upgraded to 2.4.151 and have an issue with our PyMISP apps failing with the tagging process as follows.

We've added a custom taxonomy and enabled all contained tags resulting in `Tags.local_only=1` for all these new tags. This causes our existing PyMISP app calls to `pymisp.tag()` to fail with `{"0":"Invalid Tag. This tag can only be set as a local tag."}'}`. 

**Question**: Do we need to revise all PyMISP app calls to `tag` to set `local=True` or is there a way to make these custom taxonomy tags all global when installing/enabling?

PR #7919, opened 2021-11-02
fix: [UI] restore event link on attributes index

lfortemps:
#### What does it do?
With the refactoring of the attributes view code, it seems that the link to the event disappeared. It is really useful to have it, especially in the context of search results, to be able to navigate to the parent event.

#### Questions


JakubOnderka:
Already fixed in https://github.com/MISP/MISP/pull/7902/files

PR #7931, opened 2021-11-05
Fix docblock formatting and add newer settings to README documentation

thijskh:


adulau:
Thank you! It makes sense.

PR #7939, opened 2021-11-08
Add simple background jobs

righel:
#### What does it do?
`BackgroundJobsTool` is a Tool/Utility to handle the background jobs regardless of the engine implementation (Resque/Simple).
The plan is have a lightweight engine to remove the CakeResque dependency easily in the future, while keeping support for both for engines for now.  
##### Redis
Used to keep a FIFO queue (list and rpush/lpop) of pending jobs for each queue type (default, email, etc.) and to save the job and worker statuses.
##### StartWorkerShell.php
Starts a worker that dequeues jobs from a particular queue and process them one by one.
##### Supervisor
`sudo apt install supervisor`
Manages the workers, logs, autorestart, etc.
PHP packages required to interact via API with `supervisord`:
```composer require lstrojny/fxmlrpc supervisorphp/supervisor guzzlehttp/guzzle php-http/message```

Add the following settings to the `Supervisor` conf file.
`/etc/supervisor/supervisord.conf` 
```
[inet_http_server]
port=9001
username=supervisor
password=supervisor
```

Use the following configuration as a template for the services:
`/etc/supervisor/conf.d/misp-workers.conf`
```
[group:misp-workers]
programs=default,email,cache,prio,update

[program:default]
directory=/var/www/MISP
command=/var/www/MISP/app/Console/cake start_worker default
process_name=%(program_name)s_%(process_num)02d
numprocs=1
autostart=true
autorestart=true
redirect_stderr=false
stderr_logfile=/var/www/MISP/app/tmp/logs/misp-workers-errors.log
stdout_logfile=/var/www/MISP/app/tmp/logs/misp-workers.log
directory=/var/www/MISP
user=www-data

[program:prio]
directory=/var/www/MISP
command=/var/www/MISP/app/Console/cake start_worker prio
process_name=%(program_name)s_%(process_num)02d
numprocs=1
autostart=true
autorestart=true
redirect_stderr=false
stderr_logfile=/var/www/MISP/app/tmp/logs/misp-workers-errors.log
stdout_logfile=/var/www/MISP/app/tmp/logs/misp-workers.log
directory=/var/www/MISP
user=www-data

[program:email]
directory=/var/www/MISP
command=/var/www/MISP/app/Console/cake start_worker email
process_name=%(program_name)s_%(process_num)02d
numprocs=2
autostart=true
autorestart=true
redirect_stderr=false
stderr_logfile=/var/www/MISP/app/tmp/logs/misp-workers-errors.log
stdout_logfile=/var/www/MISP/app/tmp/logs/misp-workers.log
directory=/var/www/MISP
user=www-data

[program:update]
directory=/var/www/MISP
command=/var/www/MISP/app/Console/cake start_worker update
process_name=%(program_name)s_%(process_num)02d
numprocs=1
autostart=true
autorestart=true
redirect_stderr=false
stderr_logfile=/var/www/MISP/app/tmp/logs/misp-workers-errors.log
stdout_logfile=/var/www/MISP/app/tmp/logs/misp-workers.log
directory=/var/www/MISP
user=www-data
```
`sudo service supervisor restart`
##### Config
```
  'MISP' => [
    ...,
  ],
  'SimpleBackgroundJobs' => [
    'enabled' => true,
    'redis_host' => 'localhost',
    'redis_port' => 6379,
    'redis_password' => '',
    'redis_database' => 1,
    'redis_namespace' => 'background_jobs',
    'max_job_history_ttl' => 86400,
    'track_status' => false,
    'supervisor_host' => 'localhost',
    'supervisor_port' => '9001',
    'supervisor_user' => 'supervisor',
    'supervisor_password' => 'supervisor',
  ], 
```

#### Notes
* MySQL `jobs` table should be removed imho, kept it for now for compatibility/limiting scope of the refactor. Keeping track of the jobs just using Redis is simpler and provides a single source of truth.
* Scheduled tasks (TasksController) are not supported with the new engine, as they require some extra development and to my understanding this is going to be deprecated.

### CODE REVIEW

**Comment by JakubOnderka** [comment]
This is so cool! I will write you some feedback.

File: app/Console/Command/EventShell.php [hunk 1]
```diff
@@ -414,8 +435,14 @@ public function publish_sightings()
 
         $sightingsUuidsToPush = [];
         if (isset($this->args[4])) { // push just specific sightings
-            $path = $this->args[4][0] === '/' ? $this->args[4] : (APP . 'tmp/cache/ingest' . DS . $this->args[4]);
```
JakubOnderka: This looks like bad merge.

app/Console/Command/EventShell.php
```diff
@@ -522,8 +549,9 @@ public function processfreetext()
 
         $inputFile = $this->args[0];
         $inputFile = $inputFile[0] === '/' ? $inputFile : APP . 'tmp/cache/ingest' . DS . $inputFile;
-        $inputData = FileAccessTool::readAndDelete($inputFile);
```
JakubOnderka: Also this look like just bad merge.

app/Lib/Tools/BackgroundJobsTool.php
```diff
@@ -0,0 +1,681 @@
...
+     *           'supervisor_host' => 'localhost',
+     *           'supervisor_port' => '9001',
+     *           'supervisor_user' => '',
+     *           'supervisor_password' => '',
+     *      ]
+     *
+     * @param array $settings
+     * @return void
+     */
+    public function initTool(array $settings): void
```
JakubOnderka: Why this method is not constructor?

app/Lib/Tools/BackgroundJobsTool.php
```diff
@@ -0,0 +1,681 @@
...
+                'command' => $command,
+                'args' => $args,
+                'trackStatus' => $trackStatus ?? $this->settings['track_status'],
+                'metadata' => $metadata
+            ]
+        );
+
+        $this->RedisConnection->rpush(
+            $queue,
+            json_encode($backgroundJob->jsonSerialize())
```
JakubOnderka: No need to call `jsonSerialize` method if class implements JsonSerializable.

app/Lib/Tools/BackgroundJobsTool.php
```diff
@@ -0,0 +1,681 @@
...
+            throw new InvalidArgumentException('Invalid worker name, must be one of format {queue_name}_{process_id}, example: default_00');
+        }
+
+        return true;
+    }
+
+    /**
+     * @return Redis
+     */
+    private function createRedisConnection(): Redis
```
JakubOnderka, line 586: You can use `$redis->setOption(Redis::OPT_SERIALIZER, Redis::SERIALIZER_JSON);`, so values will be automatically serialized/deserialized.

app/Lib/Tools/BackgroundJobsTool.php
```diff
@@ -0,0 +1,681 @@
...
+            new \fXmlRpc\Transport\HttpAdapterTransport(
+                new \Http\Message\MessageFactory\GuzzleMessageFactory(),
+                new \GuzzleHttp\Client($httpOptions)
+            )
+        );
+
+        return new \Supervisor\Supervisor($client);
+    }
+
+    private function getJobById(int $jobId): ?Job
```
JakubOnderka: What about to change this method to directly save process ID? Because it is only use here.

app/Lib/Tools/BackgroundJobsTool.php
```diff
@@ -0,0 +1,681 @@
...
+     * @return \Supervisor\Process
+     *
+     * @throws NotFoundException
+     */
+    private function getProcessByPid(int $pid): \Supervisor\Process
+    {
+        $procs = $this->Supervisor->getAllProcesses();
+
+        foreach ($procs as $proc) {
+            if ($proc->offsetGet('group') === self::MISP_WORKERS_PROCESS_GROUP && $proc->offsetGet('pid')  === $pid) {
```
JakubOnderka: Two spaces

app/Lib/Tools/BackgroundJobsTool.php
```diff
@@ -0,0 +1,681 @@
...
+                return $proc;
+            }
+        }
+
+        throw new NotFoundException(
+            sprintf('Worker with pid=%s not found.', $pid)
```
JakubOnderka: `"Worker with pid=$pid not found."` should be cleaner.

app/Lib/Tools/BackgroundJobsTool.php
JakubOnderka, line 586: Also you can use `$redis->setOption(Redis::OPT_PREFIX, 'myAppName:');	// use custom prefix on all keys`

app/Console/Command/StartWorkerShell.php
```diff
@@ -0,0 +1,114 @@
...
+                try {
+                    $this->BackgroundJobsTool->run($job);
+                } catch (Exception $exception) {
+                    CakeLog::error("[WORKER PID: {$this->worker->pid()}][{$this->worker->queue()}] - job id: {$job->id()} failed with exception: {$exception->getMessage()}");
+                    $job->status  = BackgroundJob::STATUS_FAILED;
+                    $this->BackgroundJobsTool->update($job);
+                }
+            }
+
+            sleep($this->sleepInterval);
```
JakubOnderka: What is the reason for this sleep? `dequeue` is blocking.

app/Console/Command/StartWorkerShell.php
```diff
@@ -0,0 +1,114 @@
...
+        return $parser;
+    }
+
+    public function main(): void
+    {
+        $this->worker = new Worker(
+            [
+                'pid' => getmypid(),
+                'queue' => $this->args[0],
+                'user' => trim(shell_exec('whoami'))
```
JakubOnderka: Maybe better to use something like this. Executing `whoami` command can be detected by SIEM tools and generate warnings. 


```php
if (function_exists('posix_getpwuid') && function_exists('posix_geteuid')) {
    $whoami = posix_getpwuid(posix_geteuid())['name'];
} else {
    $whoami = trim(shell_exec('whoami'));
}
```

app/Console/Command/MonitorWorkersShell.php
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Check workers process status
+     *
+     * @param Worker[] $workers
+     * @return void
+     */
+    private function checkWorkersProcessStatus(array $workers): void
+    {
+        foreach ($workers as $worker) {
+            if (!file_exists("/proc/{$worker->pid()}")) {
```
JakubOnderka: I think this will not work, because the result of this function is cached.

app/Console/Command/StartWorkerShell.php
righel: I'll remove it, originally tried to mimick the `--interval 5` used in CakeResque workers. But since I'm blocking until new elements, doesn't make sense.

app/Console/Command/MonitorWorkersShell.php
righel: not sure I understand, could you elaborate a bit more?
By cached you mean its only updated every `sleepInterval` seconds?

JakubOnderka: Check https://www.php.net/clearstatcache

righel: true, i'll refactor to rely only on Supervisor API for workers info/mgmt

app/Lib/Tools/BackgroundJobsTool.php
```diff
@@ -0,0 +1,647 @@
...
+    {
+        $workers = [];
+        $procs = $this->Supervisor->getAllProcesses();
+
+        foreach ($procs as $proc) {
+            if ($proc->offsetGet('group') === self::MISP_WORKERS_PROCESS_GROUP) {
+                $workers[] = new Worker([
+                    'pid' => $proc->offsetGet('pid'),
+                    'queue' => explode("_", $proc->offsetGet('name'))[0],
+                    'user' => trim(shell_exec(sprintf("ps -o uname= -p %s", (int) $proc->offsetGet('pid')))),
```
iglocska: uname= missing argument

app/Lib/Tools/BackgroundJobsTool.php
```diff
@@ -0,0 +1,647 @@
...
+     *
+     * @return Worker[] List of worker's instances.
+     */
+    public function getWorkers(): array
+    {
+        $workers = [];
+        $procs = $this->Supervisor->getAllProcesses();
+
+        foreach ($procs as $proc) {
+            if ($proc->offsetGet('group') === self::MISP_WORKERS_PROCESS_GROUP) {
```
iglocska, line 295: Check if it's running - if not then pid returned is 0 and the $proc object contains the reasons for the failure.

app/Lib/Tools/BackgroundJobsTool.php
```diff
@@ -0,0 +1,647 @@
...
+        return true;
+    }
+
+    /**
+     * Validate worker name
+     *
+     * @return boolean
+     * @throws InvalidArgumentException
+     */
+    private function validateWorkerName(string $name): bool
```
iglocska, line 570: This function receives as name the name of the queue, whilst expecting [queue]_[process_id] when starting a new worker via the UI (called by startWorker()). 

This leads to a notice error on the list() call on 539 and finally an exception thrown on 544.

iglocska:
`SimpleBackgroundJobs.enabled` and `MISP.use_simple_background_jobs` are problematic. Even though the setting in the former described that it needs the latter to be enabled, simply enabling the latter without the former (by flipping the settings one by one via the UI) makes the server settings inoperable.

Consider refactoring this into a single setting to avoid this situation, or make the checks aware of this situation being potentially the case.

app/Lib/Tools/BackgroundJobsTool.php
righel: I'll skip the worker if pid=0

app/Lib/Tools/BackgroundJobsTool.php
JakubOnderka: No, this is OK.
```
# ps -o uname= -p 1
root
# ps -o uname -p 1
USER
root
```

app/Controller/FeedsController.php
```diff
@@ -488,26 +488,30 @@ public function fetchFromFeed($feedId)
...
             );
-            $this->Job->save($data);
-            $jobId = $this->Job->id;
-            $process_id = CakeResque::enqueue(
-                'default',
-                'ServerShell',
-                array('fetchFeed', $this->Auth->user('id'), $feedId, $jobId),
-                true
+
+            $this->Feed->getBackgroundJobsTool()->enqueue(
```
iglocska, line 502: This doesn't work for the legacy background processor.

app/Model/Server.php
```diff
@@ -6935,6 +6947,95 @@ private function generateServerSettings()
...
+                'max_job_history_ttl' => [
+                    'level' => 2,
+                    'description' => __('The time in seconds the job statuses history will be kept.'),
+                    'value' => '',
+                    'test' => 'testForNumeric',
+                    'type' => 'numeric'
+                ],
+                'track_status' => [
+                    'level' => 2,
+                    'description' => __('Default behavior for tracking jobs statuses.'),
```
JakubOnderka: I don't understand what this setting means :/

app/Model/Server.php
```diff
@@ -6935,6 +6947,95 @@ private function generateServerSettings()
...
+                    'value' => '',
```
JakubOnderka: Default value should be boolean, so `true` or `false`.

app/Model/Server.php
```diff
@@ -6935,6 +6947,95 @@ private function generateServerSettings()
...
+                'redis_port' => [
+                    'level' => 2,
+                    'description' => __('The port used by the redis server to be used for background jobs.'),
+                    'value' => 6379,
+                    'test' => 'testForNumeric',
+                    'type' => 'numeric'
+                ],
+                'redis_database' => [
+                    'level' => 2,
+                    'description' => __('The database on the redis server to be used for background jobs. If you run more than one MISP instance, please make sure to use a different database on each instance.'),
```
JakubOnderka: > If you run more than one MISP instance, please make sure to use a different database on each instance.

This is not true if different namespaces are used.

app/Model/Server.php
iglocska: Track in cakeresque allows us to request the status of a running job by pid. This should always be true from a MISP perspective and doesn't need a setting.

app/Model/Server.php
iglocska: True, different namespaces could be used, though to avoid potential mess-ups (which would be on us of course), using a different db for each instance is a nice way to separate concerns. Also easier to just flush DBs when needed without affecting other instances.

app/Controller/FeedsController.php
iglocska: My bad, the issue was an upstream change from MISP:develop that broke background processes.

app/Model/Server.php
righel: removed the setting, the parameter is kept in BackgroundJobsTool::enqueue() for compatibility with CakeResque jobs.

app/Model/Server.php
righel: added some text to make it clear

iglocska:
Thanks a million, awesome stuff!

PR #7955, opened 2021-11-12
chg: [internal] Code style

JakubOnderka:
#### Questions


righel:
looks good to me

PR #8006, opened 2021-11-30
chg: [feeds] Support for sharing groups with feeds, fixes #5758

### LINKED ISSUE #5758: Support for sharing groups with feeds
The feed_generator (PyMISP) and the MISP feed system are not compatible with sharing groups.
This is needed in the use-case where communities want to synchronise two MISP instances that from not inter-connected MISP setups.  (like airgaps) 

A workaround is by implementing it with custom PyMISP scripts for export and import.
However this is a bit dirty and defeats the purpose of the feed-system. Some complexity is also needed to support updates.

cvandeplas:
#### What does it do?

this PR brings in the option to ingest feeds that have configured distribution/sharing groups. It fixes #5758 , 

#### Questions

### DIFF
--- a/app/Controller/FeedsController.php
+++ b/app/Controller/FeedsController.php
@@ -124,7 +124,9 @@ class FeedsController extends AppController
             'menuItem' => 'index'
         ]);
         $this->loadModel('Event');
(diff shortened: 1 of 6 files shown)


iglocska:
Thanks a lot!

PR #8022, opened 2021-12-10
Feature sync type filtering

### LINKED ISSUE #6765: Additional sync filters to filter unwanted data
An idea that popped up in a workshop:

- have a sync filter to exclude/include certain attribute types

we could extend it to other filters too (such as the exclusion of certain object templates, etc).

mokaddem:
Introduces a new setting `MISP.enable_synchronisation_filtering_on_type` allowing the filtering on Attribute types and Object template whenever a PUSH or PULL synchronisation mechanism is triggered.

image

Fix #6765


mokaddem:
Hey @Wachizungu!
Fixed some issues. Feel free to provide more feedback in this thread.

Wachizungu:
Current feedback (commit f8d2782 - based on tests with pull sync):

1. object template based filtering seems all fine for the tests I did (0, 1 or multiple templates filtering for 1 or 2 events syncing)
2. attribute type filtering is behaving strangely for me at the moment. If the attribute type only appears in an object of an event, nothing happens, the full object is synced. If the attribute type is only in an attribute of the event, it is correctly filtered out. If it appears in both an object and separate attribute of the event, it's chaos. Event doesn't get synced at all or the objects are all empty.
3. unrelated to the above: When you first add the object template to filter, you see the name. When you edit later you see the UUID. It would be nice if in the second case it could be UUID + name somehow.

mokaddem:
Many thanks for your feedback!
Here is mine:
1. Good news
3. Partially fixed: Done it for PUSH but not for PULL (PUSH is using known data (IDs and UUIDs), PULL is using potentially unknown data). To keep it simple code-wise, the app only displayed the object name for PUSH. Let me know if this is a blocker though.
2. This is worrisome. I tried to both PUSH and PULL between two instance for both this is what I had:
    - All events were sync
    - Attributes being blocklisted were omitted and not saved
    - Objects being blocklisted were omitted and not saved
    - ObjectAttribute blocklisted (according to their `type`, not their `object_relation`) were omitted and not saved
If you manage to reproduce issue 2 again, would you mind sharing with me the event and the server configuration?
Thanks a lot!

Wachizungu:
Sure! Will check again tomorrow and also try with some new instances.

mokaddem:
@Wachizungu Many thanks for your exhaustive tests that help us discovered few issues.
I pushed 1 fix and performance optimization for push (Only one object being filtered out) and two other fixes for pull (empty object + ignoring objectAttribute).
Waiting for your feedback!

Wachizungu:
Looking good on 4a43ddc129e1b1fcde1caac40c6b9f396b126087 ! Thank you :)

PR #8042, opened 2021-12-21
LinOTP - A little refactor

andurin:
#### What does it do?

The current 2.4 LinOTP code has some issues which will prevent an easy use. 
For example:
* Requirement for php-curl
* Forced / unconfigurable trust dependency for the TLS/SSL Connection to LinOTP
* LinOTP Config won't be saved 
* As soon the MISP User knows his own password (set in MISP) the OTP is optional
  * Caused by LinOTP first, MISP FormAuthenticate second/last
* No hints on the login page regarding a needed OTP

I addressed these issues and did a little refactoring without touching the current workflow. So the current defaults should behave like the old one except the additional "OTP" field on the login page if the plugin is enabled.

A new feature in this LinOTP Plugin is the so called "mixed mode". Mixed in the meaning of "password from MISP" + "OTP from LinOTP" which provides the capability to do the user management in MISP and add 2FA additional. While `mixedauth` is enabled the LinOTP Plugin will first check, if username+password are good against the MISP Database. If so, it will continue to authenticate the OTP against LinOTP. If LinOTP fails to authenticate a exception is thrown to enforce 2FA.

I'm curious what you and the community are thinking about this change and I would like to see this go upstream.

Regards,
Hendrik

#### Questions


adulau (approved):
+1 for the removal of an external dep.

adulau (approved):

adulau (approved):
Make sense. SSL certificate is true by default.

adulau (approved):

adulau (changes requested):
I would recommend to have a false safe for $ret to ensure a False if something goes wrong in the response.

adulau (approved):

adulau (approved):

adulau (commented):
The new mixedauth model sounds interesting. I don't have specific comment. @iglocska any feedback on the mixed mode?

andurin:
> I would recommend to have a false safe for $ret to ensure a False if something goes wrong in the response.

Makes sense - please see update : https://github.com/MISP/MISP/pull/8042/commits/8964a36b393a07550013d6660c4f4f827c1eaf4d

iglocska:
> The new mixedauth model sounds interesting. I don't have specific comment. @iglocska any feedback on the mixed mode?

It looks very good and inline with a lot of requests we're getting in general.

app/View/Users/login.ctp
```diff
@@ -94,6 +100,9 @@ function submitLoginForm() {
             var $tmpForm = $('#temp form#UserLoginForm')
             $tmpForm.find('#UserEmail').val(email)
             $tmpForm.find('#UserPassword').val(password)
+            if (Configure::read('LinOTPAuth') == true) {
```
Chalkstyle, line 103: Would change it to:
```php
if (!empty(Configure::read('LinOTPAuth'))) {
```

app/View/Users/login.ctp
```diff
@@ -82,6 +85,9 @@ function submitLoginForm() {
     var url = $form.attr('action')
     var email = $form.find('#UserEmail').val()
     var password = $form.find('#UserPassword').val()
+    if (Configure::read('LinOTPAuth') == true) {
```
Chalkstyle, line 88: Would change it to:
```php
if (!empty(Configure::read('LinOTPAuth'))) {
```

app/View/Users/login.ctp
```diff
@@ -39,6 +39,9 @@
         <?php
             echo $this->Form->input('email', array('autocomplete' => 'off', 'autofocus'));
             echo $this->Form->input('password', array('autocomplete' => 'off'));
+            if (Configure::read('LinOTPAuth') == true) {
```
mokaddem: Would change it to:
```php
if (!empty(Configure::read('LinOTPAuth'))) {
```
Same for the ones below

app/Plugin/LinOTPAuth/Controller/Component/Auth/LinOTPAuthenticate.php
```diff
@@ -21,6 +28,76 @@ public function authenticate(CakeRequest $request, CakeResponse $response)
...
+                "status" => $status,
+                "value" => $value,
+            );
+
+            if (property_exists($result, 'detail')) {
+                $ret['detail'] = $result->detail;
+            }
+
+            CakeLog::debug("user: ${user} - status: ${status} value: ${value}");
+            // CakeLog::debug(var_dump($ret));
```
mokaddem, line 96: delete line

app/Plugin/LinOTPAuth/Controller/Component/Auth/LinOTPAuthenticate.php
```diff
@@ -33,21 +110,53 @@ public function getUser(CakeRequest $request)
...
         $password = $userFields['password'];
+        $otp = $userFields['otp'];
+
         CakeLog::debug("getUser email: ${email}");
 
-        $linotp = new LinOTP(
-            Configure::read("LinOTPAuth.baseUrl"),
-            Configure::read("LinOTPAuth.realm")
-        );
+        $linOTP_baseUrl = rtrim(Configure::read("LinOTPAuth.baseUrl"), "/");
```
mokaddem, line 120: Fallback value if `LinOTPAuth.baseUrl` is unset?

mokaddem:
Hey I added some comments. Mainly nitpicking ;)

app/Plugin/LinOTPAuth/Controller/Component/Auth/LinOTPAuthenticate.php
```diff
@@ -21,6 +28,76 @@ public function authenticate(CakeRequest $request, CakeResponse $response)
...
+        // POST data
+        $data = array(
+            "user" => $user,
+            "pass" => $password,
+            "realm" => $realm,
+        );
+
+        $url = "$baseUrl/validate/check";
+
+        CakeLog::debug( "Sending POST request to ${url}");
```
iglocska, line 52: Don't think this debug statement is needed in the debug logs, it will get pretty verbose.

app/Plugin/LinOTPAuth/Controller/Component/Auth/LinOTPAuthenticate.php
iglocska: Don't think a fallback is useful - but a check before that, outright breaking the execution if it's not configured would be.

app/Plugin/LinOTPAuth/Controller/Component/Auth/LinOTPAuthenticate.php
```diff
@@ -21,6 +28,76 @@ public function authenticate(CakeRequest $request, CakeResponse $response)
...
+        $results = $HttpSocket->post($url, $data);
+        $response = json_decode($results->body());
```
iglocska, line 57: Before this line I'd check if the response was succesful, you can do this via $results->code() or $results->isOk() iirc. This will avoid accidentally parsing / trying to json_decode() unwanted responses.

iglocska:
Just a few minor comments from me too. Looks really good, thanks for the awesome work. Also an additional thank you for ensuring that the new version will work out of the box without any modifications to the prior configuration, allowing for a smooth upgrade for users.

app/Plugin/LinOTPAuth/Controller/Component/Auth/LinOTPAuthenticate.php
andurin: rm_line() :)

app/Plugin/LinOTPAuth/Controller/Component/Auth/LinOTPAuthenticate.php
andurin: Thanks for the note - back to xdebug to check if all is fine...

adulau:
If you also fix the CodeFactor issue that would be great! Thanks again for your contribution.

app/View/Users/login.ctp
andurin: Done

andurin:
Next round please :)

adulau:
Thanks a lot for the changes. That's awesome. If the CI is passing, I'll merge it except if the Grandmaster flash @iglocska refuses ;-)

iglocska:
That looks great, thanks for all the work and the rapid fixes! Merged into develop and it will be included in the next release.

PR #8045, opened 2021-12-22
Allow sightings to be pushed upstream

RichieB2B:
The MISP event distribution level is decreased from 2 -> 1 -> 0 at each remote MISP server the event is passed on to:

4: Sharing group
3: All communities
2: Connected communities
1: This community
0: Your organisation only

Sightings however travel in both directions: upstream and downstream. Since 8830696c4db4a97bd6a7ea6a8f2a55cf0e018882 sightings are no longer pushed upstream. This patch makes that possible again.

The `push_sightings` flag for sync servers determines where to push sightings to. Only allowing sightings to be pushed to a remote server when the event distribution is set to "All communities" or "Connected communities" makes no sense.

Limiting sightings for "Sharing group" events to servers in the sharing group does make sense. This code is untouched.


iglocska:
Nice catch, thanks a lot!

PR #8055, opened 2021-12-24
fix: Fix typo and unify variable name with other.

### LINKED ISSUE #7929: Return uuid on duplicate object error.
Hello,

When pushing an object that already exist, this error is returned:
```
Something went wrong (403): {'saved': False, 'name': 'Could not add Object', 'message': 'Could not add Object', 'url': '/objects/add', 'errors': 'Object could not be saved.\nvalue: Duplicate object found (id: 80206). Since breakOnDuplicate is set the object will not be added.'}
```

Would it be possible to return an uuid instead of an id? (or maybe both)

The method in question to seems to be this one:
https://github.com/MISP/MISP/blob/1e930903b27c802df093787ee5674ca33a84340b/app/Model/MispObject.php#L362

nyx0:
This PR fixes a typo introduced in ed9ab9d7c2d6ca7792700151c2e442d09c8d67fd, I took the opportunity to rename the variable to something close to it's sibling (`$duplicatedObjectID`).

However even with this fix, it does not fix #7929, the object's UUID is not displayed as showed below.

```
pymisp/api.py(561)add_object()
    560         params = {'breakOnDuplicate': True} if break_on_duplicate else {}
    561         r = self._prepare_request('POST', f'objects/add/{event_id}', data=misp_object, kw_params=params)
--> 562         new_object = self._check_json_response(r)

> print(r.content)
b'{"saved":false,"name":"Could not add Object","message":"Could not add Object","url":"/objects/add","errors":"Object could not be saved.\\nvalue: Duplicate object found (id: 120025, uuid: ). Since breakOnDuplicate is set the object will not be added."}'
```

### DIFF
--- a/app/Model/MispObject.php
+++ b/app/Model/MispObject.php
@@ -359,7 +359,7 @@ class MispObject extends AppModel
         }
     }
 
-    public function checkForDuplicateObjects($object, $eventId, &$duplicatedObjectID, &$duplicateObjectUuid)
+    public function checkForDuplicateObjects($object, $eventId, &$duplicatedObjectID, &$duplicatedObjectUUID)
     {
         $newObjectAttributes = array();
         if (isset($object['Object']['Attribute'])) {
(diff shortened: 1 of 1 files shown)


nyx0:
@iglocska Do you think you could take a look at this? thank you :)

JakubOnderka:
Thanks for contribution, with small changes merged as https://github.com/MISP/MISP/pull/8348

Issue #8064, opened 2021-12-29
Bug: unable to start worker of type 'supervisord_status'


K0nfucius:
### Expected behavior

I'm using MISP (v2.4.152) with supervisord. I expect one or more running workers of type 'supervisord_status'.

### Actual behavior

I cannot see any running 'supervisord_status' workers at '/servers/serverSettings/workers'. Instead there is a red box 'worker is not running!'. When I click the box 'Start a worker' I get an error message 'An Internal Error Has Occurred.`

All other worker types are running without any problems, supervisord.service is running without problems. 

### Steps to reproduce

1. start your misp host and configured supervisord.service.

### Version

2.4.152

### Operating System

openSUSE Leap

### Operating System version

15.2

### PHP version

7.4.2

### Relevant log output

```shell
2021-12-29 18:28:46 Error: [InvalidArgumentException] Invalid background job queue supervisord_status, must be one of: [default, email, cache, prio, update]
Request URL: /servers/startWorker/supervisord_status
Stack Trace:
#0 /services/inst/misp/app/Lib/Tools/BackgroundJobsTool.php(373): BackgroundJobsTool->validateQueue()
#1 /services/inst/misp/app/Controller/ServersController.php(1269): BackgroundJobsTool->startWorkerByQueue()
#2 [internal function]: ServersController->startWorker()
#3 /services/inst/misp/app/Lib/cakephp/lib/Cake/Controller/Controller.php(499): ReflectionMethod->invokeArgs()
#4 /services/inst/misp/app/Lib/cakephp/lib/Cake/Routing/Dispatcher.php(193): Controller->invokeAction()
#5 /services/inst/misp/app/Lib/cakephp/lib/Cake/Routing/Dispatcher.php(167): Dispatcher->_invoke()
#6 /services/inst/misp/app/webroot/index.php(99): Dispatcher->dispatch()
#7 {main}
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

righel:
Hello @K0nfucius ,
Thanks for reporting, `supervisord_status` is not a valid worker, the bug is showing it there in the UI, so it should not affect the usage of the tool.
Let me know if you have any other issues with the workers.
 
I created a PR to fix this problem: https://github.com/MISP/MISP/pull/8067

PR #8067, opened 2021-12-30
fix: supervisord_status showing as a worker when its not

righel:
#### What does it do?
removes `supervisord_status` worker entry from workers diagnostics page as it's a auxiliary key to evaluate the status of `supervisord` and not a valid worker queue.

Fixes issue #8064

#### Questions


iglocska:
Thanks a lot!

PR #8092, opened 2022-01-13
LinOTP minor fixes

andurin:
#### What does it do?

1. Improved error handling/logging if SSL/TLS socket could not be established.
2. Added a link to login.ctp to visit the LinOTP Selfservice portal (if LinOTP is enabled).
3. Make LinOTP config via web-ui + cli interface.

#### Questions

- [ ] Does it require a DB change? - NO
- [x] Are you using it in production? - Yes
- [ ] Does it require a change in the API (PyMISP for example)? - NO

### DIFF
--- a/app/Model/Server.php
+++ b/app/Model/Server.php
@@ -1388,6 +1388,7 @@ class Server extends AppModel
             'misc' => 'Security',
             'Security' => 'Security',
             'Session' => 'Security',
+            'LinOTPAuth' => 'Security',
             'SimpleBackgroundJobs' => 'SimpleBackgroundJobs'
         );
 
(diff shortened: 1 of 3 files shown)


3c7:
Great additions!

app/Model/Server.php
```diff
@@ -7135,6 +7136,37 @@ private function generateServerSettings()
                 'type' => 'boolean',
                 'null' => true
             ),
+            'LinOTPAuth' => array(
+                'branch' => 1,
+                'baseUrl' => array(
+                    'level' => 2,
+                    'description' => __('The default LinOTP URL.'),
+                    'value' => 'https://linotp',
```
adulau: Can this be abused if a default value without domain name is controlled by an attacker locally? Just wondering if it's better to keep it empty by default

andurin: What would you think about:

`'value' => 'https://<your-linotp-baseUrl>',` ?

andurin:
Question to the Devs - does this failed build has relevance to this change? Guess not...
https://github.com/MISP/MISP/runs/4837101554?check_suite_focus=true#step:25:271

app/Model/Server.php
adulau: Perfect thanks!

adulau:
Thank you very much!

Issue #8108, opened 2022-01-21
Bug: Download as STIX1 XML invalid XML when you have a space in your org name


kara-1234:
### Expected behavior

STIX 1 XML export will correctly add an underscore in the org name, instead it will create two tags at the start of the XML Export.
```
<stix:STIX_Package 
        xmlns:TEST_ORG="https://exmaple.com"
        xmlns:cyboxCommon="http://cybox.mitre.org/common-2"
        xmlns:cybox="http://cybox.mitre.org/cybox-2"
	xmlns:cyboxVocabs="http://cybox.mitre.org/default_vocabularies-2"
	xmlns:AccountObj="http://cybox.mitre.org/objects#AccountObject-2"
	xmlns:ArtifactObj="http://cybox.mitre.org/objects#ArtifactObject-2"
	xmlns:ASObj="http://cybox.mitre.org/objects#ASObject-1"
	xmlns:AddressObj="http://cybox.mitre.org/objects#AddressObject-2"
	xmlns:PortObj="http://cybox.mitre.org/objects#PortObject-2"
	xmlns:DomainNameObj="http://cybox.mitre.org/objects#DomainNameObject-1"
```

### Actual behavior

Instead the export will add line three where there is a space in the org name, when this happens the file fails XML validation. 
```
<stix:STIX_Package 
        xmlns:TEST_ORG="https://exmaple.com"
        xmlns:TEST ORG="https://example.com"
        xmlns:cyboxCommon="http://cybox.mitre.org/common-2"
        xmlns:cybox="http://cybox.mitre.org/cybox-2"
	xmlns:cyboxVocabs="http://cybox.mitre.org/default_vocabularies-2"
	xmlns:AccountObj="http://cybox.mitre.org/objects#AccountObject-2"
	xmlns:ArtifactObj="http://cybox.mitre.org/objects#ArtifactObject-2"
	xmlns:ASObj="http://cybox.mitre.org/objects#ASObject-1"
	xmlns:AddressObj="http://cybox.mitre.org/objects#AddressObject-2"
	xmlns:PortObj="http://cybox.mitre.org/objects#PortObject-2"
	xmlns:DomainNameObj="http://cybox.mitre.org/objects#DomainNameObject-1"
```

### Steps to reproduce

Change your org name in settings to something with a space in it and then go to any event and click download as STIX 1 XML.  If you do not have a space in your org name it will export as expected. 

### Version

2.4.152

### Operating System

CentOS

### Operating System version

7

[...]

PR #8111, opened 2022-01-24
fix : Update Stix1Export.php to fix issue #8108

kara-1234:
Not sure if this is the best solution to #8108, but replacing the whitespace in the orgname here fixed the double tag problem. It also eliminates the need to have several sections of code in the misp_stix_converter.py where you are doing  org = re.sub('[\W]+', '', org.replace(" ", "_"))

#### What does it do?

Fixes issue #8108

#### Questions

- [ ] Does it require a DB change?  no
- [x] Are you using it in production? yes
- [ ] Does it require a change in the API (PyMISP for example)? no

### DIFF
--- a/app/Lib/Export/Stix1Export.php
+++ b/app/Lib/Export/Stix1Export.php
@@ -16,7 +16,7 @@ class Stix1Export extends StixExport
             '-s', $this->__scope,
             '-v', $this->__version,
             '-n', Configure::read('MISP.baseurl'),
-            '-o', Configure::read('MISP.org'),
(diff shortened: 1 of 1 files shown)


adulau:
Thanks a lot for the PR. I think @chrisr3d just did a fix. 

https://github.com/MISP/misp-stix/commit/64d61aa7baf4ca8179998818e47cc13640f265f5

@chrisr3d feel free to merge on 2.4

kara-1234:
Closing as the issue was fixed inside the python library. Also the misp-stix library fixed some other places where the orgname with a space would have caused issues.

PR #8141, opened 2022-02-12
Preserve Session.* configuration in serverSettingsSaveValue

folbricht-stripe:
#### What does it do?

Fixes an issue where running `/var/www/MISP/app/Console/cake live 1` would remove any configuration under `Session` from config.php.

Repro:

1) Set `Session.defaults = 'database'` by running `/var/www/MISP/app/Console/cake Admin setSetting "Session.defaults" "database"`
This correctly inserts the setting into config.php

2) Run `/var/www/MISP/app/Console/cake live 1`. This removes the setting from config.php as it checks to see if `Session.defaults` exists in a hash that has a `Session` as key.

Without this, logging into multi-node installation fails with `You have tripped the cross-site request forgery protection of MISP` as sessions are local instead of in the DB.

#### Questions

### DIFF
--- a/app/Model/Server.php
+++ b/app/Model/Server.php
@@ -2308,10 +2308,9 @@ class Server extends AppModel
 
         $settingsToSave = array(
             'debug', 'MISP', 'GnuPG', 'SMIME', 'Proxy', 'SecureAuth',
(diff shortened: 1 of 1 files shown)


iglocska:
Nice catch, thanks a lot!

PR #8167, opened 2022-02-23
Contribute TAXII push script

chisholm:
This PR contributes the taxii_push.py script.  It operates as a standalone script with a commandline interface:

```
usage: taxii_push.py [-h] --dir DIR --collection COLLECTION [--log_level {fatal,error,warning,info,debug}]

Translate MISP content to STIX 2.1 and push it to a TAXII 2.1 server.

optional arguments:
  -h, --help            show this help message and exit
  --dir DIR             A directory with files containing JSON MISP events.
  --collection COLLECTION
                        A URL to a TAXII 2.1 collection to push content to
  --log_level {fatal,error,warning,info,debug}
                        Set logging verbosity level. Default: warning

This tool reads all files from the given directory and assumes they contain JSON, not just those named as *.json.
```

The interface is a bit different from what you presented: instead of `--api_root`, it uses `--collection`.  That's because we need the URL to a TAXII collection as a target for pushing STIX content.  Script output is done via the logging system, so I also added a commandline parameter to control output verbosity.  It uses the standard StreamHandler, which logs messages to stderr.

A TAXII API root includes a restriction on the maximum HTTP request (body) size it will accept, in bytes.  One driver in the design of this system is the potentially large quantity of MISP content which may be pushed at once.  Therefore, it seemed prudent to query and honor the API root's max content length restriction, to avoid failures.  That means the script will generally need to re-segment the STIX content to ensure each post doesn't exceed the limit.

Re-segmenting the STIX content decouples the HTTP requests from the way the MISP content was split amongst the files: a large file may be pushed in several parts.  The way it was implemented also has an "inverse" consequence: converted content from multiple MISP files may be coalesced into a single HTTP request.  So there can be both splits and merges.  The script tries to build up as much content as it can without exceeding the limit, so each request winds up being maximally sized.

It introduces a new failure mode as well: a STIX object may be so large that it doesn't fit within the max content length restriction by itself.  In that case, it is impossible to push the object.  If that happens, the script currently logs an error and skips the object.  I hope max content lengths are large enough and STIX objects small enough that this is unlikely to occur.

I have not done integration testing with MISP, but since it's a standalone script, it can be run by itself to test it.  So I've only done some unit tests.  Any bugs or problems, feel free to report them.  Or questions or comments too.

### DIFF
--- a/app/files/scripts/taxii/taxii_push.py
+++ b/app/files/scripts/taxii/taxii_push.py
@@ -99,9 +99,8 @@ def parse_args():
     )
 
     parser.add_argument(
-        "--api_root",
-        help="A URL to a TAXII 2.1 collection to push content to (we"
-             " need a collection URL at minimum, not an API root)",
+        "--collection",
+        help="A URL to a TAXII 2.1 collection to push content to",
         required=True
     )
 
@@ -407,7 +406,7 @@ def main():
 
     try:
 
-        push_content(args.dir, args.api_root)
+        push_content(args.dir, args.collection)
 
     except Exception:
         log.fatal(
--- a/app/files/scripts/taxii/taxii_push.py
+++ b/app/files/scripts/taxii/taxii_push.py
@@ -1,0 +1,425 @@
+"""
+Read MISP JSON content from files in a directory, convert it to STIX, and
+push the content to a TAXII server.
+"""
+import argparse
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks a lot, lovely!

Issue #2917, opened 2018-02-09
User account Roles with "Tag Editor" permission cannot edit tags except for admin


darienhuss:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | ubuntu
| OS version (client)       | OS Independent (Linux, Windows, OS X)
| PHP version               | 7.0
| MISP version / git hash   | 2.4.87
| Browser                   | Firefox, Chrome, IE

### Expected behavior
User accounts with the "Tag Editor" permission should be able to edit tags on the List Tags page (I think?)

### Actual behavior
User accounts with the "Tag Editor" permission cannot edit tags, including their own. For instance, user accounts assigned to the default role "Researcher" cannot edit tags.

### Steps to reproduce the behavior
1. Install MISP
2. Add user account with "Researcher" roll
3. Log in to Researcher account
4. Create a new tag
5. View "List Tags" page
6. Actions on the far right of the page (View Graph, Edit tag, Delete tag) are all missing on our MISP instance

adulau:
Indeed. It's a bug. We need to fix it. Thank you!

iwitz:
We have the same issue on MISP 2.4.105 on RHEL7. Clients use Firefox and Chrome exclusively and are all affected.
It also affects template edition.

github-germ:
2.4.114

could the same root cause be causing `ExpandedPyMisp.update_tag` connected with an authkey of a user with a role which has `Tag Editor Granted` be causing this error?

```
20190913-18:16:55 DEBUG Starting new HTTPS connection (1): localhost
20190913-18:16:55 DEBUG https://localhost:443 "POST /tags/edit/1386 HTTP/1.1" 405 157
ERROR [aping.py:2039 - _check_response() ] Something went wrong (405): {'name': 'You do not have permission to use this functionality.', 'message': 'You do not have permission to use this functionality.', 'url': '/tags/edit/1386'}
20190913-18:16:55 ERROR Something went wrong (405): {'name': 'You do not have permission to use this functionality.', 'message': 'You do not have permission to use this functionality.', 'url': '/tags/edit/1386'}
```

Same code using authkey of a user with an `Admin` role updates that same tag without an error.

github-germ:
Also, when using `PyMISP.edit_tag` with the same code and authkey as above, the tag revision succeeded without error.

[...]

PR #8170, opened 2022-02-25
fix: Allow Tag_Editors to edit tags

marjatech:
#### What does it do?
Allows users with permisssion Tag_Editor to not only create but also edit Tags.

#### Questions


3c7:
Edit: just saw the issue related to this #2917.

cvandeplas:
Thank you @3c7  . For now this PR cannot be merged, as explained in #2917 allowing tags to be renamed  is not a good idea.
However some other features could be useful such as changing the name of the permission and allowing the permission to hide/unhide existing tags.

PR #8172, opened 2022-02-25
import json file from spiderfoot 

FoxSca:
Are possible integrate spiderfoot maltego torbot result on misp ?


adulau:
I would recommend to open issue on MISP not a pull-request if you have specific question. Thank you.

PR #8216, opened 2022-03-17
Update OidcAuth readme

3c7:
#### What does it do?
Replaced required dependency in the readme as mentioned by @JakubOnderka.

#### Questions


adulau:
Thank you!

PR #8217, opened 2022-03-17
[chg] LinOTP error exceptions up to the ui

andurin:
#### What does it do?

Current situation would only throw a common unspecified "Internal Error Message...." when something went wrong while authenticating against LinOTP.
With this PR it throws some more meaningful error messages back to the UI - regardless of debug enabled.  

#### Questions


adulau:
Thank you!

PR #8219, opened 2022-03-17
[chg] LinOTP now with enable/disable as config feature

andurin:
#### What does it do?

Give a new LinOTP Configuration option (cake cli AND web-ui) to enable/disable LinOTP functionality.

#### Questions


iglocska:
Thanks a lot!

PR #8277, opened 2022-04-13
Add in new RestAPI parameter to filter by sharing group on Event or Attribute search

tomking2:
#### What does it do?

This adds the ability to filter events or attributes by their sharing group ID when using the RestAPI. Currently this feature is missing from the RestAPI and is a beneficial addition to allow for searching for attributes or events specific to a sharing group(s).

It uses the parameter 'sharinggroup', to match the same parameter syntax used in the `searchIndex` endpoint, so that fields are kept consistent.

Once merged, I'll also follow up with a PR in PyMISP to add the new parameter.

#### Questions

- [x] Does it require a change in the API (PyMISP for example)? (Beneficial **not** necessary)


righel:
Thanks for the contribution. Merging to `develop` branch.

PR #8292, opened 2022-04-21
fix: Enable sharing group filter for Event controller not just attribute

tomking2:
#### What does it do?

This fixes the ability to filter by sharing group on an event, as this was accidentally omitted from the previous commit. 

#### Questions


iglocska:
Thanks a lot!

PR #8331, opened 2022-05-04
chg: add index for attributes.timestamp, show index diagnostics in di…

### LINKED ISSUE #8329: Feature Request: Add attributes.timestamp database index for performance improvement

We do a lot of PyMISP client apps looking at `Attributes` qualified by their `timestamp`. Some of these apps take a very long time to process.

We see that `attributes.timestamp` is not a database index.

Testing the query time for a database query like the following we see:
**a 50 fold improvement with the index!!**
- 5 minutes: no index (with `db_version=85` now in 2.4.158)
- 5 **seconds**: with an index

```
SELECT COUNT(*) AS `count` FROM `misp`.`attributes` AS `Attribute` 
LEFT JOIN `misp`.`events` AS `Event` ON (`Attribute`.`event_id` = `Event`.`id`) 
LEFT JOIN `misp`.`objects` AS `Object` ON (`Attribute`.`object_id` = `Object`.`id`) 
WHERE ((`Attribute`.`timestamp` >= 1651331183) AND (`Attribute`.`deleted` = '0'))
```

Can the `timestamp` column in the `attributes` table be an index?

_No response_

### Additional context

_No response_

### Code of Conduct

righel:
#### What does it do?

Add index for `attributes.timestamp` column, show index diagnostics on index diagnostics page when db versions mismatch.

fixes #8329

#### Questions

### DIFF
--- a/app/Model/AppModel.php
+++ b/app/Model/AppModel.php
@@ -1686,6 +1686,7 @@ class AppModel extends Model
                 break;
             case 86:
                 $this->__addIndex('attributes', 'timestamp');
+                break;
             case 'fixNonEmptySharingGroupID':
                 $sqlArray[] = 'UPDATE `events` SET `sharing_group_id` = 0 WHERE `distribution` != 4;';
                 $sqlArray[] = 'UPDATE `attributes` SET `sharing_group_id` = 0 WHERE `distribution` != 4;';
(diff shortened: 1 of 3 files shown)


app/Model/AppModel.php
```diff
@@ -1684,6 +1684,8 @@ public function updateDatabase($command)
                 $this->__addIndex('cryptographic_keys', 'parent_type');
                 $this->__addIndex('cryptographic_keys', 'fingerprint');
                 break;
+            case 86:
+                $this->__addIndex('attributes', 'timestamp');
```
JakubOnderka, line 1688: break is missing

righel: thanks!

github-germ: much appreciated!

PR #8361, opened 2022-05-11
Optimize event view tags

righel:
#### What does it do?
For Events with a lot of Attributes or Objects and several tags applied to them `/events/view/[id]` takes a long time to load due to the repetitive database queries to fetch a Taxonomy give an a Tag name. Proposed solution caches this function for a short period to reduce the load in the database.
**Solution:**
Use Cake Cache wrapper to cache `Taxonomy:getTaxonomyForTag()`.

**Benchmark:**
Created an Event with ~45.000 Attributes, each one with 7 Tags.
No cache: ~2.6 minutes
Cache, File: ~14s
Cache, Redis: ~19s
Cache, Apcu: ~10s 
Redis adhoc cache: ~17s
Disable checks: ~8s

#### Questions

### DIFF
--- a/app/Model/Feed.php
+++ b/app/Model/Feed.php
@@ -486,7 +486,8 @@ class Feed extends AppModel
 
             if (in_array($attribute['type'], $compositeTypes, true)) {
                 $parts = explode('|', $attribute['value']);
-                list($value1, $value2) = count($parts) === 1 ? [$parts[0], null] : $parts;
+                $value1 = $parts[0];
+                $value2 = $parts[1] ?? null;
                 $parts = [$value1];
 
                 if (!in_array($attribute['type'], Attribute::PRIMARY_ONLY_CORRELATING_TYPES, true)) {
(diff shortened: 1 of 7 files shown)


app/Model/Feed.php
```diff
@@ -485,7 +485,8 @@ public function attachFeedCorrelations(array $attributes, array $user, array &$e
             }
 
             if (in_array($attribute['type'], $compositeTypes, true)) {
-                list($value1, $value2) = explode('|', $attribute['value']);
+                $parts = explode('|', $attribute['value']) ?? [null, null];
```
JakubOnderka: When this problem occurs?

righel: indeed this error was due to improper attribute values that were ingested directly in the DB, should never happen

adulau:
Thanks for 5db8fe8b4da38967b82b96a4336fa37c7c4a1cbb following our discussion. @iglocska @JakubOnderka should we merge?

PR #8367, opened 2022-05-13
BEAM SOC MISP Community added.

kdrypr:
#### What does it do?

BEAM information added.


adulau:
Thank you for the contribution. It might useful to add an email in the future if the member would like to contact you and if you want to get the request automatically.

kdrypr:
I added. Thank you @adulau

PR #8385, opened 2022-05-18
Updated instruction on how to configure Azure AD Plugin

noodlemctwoodle:
# What does it do?

Full configuration end to end for configuring Auzre AD SSO using the AadAuth plugin with explanations and images.


adulau:
Great stuff! Thank you very much for your contribution.

PR #8412, opened 2022-05-28
fix: do not fetch tag related entities

righel:
#### What does it do?

Do not fetch recursivly records related to Tags, in some scenarios the recursive query can return millions of related records, causing the script to run out of memory.

Thanks a lot @ag-michael for reporting and helping debug the issue.

#### Questions


JakubOnderka:
@righel Nice catch 👍

PR #8414, opened 2022-05-28
chg: fix and fix default feed link

Applenice:
#### What does it do?

default feed metadata: 
- https://bazaar.abuse.ch/export/txt/md5/full/
- https://urlhaus.abuse.ch/downloads/csv/
The above two links will download zip files, not CVS files, resulting in not pulling feed data properly, I modified them to recent data links and added the recent data links for threatfox.

Also standardized the case of provider names.


adulau:
Thank you very much!

PR #8417, opened 2022-05-29
new: [UI] Use chosen for query builder

JakubOnderka:
#### Questions


adulau:
Do you have any specific reason to use minified version of the library? IMHO, I prefer to keep a parseable version especially for diff and like.

PR #8422, opened 2022-05-30
Fixed a logic error in Tag::quickEdit()

obert01:
Tag::quickEdit(): Fixed a logic error in this method that was causing tags to always be set to "local_only", wherever not intended.

I found this issue because after calling pymisp.enable_taxonomy_tags(), all my tags were systematically changed to local_only.


adulau:
Nice catch!

PR #8464, opened 2022-06-28
chg: [internal] Remove old way for putting API key to rest search

JakubOnderka:
#### Questions


adulau:
After a quick discussion with @iglocska are the others settings which could be in the session still kept in some ways?

Issue #8403, opened 2022-05-24
Support: Azure AD authentication


LOMA-TheCollectiveConsulting:
### Support Questions

Hi

I'm currently having some issue's getting the Azure AD authentication plugin to work properly. 

I implemented the following guide: https://github.com/MISP/MISP/tree/2.4/app/Plugin/AadAuth
I created users with the same username as their upn in Azure AD. 
They are created as a regular User.  
I have set no password when creating the MISP user, as I expect it to use the auth token provided by Azure AD. 

I have added the user to the site admin group in Azure AD. (The group that is also configured in the config file in MISP)
I assume that if I have added the user to the Site admin group in AD that the assigned permissions on the MISP instance are overwritten? As I can't create a MISP user without any permissions. 

My experience:
When I use the "login with azure AD" the authentication flows start. 
I'm forwarded the tenant Azure Ad sign-in experience, confirm with MFA. 
In Azure Ad, I can see lots of successful sign-in with the app registration I created. 

On completion, I'm being forwarded to the sign-in  page of the MISP instance with a token in the URL bar. 
But I'm not signed in. 

I have also included some error logs from the MISP vm. 

### MISP version

 v2.4.158 (2564a90dd9e43104f94cb4e5c3a53ed16ffc8907)

### Operating System

Ubuntu

### Operating System version

20.04

### PHP version

7.4.3 

### Relevant log output

```shell
error log output:
2022-05-24 16:03:07 Warning: Error received during Bearer token fetch (context).
2022-05-24 16:03:07 Warning: Warning (2): file_get_contents(https://login.microsoftonline.com/<tenantid>/oauth2/v2.0/token) [<a href='http://php.net/function.file-g
et-contents'>function.file-get-contents</a>]: failed to open stream: HTTP request failed! HTTP/1.1 400 Bad Request
 in [/var/www/MISP/app/Plugin/AadAuth/Controller/Component/Auth/AadAuthenticateAuthenticate.php, line 204]
Trace:
ErrorHandler::handleError() - APP/Lib/cakephp/lib/Cake/Error/ErrorHandler.php, line 230
file_get_contents - [internal], line ??
AadAuthenticateAuthenticate::_getUserAad() - APP/Plugin/AadAuth/Controller/Component/Auth/AadAuthenticateAuthenticate.php, line 204
AadAuthenticateAuthenticate::getUser() - APP/Plugin/AadAuth/Controller/Component/Auth/AadAuthenticateAuthenticate.php, line 143
[...]

PR #8478, opened 2022-07-06
chg: improve logging, use HttpSocket instead of file_get_contents() f…

righel:
#### What does it do?
Improve AadAuth plugin logging.

triggered by #8403

#### Questions


righel:
@cudeso : It would be awesome if you could take a look at this.

cudeso:
Hello @righel , great work! I've put a reminder in my schedule to have a look in the end of the week of the 18-Jul (no availabilities earlier) ; I'll check and provide feedback!

MarkJaroski:
I had hoped to see that this would fix authentication for installations which sit behind a forward proxy, but apparently that's not the case. It would be nice to add the proxy configuration from the config file to each instantiation of the HttpSocket, or perhaps even creating a single instance.

cudeso:
@righel For me this looks very good. The code certainly looks cleaner and the debug messages help a lot when you setup the plugin (I did a new install on a demo environment to test it from scratch). Thank you for the nice work!

righel:
Thanks a lot @cudeso for taking a look !

PR #8481, opened 2022-07-07
fix: [internal] Faster changing IDS flag for CIDR

JakubOnderka:
#### Questions


app/Model/Attribute.php
```diff
@@ -2044,7 +2044,6 @@ public function fetchAttributesSimple(array $user, array $options = array())
         $params = array(
             'conditions' => $this->buildConditions($user),
             'fields' => array(),
-            'recursive' => -1,
```
iglocska, line 2047: Is this on purpose? This can have some pretty heavy consequences...

iglocska: Never mind, I see it's a duplicate setting. All good, feel free to ignore it ;)

PR #8487, opened 2022-07-08
Styx Import Error

iRymden:
Styx import internal error caused by argument used by old python version, causes script to exit with error code

#### What does it do?
Fixes an issue when importing styx files via ui.
currently has a statement not needed in modern versions of python, causes the script to exit with error


adulau:
Thanks for the PR. I assume this was solved with a recent version of Python and the latest changes in misp-stix.

PR #8497, opened 2022-07-18
fix: [diag] Correctly set DB session errorCode

brlogan:
With the recent changes to session diagnostics in PR https://github.com/MISP/MISP/pull/8393, when using the database session handler, the test reports "Test failed" when nothing is wrong. This is because errorCode is set to 9 ("Test failed") at the start of the function, but then never changed after success. This fix sets the errorCode to 0 ("OK") after sessionCount is successfully retrieved from the DB.


righel:
Nice catch, thanks!

PR #8501, opened 2022-07-19
Working  (in our environment) version of AzureAD integration based on a patch by @righel, but with user creation and group enumeration

MarkJaroski:

#### What does it do?

#### Questions


peritz:
Is there a reason (other than the merge conflicts) why this PR hasn't been resolved more than a year after it was opened? I'd be keen on getting this integrated, saves a lot of administrative hassle 😅

tbethe:
Just like the others I also ran into this problem. Just here to let you know this is still an issue that people run into and that it is worth solving.

PR #8502, opened 2022-07-20
new: add LightPaginator when total page count is not needed/to expensive

righel:
#### What does it do?
Adds `LightPaginator` for addressing performance issues when paginating over big tables.
1. Add the following line to the Model you want to use the `LightPaginatorBehavior`.
    ```php
    public $actsAs = ['LightPaginator'];
    ```
    * This behavior will prevent `$Model::paginate()` to query the db for the `count(*)` of matching records. 
2. When adding pagination buttons in the views, use `LightPaginatorHelper` instead of the default one, for example:
    ```php
    <div class="pagination">
        <ul>
            <?php
            echo $this->LightPaginator->prev('&laquo; ' . __('previous'), array('tag' => 'li', 'escape' => false), null, array('tag' => 'li', 'class' => 'prev disabled', 'escape' => false, 'disabledTag' => 'span'));
            echo $this->LightPaginator->numbers(array('modulus' => 20, 'separator' => '', 'tag' => 'li', 'currentClass' => 'active', 'currentTag' => 'span'));
            echo $this->LightPaginator->next(__('next') . ' &raquo;', array('tag' => 'li', 'escape' => false), null, array('tag' => 'li', 'class' => 'next disabled', 'escape' => false, 'disabledTag' => 'span'));
            ?>
        </ul>
    </div>
    ```
3. If using `genericElements/IndexTable/index_table` view component, you can pass `light_paginator` if the model you are indexing has the `LightPaginatorBehavior`, example:
    ```php
    echo $this->element('genericElements/IndexTable/index_table', [
      'data' => [
          'light_paginator' => 1,
          'data' => $data
   ]);
    ```

#### Questions


iglocska:
Thanks a lot, this should make the log index usable again!

PR #8515, opened 2022-07-29
Update AttributeValidationTool.php

### LINKED ISSUE #8514: Bug: E-mail attributes containing a space are accepted as valid e-mail addresses
### Expected behavior

Getting the message "Email address has an invalid format. Please double check the value or select type "other"" when adding an attribute 'blah blah@blah.com' as email-src

### Actual behavior

An attribute with 'blah blah@blah.com' is stored as email-src without problems.

### Steps to reproduce

Create an attribute 'blah blah@blah.com'.

The preg in https://github.com/MISP/MISP/blob/2.4/app/Lib/Tools/AttributeValidationTool.php#L413 is not correct.
Instead of the existing `preg_match("#^.*\@.*\..*$#i", $value)` this should be

```
if (preg_match("#^.[^\s]*\@.*\..*$#i", $value)) {
   return true;
}
return __('Email address has an invalid format. Please double check the value or select type "other".');
```

### Version

2.4.159

### Operating System

Ubuntu

### Operating System version

20.04

### PHP version

7.4

### Browser

_No response_

### Browser version

_No response_

### Relevant log output

_No response_

### Extra attachments

_No response_

### Code of Conduct

cudeso:
#### What does it do?

Fixes https://github.com/MISP/MISP/issues/8514

#### Questions

### DIFF
--- a/app/Lib/Tools/AttributeValidationTool.php
+++ b/app/Lib/Tools/AttributeValidationTool.php
@@ -407,7 +407,7 @@ class AttributeValidationTool
             case 'dns-soa-email':
             case 'jabber-id':
                 // we don't use the native function to prevent issues with partial email addresses
-                if (preg_match("#^.*\@.*\..*$#i", $value)) {
+                if (preg_match("#^.[^\s]*\@.*\..*$#i", $value)) {
                     return true;
                 }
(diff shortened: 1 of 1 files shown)


adulau:
Nice catch! Thank you!

PR #8519, opened 2022-08-01
new: [internal] Add option to log last API request

tomking2:
#### What does it do?

This PR adds in functionality to allow users to optionally turn on storing of the last time an API request was made against an account, similar to the last_login field.

This adds a new field under DB version 87 for `last_api_access`, updating each time a successful API authentication request occurs, and can be configured from within the MISP settings.

This is a useful change for us (and I assume others), as it allows you to get a better understanding of the activity of all user accounts, not just those that login on the web.

#### Questions

- [X] Are you using it in production? (Not yet - plan to!)

### DIFF
--- a/app/Config/config.default.php
+++ b/app/Config/config.default.php
@@ -71,6 +71,7 @@ $config = array(
         'enableOrgBlocklisting'          => true,
         'log_client_ip'                  => false,
         'log_auth'                       => false,
+        'store_api_access_time'          => false,
         'disableUserSelfManagement'      => false,
         'disable_user_login_change'      => false,
         'disable_user_password_change'   => false,
--- a/app/Controller/AppController.php
+++ b/app/Controller/AppController.php
@@ -434,6 +434,10 @@ class AppController extends Controller
                         );
                         $this->Log->save($log);
                     }
+                    $storeAPITime = Configure::read('MISP.store_api_access_time');
+                    if (!empty($storeAPITime) && $storeAPITime) {
+                        $this->User->updateAPIAccessTime($user);
+                    }
                     $this->Session->renew();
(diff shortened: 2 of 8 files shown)


github-germ:
+1

tomking2:
Not sure why the tests are failing, complaining that the new field in the DB is missing (maybe the DB upgrade hasn't happened in the tests?) New functionality confirmed to work locally.

Any idea @iglocska?

iglocska:
I've had a quick look: 

- new db changes
```
            case 86:
                $sqlArray[] = sprintf("ALTER table users MODIFY description text;");
                break;
            case 87:
                $sqlArray[] = "ALTER TABLE users ADD `last_login` INT(11) DEFAULT 0;";
```

Field that the code relies on: `last_api_access`

https://github.com/MISP/MISP/pull/8519/files#diff-34468d89c8edd011225c0d3577502650bf96ed235857dff317c97cc203f64d97R1382

Make sure you fix the update script ;)

JakubOnderka:
Hello, the similar thing can be done by setting `MISP.log_user_ips` and `MISP.log_user_ips_authkeys` to `true`. Then when you have enabled advanced authkeys, every key access will be logged to Redis and available from web interface.

tomking2:
@iglocska Doh! I must've replaced the upgrade, copying across the field from a previous example and not replaced the column! Thank you! All sorted :crossed_fingers: 

@JakubOnderka thanks for the tip, wasn't aware of that! I think this still has merit, as this provides a quick way to get the last activity through querying the MISP users, instead of having to look through the UI or logs.

iglocska:
Thanks a lot, merged into develop!

PR #8524, opened 2022-08-08
Default correlation cleanup

JakubOnderka:
#### Questions


iglocska:
Hi @JakubOnderka - nice changes, good call on dropping all those indexes, initially those were queried rather than looped on after the fetch, so that's a lovely catch!

iglocska:
Ended up doing some changes too, sorry for diverging.

Issue #6379, opened 2020-10-02
Email attribute leads to duplicate SIDs in NIDS export


baderj:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Bug
| OS version (server)       | Ubuntu 20.04
| OS version (client)       | Ubuntu 20.04
| PHP version               | 5.6
| MISP version / git hash   | 2.4.132, 90517f591b6f37960e78dcdf7d4c6245baf78821
| Browser                   | no applicable

### Expected behavior
The NIDS rules export should generate distinct ``sid``s for all rules.

### Actual behavior
An attribute can lead to multiple rules. For example ``network-activity`` / ``email``, will produce a rule for *Source Email Address* and one for *Destination Email Address*. However, both rules have the same ``sid``. This leads to an error like this:
```
ERROR: Rule/Rev combination "153612345 1" count: 2
```

### Steps to reproduce the behavior
- create a new event
- add an attribute with category ``network-activity`` and type ``email``
- add an email as value ``mail@example.com``
- set the IDS flag on the attribute
- generate rules using the REST API for the NIDS export.
- there will be two rules for the email address (correct), but they share the same ``sid``.

### Logs, screenshots, configuration dump, ...
Example export for the above event. Notice that both rules have the same ``sid``:
```
misp-test.rules:alert tcp $EXTERNAL_NET any -> $SMTP_SERVERS 25 (msg:
"MISP-test e1 [tag1,tag2,tlp:amber] Source Email Address:
mail@example.com"; flow:established,to_server;
content:"MAIL FROM|3a|"; nocase;
content:"mail@example.com"; fast_pattern; nocase;
content:"|0D 0A 0D 0A|"; within:8192; tag:session,600,seconds;
classtype:trojan-activity; sid:153612345; rev:1; priority:1;
reference:url,https://misp.example.com/events/view/1;)

misp-test.rules:alert tcp $EXTERNAL_NET any -> $SMTP_SERVERS 25 (msg:
"MISP-test e1 [tag1,tag2,tlp:amber] Destination Email
Address: mail@example.com"; flow:established,to_server;
content:"RCPT TO|3a|"; nocase; content:"mail@example.com";
fast_pattern; nocase; content:"|0D 0A 0D 0A|"; within:8192;
tag:session,600,seconds; classtype:trojan-activity; sid:153612345;
rev:1; priority:1;
```

PR #8538, opened 2022-08-11
Create unique SIDs for email attributes in NIDS export - Second Attempt

baderj:
The NIDS export creates two rules for attributes with type `email` (a *src* and *dst* rule). However, the same SID is used for both rules. Since SIDs must be unique for a ruleset, this will be logged as an error by Suricata and the rule is not loaded (see issue #6379).

This fixes the issue by incrementing the SID before creating the second email rule.

This pull request does not require a DB change or a change in API. I'm not using the fix in production, but successfully applied it to a test instance running 2.4.161.

A similar pull request had already been merged https://github.com/MISP/MISP/pull/8433, but a later overwritten by https://github.com/MISP/MISP/commit/4ab4121d053915fc32afd5585204b5e5a8a95074

### DIFF
--- a/app/Lib/Export/NidsExport.php
+++ b/app/Lib/Export/NidsExport.php
@@ -204,6 +204,7 @@ class NidsExport
                         break;
                     case 'email':
                         $this->emailSrcRule($ruleFormat, $item['Attribute'], $sid);
+			$sid++;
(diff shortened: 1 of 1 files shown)


adulau:
Oops... thank you for re-doing the PR!

baderj:
No problem at all, thanks for the quick merge.

PR #8543, opened 2022-08-16
Fix markdown formatting in INSTALL.rhel7.md

ghost:
Just a simple change to the formatting of this markdown file - there were a lot of null bytes at the end of the file that was preventing the markdown text from loading properly, at least in Firefox.


adulau:
Thank you!

PR #8548, opened 2022-08-18
fix: Undefined index: Tag in Feed.php #8547

### LINKED ISSUE #8547: Bug: Tag already added check not working on feed pull
### Expected behavior

There is a check during feed pull, if the feed tag (if configured) is already added to the imported events. This check doesn't work, the result is ok, because the Tag will be added if the check fails, but the error is logged for each event.

### Actual behavior

The feed object only includes tag_id and not the tag name, so the evaluation of `$feed['Tag']['name']` in line 1013 (`app/Model/Feed.php`) results in `Undefined index: Tag`.

### Steps to reproduce

Enable an (MISP) feed (e.g. CIRCL OSINT Feed) and define a default tag in feed settings. During the execution of `/var/www/MISP/app/Console/cake Server fetchFeed 1 all` the error is logged multiple times.

### Version

2.4.161

### Operating System

CentOS

### Operating System version

8

### PHP version

 7.4.19

### Browser

_No response_

### Browser version

_No response_

### Relevant log output

```shell
Notice Error: Undefined index: Tag in [/var/www/MISP/app/Model/Feed.php, line 1013]

2022-08-18 09:39:37 Notice: Undefined index: Tag in [/var/www/MISP/app/Model/Feed.php, line 1013]
```

### Extra attachments

_No response_

### Code of Conduct

Benni0:
#### What does it do?

Fixes: #8547

#### Questions

### DIFF
--- a/app/Model/Feed.php
+++ b/app/Model/Feed.php
@@ -1008,16 +1008,17 @@ class Feed extends AppModel
             if (!isset($event['Event']['Tag'])) {
                 $event['Event']['Tag'] = array();
             }
-            $found = false;
-            foreach ($event['Event']['Tag'] as $tag) {
-                if (strtolower($tag['name']) === strtolower($feed['Tag']['name'])) {
-                    $found = true;
-                    break;
+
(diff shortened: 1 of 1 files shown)


mokaddem:
Hello,
Thanks a bunch for the fix!
Cheers!

Issue #8516, opened 2022-07-29
Bug: AadAuth with Group Auth enabled fails if MISP group not in first 100 groups


ntnco:
### Expected behavior

User gets authenticated

### Actual behavior

User does not get authenticated

### Steps to reproduce

- Create at least 100 Azure AD groups. 
- Add another group for `misp_siteadmin` and assign it to your user.
- Enable AadAuth by following the [documentation](https://github.com/MISP/MISP/tree/2.4/app/Plugin/AadAuth).
- Try to login with `AzureAD` button and observe that it fails.

### Version

2.4.159

### Operating System

Ubuntu

### Operating System version

20.04

### PHP version

7.4.3

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ntnco:
Thanks to @righel 's [logging improvements](https://github.com/MISP/MISP/commit/8b7b53975d45899ede6560633acbfcd287d46a26), I was able to pinpoint that the 100+ groups were causing the bug, because MS Graph's results are paginated.

I was able to fix the bug by implementing a simple while loop. I will submit a PR

righel:
thanks for the fix, https://github.com/MISP/MISP/pull/8552 was merged.

PR #8552, opened 2022-08-19
fix: traverse paginated Aad Roles

ntnco:
Only the first 100 Roles appear on the first page of Roles given by MS Graph. Roles beyond 100th were missed as seen in issue `#8516`

#### What does it do?

fixes `#8516`

#### Questions


adulau:
Just some notes about the pagination aspect from Microsoft: https://docs.microsoft.com/en-us/graph/paging just wondering if something goes wrong on the API return.

righel:
looks good to me.

adulau:
Thank you!

PR #8560, opened 2022-08-30
fix: Removes unnecessary escape character

00willo:
Fixes a validation error - found unknown escape character

#### What does it do?

This fixes an issue identified in https://github.com/Shuffle/openapi-apps/issues/64

#### Questions


adulau:
Thank you very much. and thanks for the fix on the other repos ;-)

PR #8570, opened 2022-09-07
Allow new authentication keys to be replaced

ostefano:
#### What does it do?

Allow new authentication keys to be replaced rather than just reset.
This means that it is now possible to run the following command:
```
cake user change_authkey 1 <new auth_key>
```

#### Questions

### DIFF
--- a/app/Console/Command/UserShell.php
+++ b/app/Console/Command/UserShell.php
@@ -78,6 +78,7 @@ class UserShell extends AppShell
             'parser' => [
                 'arguments' => [
                     'userId' => ['help' => __('User ID or e-mail address.'), 'required' => true],
+                    'authKey' => ['help' => __('Optional new authentication key.'), 'required' => false],
                 ],
             ],
         ]);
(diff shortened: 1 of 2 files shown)


ostefano:
Note that I erred on the side of reducing the duplication of code, and I didn't create new methods but just added some optional parameters to existing ones. 

Let me know if you prefer to duplicate the two affected functions.

ostefano:
@iglocska @Kagee PR is now ready for review.

Tested on my installation with a client using pyMISP.

adulau:
Security wise, I'm not a big fan of this. This open the door of weak or hardcoded API key which was something we wanted to avoid.

iglocska:
It is for a specific use-case of automatic re-deployments of the same instance in dockerised environments from what I understood. Also, the default behaviour is maintained.

ostefano:
> It is for a specific use-case of automatic re-deployments of the same instance in dockerised environments from what I understood. Also, the default behaviour is maintained.

That would be exact. Secrets are managed separately.

adulau:
> That would be exact. Secrets are managed separately.

I hope that the generation of secret is strong enough ;-)

iglocska:
The problem that @adulau has with this that it does give folks some rope to hang themselves with - his worry is that we'll see more and more guides/deployments simply reusing hard coded keys because our industry at times makes you want to curl up into a ball in a corner and cry.

ostefano:
Hi @adulau , 

the issue is that we want to use a MISP instance in a semi-production environment deployed using ansible and docker.
Secrets management is a well defined functionality when deploying things at scale or in infrastructures with interdependent services. The fact that the user is not involved in the provisioning of those secrets is actually what makes it easy to handle secrets with arbitrary complexity.

Is there anything that we can do to make this PR go ahead?
Maybe adding some additional warnings?

PR #8575, opened 2022-09-08
[feature] Periodic notifications

mokaddem:
# New feature: Periodic notifications
- Added periodic notification setting interface to let users subscribe to the notification period of their choice (daily, weekly or monthly), as well as the filtering to be applied on collected events
- Added new page showing a summary of the compiled data for the chosen period
- Added CLI tool allowing to simply send out reports to users that subscribed to a period
- Added possibility to override / improve part of the report by extending the base template enabling full customization for each periods

# Other changes and improvement
- New `/events/restSearch` parameter: `event_tags` allowing to retrieve only the events matching the provided tags
- Slightly Improved generation and UI of `context` restsearch export
- Added new tool to compute trends on tags based on the provided rolling window settings
- Possibility to attach `base_score` to an event only based on its directly attached tags (parameter `includeBaseScoresOnEvent`)
- Added `picker` support for `genericForm` dropdown field
- Added `genericForm` `tagsPicker` field
- Added `static` option for `galaxy_matrix` restsearch export to generate JS-free HTML output


app/Console/Command/ServerShell.php
```diff
@@ -578,6 +578,37 @@ public function enqueuePush()
...
+            }
+        }
+        echo __('All reports sent. Task took %s secondes', time() -  $start_time) . PHP_EOL;
+    }
+
+    private function __getPeriodsForToday(): array
+    {
+        $today = new DateTime();
+        $periods = ['daily'];
+        if ($today->format('j') == 1) {
```
iglocska, line 603: For the future, perhaps it would be nice to have the days configurable. Nothing we need urgently though

app/View/Users/admin_edit.ctp
```diff
@@ -96,6 +96,12 @@
         echo $this->Form->input('disabled', array('type' => 'checkbox', 'label' => __('Immediately disable this user account')));
         echo '</div>';
     ?>
+    <h5><?= __('Subscribe to the following notification periods:') ?></h5>
```
iglocska, line 99: This should also be part of the Users/edit.ctp

app/View/Users/notification_settings.ctp
```diff
@@ -0,0 +1,87 @@
+<?php
```
iglocska, line 1: What's with the sad field widths? :(

iglocska: image

PR #8576, opened 2022-09-08
[fix] Cleans-up UUID duplicates and add unique constraints

mokaddem:
# Changes
- Added functionality to clean-up any UUID duplicates and then enforce unique constraint on the following models:
  - Event
  - Attribute
  - Object
  - Sighting
  - Dashboard
  - Inbox
  - TagCollection
- Deletions are log (verbosely if needed)
- The clean-up and alter table automatically apply once the update starts


app/Model/AppModel.php
```diff
@@ -2180,6 +2193,18 @@ private function __addIndex($table, $field, $length = null, $unique = false)
         return $additionResult;
     }
 
+    private function __checkIndexExists($table, $column_name, $is_unique = false): bool
+    {
+        $query = sprintf(
```
iglocska, line 2198: This could be dangerous in the future. Not an issue in this context, but we need to make sure that column_name and table name are not arbitrarily entered ever (unless we move the function to use prepared statements)

PR #8578, opened 2022-09-12
Minor improvements

LM-CT:
#### What does it do?

Adds support to adding multiple clusters via API using `attachMultipleClusters` and a minor fix in the `sub.py` script in the misp-zmq tools to cleanly exit after a KeyboardInterrupt by the user.

#### Questions


LM-CT:
Changed `.gitignore` to ignore gzip compressed cache (`*.cache.gz`), tags (`*.etag`) and loose **files** in `app/tmp/` (`!/app/tmp/*/`).

LM-CT:
Sorry for the messy git history, I'm new to this _dev_ thing... 😁

Issue #8616, opened 2022-09-26
Support: Split frontend/background workers


pburgisser:
### Support Questions

Hi folks,

In a fork of the misp-docker from CoolAcid, I'm trying the split the workers from the frontend.
From version 2.4.151, using the SimpleBackgroundJobs configuration, would it possible to run the workers away from the rest of the frontend ?

I already tried but I'm getting PHP errors (see below)
Thanks for your help

### MISP version

2.4.161

### Operating System

Ubuntu

### Operating System version

bullseys

### PHP version

7.4

### Browser

Chrome

### Relevant log output

```shell
2022-09-26 20:44:18 Error: [TypeError] Return value of Supervisor\Process::offsetGet() must be an instance of Supervisor\mixed, string returned
Request URL: /servers/serverSettings
Stack Trace:
#0 /var/www/MISP/app/Lib/Tools/BackgroundJobsTool.php(294): Supervisor\Process->offsetGet()
#1 /var/www/MISP/app/Model/Server.php(7565): BackgroundJobsTool->getWorkers()
#2 /var/www/MISP/app/Model/Server.php(3559): Server->getWorkers()
#3 /var/www/MISP/app/Controller/ServersController.php(1204): Server->workerDiagnostics()
#4 [internal function]: ServersController->serverSettings()
#5 /var/www/MISP/app/Lib/cakephp/lib/Cake/Controller/Controller.php(499): ReflectionMethod->invokeArgs()
#6 /var/www/MISP/app/Lib/cakephp/lib/Cake/Routing/Dispatcher.php(193): Controller->invokeAction()
#7 /var/www/MISP/app/Lib/cakephp/lib/Cake/Routing/Dispatcher.php(167): Dispatcher->_invoke()
#8 /var/www/MISP/app/webroot/index.php(99): Dispatcher->dispatch()
#9 {main}
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

pburgisser:
I have the same issue when running both the frontend and the workers within the same container.

righel:
Yes, it should be possible to move the workers to a backend.

The error you are getting is strange, could it be that you have installed  supervisor:^5.0 ? version 5 is for PHP 8, won't work on PHP 7.4. It should be `supervisorphp/supervisor:^4.0` (< 5.0)
 
[...]

PR #8651, opened 2022-10-10
chg: [jobs] Store job data in Redis when SimpleBackgroundJobs are enabled

JakubOnderka:
Will help to solve #8616

#### Questions

### DIFF
--- a/app/Console/Command/AppShell.php
+++ b/app/Console/Command/AppShell.php
@@ -17,6 +17,7 @@
  */
 
 App::uses('AppModel', 'Model');
+App::uses('BackgroundJobsTool', 'Tools');
 
 /**
  * Application Shell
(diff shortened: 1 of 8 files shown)


pburgisser:
When this merge will available in the MISP ?

hodgigre:
Hi @pburgisser and all.  I'm not much of a dev so I hope this is not a misdirected question.  Is this issue still outstanding or has it perhaps been resolved in some other change?  In short, thank you so much for all the efforts here and in this somewhat dated but informative post. [https://dev.to/pburgisser/misp-at-scale-on-kubernetes-2k31](https://dev.to/pburgisser/misp-at-scale-on-kubernetes-2k31)  Have some or all of the issues identified in running MISP in containers with horizontal scaling been overcome?  I've found some other projects running MISP on EKS.  I am interested in running on GKE.  Any insight into the general current version readiness of the https://github.com/MISP/misp-docker to run with horizontal scaling would be appreciated (to perhaps understand a little better what dev knowledge/effort might be required in any attempt to address), as would sharing any knowledge of work on GKE if you were aware of the same.  Thanks!!!

PR #8709, opened 2022-10-26
Update background-jobs-migration-guide.md

lnfernux:
Update config guidance to account for an issue where workers will error instantly: https://github.com/Supervisor/supervisor/issues/212

#### What does it do?

#### Questions

### DIFF
--- a/docs/background-jobs-migration-guide.md
+++ b/docs/background-jobs-migration-guide.md
@@ -54,6 +54,7 @@ Run on your MISP instance the following commands.
     stdout_logfile=/var/www/MISP/app/tmp/logs/misp-workers.log
     directory=/var/www/MISP
     user=www-data
+    startsecs = 0
 
     [program:prio]
     directory=/var/www/MISP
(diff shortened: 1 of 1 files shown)


righel:
Hello,
This should not be necessary and can cause other issues, as adding `startsecs = 0`, will cause to endlessly try to restart the processes if there is an error with the cake php worker scripts (such as missing dependencies).

If supervisor shows the processes as in an erroneous state there must be something else going on. I suggest you to check the supervisor/programs logs.

Did this fix your issue? Are you sure your background jobs are working after this change?

lnfernux:
> Hello, This should not be necessary and can cause other issues, as adding `startsecs = 0`, will cause to endlessly try to restart the processes if there is an error with the cake php worker scripts (such as missing dependencies).
> 
> If supervisor shows the processes as in an erroneous state there must be something else going on. I suggest you to check the supervisor/programs logs.
> 
> Did this fix your issue? Are you sure your background jobs are working after this change?

Yes, this fixed my issue. All my workers are running and have been since last reboot as far as I can tell and all background jobs are completing. Without this setting our workers would be stuck in a boot/error loop.

PR #8743, opened 2022-11-08
fix: typo in exists (does not exists => does not exist)

CriimBow:
#### What does it do?

This fix corrects a spelling error throughout the code (commands, controllers, locales, etc.).

#### Questions


iglocska:
Thanks a lot!

PR #8749, opened 2022-11-11
new: [logging] Access log in database

JakubOnderka:
#### Questions


github-germ:
This looks interesting as we ae required to monitor MISP user access and provide metrics. We do have a PyMISP client scraping the `logs` table which is very slow since there are no indexes, etc. Looking forward to learning more. Thanks.

JakubOnderka:
@github-germ Merged to develop, so you can test it and provide feedback.

PR #8751, opened 2022-11-14
new: [UI] Add ability to disable discussion

JakubOnderka:
#### Questions


Wachizungu:
A bit late (just saw it in changelog) but:
Thanks a lot for this <3, it will help us quite a bit to have this enabled on one of our public instances!

PR #8812, opened 2022-12-09
Update correlations.ctp

StefanKelm:
tiny typo ;-)


adulau:
Thank you! nice catch.

PR #8870, opened 2023-01-26
chg: [internal] allow site admins ability to view event_creator_email for all events in export

goodlandsecurity:
#### What does it do?

The comment on line 2097 states:

> Include information about event creator user email. This information is included for: 
>  - users from event creator org
>  - site admins
>  In export, this information will be included in `event_creator_email` field just for auditors of event creator org.

`$event['Event']['event_creator_email']` is not set for site admins, only for users from the same org that have a role assigned with audit permission. 

This change will allow a site admin the ability to gather more granular metrics on which users are actively sharing to a community MISP instance via the MISP API.

#### Questions

### DIFF
--- a/app/Model/Event.php
+++ b/app/Model/Event.php
@@ -2097,7 +2097,7 @@ class Event extends AppModel
             // Include information about event creator user email. This information is included for:
             // - users from event creator org
             // - site admins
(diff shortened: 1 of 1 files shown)


iglocska:
I don't see why not, good idea, thanks!

PR #8946, opened 2023-03-04
new: [ApacheAuthenticate] Add STARTTLS support for LDAP connection

Kagee:
Controlled by setting `ApacheSecureAuth.starttls`. Default (`ApacheSecureAuth.starttls undefined`) is `false`, since it is a new feature.

config.default.php is updated with `ApacheSecureAuth.starttls = true` as default and extra explanations.

### DIFF
--- a/app/Controller/Component/Auth/ApacheAuthenticate.php
+++ b/app/Controller/Component/Auth/ApacheAuthenticate.php
@@ -38,10 +38,10 @@ class ApacheAuthenticate extends BaseAuthenticate
         }
         return $returnCode;
     }
-	
+
     private function getEmailAddress($ldapEmailField, $ldapUserData)
     {
-	// return the email address of an LDAP user if one of the fields in $ldapEmaiLField exists
+        // return the email address of an LDAP user if one of the fields in $ldapEmaiLField exists
         foreach($ldapEmailField as $field) {
             if (isset($ldapUserData[0][$field][0])) {
                 return $ldapUserData[0][$field][0];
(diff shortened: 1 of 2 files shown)


Kagee:
The failed build tests does not appear to relate to my changes, they looks like a error in the test setup?

Kagee:
To make something a bit clearer: ApacheSecureAuth supported LDAPS (LDAP over TLS) from before. However, the only combination of LDAP and TLS that is an official standard is LDAP with STARTTLS.

app/Controller/Component/Auth/ApacheAuthenticate.php
```diff
@@ -73,6 +73,13 @@ public function authenticate(CakeRequest $request, CakeResponse $response)
         ldap_set_option($ldapconn, LDAP_OPT_PROTOCOL_VERSION, Configure::read('ApacheSecureAuth.ldapProtocol'));
         ldap_set_option($ldapconn, LDAP_OPT_REFERRALS, Configure::read('ApacheSecureAuth.ldapAllowReferrals', true));
 
+        if (Configure::read('ApacheSecureAuth.starttls', false) === true) {
```
mokaddem: I'd use
```php
if (!empty(Configure::read('ApacheSecureAuth.starttls'))) {
```
If someone configure the entry like this
```php
'ApacheSecureAuth.starttls' => 1
```
It will be parsed as false

Kagee: But, then you would not be able to (technically) turn off starttls without removing the setting?

Kagee: But i should change from === to ==, then both true, 1, "-1", etc would work. https://www.php.net/manual/en/types.comparisons.php

Kagee: Force pushed the above change.

Kagee:
Rebased on current 2.4 tip.

PR #8948, opened 2023-03-05
[new]: [ApacheSecureAuth]: Logout support endpoint and minimal documentation

Kagee:
If using ApacheSecureAuth and capturing the user's credentials using HTTP Basic Auth, it can be difficult to make
the browser forget these. 

There is no common or properly defined way of "logging out" after logging in with HTTP Basic Auth.

If the user presses the GUI "Log out" link, this can result in a logout-login loop, where the user
is logged out, but then immediately loggged back in by means of the browsers cached HTTP Basic Auth
credentials. This can be observed when a user presses "Log out", for then to be returned to the
events view with two flash messages - one about a successful logout, and one "Welcome back" login-message.

This PR contains a logout endpoint (`/users/logout401`) and documentation that can help resolve this logout problem.

#### Questions

### DIFF
--- a/app/View/Users/logout401.ctp
+++ b/app/View/Users/logout401.ctp
@@ -1,0 +1,41 @@
+<div style="width:100%;">
+    <script>
+        // Chrome/Edge will log you out once it sees the HTTP401.
+        // We need to be extra hacky to properly log out on i.e. Firefox.
+        <?php
+        $split_baseurl = array();
+        # We split the baseurl, since we need to add data between the
+        # schema and hostname. We *could* use parse_url here, but then
+        # we would need a lot of code to rebuild it
+        if (preg_match("/(https?:\/\/)(.*)/", $baseurl, $split_baseurl)):
+        ?>
+            let logoutxhr401 = new XMLHttpRequest()
+            logoutxhr401.open("GET", "<?php echo $split_baseurl[1]; ?>logout:@<?php echo $split_baseurl[2]; ?>/users/login")
+            logoutxhr401.send()
+        <?php
+        else:
+          echo "// We failed to parse baseurl";
+        endif;
+        ?>
+    </script>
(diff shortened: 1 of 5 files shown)


app/View/Users/logout401.ctp
```diff
@@ -0,0 +1,41 @@
...
+        // We need to be extra hacky to properly log out on i.e. Firefox.
+        <?php
+        $split_baseurl = array();
+        # We split the baseurl, since we need to add data between the
+        # schema and hostname. We *could* use parse_url here, but then
+        # we would need a lot of code to rebuild it
+        if (preg_match("/(https?:\/\/)(.*)/", $baseurl, $split_baseurl)):
+        ?>
+            let logoutxhr401 = new XMLHttpRequest()
+            logoutxhr401.open("GET", "<?php echo $split_baseurl[1]; ?>logout:@<?php echo $split_baseurl[2]; ?>/users/login")
```
mokaddem, line 19: I'm not sure I understand what's going on here. Do you mind to explain why you need to perform a self call that would look like this `https://logout:@misp-hostname.org/users/login` ?
Is LDAP relying on the additional `logout:@` data?

mokaddem: Also, this operation could probably be done from the controller function `logout401` instead of the view.

Kagee: This is required, in combination with the webserver setup in the accompanying docs, to properly log out Firefox. This is in the view, and as Javascript since the call has to be done *in the users browser* to log out the browser, and not serverside. I will update the commends to make this more clear.

Kagee: Force pushed code with better inline docs and explanation.

Kagee:
Rebased on current 2.4 tip.

PR #8965, opened 2023-03-19
Feature/api log and pin ip

cvandeplas:
#### What does it do?

[x] Logs the unique IPs used with API keys and shows them to the user
[x] Update the SQL db schema  (default + change on upgrade)
[x] With the press of a button on one of the IPs, "pin" the selected IP as "allowed_ip" 

#### Questions

- [x] Does it require a DB change?  Yes, additional field in the `AuthKeys` table

#### Screenshots
The AuthKeys index page, with the Pin icon to assign the IP with a click of a button.
image

The view page:
image
 
(updated on 23 March: with the Pin icon)

### DIFF
--- a/INSTALL/MYSQL.sql
+++ b/INSTALL/MYSQL.sql
@@ -88,6 +88,12 @@ CREATE TABLE IF NOT EXISTS `attribute_tags` (
   INDEX `tag_id` (`tag_id`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
+-- -------------------------------------------------------
+
+--
+-- Table structure for table `auth_keys`
+--
+
 CREATE TABLE IF NOT EXISTS `auth_keys` (
     `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
     `uuid` varchar(40) COLLATE utf8mb4_unicode_ci NOT NULL,
@@ -98,6 +104,8 @@ CREATE TABLE IF NOT EXISTS `auth_keys` (
     `expiration` int(10) unsigned NOT NULL,
     `user_id` int(10) unsigned NOT NULL,
     `comment` text COLLATE utf8mb4_unicode_ci,
+    `allowed_ips` text COLLATE utf8mb4_unicode_ci,
+    `unique_ips` text COLLATE utf8mb4_unicode_ci,
     PRIMARY KEY (`id`),
     KEY `authkey_start` (`authkey_start`),
     KEY `authkey_end` (`authkey_end`),
--- a/app/Model/AuthKey.php
+++ b/app/Model/AuthKey.php
@@ -104,6 +104,12 @@ class AuthKey extends AppModel
             if (isset($val['AuthKey']['allowed_ips'])) {
                 $results[$key]['AuthKey']['allowed_ips'] = JsonTool::decode($val['AuthKey']['allowed_ips']);
             }
+            if (isset($val['AuthKey']['unique_ips'])) {
+                $results[$key]['AuthKey']['unique_ips'] = JsonTool::decode($val['AuthKey']['unique_ips']);
+            } else {
+                $results[$key]['AuthKey']['unique_ips'] = [];
+            }
+            
         }
         return $results;
     }
@@ -115,6 +121,13 @@ class AuthKey extends AppModel
                 $this->data['AuthKey']['allowed_ips'] = null;
             } else {
                 $this->data['AuthKey']['allowed_ips'] = JsonTool::encode($this->data['AuthKey']['allowed_ips']);
+            }
+        }
+        if (isset($this->data['AuthKey']['unique_ips'])) {
+            if (empty($this->data['AuthKey']['unique_ips'])) {
+                $this->data['AuthKey']['unique_ips'] = null;
+            } else {
+                $this->data['AuthKey']['unique_ips'] = JsonTool::encode($this->data['AuthKey']['unique_ips']);
             }
         }
         return true;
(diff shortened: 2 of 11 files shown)


app/Model/Sighting.php
cvandeplas, line 1: Where did this come from? Please ignore this change, probably an issue due to concurrent git changes

INSTALL/MYSQL.sql
mokaddem, line 1: There is no need to update the sql file.
When you first log-in, MISP will perform all migrations starting from version `61`.
https://github.com/MISP/MISP/blob/cb3b4b43dc1ba472be72c11f809e5611caf6ba55/INSTALL/MYSQL.sql#L1475-L1476
And thus, this one will be run
https://github.com/MISP/MISP/blob/cb3b4b43dc1ba472be72c11f809e5611caf6ba55/app/Model/AppModel.php#L1947-L1948

app/Model/AuthKey.php
```diff
@@ -162,12 +175,27 @@ public function getAuthUserByAuthKey($authkey, $includeExpired = false)
...
             'conditions' => $conditions,
         ]);
         $passwordHasher = $this->getHasher();
         foreach ($possibleAuthkeys as $possibleAuthkey) {
-            if ($passwordHasher->check($authkey, $possibleAuthkey['AuthKey']['authkey'])) {
+            if ($passwordHasher->check($authkey, $possibleAuthkey['AuthKey']['authkey'])) {  // valid authkey
+                // store IP in db if not there yet
+                $remote_ip = $this->_remoteIp();
+                $update_db_ip = true;
+                if (is_array($possibleAuthkey['AuthKey']['unique_ips']) && in_array($remote_ip, $possibleAuthkey['AuthKey']['unique_ips'])) {
```
mokaddem: The check
```php
is_array($possibleAuthkey['AuthKey']['unique_ips'])
```
in combination with this in the `else` branch is a bit weird to me.
```php
$possibleAuthkey['AuthKey']['unique_ips'][] = $remote_ip;
```
Why not do directly do ?
```php
$possibleAuthkey['AuthKey']['unique_ips'] = [$remote_ip];
```

app/Model/AuthKey.php
```diff
@@ -162,12 +175,27 @@ public function getAuthUserByAuthKey($authkey, $includeExpired = false)
...
+                    $update_db_ip = false;  // IP already seen, skip saving in DB
+                } else {   // first time an IP is seen for this API key
+                    $possibleAuthkey['AuthKey']['unique_ips'][] = $remote_ip;
+                }
+                if ($update_db_ip) {
+                    // prevent double entries due to race condition
+                    $possibleAuthkey['AuthKey']['unique_ips'] = array_unique($possibleAuthkey['AuthKey']['unique_ips']);
+                    // save in db
+                    $this->save($possibleAuthkey);
```
mokaddem: This is minor but I like to make it evident which field we want to save by using the `fieldList` parameter.
For example:
```php
$this->save($possibleAuthkey, ['fieldList' => ['unique_ips']);
```

mokaddem:
In regard to your comment
> [ ] With the press of a button on one of the IPs, "pin" the selected IP as "allowed_ip"

Just to make sure I understand correctly, this should only be usable once right? Because if you add an IP to the allowedlist this way, you shouldn't have other entries than this one showing up in the seen IPs.

INSTALL/MYSQL.sql
cvandeplas: ack

app/Model/AuthKey.php
cvandeplas: I removed the `is_array()` , that's indeed not needed (anymore). Was needed before I did the current `afterFind()` who initiates the array in any case.

For the  else`: I corrected the comment "first time **this** IP is seen" 
It's to append the new IP to the array, as the array could have contained other IPs from previous connections.

app/Model/AuthKey.php
cvandeplas: excellent remark !

cvandeplas:
> In regard to your comment
> 
> > [ ] With the press of a button on one of the IPs, "pin" the selected IP as "allowed_ip"
> 
> Just to make sure I understand correctly, this should only be usable once right? Because if you add an IP to the allowedlist this way, you shouldn't have other entries than this one showing up in the seen IPs.

Yes, indeed. The idea is to motivate people to add source IP filtering as it's "just a click away".

cvandeplas:
I'll start working on the "one click pin IP" feature.

cvandeplas:
I've updated the screenshots with the "pin" icon
image

Tooltip is set, and users are prompted for confirmation.

There is however one very annoying bug related to the following line in the `AuthKeysController::pin()` function:
```
$this->AuthKey->save($authKey, ['fieldList' => ['allowed_ips']]);
```
This line seems to also log out the user.
I could not (yet) find a reason why, input is very welcome.

mokaddem:
> ```
> $this->AuthKey->save($authKey, ['fieldList' => ['allowed_ips']]);
> ```
> 
> This line seems to also log out the user. I could not (yet) find a reason why, input is very welcome.

For me, it doesn't trigger the log out when I pin an IP.
However, the reload of the user session might be the cause:
1. The user gets logged out because this function is called when the authkey is updated.
https://github.com/MISP/MISP/blob/13d99cfacac93ec81d186d29206a0834f7ac63a9/app/Model/AuthKey.php#L334-L339
2. Which in turn, call this other function because the date has been modified.
https://github.com/MISP/MISP/blob/13d99cfacac93ec81d186d29206a0834f7ac63a9/app/Controller/AppController.php#L507-L510
3. Finally, the session refresh happens.
https://github.com/MISP/MISP/blob/13d99cfacac93ec81d186d29206a0834f7ac63a9/app/Controller/AppController.php#L1373

cvandeplas:
Very weird logout issue. 
With xdebug tracing it happens much less. Very tough to locate the origin. 
Finally, I changed the PHP session management to redis, and then I am unable to reproduce the issue.
I therefore believe this is something PHP related, probably something wrong on my testing VM that causes me to get logged out.

I'll merge the PR now. Thanks again for your review @mokaddem  !

cvandeplas:
bummer, I merged to 2.4, reverting and re-doing

Issue #8974, opened 2023-03-22
Feature Request: Add user id to sys logged actions


dragsu:
### Is your feature request related to a problem? Please describe.

When a user performs an action, the user id is not logged with it.  As a result, when multiple users are doing changes in MIPS it is hard to track which user did what.
 

### Describe the solution you'd like

Prepend the user id to sys logged actions

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #8975, opened 2023-03-22
chg: [logging] Add user id to sys logged actions

dragsu:
When a user performs an action, the user id is not logged with it. As a result, when multiple users are doing changes in MIPS it is hard to track which user did what.

This change will prepend the user id to sys logged actions.

#### What does it do?

If it fixes an existing issue, please use github syntax: #8974

#### Questions

### DIFF
--- a/app/Model/AuditLog.php
+++ b/app/Model/AuditLog.php
@@ -275,7 +275,7 @@ class AuditLog extends AppModel
             }
         }
         if ($this->syslog) {
-            $entry = $data['AuditLog']['action'];
+            $entry = "User (" . $data['AuditLog']['user_id'] . ") -- " . $data['AuditLog']['action'];
             $title = $this->generateUserFriendlyTitle($data['AuditLog']);
             if ($title) {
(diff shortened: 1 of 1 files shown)


adulau:
Thanks for the PR. It's a good idea. I'm just worried about the change of format as maybe some tools rely on it.  @iglocska @mokaddem @righel what do think of it?

dragsu:
Pipeline failing due to PyMISP version mismatch.

`self.assertEqual(self.user_misp_connector.version, self.user_misp_connector.pymisp_version_master)
AssertionError: {'version': '2.4.169'} != {'version': '2.4.169.2'}`

dragsu:
@adulau Can we please revisit this?

JakubOnderka:
I would recommend new ECS logging – it generates logs in structured JSON [Elastic Common Schema](https://www.elastic.co/guide/en/ecs/current/ecs-reference.html) format, which contains much more information and also user ID and user e-mail.

Issue #5257, opened 2019-10-02
Add PyMISP argument to ignore dup indicators during ingestion with MISP server-side support


github-germ:
2.4.114

In order to increase performance in a `ExpandedPyMISP` client that is processing a large feed ingestion particularly into existing events that are large, it is more efficient to not first retrieve the entire event into the app to look for dups. Rather simple add the attribute, and deal with the error if it occurs.

I refactored two major feed ingestion clients with this approach and the processing time went for hours to minutes; so I was quite happy to deal with the error return. However, now the `logs` table is growing rather quickly.

This method does trigger an error return to the `ExpandedPyMISP` client, and also  the insertion of a row into the database `logs` table -- example below.

Might it make sense to not trigger an error in this case, possibly adding an optional boolean argument to `pymisp.add_attribute` requesting the error but defaulted to `False`.

```
MariaDB [misp]> select * from logs where id=91023522;
+----------+--------------------------------------------------------------------------------------------------+---------------------+-----------+----------+--------+---------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------+---------+-------------+-----+
| id       | title                                                                                            | created             | model     | model_id | action | user_id | change                                                                                                                                                                                                                                                                                                                                    | email                      | org     | description | ip  |
+----------+--------------------------------------------------------------------------------------------------+---------------------+-----------+----------+--------+---------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------+---------+-------------+-----+
| 91023522 | Attribute dropped due to validation for Event 41069 failed: Network activity/ip-src 154.68.5.145 | 2019-10-02 15:05:22 | Attribute |        0 | add    |      45 | Validation errors: {"value":["A similar attribute already exists for this event."]} Full Attribute: {"uuid":"f2abb7ea-5a1b-43f9-b102-79893cf0ef52","type":"ip-src","value":"154.68.5.145","category":"Network activity","to_ids":1,"disable_correlation":0,"comment":"[10,2019100215]","event_id":"41069","object_id":0,"distribution":5} | api@local.net | X-Feed |             | ::1 |
[...]

PR #8980, opened 2023-03-23
[new]: add breakOnDuplicate option to attributes/add

righel:
#### What does it do?
Fixes https://github.com/MISP/MISP/issues/5257

* Add support for a `breakOnDuplicate` named parameter on `/attributes/add` endpoint, default value is `true` which keeps the current behavior of throwing an error when the user tries to add duplicate attribute to an event.
When set to `false` the endpoint will work as an upsert, updating the attributes `timestamp` and any other properties provided in the payload, no error logs will be written.
* Minor fixes to objects breakOnDuplicate logic.

**Example 1:**
`/attributes/add/1/breakOnDuplicate:0`
**Body:**
```json
{
  "value": "10.0.0.11",
  "type": "ip-dst"
}
```

PyMISP related PR: https://github.com/MISP/PyMISP/pull/958

#### Questions

### DIFF
--- a/app/Model/Attribute.php
+++ b/app/Model/Attribute.php
@@ -736,7 +736,7 @@ class Attribute extends AppModel
             return true;
         }
 
-        $existingAttribute = $this->findAttributeByValue($fields['value']);
+        $existingAttribute = $this->findAttributeByValue($this->data['Attribute']);
 
         return empty($existingAttribute);
     }
@@ -2482,8 +2482,9 @@ class Attribute extends AppModel
         if (!isset($attribute['distribution'])) {
             $attribute['distribution'] = $this->defaultDistribution();
         }
+        $breakOnDuplicate = true;
         if (isset($params['breakOnDuplicate'])) {
-            $this->data['Attribute']['breakOnDuplicate'] = (bool)$params['breakOnDuplicate'];
+            $breakOnDuplicate = (bool)$params['breakOnDuplicate'];
         }
         $params = array(
             'fieldList' => self::CAPTURE_FIELDS,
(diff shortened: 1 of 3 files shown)


app/Model/Attribute.php
```diff
@@ -2515,6 +2498,10 @@ public function captureAttribute($attribute, $eventId, $user, $objectId = false,
                 unset($attribute['sharing_group_id']);
             }
         }
+        if (isset($attribute['breakOnDuplicate']) && $attribute['breakOnDuplicate'] === false) {
+            $this->data['Attribute'];
```
mokaddem: Doesn't seems to do anything. Delete the line?

app/Model/Attribute.php
```diff
@@ -3526,4 +3513,66 @@ private function generateTypeDefinitions()
...
+
+    /**
+     * If there is no id set but attribute value is set, check if the attribute already exists (by value) 
+     *
+     * {@inheritdoc }
+     *
+     * @param int|string $id ID of record to check for existence
+     * @return bool True if such a record exists
+     */
+    public function exists($id = null)
```
mokaddem: Why is the override needed?

app/Model/Attribute.php
```diff
@@ -3526,4 +3513,66 @@ private function generateTypeDefinitions()
...
+            return parent::exists($id);
+        }
+
+        // find existing attribute by value
+        if (empty($id) && isset($this->data['Attribute']['value'])) {
+            $existingAttribute = $this->findAttributeByValue($this->data['Attribute']['value']);
+            if (!empty($existingAttribute)) {
+                $this->id = $existingAttribute['Attribute']['id'];
+                $this->data['Attribute']['id'] = $existingAttribute['Attribute']['id'];
+                $this->data['Attribute']['uuid'] = $existingAttribute['Attribute']['uuid'];
```
mokaddem: 1. Does it still work if `timestamp`  is not set?
2. Are there other functions using `exists` that would have their behavior changed due to the override (couldn't find any this question about the override)?

app/Model/Attribute.php
righel: good catch

app/Model/Attribute.php
righel: `cakephp` uses this function to check if it's a new or existing entity, this happens in `Model::_doSave()`, before `beforeSave()` is called.
The alternative would be add additional checks to `Attribute:captureAttribute()` but this won't cover all the other places `Attribute->save()` is called directly

app/Model/Attribute.php
righel: 1. It sets a new  `timestamp`, if it's present it sets the provided value, no changes in behavior here.
2. Direct calls in `Attribute::restore()`, but `$this->id` is set so the new code won't execute (regardless the `breakOnDuplicate` value). By the framework in `Model::_doSave()` and `Model::delete()`.

PR #8989, opened 2023-03-29
Fix typo in OPTIMIZE MySQL commands

dnso86:
#### What does it do?

I believe there is a typo in both of these queries. Tested with 

- [MariaDB](https://mariadb.com/kb/en/optimize-table/) (`10.11.2`)
- [MySQL](https://dev.mysql.com/doc/refman/8.0/en/optimize-table.html) (`8.0.32`)

and they seem to only work with `OPTIMIZE TABLE ...` .

```
MariaDB [mysql]> optimise table user;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'optimise table user' at line 1
MariaDB [mysql]> optimize table user;
+------------+----------+----------+------------------------------------------+
| Table      | Op       | Msg_type | Msg_text                                 |
+------------+----------+----------+------------------------------------------+
| mysql.user | optimize | Error    | 'mysql.user' is not of type 'BASE TABLE' |
| mysql.user | optimize | status   | Operation failed                         |
+------------+----------+----------+------------------------------------------+
2 rows in set (0.000 sec)
```

```
mysql> optimize table user;
+------------+----------+----------+-------------------------------------------------------------------+
| Table      | Op       | Msg_type | Msg_text                                                          |
+------------+----------+----------+-------------------------------------------------------------------+
| mysql.user | optimize | note     | Table does not support optimize, doing recreate + analyze instead |
| mysql.user | optimize | status   | OK                                                                |
+------------+----------+----------+-------------------------------------------------------------------+
2 rows in set (0.75 sec)

mysql> optimise table user;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'optimise table user' at line 1
```

#### Questions

### DIFF
--- a/app/Console/Command/AdminShell.php
+++ b/app/Console/Command/AdminShell.php
@@ -845,7 +845,7 @@ class AdminShell extends AppShell
         ]);
 
         foreach ($tables as $table) {
-            $dataSource->query('OPTIMISE TABLE ' . $dataSource->name($table));
+            $dataSource->query('OPTIMIZE TABLE ' . $dataSource->name($table));
             $progress->increment();
             $progress->draw();
         }
--- a/app/Model/AccessLog.php
+++ b/app/Model/AccessLog.php
@@ -165,7 +165,7 @@ class AccessLog extends AppModel
         $deleted = $this->getAffectedRows();
         if ($deleted > 100) {
             $dataSource = $this->getDataSource();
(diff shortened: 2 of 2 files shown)


adulau:
Good catch! Thank you very much.

PR #9006, opened 2023-04-11
Add `startsecs=0` to worker config

alekwisnia:
In my case without this it was throwing an error: misp-workers:cache_00            FATAL     Exited too quickly (process log may have details)

#### What does it do?

#### Questions


righel:
Are you sure your workers are running OK or just displaying as healthy in the supervisor monitor?
See https://github.com/MISP/MISP/pull/8709#issuecomment-1296999836

Workers should be show as healthy with the current config.

PR #9021, opened 2023-04-15
new: [misp-workflow-modules] Telegram alert module

vincenzocaputo:
#### What does it do?

This pull requests add a new workflow action module that can be used to send message alerts to a public/private Telegram channel. 

The module extends the Webhook module in order to make the "sendMessage" request and use Jinja template rendering for the messages. Note, however, that not all HTML tags are supported by Telegram API (more details at https://core.telegram.org/bots/api#formatting-options).

#### Questions


adulau:
Thank you very much for the contribution!

PR #9056, opened 2023-05-03
fix: [internal] Warning when searchvalue is not defined

JakubOnderka:
#### Questions


github-germ:
Thx!

PR #9062, opened 2023-05-08
Run misp with postgresql

JustKiddingCode:
We are trying to setup a new instance of MISP with a postgresql database.
We will document needed changes in this PR.


baonq-me:
+1 vote

baonq-me:
@JustKiddingCode Have you used the PR in production yet?

JustKiddingCode:
> @JustKiddingCode Have you used the PR in production yet?

No. It is not possible to use MISP with postgres. Even if you set everything up, change every backtick, in the next release everything will break because the db migrations are run  are in the AppModel.php.
But a good reminder to close this pr.

Issue #9072, opened 2023-05-11
$technique and $HttpSocket undefined before use in app/Console/Command/ServerShell.php


kindtrojan:
https://github.com/MISP/MISP/blame/305b456477b449d19c291b2c9a2a2e07fe5d34d6/app/Console/Command/ServerShell.php#LL810-L811

taxii push from MISP results in "Notice" logs being seen on MISP that point to this issue.
`2023-05-11 18:33:28 Notice: Undefined variable: technique in [/var/www/MISP/app/Console/Command/ServerShell.php, line 810]
2023-05-11 18:33:28 Notice: Undefined variable: HttpSocket in [/var/www/MISP/app/Console/Command/ServerShell.php, line 810]
`

kindtrojan:
Possible fix : https://github.com/MISP/MISP/pull/9073 to alleviate this issue.
But, still doesnt fix the taxii push issues.

PR #9073, opened 2023-05-11
missing variable assignments before reference: ServerShell.php - issue : 9072

kindtrojan:
missing variable assignments before reference.

#### What does it do?

github issue: `#9072`

#### Questions

### DIFF
--- a/app/Console/Command/ServerShell.php
+++ b/app/Console/Command/ServerShell.php
@@ -800,13 +800,19 @@ class ServerShell extends AppShell
         $userId = $this->args[0];
         $user = $this->getUser($userId);
         $serverId = $this->args[1];
+        $technique = empty($this->args[2]) ? 'full' : $this->args[2]; 
         if (!empty($this->args[3])) {
             $jobId = $this->args[3];
         } else {
             $jobId = $this->Job->createJob($user, Job::WORKER_DEFAULT, 'push_taxii', 'Server: ' . $serverId, 'Pushing.');
         }
         $this->Job->read(null, $jobId);
-
+       
+        App::uses('SyncTool', 'Tools');
+        $syncTool = new SyncTool(); 
+        $server = $this->getServer($serverId);
+        $HttpSocket = $syncTool->setupHttpSocket($server); 
+        
         $result = $this->TaxiiServer->push($serverId, $technique, $jobId, $HttpSocket, $user);
 
         if ($result !== true && !is_array($result)) {


app/Console/Command/ServerShell.php
```diff
@@ -800,13 +800,19 @@ public function push_taxii()
...
         }
         $this->Job->read(null, $jobId);
-
+       
+        App::uses('SyncTool', 'Tools');
+        $syncTool = new SyncTool(); 
+        $server = $this->getServer($serverId);
+        $HttpSocket = $syncTool->setupHttpSocket($server); 
+        
         $result = $this->TaxiiServer->push($serverId, $technique, $jobId, $HttpSocket, $user);
```
VincentEggerling, line 816: Looking at `app/Model/TaxiiServer.php:push()`, I think the fix should actually be to change the arguments to 
`$result = $this->TaxiiServer->push($serverId, $user, $jobId);`

With that, I managed to make the execution of a Taxii push go further, but I'm now hitting another exception
`Error Call to a member function saveField() on null`....
EDIT: I think this is because the `Job` is not created because calling `push()` skips the call `pushRouter()` which does create the Job.

EDIT2: Actually not specifying the `$jobId` and let it default to `null` seems to work better since this will force the creation of a new Job.

kindtrojan: 'Thanks @VincentEggerling , maybe you are right. 
I was skeptical of changing the function signature.
I will take a look again.

How are you debugging the MISP run? is there a debug switch/config somewhere that I couldnt find?

VincentEggerling: You can "in live" tweak the .php files and add something like `$this->log("Some var = " . $var, LOG_WARNING);` Then have a look in `app/tmp/logs` and sort by most recently modified. You should see your log appear somewhere.
I use _warning_ level to be sure the log appear because sometimes lower level log are ignored.

With this change, I was able to push some events to my TAXII server. However you might need some more tweaks such as: 
-  https://github.com/MISP/MISP/issues/9089 . You'll need to edit manually the collection for now.
- Make sure your TAXII server advertise API-root without any path prefix. This works `http://host:port/149acdaa-dde3-4053-a296-18ac9d7b8d79`  but this doesn't  `http://host:port/someprefix/149acdaa-dde3-4053-a296-18ac9d7b8d79` . This boils down to this part https://github.com/MISP/MISP/blob/6243e03e5e972a120c02696ca3896331a4775535/app/Controller/TaxiiServersController.php#L131
- You need to `pip install misp-stix taxii2-client` on your MISP server.


You might still get errors coming from the conversion MISP->STIX2.1, currently I'm fighting with error of conversion when some GalaxyCluster are involved. (This will appear in the log `exec-errors.log`).
-  `pip install --upgrade stix2` could maybe help ?
- You can also tweak the conversion file `app/files/scripts/taxii/taxii_push.py` and add some `log.warning()`. 
- You might want to comment out the reraise part https://github.com/MISP/MISP/blob/6243e03e5e972a120c02696ca3896331a4775535/app/files/scripts/taxii/taxii_push.py#L378 so that the conversion will still proceed even if one file fail to be converted.

This is really hacky for now. For some reason MISP creates two jobs when you click the "Push to TAXII" icon. One of them fails immediately, while the other if you're lucky will end up in "Completed" but still with a grey progression bar.
I think this is because the button triggers two different execution path that both create a job. (See Administration -> Jobs)

kindtrojan: That's a very detailed explanation. Thank you very much.

I will give this a try.
I had shelved the MISP <-> TAXII integration efforts due to the bugs.

PR #9085, opened 2023-05-19
Time-based and Single Use One-time password support (TOTP / HOTP)

cvandeplas:
## What does it do?

Adds TOTP support, build in MISP in 2 modes:
- (default) optional (T/H)OTP for users (when required libraries are installed)
- (optional) mandatory (T/H)OTP for all users

When logging in the user can enter either the TOTP or the HOTP (one time paper token)
OTP attempts are also limited by the Bruteforce component. So multiple failed attempts will result in a temporary blocking.

## Screenshots
User can generate TOTP through their Profile page:
image

A QR code is generated and they need to fill in the code once to confirm all is well:
image
Then they get directed to the page containing their next 50 HOTP/paper tokens:
image

Their profile then shows they have a token, they can also check again what their paper tokens are:
image
So does the admin page (the phone icon)
image
(org)Admins can delete the secret of a user:
image

When they have their TOTP secret, after user/pass window they are prompted to enter the or the HOTP.
image

Logging is also generated:
image

The `MISP.totp_required` security setting allows enforcing TOTP for the whole MISP instance.
In this case users are invited to store their TOTP at next login. They cannot access any other page until they validated the TOTP. The server wide parameter has a beforeHook to ensure the required PHP libraries are installed, as otherwise the admin might lock themselves out.

Requires 2 additional PHP libraries to be installed through composer: 
- "spomky-labs/otphp"
- "bacon/bacon-qr-code"

## Questions

- Does it require a DB change?   Yes
- Does it require a change in the API (PyMISP for example)?  No


app/Controller/UsersController.php
```diff
@@ -588,17 +590,7 @@ public function admin_view($id = null)
             unset($user['User']['authkey']);
         }
         if ($this->_isRest()) {
-            $user['User']['password'] = '*****';
```
cvandeplas, line 597: @iglocska Can you check if I break things here?  Looks like filtering in `admin_view()` and `view()` was different. This change unifies the filtering.

cvandeplas:
Might be interesting to also include the code for #9107 HTOP support. On in a later commit to allow fast merging.

cvandeplas:
hmm, perhaps the TOTP is missing an anti-bruteforce. Or am I getting too paranoid ?
(2 minutes later): wow that was easy to implement

cvandeplas:
Commit cb74ad507f0aa2709cf7e9ad1bdb9053e932be58 brings support for HOTP:
image

After generating the TOTP the user is directed to a page with a list of the next 50 HOTP tokens.
image


At login the user can either give the TOTP or the HOTP.
image

cvandeplas:
There's one more thing I'm wondering: right now the db holds the secret.
Perhaps it'd be better to store the provisioning URI instead, [as the provisioning Uri](https://github.com/Spomky-Labs/otphp/blob/10.0.x/doc/QA.md) 

> contains all parameters including the custom ones, then only the new provisioning will be affected by your new security policy.

What are your thoughts?

iglocska:
Don't think that the db storing the secret in this case is an issue. If you get access to the db, you have bigger issues to worry about.

As for moving to the url approach, I don't expect the values to be tweaked by users. I think we can go for a sane set of defaults for now and if we ever absolutely have to change the security policy we will do a mass revocation on update. I don't expect this to happen frequently (we've had one internal password policy change for example, moving from salted sha1 to bcrypt, I expect this to be similar). Also from what I understand it's a hassle for HOTP.

iglocska:
Thanks for the lovely feature, I've merged it into develop. I've made some tiny changes to avoid some potential issues (updates not being executed yet blocking you from logging in, potential minor security issue with org admins deleting site admin keys, unprompted post requests to delete totp for a user). 

Some potential ideas for improvements:
- when requesting the hotp list, signal the current next key to the user (we have the counter in the user table anyway)
- creating a separate log action for otp failures rather than reusing login_fail

iglocska:
Another super nice to have idea, based on some external discussions: 

Users should be able to remove their own totp, provided they can reauthenticate using totp as part of the deletion process.

iglocska:
For posterity: The idea to not be able to remove your own totp was my idea, apologies for being an arse :)

cvandeplas:
Thanks for the review and the security issues identified before being implemented. I reviewed the changes and see where I made the mistakes.  

> when requesting the hotp list, signal the current next key to the user (we have the counter in the user table anyway)

Actually, the generated list of HOTP starts at the next key (you can see this in the screenshot above) so no need to mention the sequence number to the user. The login page mentions which key to use though.

I'll look into allowing the authenticated user to remove the TOTP.

PR #9086, opened 2023-05-20
Explicitly add dependency to 'php-http/message-factory'

ostefano:
This is needed because `php-http/message-factory` is not an explicit dependency of `php-http/message` since https://github.com/php-http/message/releases/tag/1.16.0

CC @garanews @iglocska


iglocska:
Thanks a lot!

garanews:
Thanks to you , what a team: one founds, one fixes and one brings in production! ❤️

PR #9140, opened 2023-06-15
fix: update composer and fix dependencies

ostefano:
Changes:
- update composer.phar to latest stable (2.5.8)
- rollback pinning indirect dependencies


iglocska:
Much appreciated, merged into develop

PR #9153, opened 2023-06-20
upd: add jammy release for arm64.

nyx0:
I didn't find where the install script is generated so I will leave it there.
Here is a small update to add ubuntu-jammy aarch64.


iglocska:
Thanks!

PR #9158, opened 2023-06-24
Fix search galaxy clusters

ostefano:
Now working as expected.

Basically made it consistent with what other controllers do.


ostefano:
^ @iglocska

iglocska:
Thanks a lot!

PR #9163, opened 2023-06-29
Update INSTALL.sh

ajb3932:
#### What does it do?

The current command in the RHEL install adds the line "Listen 443" after the line containing "Listen 80" even if "Listen 443" already exists.

In my update, the "Listen 443" line will only be added if it doesn't already exist in the file.

#### Questions


adulau:
Good catch! Thanks for the contribution.

PR #9166, opened 2023-06-30
fix: No events returned when one of many orgs in filter does not exist

tomking2:
#### What does it do?
Discovered that when using the search_index endpoint (or filtering through UI), if you pass a list of organisations in the `org` parameter, and at least **one** of the orgs does not exist (either by ID, UUID or Name), then no results are returned. It did not matter if other orgs provided in the parameter exist.

To replicate:
The following search returns a result:
```
pm.search_index(org=["CIRCL"], limit=1)
```

The following does not:
```
pm.search_index(org=["CIRCL", "CIRCL_____MISSING"], limit=1)
```

Fixed by handling where at least one org was found, and also skipping including an empty `$test` in the conditions (for code cleanup) 

#### Questions

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -427,6 +427,7 @@ class EventsController extends AppController
                     // if the first character is '!', search for NOT LIKE the rest of the string (excluding the '!' itself of course)
                     $pieces = is_array($v) ? $v : explode('|', $v);
                     $test = array();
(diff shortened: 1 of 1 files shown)


tomking2:
I noticed this is still an issue on 2.4 + 2.5. @iglocska Just checking this isn't intended activity? If not, probably a worthwhile fix

Issue #9004, opened 2023-04-07
Feature Request: Add logs for module enrichment execution


vincenzocaputo:
### Is your feature request related to a problem? Please describe.

I cannot find any logs in MISP concerning the execution of the enrichment modules. They can be useful when a user executes a module but does not complete the attribute submission process.

### Describe the solution you'd like

Additional logs with this information
- name of the module
- when it was run
- user who run the module

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #9170, opened 2023-07-01
Add enrichment modules execution logs

vincenzocaputo:
#### What does it do?

This pull request adds logging for enrichment module executions, see [`#9004`](https://github.com/MISP/MISP/issues/9004)

A log entry is created whenever an enrichment module is executed, even if the user does not submit the enrichment results at the end. I think this could be useful for keeping track of enrichment module usage.

#### Questions


mokaddem:
Looks good to me. I'm just wondering if we shouldn't make this behavior configurable, especially for users that systematically enrich all incoming data.
What do you think @iglocska ?

Issue #9175, opened 2023-07-06
Bug: When attributes are fetched using pagination pages can be empty or not full when items are skipped


Benni0:
### Expected behavior

When artifacts are fetched and enough attributes are available for the selected time, each page should be full except the last one.
This behavior is also expected when some attributes are skipped due enforceWarninglist or excludeDecayed.

### Actual behavior

The amount of specified items for the specific page are fetched from the database and in the for loop each attribute which should be filtered is skipped. So the amount of returned attributes may be lower as the requested page size while matching attributes may exist on the next page.

### Steps to reproduce

Fetch attributes using the  REST API and use one of the filter, if there are some attributes which are skipped on one of the pages, the affected page has not as much attributes as requested.

### Version

2.4.172

### Operating System

CentOS

### Operating System version

8

### PHP version

7.4.30

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

iglocska:
The current behaviour is intentional as certain elements are removed in the post-processing. Basically we pass back a header that signals whether there is more data to fetch (based on the intrnal fetch, so before the post-processing fires) so relying on that would be best. Looping would potentially be a massive performance overhead.

With that said, if you could reimplement the support for the looping to be an optional feature that needs to be passed in a parameter (such as ensurePageLimit or something to that effect) then we could merge it.

Benni0:
Hi @iglocska,
sorry for the late answer.
I recommend to not merge this pull request, as it would implement a mechanism which might have an unpredictable runtime behavior. I've created it as an prof of concept more less. 

But I'm currently struggling with this problem again. For my use-case client side looping would be ok, which should be possible with the mentioned header, but I can't find this header. So my questions are, what is the name of this header and where is it set? I think, this header should be set within the `foreach` loop in the `fetchAttributes` function when an attribute got skipped, but it isn't there.

To clarify what I want to achieve with this header: I have implemented a paginated polling importer and need a valid termination criterion. I can't break when the amount of items is lower than the limit and also not if no items are returned, because there can be further attributes on the next page.
As a workaround I break up when the `X-Result-Count - 1 % limit == 0` and no Items are returned, which only works due to a bug in  `X-Result-Count` calculation (#9422).

Many thanks for your support!

Benni0:
As it is now possible to check if items are skipped due to post processing on client side, using the X-Skipped-Elements-Count header, I close this Issue.

PR #9176, opened 2023-07-06
Fix pagination issue

Benni0:
#### What does it do?

Fixes Issue: #9175
Bug: When attributes are fetched using pagination pages can be empty or not full when items are skipped

#### Questions

#### Note
This fix should be considered as an example because of the fact that this implementation has an impact on performance. The main impact is, that all previous pages must be fetched in any case, otherwise the exact beginning of a page could not be determined.

A better solution would be, to do the filtering on SQL-level but this wouldn't be a quick fix.

Maybe we should also consider to avoid the n+1 select (indirect recursive call) when decay-score is included:
1. [attachScoresToAttribute](https://github.com/MISP/MISP/blob/648c1c9ea21b85e2d360033a65882fe33c8b9bbe/app/Model/Attribute.php#L1846C48-L1846C96)
2. [getScore](https://github.com/MISP/MISP/blob/648c1c9ea21b85e2d360033a65882fe33c8b9bbe/app/Model/DecayingModel.php#L625C33-L625C33)
3. [fetchAttributes](https://github.com/MISP/MISP/blob/648c1c9ea21b85e2d360033a65882fe33c8b9bbe/app/Model/DecayingModel.php#L691)

To avoid this it could be possible to add the most recent sighting date to the attribute like the last seen date.

### DIFF
--- a/app/Model/Attribute.php
+++ b/app/Model/Attribute.php
@@ -1773,7 +1773,8 @@ class Attribute extends AppModel
             if (($params['limit'] != 0) && ($options['enforceWarninglist'] || $proposals_block_attributes || $options['excludeDecayed'])) {
                 $loopLimit = $params['limit']; // optimistic approach
                 $loop = true;
-                $preventUnfinishedPages = True;                
+                $preventUnfinishedPages = True;
+                $params['page'] = 1;
             } else {
                 $loop = false;
             }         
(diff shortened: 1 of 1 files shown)


adulau:
Could you update the PR as proposed by @iglocska in https://github.com/MISP/MISP/issues/9175#issuecomment-2058889503 that would be awesome.

Benni0:
As I've already mentioned in #9175, this implementation was a proof of concept, more less.
This implementation would might have an unpredictable runtime behavior.

Issue #8147, opened 2022-02-17
Bug: Connector tags for custom MISP galaxies use UUID rather than name/value


JamesSOC080:
### Expected behavior

The "connector" tag used for default galaxies/clusters (i.e. those created in the cluster json files stored on the in the webroot directory) uses the cluster name/value. 

For example:

Connector tag misp-galaxy:mitre-intrusion-set="APT-C-36 - G0099"

When creating a custom cluster within the same galaxy, e.g. Aquatic Panda, I would expect the connector tag to take the form:

Connector tag misp-galaxy:mitre-intrusion-set="AquaticPanda"

### Actual behavior

Instead, custom clusters use the galaxy UUID instead of the value/name in hte connector tag. Therefore, for the above custom AquaticPanda cluster, the connector tag is:

Connector tag misp-galaxy:mitre-intrusion-set="7f7091ba-58f3-447e-b7a6-039fafd24e37"

This is causing us issues when ingesting event metadata into our SIEM - as our lookup utilises the connector tags to parse the galaxy clusters that are attached to MISP events.

### Steps to reproduce

1. Create a new custom galaxy
2. Open the custom galaxy and review the Connector tag field

### Version

2.4.153

### Operating System

Ubuntu

### Operating System version

18.04

### PHP version

7.1

### Browser

Chrome

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

UFOSmuggler:
Just realised I reported this issue here as well: https://github.com/MISP/MISP/issues/8707

PR #9178, opened 2023-07-06
Fix custom galaxy cluster connector tag value

vincenzocaputo:
#### What does it do?

Fix Issue: [#8147](https://github.com/MISP/MISP/issues/8147)

This pull request fixes the connector tag value of custom Galaxy Clusters, using the Cluster name rather then the UUID.

A check on the uniqueness of the connector tag was also added

#### Questions


mokaddem:
Hey,
Thanks for your PR but this is the intended behavior. Any custom galaxy clusters have their connector tag with their `UUID`.
This is mainly for two reasons:
1. Allow duplication of cluster (by forking or if you have different cluster-elements or relationships)
2. To allow the synchronisation of attached tag without leaking the cluster if you're not allowed to see it

Once you have access to the cluster, the connector tag will be replaced by the actual cluster in the UI and API.

Does it make sense?

PR #9179, opened 2023-07-06
Fix internal error caused by missing entries in warninglist

vincenzocaputo:
#### What does it do?

Fix internal error caused by not inserting entries during WarningList creation or update. If the user does not provide any values, an "An Internal Error Has Occurred" page is displayed. Wiith this fix a kinder error message is shown to prompt the user to fill out the field.

#### Questions


mokaddem:
Hey,
Thanks for the feedback and work!
Your patch wasn't working for me as it was missing few things.
It's fixed in 786ad2dde8cf948a993506e107ba84c24099c8a5.
Cheers!

Issue #9088, opened 2023-05-22
Bug: MISP Diagnostic page is not loading fully


dragsu:
### Expected behavior

I can load the Diagnostic page.

### Actual behavior

The diagnostic page is only loading partially.

### Steps to reproduce

When I try to load the Diagnostic page it only loads partially and gives the below error. This happens when it tries to show PHP Dependencies and I did not have this issue in the prior versions.  

I'm running a containerised version of MISP.

```
PHP Dependencies
   
An Internal Error Has Occurred.

Error: An Internal Error Has Occurred. Please try your action again. If the problem persists, please contact administrator.
```

### Version

2.4.169, 2.4.170, 2.4.171

### Operating System

RHEL

### Operating System version

8.8

### PHP version

7.4.3

### Browser

Firefox

### Relevant log output

```shell
cat /var/www/MISP/app/tmp/logs/error.log


2023-05-22 15:09:30 Error: [Error] Call to undefined function _()
Request URL: /servers/serverSettings/diagnostics
Stack Trace:
#0 /var/www/MISP/app/Lib/cakephp/lib/Cake/View/View.php(971): include()
#1 /var/www/MISP/app/Lib/cakephp/lib/Cake/View/View.php(933): View->_evaluate()
#2 /var/www/MISP/app/Lib/cakephp/lib/Cake/View/View.php(1224): View->_render()
#3 /var/www/MISP/app/Lib/cakephp/lib/Cake/View/View.php(418): View->_renderElement()
#4 /var/www/MISP/app/View/Servers/server_settings.ctp(11): View->element()
#5 /var/www/MISP/app/Lib/cakephp/lib/Cake/View/View.php(971): include('/var/www/MISP/a...')
#6 /var/www/MISP/app/Lib/cakephp/lib/Cake/View/View.php(933): View->_evaluate()
#7 /var/www/MISP/app/Lib/cakephp/lib/Cake/View/View.php(473): View->_render()
#8 /var/www/MISP/app/Lib/cakephp/lib/Cake/Controller/Controller.php(968): View->render()
#9 /var/www/MISP/app/Lib/cakephp/lib/Cake/Routing/Dispatcher.php(200): Controller->render()
#10 /var/www/MISP/app/Lib/cakephp/lib/Cake/Routing/Dispatcher.php(167): Dispatcher->_invoke()
#11 /var/www/MISP/app/webroot/index.php(99): Dispatcher->dispatch()
#12 {main}
```
```


### Extra attachments

[...]

PR #9180, opened 2023-07-07
fix: Localisation workflow typo

dragsu:
#### What does it do?

Fixes a typo in `editor.ctp` that breaks the loading of the workflow page.

#### Questions

- [x]  Are you using it in production?
- [ ]  Does it require a change in the API (PyMISP for example)?


dragsu:
@JakubOnderka a typo fix similar to https://github.com/MISP/MISP/issues/9088.

iglocska:
Thanks a lot!

PR #9192, opened 2023-07-13
fix: [UI] use acl to determine whether to show "audit logs" and "sear…

### LINKED ISSUE #8949: Bug: Audit Logs menu not shown for user with Audit Actions
### Expected behavior

Users with "Audit Actions" permissions should be able to view Application, Audit, Access, Search logs menu items.

### Actual behavior

Users with "Audit Actions" cannot view all Logs menu items. Only Application logs link is displayed
Accessing {URL}/admin/audit_logs/index directly works and provides the required information, so the issue is with GUI permissions and not backend user account permissions.

### Steps to reproduce

Add an user with "Audit Actions" permission and try to access Logs. Only Application Logs link is available.
Type in address bar {URL}/admin/audit_logs/index or {URL}admin/access_logs/index or {URL}admin/logs/search and the pages are available.

### Version

2.4.168

### Operating System

RedHat

### Operating System version

7.9

### PHP version

7.3

### Browser

_No response_

### Browser version

_No response_

### Relevant log output

_No response_

### Extra attachments

_No response_

### Code of Conduct

Wachizungu:
…ch logs" buttons

#### What does it do?

fix #8949

Please note that this doesn't change anything for access logs, because it doesn't look like perm_audit grants access to that one.

It's somewhat odd that these pages under /admin are accessible by a normal user with audit permissions, but on the other hand it makes sense (since audit permission is there to check the audit logs...). I don't know what the intended design is here in the end, but this commit should at least align the UI with the allowed access.

#### Questions

### DIFF
--- a/app/View/Elements/global_menu.ctp
+++ b/app/View/Elements/global_menu.ctp
@@ -465,7 +465,7 @@
                     array(
                         'text' => __('Audit Logs'),
                         'url' => $baseurl . '/admin/audit_logs/index',
-                        'requirement' => Configure::read('MISP.log_new_audit') && $isAdmin,
+                        'requirement' => Configure::read('MISP.log_new_audit') && $this->Acl->canAccess('auditLogs', 'admin_index'),
                     ),
                     array(
                         'text' => __('Access Logs'),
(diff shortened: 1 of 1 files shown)


cvandeplas:
thank you Jeroen !

PR #9194, opened 2023-07-14
NATO MISP only for cyber defense

JakubOnderka:
According to NATO MISP terms of use, NATO MISP is open only for cyber defense related governmental entities, not to all governmental entities.

#### Questions


adulau:
Thanks for the update!

PR #9195, opened 2023-07-17
explicitly set loginAction with baseurl

### LINKED ISSUE #2876: BASEURL protocol is not used in constructing 302 redirects
This is possibly a lack of understanding on my behalf of how the software is supposed to react, but when navigating to the base URL my browser will receive a 302 redirect to http://example.com/users/login however my BASEURL in the config.php is set to https://example.com/. given that port 80 isn't available this results in the browser hanging.

matletix:
Fix https://github.com/MISP/MISP/issues/2876

Got issues with *http* instead of *https* 302 redirects to http://example.com/users/login when accessing https://example.com, while the baseurl starts with *https*.

Here, the redirect uses the default `loginAction` :
https://github.com/MISP/MISP/blob/2.4/app/Controller/AppController.php#L1184

Explicitly defining `loginAction` with the baseurl fixes the issue and I get correct *https* 302 redirects to https://example.com/users/login.

### DIFF
--- a/app/Controller/AppController.php
+++ b/app/Controller/AppController.php
@@ -111,6 +111,7 @@ class AppController extends Controller
 
         $this->_setupBaseurl();
         $this->Auth->loginRedirect = $this->baseurl . '/users/routeafterlogin';
+        $this->Auth->loginAction = $this->baseurl . '/users/login';
 
         $customLogout = Configure::read('Plugin.CustomAuth_custom_logout');
         $this->Auth->logoutRedirect = $customLogout ?: ($this->baseurl . '/users/login');


iglocska:
That makes sense and encountered the issue too, nice catch, merged into develop.

Thanks a lot!

righel:
@matletix could you please check if this change fixed your issue?
https://github.com/MISP/MISP/pull/9212

your proposed fix introduced other problems

matletix:
@righel nop, ` $this->Session->write('Auth.redirect', $this->baseurl . '/users/login');` doesn't work. If we want to keep `loginAction` in the Array format instead of the string format in order to keep the `admin => false` field, then maybe the fix must be done in the `bootstrap.php` file as `App.fullBaseUrl` should be used in this case ?

righel:
`App.fullBaseUrl` is already set with the same value of `MISP.baseurl` in my tests. I cannot reproduce your issue, can you elaborate more on your settings? proxy, apache conf, etc? 
Reverted the change .

matletix:
@righel I use the [coolacid/docker-misp](https://github.com/coolacid/docker-misp) image to run the MISP in a custom kubernetes statefulset. TLS termination is done by the istio service mesh in the kubernetes cluster.
I'll try and post an MRE here when I find the time to do so.

PR #9203, opened 2023-07-23
Add attributes to warninglist workflow module

vincenzocaputo:
#### What does it do?

This pull request adds a new workflow module for adding attributes to a custom (non-default) warninglist.

Possible use cases:

- Add an attribute to a warninglist when soft-deleted
immagine

- Add an attribute to a warninglist when tagged with a specific tag

- Add attributes with false-positive sightings when an event is published/saved
immagine

During tests I noticed that if appropriate logic checks are not added to the workflow, attributes not matching the warninglist type (e.g.: string in cidr warninglist) will be added without errors. At the moment, it is up to the user to properly add logic modules and filters to the workflow to avoid this. 

#### Questions

### DIFF
--- a/app/Model/WorkflowModules/action/Module_add_to_warninglist.php
+++ b/app/Model/WorkflowModules/action/Module_add_to_warninglist.php
@@ -1,0 +1,85 @@
+<?php
+include_once APP . 'Model/WorkflowModules/action/Module_attribute_edition_operation.php';
+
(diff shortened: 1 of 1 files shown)


mokaddem:
Hey,

Many thanks to your contribution! It's on develop.
I took the opportunity to fix one or two things that either wasn't working or was bothering me :wink: 

Cheers!

PR #9210, opened 2023-07-26
Update defaults.json

milannshrestha:
Adding Phishfort blocklist to MISP. 26-Jul-2023


adulau:
The json input per se doesn't exist in the feed import. Could you update to match an existing format such as MISP, CSV or freetext? Thank you very much for the contribution

PR #9218, opened 2023-07-27
Added James Brine Bruteforce IPs to feed-metadata defaults json

referefref:
Added freetext feed endpoint for Bruteforce IPV4 addresses

#### What does it do?

Adds a IPV4 feed location

#### Questions


adulau:
Thank you very much for the PR. As the file is in CSV format, could you update it to be in the csv format and add the correct delimiter parameter in the JSON.

referefref:
> Thank you very much for the PR. As the file is in CSV format, could you update it to be in the csv format and add the correct delimiter parameter in the JSON.

Done and tested.

adulau:
Thank you very much for the updates and the contribution!

PR #9229, opened 2023-08-02
Kagee remove dead symlink

Kagee:
- chg: [doc] Remove symlink to non-exsisting ubuntu 16.04 docs
- chg: [doc] Fix python naming swap (based on example and practice) and stray whitespace.

#### What does it do?


iglocska:
Thanks a lot!

PR #9230, opened 2023-08-02
Improve diagnostics when instance does not have internet or does not use self-update.

Kagee:
new [diag]: Improve diagnostics when instance does not have internet or does not use self-update.

Introduces two new settings:
* `MISP.self_update` allows to enable/disable the GUI button for MISP self-update on the Diagnostics page.
* `MISP.online_version_check` allows to enable/disable the online MISP version check when loading the Diagnostics page.

These settings are useful for 1. container installations that should not be updated using self-update, and 2. installation that have no direct or proxy internet access.

There are also improvements on the Diagnostics page, primarily the MISP version area. Font color has been replace with classes, this allows the use of the `bold` class, not just colors, and possible combination with the red/green/orange colour classes.

The info/status/warning/error texts have been changed to take into account the status of `MISP.self_update` and
`MISP.online_version_check`.

Settings text, level and position.

2023-08-02-201707_1200x136_scrot

Default, `MISP.self_update` = `true`, `MISP.online_version_check` = `true`

2023-08-02-201541_1201x317_scrot

`MISP.self_update` = `true`, `MISP.online_version_check` = `false`

2023-08-02-201508_1223x318_scrot

`MISP.self_update` = `false`, `MISP.online_version_check` = `true`

2023-08-02-201605_1169x317_scrot

`MISP.self_update` = `false`, `MISP.online_version_check` = `false`

2023-08-02-201622_1173x315_scrot

### DIFF
--- a/app/Model/Server.php
+++ b/app/Model/Server.php
@@ -4135,7 +4135,11 @@ class Server extends AppModel
         $current = implode('.', $version_array);
 
         $upToDate = version_compare($current, substr($newest, 1));
-        if ($upToDate === 0) {
+        if ($newest === null && Configure::read('MISP.online_version_check')) {
+            $upToDate = 'error';
+        } elseif ($newest === null && !Configure::read('MISP.online_version_check')) {
+            $upToDate = 'disabled';
+        } elseif ($upToDate === 0) {
             $upToDate = 'same';
         } else {
             $upToDate = $upToDate === -1 ? 'older' : 'newer';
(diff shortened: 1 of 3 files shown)


ostefano:
@Kagee 

I would call the options `allow_self_update` and `check_version_online`.

Also in the second and fourth case I would just write (black and bold) "You are not on a branch" without explaining things further.

In conclusion, should the first case and show the latest available version?

Kagee:
I agree on the "You are not on a branch", that is probably enough, as long as it is not colored any way.

That last point is interesting, the test images are from a MISP w/o internet access, i have not actually testet it on a MISP w. internet access.

I'll hold back on changing the settings name until i have more feedback on it :)

PR #9233, opened 2023-08-03
Fix /taxonomies/view string filter

### LINKED ISSUE #8875: Bug: taxonomies/view Filter broken
### Expected behavior

The Filter on the View Taxonomy WebUI page for a given taxonomy no longer returns with matching rows.

It's working in 2.4.164, but not in 2.4.167.

### Actual behavior

 The `Loading` spinning wheel appears, but the resulting list is not limited to only matches as per the filter textbox.

### Steps to reproduce

Test the `Filter` on `taxonomies/view/nnn`

### Version

2.4.167

### Operating System

RH

### Operating System version

8.6

### PHP version

7.4.33

### Browser

_No response_

### Browser version

_No response_

### Relevant log output

_No response_

### Extra attachments

_No response_

### Code of Conduct

righel:
#### What does it do?
re-implements `/taxonomies/view` string filter removed by 5da9497efdb5bde654811641523c23fce752f5b5
fixes #8875

#### Questions

### DIFF
--- a/app/Controller/TaxonomiesController.php
+++ b/app/Controller/TaxonomiesController.php
@@ -77,9 +77,6 @@ class TaxonomiesController extends AppController
         $urlparams = '';
         App::uses('CustomPaginationTool', 'Tools');
         $filter = isset($this->passedArgs['filter']) ? $this->passedArgs['filter'] : false;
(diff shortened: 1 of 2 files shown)


adulau:
Thank you!

PR #9254, opened 2023-08-16
new: [Users] add last password change timestamp for users

Wachizungu:
Creating PR for an initial review / input if possible. Don't think it's ready as-is.
Still testing myself, note that I didn't test the db update part with MISP update procedure yet.

If you see an opportunity to finish / fix this from your end please feel free.

#### What does it do?

Adds a last_pw_change timestamp for users.
Idea is to use it to trigger a password change on next login after a specific time (will create a cli command and/or pymisp script for this).

For existing users, the last_pw_change timestamp would initially be set to date_modified. This feels like the best option at the moment.

#### Questions

### DIFF
--- a/app/Controller/UsersController.php
+++ b/app/Controller/UsersController.php
@@ -301,6 +301,7 @@ class UsersController extends AppController
             // What fields should be saved (allowed to be saved)
             $user['User']['change_pw'] = 0;
             $user['User']['password'] = $this->request->data['User']['password'];
+            $user['User']['last_pw_change'] = time();
             if ($this->_isRest()) {
                 $user['User']['confirm_password'] = $this->request->data['User']['password'];
             } else {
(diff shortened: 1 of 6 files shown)


Wachizungu:
Added line to also update last_pw_change timestamp when doing a password reset.

Wachizungu:
Hi,

I tried to optimize the tests. I can't get them to fail on my own instance anymore but they seem to inconsistently fail on the test instance here. Any advice or idea?

Alternatively, they could be used during the review and we can add @unittest.skip to skip them during automated actions for now.

iglocska:
Thanks a lot, and also a massive thanks for going the extra mile with the tests!

iglocska:
Will check why it fails on the CI and apply the appropriate fixes, my guess is the schema being slightly different.

PR #9259, opened 2023-08-22
fix: check for existence of push_rules in /server/edit requests before parsing JSON

TomOgs:
#### What does it do?

Fixes an issue with the /server/edit/ API Endpoint where 'Push_Rules' are required for the query to work as discussed in MISP/PyMISP#861

Used a couple of `If` statement to preform some checks to see if the field contains data before executing code that parses JSON from that field. 

#### Questions


righel:
Thanks for the contribution :+1:

PR #9275, opened 2023-09-06
Disable submodule update section when MISP.self_update is disabled, to allow not carrying git dependencies in docker

oivindoh:
#### What does it do?

Disable nonfunctional git based submodule checks when MISP.self_update is disabled.

This allows us to not expose functionality that doesn't work well in a docker scenario, and also allows us to build images without git to keep image size smaller.

Screenshot 2023-09-14 at 14 42 35

#### Questions


ostefano:
@Kagee thoughts?

iglocska:
Thanks a lot, nice catch!

PR #9285, opened 2023-09-12
Pin python dependencies

ostefano:
Pin python dependencies (minimum versions as per diagnostic page and/or urgent fixes)

Open questions:
- Should I remove pip from list of dependencies used (theoretically) by tests?
- Should I change the doc and edit the install script?
- Should I split requirements.txt into two files (so the two sets of dependencies are separate)?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,10 +1,19 @@
-pyzmq
-coveralls
-codecov
-stix
-pymisp
-requests-mock
-pip
-nose
+cybox>=2.1.0.21
 jsonschema
-plyara >= 2.0.2
+lief>=0.13.1
+maec>=4.1.0.17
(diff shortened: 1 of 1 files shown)


ostefano:
I would remove `pip` from list of deps used (theoritically) by tests.

ostefano:
CC @iglocska @Rafiot

ostefano:
Should I change the doc and edit the install script?

ostefano:
Also, happy to split requirements.txt into two files (so the two sets of deps are separate)

oivindoh:
Probably want to pin specific or major.minor rather than greater than - bit these pins will surely improve delivery quality

iglocska:
Thanks a lot!

PR #9296, opened 2023-09-19
fix: [RestSearch] allow filtering on eventinfo for events and attributes again

Wachizungu:
#### What does it do?

You might want to include this in 2.4 directly.
It used to be possible to filter on eventinfo when using /events/restsearch and /attributes/restsearch. Now the filter is simply ignored (which can lead to some bad stuff, if the results are used for other operations).
I'm guessing 37ecf81b84a01baa4d4b1fade4de94a9018c32ed broke it. Not (yet) sure if some other filters are affected as well, would be nice if someone can double check and fix those if necessary.

#### Questions


iglocska:
Thanks a lot!

PR #9301, opened 2023-09-27
[users] fix user view totp delete checks

Wachizungu:
#### What does it do?

fixes:

- Org admins can delete totp for users, but the link doesn't show up in the view.
screenshot after the change:
2023-09-27 14_59_41-misp-dev - VMware Workstation

- The delete link doesn't show up on your own profile view when you are site admin or org admin

Granted, the my profile page doesn't look very nice this way when you have totp (two links next to each other). If you have an easy fix to make it look better... would be welcome:
2023-09-27 14_35_18-misp-dev - VMware Workstation

#### Questions


Wachizungu:
With conditional space :)

my profile view (admin with TOTP set up):
image

And admin user view:
image

cvandeplas:
thank you  !

iglocska:
Thanks a lot, nice catch!

PR #9303, opened 2023-09-28
[users:totp] set correct rest response action for totp_delete

Wachizungu:
#### What does it do?

Fixes likely typo in the rest response action for totp_delete. As far as I can see the action is totp_delete and not admin_totp_delete.

perhaps @cvandeplas can review this / confirm.

Doesn't really cause issues though it just changes the resulting message:
image

#### Questions


cvandeplas:
thank you @Wachizungu

PR #9308, opened 2023-10-02
fix: [Taxonomies] fix enabling of individual taxonomy tags for tags w…

### LINKED ISSUE #9300: Bug: Some Taxonomies are not creating tags after enabling from the taxonomy page
### Actual behavior

On most taxonomies (in this case I am using workflow but I tested, TOR, admiralty and few others) the server does not create a tag once the + sign in the taxonomy page is clicked. Steps: 
1. Enable taxonomy
2. Click the plus sign on the tag that want to be created 
3. Notification messsage shows up, click "Submit" 
image

4. Nothing happens. No error and the tag is not created. 
image

Additionally, if "Enable all" button is used then all the tags are created. However, the "Stop" button does not seem to do anything. 

### Expected behavior

For example following the same workflow for the TLP taxonomy: 
1. Enabled taxonomy
2. On the taxonomy page I clicked + on tlp clear
image

3. The tlp clear tag was created: 
image

### Steps to reproduce

Steps:
1. Performed clean installation of MISP 2.4.177 using the installation script on Ubuntu 22.04. 
2. Event Actions -> List Taxonomies -> Select Taxonomy (like Workflow) -> Enabled taxonomy
4. From the taxonomy page "/taxonomies/view/147" click the plus sign on the tag that want to be created
5. Notification message shows up, click "Submit"
7. Nothing happens. No error and the tag is not created.

### Version

2.4.177 

### Operating System

Ubuntu

### Operating System version

22.04

### PHP version

7.4.33

### Browser

Chrome, Firefox, Saphari

### Browser version

latest

### Relevant log output

```shell
I could not find any error logs that could be related to it. The app says the tag is created but it is not showing it.
```

### Extra attachments

n/a

### Code of Conduct

Wachizungu:
…ith special chars. fixes #9300

#### What does it do?

proposal to fix #9300 , or at least the part about enabling individual taxonomy tags.

IndexTable actions view (actions.ctp) has this line for onclick params:
$temp[] = h($k) . ':' . urlencode(h($extracted_value[0]));

resulting in urls like these for the onclick action in case of tags with double quotes " (note the \&quot; for quotes): /taxonomies/addTag/taxonomy_id:166/name:workflow%3Astate%3D%26quot%3Bcomplete%26quot%3B

Basically MISP is/was comparing a html encoded version with the non encoded one stored in db, which results in no match.
Added a second check so these values are accepted by taxonomy addTags as well. Modified 2 other lines so that when there is resulting match, it uses the _taxonomy_ version of the tag name (those end up in $entry['tag'].

#### Questions

### DIFF
--- a/app/Model/Taxonomy.php
+++ b/app/Model/Taxonomy.php
@@ -445,11 +445,11 @@ class Taxonomy extends AppModel
             }
             if ($tagList) {
                 foreach ($tagList as $tagName) {
-                    if ($tagName === $entry['tag']) {
+                    if ($tagName === $entry['tag'] || $tagName === h($entry['tag'])) {
                         if (isset($tags[strtoupper($entry['tag'])])) {
-                            $this->Tag->quickEdit($tags[strtoupper($entry['tag'])], $tagName, $colour, 0, $numerical_value);
+                            $this->Tag->quickEdit($tags[strtoupper($entry['tag'])], $entry['tag'], $colour, 0, $numerical_value);
                         } else {
-                            $this->Tag->quickAdd($tagName, $colour, $numerical_value);
+                            $this->Tag->quickAdd($entry['tag'], $colour, $numerical_value);
                         }
                     }
                 }


iglocska:
Thanks!

PR #9310, opened 2023-10-03
Regression - Rest search with 'attributes' controller no longer filters by sharing group ID

tomking2:
#### What does it do?
Fixes a regression that occurred during REST API improvements that stopped the 'attributes' filtering by Sharing Group (using the `sharinggroup` parameter)

Appears to be as simple as exposing `sharinggroup` as an allowed parameter for an Attribute REST.

#### Questions

### DIFF
--- a/app/Controller/Component/RestSearchComponent.php
+++ b/app/Controller/Component/RestSearchComponent.php
@@ -52,6 +52,7 @@ class RestSearchComponent extends Component
             'first_seen',
             'last_seen',
             'eventinfo',
+            'sharinggroup',
             'allow_proposal_blocking',
             'flatten',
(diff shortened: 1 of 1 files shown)


Wachizungu:
Hey,

Thanks for the contribution already. Just one note from my side:
I believe this list is also used to determine the order for ordered url params, which means it's best to add new entries at the end (I'm guessing you put it where you did to align with the event param list).

I think whoever will potentially merge this from the core dev team can just put it at different spot though, so not sure it's necessary to do a new push.

iglocska:
Looks good to me, thanks! If you find more cases of those failing let me know!

PR #9312, opened 2023-10-04
new: [User] Add setting to limit site admin roles to instance's host …

Wachizungu:
…org.

#### What does it do?

Adds a setting which, if set to true, will prevent assigning a site admin role to users that do not belong to the host org of the instance. This is mainly to prevent some potential oopsies.
I made it a recommended setting because my assumption is that on most instances, site admins belong to the host org.

I didn't make the role selection list conditional for the form, because:

1. I think it might be nicer that you actually get a message indicating that it's a setting on this instance that prevents you from assigning site admin role to specific user.
2. It's not something I know how to do straightaway at the moment (still getting back into php/js coding, getting more familiar with cakephp is on my todo as well)

That being said, I think it would be nice to get a warning message on the form, when about to grant site admin role to a non host org user. So I'll keep that in the back of my mind for a future improvement.

#### Questions

### DIFF
--- a/app/Model/Server.php
+++ b/app/Model/Server.php
@@ -6407,6 +6407,14 @@ class Server extends AppModel
                     'type' => 'boolean',
                     'null' => true
                 ),
+                'limit_site_admins_to_host_org' => array(
+                    'level' => self::SETTING_RECOMMENDED,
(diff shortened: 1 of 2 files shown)


iglocska:
Sounds like a good idea, indeed, whilst there are exceptions this is the most common use-case.

PR #9313, opened 2023-10-04
chg: [user] Update store api access time setting description

Wachizungu:
Api access time is stored once per hour by default (since commit a5f5a4e113872a77d4e6c2b1a125f03ee89773c2), making the old description of this setting incorrect.

#### What does it do?

Changes the relevant setting description.

@cvandeplas , if my understanding of commit a5f5a4e113872a77d4e6c2b1a125f03ee89773c2 is correct, last_api_access is logged for users by default (once per hour max), so this description of the setting would be better in my opinion.

Unfortunately the name of the setting is now misleading, but I guess there is no way to change it without affecting people that already use it?

#### Questions

### DIFF
--- a/app/Model/Server.php
+++ b/app/Model/Server.php
@@ -5615,7 +5615,7 @@ class Server extends AppModel
                 ),
                 'store_api_access_time' => array(
                     'level' => 1,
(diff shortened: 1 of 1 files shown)


cvandeplas:
Excellent clarification. Thank you @Wachizungu !

PR #9314, opened 2023-10-04
chg: [RestClient] use http_method value from template if available

Wachizungu:
#### What does it do?

Some templates for the restclient have a http_method defined, but they are not set properly when using the query builder.

An example is Feed previewIndex with http_method GET.
Still keeping POST as default for the ones that don't have it specifically defined.

#### Questions


adulau:
Indeed good one! Thank you.

PR #9315, opened 2023-10-04
new: [RestClient] Add user totp_delete to query builder

Wachizungu:
#### What does it do?

Adds user totp delete to restclient query builder

#### Questions


adulau:
Thank you!

PR #9329, opened 2023-10-12
fix: [misp-zmq] Include ZMQ support for connecting to Redis over TLS

peritz:
#### What does it do?

Adds support for ZeroMQ to connect to Redis backend via TLS. This is implemented by checking the start of the `redis_host` setting for `tls://`. We check for this because if you want to connect to redis over TLS in CakePHP (i.e. for any other setting in MISP using redis) you prepend the redis hostname with `tls://`.

If it does begin with `tls://` then we remove the scheme, otherwise it throws an error as below:

```
redis.exceptions.ConnectionError: Error -2 connecting to tls://redis_host:6379. Name or service not known.
```

#### Questions

### DIFF
--- a/app/files/scripts/mispzmq/mispzmq.py
+++ b/app/files/scripts/mispzmq/mispzmq.py
@@ -80,9 +80,14 @@ class MispZmq:
         with open((self.tmp_location / "mispzmq_settings.json").as_posix()) as settings_file:
             self.settings = json.load(settings_file)
         self.namespace = self.settings["redis_namespace"]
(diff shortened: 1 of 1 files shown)


adulau:
It's indeed a very good idea. Thanks for the PR. 

I'm just wondering about the remaining configuration for the RedisSSL like `ssl_certfile`, `ssl_keyfile`, `ssl_cert_reqs` or `ssl_ca_certs` which could be used for the MISP configuration.

PR #9339, opened 2023-10-19
Setting to disable IP retention in DB on Authkey use

sid1033224:
#### What does it do?

Adds a new setting - `MISP.remember_seen_ips_authkeys` - to control whether IP addresses used to access the API with an Authkey are retained in the `unique_ips` field of the `auth_keys` table.  Since TEXT columns in MySQL are at most 64KB long, this field can overflow if the same authkey accesses the API from a sufficient number of remote endpoint addresses.  Upon such overflow, API calls fail with 5xx errors.

If this setting is set to false, the "Seen IPs" column in the "Authentication Key Index" UI will be hidden, as it will likely not contain complete, meaningful data.

#### Questions

### DIFF
--- a/app/Locale/ara/LC_MESSAGES/default.po
+++ b/app/Locale/ara/LC_MESSAGES/default.po
@@ -6997,6 +6997,10 @@ msgstr ""
 msgid "Log user IP and key usage on each API request. All logs for given keys are deleted after one year when this key is not used."
 msgstr ""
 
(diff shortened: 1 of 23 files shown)


iglocska:
Thanks a lot, merged on develop

iglocska:
FYI, I've changed it slightly so that we do log by default, but can be disabled on demand rather than having it as an opt-in functionality.

PR #9347, opened 2023-10-21
Fix API url in Telegram workflow module

### LINKED ISSUE #9346: Support: Sending messages through Telegram
### Support Questions

Hi all,

I'm trying to set a workflow to send a message through Telegram. I saw that, from the actions available, there is the action "Telegram send alert". I set the chat id and the telegram token and a simple textual message. After I publish an event, trigger of the workflow, I don't receive any message. I'm sure that the telegram token and chat id are correct because I tryed it with the python-telegram-bot python module. I'm also sure that the workflow is successfull from the logs. 

Any advice?

image

### MISP version

v2.4.176

### Operating System

Ubuntu

### Operating System version

20.04

### PHP version

7.4

### Browser

_No response_

### Browser version

_No response_

### Relevant log output

_No response_

### Extra attachments

_No response_

### Code of Conduct

vincenzocaputo:
#### What does it do?

Fixes #9346 

#### Questions

### DIFF
--- a/app/Model/WorkflowModules/action/Module_telegram_send_alert.php
+++ b/app/Model/WorkflowModules/action/Module_telegram_send_alert.php
@@ -51,11 +51,11 @@ class Module_telegram_send_alert extends Module_webhook
 		'parse_mode' => "HTML",
 	];
 
-	$url = $this->telegram_url . $bot_token . "/sendMessage";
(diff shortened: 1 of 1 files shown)


adulau:
Thank you!

PR #9373, opened 2023-11-15
Accessibility: Improved tag reporting in Event view

obert01:
Added a "role" attribute so the global/local nature of tags are read correctly by all screen readers.

Without "role" attributes, JAWS ignore the whole "<span><i> ... </i></span>", even if aria-label is specified.

### DIFF
--- a/app/View/Elements/rich_tag.ctp
+++ b/app/View/Elements/rich_tag.ctp
@@ -35,7 +35,7 @@ if (isset($tag_display_style)) {
 }
 $aText = h($aText);
 $span_scope = !empty($hide_global_scope) ? '' : sprintf(
-    '<span class="%s" title="%s" aria-label="%s"><i class="fas fa-%s"></i></span>',
+    '<span class="%s" title="%s" role="img" aria-label="%s"><i class="fas fa-%s"></i></span>',
     'black-white tag',
     !empty($tag['local']) ? __('Local tag') : __('Global tag'),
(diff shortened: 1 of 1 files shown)


iglocska:
Thank you as always @obert01! It's merged into develop and will go into the release later today.

iglocska:
By the way, how usable is MISP for you with JAWS? Do you see anything we should be doing to make it more usable?

obert01:
Hi @iglocska,

Generally speaking, MISP is rather accessible with JAWS and NVDA, the two major screen readers available for Windows. I use MISP in my daily job at Cert-IST.

There are some small HTML details in a few views, but I don't have them in mind. I will prepare PRs when time permits. 

Anyway, the most significant hidashes for the moment are:
1. The multi-select / autocomplete combo list component you use for choosing clusters / tags / objects / attribute types. You are using the third-party js library Chosen which lacks the accessibility features. There is an official fix we are using at Cert-IST's, which is available here:
https://github.com/harvesthq/chosen/pull/2047#issuecomment-1333301562
Unfortunately, the author of Chosen hasn't shown any interrest recently for accessibility. 
2. The way the "server settings" screens works. To modify a given setting, I have to route the mouse pointer to the location where my screen reader's virtual cursor is, and then simulate a double-click. I am used to these kind of things, but this remains a bit tricky. A good alternative would be to have a bouton (even an ARIA-only button) for each setting which when clicked, switches the given setting in edit mode.

PR #8964, opened 2023-03-18
Feature/user login profiles

cvandeplas:
Please have a look at the following PR, note it is still a work in progress.
- review code 
- review feature 
Some more work is needed before being merged completely, but input would be very welcome now. 

#### What does it do?

This new feature adds a way to inform users (email) when an interactive login happens with their account from an unrecognized browser profile / userloginprofile.
The user can then confirm the login was legitimate. 
This information is then stored in the database and allows users to be alerted (email) when a suspicious interactive login happens.

How is the userloginprofile / browser profile built? 
- user_agent (ua_platform/OS, ua_browser, ua_pattern) - using browscap (if installed) or primitive function (if not installed)
- ip & geoip (if installed, the geoipdb is delivered with this PR)
- accept_lang browser variable

At every (interactive/api) login the log table is populated with this profile information.
This info is then used to build a login history table, where users can mark logins as legitimate/evil.

#### UI changes:
User view page - addition of a "view login history"
image

Login history: lists grouped logins based on loginprofile
image

Users can press the buttons on the right to mark something good/evil: 
image

They can review the confirmed userloginprofiles on the recent login attempts page (at the bottom). (site/org) Admins can delete entries from this list.
image

#### DB changes
- logs the userloginprofile to the logs  (no real db change)
- Adds a table userloginprofiles

#### Questions

- [x] Does it require a DB change?
- [ ] Are you using it in production?
- [x] Does it require a change in the API (PyMISP for example)?   APIaccess not tested, more dev is needed to allow API access to all, if we deem this necessary.

#### Limitations and Future plans

- The alerting currently only works with interactive logins, no API logins. (for performance reasons and mass-mailing-issues)
- Right now the feature does NOT prevent 'malicious' authentications to log in. An email alert is sent to the user and a log is generated so that the org_admin and site admin can be aware.
- Requires the installation of 2 additional composer packages: browscap and geopip
- The UI is not the most userfriendly yet
- The current "is it suspicious" logic is relatively primitive
- Build something for API logins, such as storing the IPs used and allowing the user to easily see which source IP was used with an API key, and pin it to mark it as "only allow logins from there"

PR #9379, opened 2023-11-22
Feature/user login profiles2

cvandeplas:
(replaces: https://github.com/MISP/MISP/pull/8964 ) 

#### What does it do?

This new feature adds a way to inform users (email) when an interactive login happens with their account from an unrecognized browser profile / userloginprofile.
The user can then confirm the login was legitimate. Subsequent logins using a similar loginprofile will not generate an email alert.

How is the userloginprofile / browser profile built? 
- user_agent (ua_platform/OS, ua_browser, ua_pattern) - using browscap (if installed) or primitive function (if not installed)
- ip & geoip (if installed, the geoipdb is delivered with this PR)
- accept_lang browser variable

At every (interactive/api) login the log table is populated with this profile information.
This info is then used to build a login history table, where users can mark logins as legitimate/evil.

#### UI changes:
User view page - addition of a "view logins" button
image

Login history: lists grouped logins based on loginprofile
image

Users can press the buttons on the right to mark something good/evil with the buttons. 

They can review the confirmed userloginprofiles on the recent login attempts page (at the bottom). (site/org) Admins can delete entries from this list.
image

image

#### DB changes
- logs the `userloginprofile` to the logs  (no real db change)
- Adds a table `userloginprofiles`

#### Questions

- [x] Does it require a change in the API (PyMISP for example)?   API access not tested, more dev is needed to allow API access to all, if we deem this necessary.

#### Limitations and Future plans

- The alerting currently only works with interactive logins, no API logins. (for performance reasons and mass-mailing-issues).
- Right now the feature does NOT prevent 'malicious' authentications to log in. An alert is generated in the logs so that the org_admin and site admin can be aware. User-email alerts are not done considering the risks of one user report creating collateral damage for many other users.
- Requires the installation of 2 additional composer packages: browscap and geopip
- The current "is it suspicious" logic is relatively primitive


INSTALL/MYSQL.sql
```diff
@@ -1424,6 +1424,31 @@ CREATE TABLE IF NOT EXISTS `users` (
...
+  `geoip` varchar(255) DEFAULT NULL,
+  `ua_platform` varchar(255) DEFAULT NULL,
+  `ua_browser` varchar(255) DEFAULT NULL,
+  `ua_pattern` varchar(255) DEFAULT NULL,
+  PRIMARY KEY (`id`),
+  KEY `ip` (`ip`),
+  KEY `status` (`status`),
+  KEY `geoip` (`geoip`),
+  INDEX `user_id` (`user_id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```
iglocska: Option1: 
For all varchars -> varchar(191)
table collate = utf8mb4

Option2:
user_agent collate = utf8mb4
user_agent varchar(191)

app/Model/AppModel.php
```diff
@@ -1980,6 +1981,26 @@ public function updateDatabase($command)
             case 116:
                 $sqlArray[] = "ALTER TABLE `event_reports` modify `content` mediumtext";
                 break;
+            case 117:
+                $sqlArray[] = "CREATE TABLE IF NOT EXISTS `user_login_profiles` (
```
iglocska: See changes in the schema

cvandeplas: done

INSTALL/MYSQL.sql
cvandeplas: done

jheller:
The line 
CREATE TABLE `user_login_profiles`
should be 
CREATE TABLE IF NOT EXISTS `user_login_profiles`

This has broken our install automation.

adulau:
@jheller Very good point.

adulau:
Thanks it's fixed in develop https://github.com/MISP/MISP/commit/aac29ad6af83fe7d617628e26725f94e30ede4a1

PR #9457, opened 2023-12-21
Banco do Brasil public feed

threatintelBB:
Added Banco do Brasil public feed

### DIFF
--- a/app/files/feed-metadata/defaults.json
+++ b/app/files/feed-metadata/defaults.json
@@ -1737,7 +1737,7 @@
       "lookup_visible": true,
       "caching_enabled": true
     }
-  }
+  },
   {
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks a lot!

iglocska:
Hmm, looks like the feed is unavailable, could you double check the URL?

cgoncalves1:
> Hmm, looks like the feed is unavailable, could you double check the URL?

Hello I'm from Banco do Brasil's CTI team. The URL is correct. Maybe there is a firewall rule blocking access, we'll check it out.

threatintelBB:
Hello, the feed URL is now available. Thanks!!!

PR #9481, opened 2024-01-09
chg: [totp] add clarifications to totp setup view

Wachizungu:
#### What does it do?

Adds some clarifications to the totp setup page. We've noticed some people get confused when they see this page (especially if totp was mandated on the instance).

#### Questions


adulau:
Thank you! It's indeed clearer.

PR #9485, opened 2024-01-10
chg: [doc/openapi] clarify 'deleted' restsearch filter

### LINKED ISSUE #9437: Bug: `deleted` param in /attributes/restSearch endpoint behaves differently than expected
### Actual behavior

When `"deleted": true` is passed in the POST body to the `/attributes/restSearch` endpoint, the results _only_ include deleted attributes (attributes where a proposal has been made and accepted to "soft-delete" the attribute). 

### Expected behavior

Based on the description of the `deleted` parameter [here](https://github.com/MISP/MISP/blob/2.4/app/Controller/Component/RestResponseComponent.php#L1055) and [here](https://github.com/MISP/MISP/blob/2.4/docs/API_Doc.md?plain=1#L119) I would expect the `deleted` param to include attributes marked as deleted:true AND deleted:false. 

If that is not desired, I would suggest changing the docs and tooltip to read more like that for the `to_ids` param, e.g., "the state of the `deleted` flag. Further, I would then request something like an `includeDeleted` param that functions as I described returning attributes marked both deleted and not. 

Just a tangential note: I would have expected deleted Events to also get soft-deleted but it appears they are actually removed from the database. Therefore, searching `/attributes/restSearch` with `deleted: true` with the expectation to find attributes included in deleted Events will not yield anything. 

### Steps to reproduce

```
HTTP/1.1 POST /attributes/restSearch

{
    "returnFormat": "json",
    "enforceWarninglist": "true",
    "last_seen": ["30d", "0s"],
    "limit": "1000",
    "deleted": true
}
```

### Version

2.4.179

### Operating System

CentOS

### Operating System version

Stream 8

### PHP version

7.4

### Browser

Edge, Chrome, Firefox

### Browser version

latest

### Relevant log output

_No response_

### Extra attachments

_No response_

### Code of Conduct

Wachizungu:
#### What does it do?

Fix documentation for /events/restSearch and /attributes/restSearch 'deleted' filter.

fix #9437

I am not sure whether these app/Locale files should be updated as well, but included them in the PR just in case.

#### Questions

### DIFF
--- a/app/Controller/Component/RestResponseComponent.php
+++ b/app/Controller/Component/RestResponseComponent.php
@@ -1052,7 +1052,7 @@ class RestResponseComponent extends Component
                 'input' => 'radio',
                 'type' => 'integer',
                 'values' => array(1 => 'True', 0 => 'False' ),
-                'help' => __('Include deleted elements')
+                'help' => __('Default value 0. If set to 1, only soft-deleted attributes will be returned. If set to [0,1] , both deleted and non-deleted attributes wil be returned')
             ),
             'delta_merge' => array(
                 'input' => 'radio',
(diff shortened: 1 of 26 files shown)


cvandeplas:
Thank you very much!

Issue #9503, opened 2024-01-18
Feature Request: Support first_seen, last_seen filters for pymisp.search(controller='objects')


github-germ:
### Is your feature request related to a problem? Please describe.

Add server support to search objects by `first_seen` and/or `last_seen`.

### Describe the solution you'd like

Enhance server `MISPObject.php` filter for `restSearch` to support `first_seen` and/or `last_seen`. For example,`pymisp.search(controller='objects', first_seen='24h')` should retrieve only those objects with `first_seen` revised in the past 24 hours. Currently, these are ignored and ALL objects independent of the arg are returned from the server since those are not included in the filter rule.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

github-germ:
@iglocska Looks like https://github.com/MISP/MISP/pull/9564/commits/bf6a148bc8bea67f72215d91d4e05d8d3b464e4d enabled support for `first_seen`. Can we get `last_seen` also?

github-germ:
@iglocska `first_seen` filter works, `last_seen` not working in 2.4.200. This would still be nice to have. Thanks.

iglocska:
> @iglocska Looks like [bf6a148](https://github.com/MISP/MISP/commit/bf6a148bc8bea67f72215d91d4e05d8d3b464e4d) enabled support for `first_seen`. Can we get `last_seen` also?

Hmm this should work in theory, the change just catches the first_seen parameters on both scopes in the fix, last seen should fall into the "else" case. But I'll have a look!

iglocska:
Confirmed not working, what the hell.

PR #9564, opened 2024-02-14
fix: fix objects restsearch first_seen filter

Wachizungu:
#### What does it do?

Makes objects restsearch's first_seen filter work the same way as attribute first_seen (meaning if you use "2d" it should return those with a first_seen timestamp >= current time - 2d

Looks like a small oversight in commit 713a9f4 , which should do some of the lifting for #9503 I believe?

#### Questions

### DIFF
--- a/app/Model/Attribute.php
+++ b/app/Model/Attribute.php
@@ -2428,7 +2428,7 @@ class Attribute extends AppModel
             $conditions['AND'][] = array($scope . ' <=' => $timestamp[1]);
         } else {
             $timestamp = intval($this->resolveTimeDelta($timestamp)) * 1000000; // seen in stored in micro-seconds in the DB
-            if ($scope == 'Attribute.first_seen') {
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks a lot!

github-germ:
@iglocska Maybe I'm missing something here:
1. I don't think the addition in this PR is in the 2.4.196 code when I look at https://github.com/MISP/MISP/blob/v2.4.196/app/Model/Attribute.php#L2493
2. Can `last_seen` also be a condition?

Wachizungu:
Hi,

1:
It is I think, it's just the function below:
https://github.com/MISP/MISP/blob/v2.4.196/app/Model/Attribute.php#L2523
or permalink:
https://github.com/MISP/MISP/blob/4e8690e6d5e5c41b104ff65d64564604ac19668d/app/Model/Attribute.php#L2523
2:
Should be yes.

Does something look broken?

PR #9571, opened 2024-02-18
fix: [workflow] fix attribute edit module actions

### LINKED ISSUE #9547: Bug: Workflows: Unable to add attribute comment via event publish trigger
### Actual behavior

Attribute comments are not changed although they are of the correct type and the trigger is triggered.

### Expected behavior

I'm trying to setup a simple workflow that updates attributes of type hostname comments on each publish of an event.
Maybe I'm misunderstanding the logic of the trigger, and building blocks?
Setup:
- the attribute comment operation is directly linked to the event publish trigger
- I use the selector `Event._AttributeFlattened.{n}` in attribute comment operation
- and filter on `hostname` equals `type` (this should select all attributes of hostname)

The logs indicate "partial-success", but I don't immediately see a cause for the "partial".

### Steps to reproduce

- Event Publish trigger
- Add Action "Attribute comment operation"
- Edit comment, and add filter as described above

Unpublish and publish an event.

Note, the same **problem** also occurs when using the **Attribute IDS Flag operation** (*remove* IDS)

### Version

2.4.183

### Operating System

Ubuntu

### Operating System version

20.04

### PHP version

7.4

### Browser

_No response_

### Browser version

_No response_

### Relevant log output

```shell
Save the trigger as blueprint, this is the content


{
    "id": 8,
    "name": "Attribute comment operation",
    "data": {
        "node_uid": "zc13llln1i0eofptur8rn8",
        "indexed_params": {
            "comment": "My comment for an attribute "
        },
        "saved_filters": {
            "selector": "Event._AttributeFlattened.{n}",
            "value": "hostname",
            "operator": "equals",
            "path": "type"
        },
        "module_type": "action",
        "id": "Module_attribute_comment_operation",
        "name": "Attribute comment operation",
        "multiple_output_connection": false,
        "previous_module_version": "0.1",
        "module_version": "0.1"
    },
    "class": "block-type-default expect-misp-core-format",
    "typenode": false,
    "inputs": {
        "input_1": {
            "connections": [
                {
                    "node": "1",
                    "input": "output_1"
                }
            ]
        }
    },
    "outputs": {
        "output_1": {
            "connections": []
        }
    },
    "pos_x": -640,
    "pos_y": -552
}
```
```


### Extra attachments

_No response_

### Code of Conduct

Wachizungu:
#### What does it do?

Draft attempt to fix #9547

If someone could do a step by step review that would be nice though as I saw there is multiple editAttribute functions, editAttributeBulk as well.

#### Questions

### DIFF
--- a/app/Model/WorkflowModules/action/Module_attribute_edition_operation.php
+++ b/app/Model/WorkflowModules/action/Module_attribute_edition_operation.php
@@ -39,22 +39,17 @@ class Module_attribute_edition_operation extends WorkflowBaseActionModule
     protected function __saveAttributes(array $attributes, array $rData, array $params, array $user): array
     {
         $success = false;
-        $attributes = [];
         $newAttributes = [];
         foreach ($attributes as $k => $attribute) {
             $newAttribute = $this->_editAttribute($attribute, $rData, $params);
-            $newAttributes[] = $newAttribute;
             unset($newAttribute['timestamp']);
-            $result = $this->Attribute->editAttribute($newAttribute, $rData, $user, $newAttribute['object_id']);
-            if (is_array($result)) {
-                $attributes[] = $result;
-            }
+            $newAttributes[] = $newAttribute;
         }
-        $this->Attribute->editAttributeBulk($attributes, $rData, $user);
+        $this->Attribute->editAttributeBulk($newAttributes, $rData, $user);
         foreach ($attributes as $k => $attribute) {
             $saveSuccess = empty($this->Attribute->validationErrors[$k]);
             if ($saveSuccess) {
-                $rData = $this->_overrideAttribute($attribute, $newAttribute, $rData);
+                $rData = $this->_overrideAttribute($attribute, $newAttributes[$k], $rData);
             }
             $success = $success || !empty($saveSuccess);
         }


Wachizungu:
While this change seems to make the edit operation work, subsequent tag operation no longer works as expected after this change (I'm testing with 1 separate attribute and 2 attributes in a file object in 1 event)

mokaddem:
Many thanks for the PR.
In addition to your fix, I [restored](https://github.com/MISP/MISP/commit/752807ef37cc351e8d6e6bb8e1913b7444dd8fa8) the call to `Attribute->editAttribute` to mimic what `Event->_edit` does.

Wachizungu:
Thanks! Works for me!

One note:

https://github.com/MISP/MISP/blob/752807ef37cc351e8d6e6bb8e1913b7444dd8fa8/app/Model/WorkflowModules/action/Module_attribute_edition_operation.php#L49

$attributes is not used anymore later

I'm guessing if it should be like in event->_edit then that line should be $newAttributes[] = $result;
and this line to be deleted:
https://github.com/MISP/MISP/blob/752807ef37cc351e8d6e6bb8e1913b7444dd8fa8/app/Model/WorkflowModules/action/Module_attribute_edition_operation.php#L46

PR #9575, opened 2024-02-20
fix: [UI] Custom logos

JakubOnderka:
#### What does it do?

After moving images to different folder, there is a lot of broken functionalities. This should fix them.

#### Questions


matthijsvp:
Hi, is this merge in the current release (.189)? I'm still having issues with uploaded files – the web UI throws errors:
image

PR #9582, opened 2024-02-23
Minor documentation changes; add example to create users via REST API

cudeso:
#### What does it do?

- Minor documentation changes
- Add example to create users via REST API

#### Questions


adulau:
Thank you!

PR #9589, opened 2024-02-25
new: [ Workflows ] Add shadow attribute before save trigger for Workflows

vincenzocaputo:
#### What does it do?

This pull request adds a new workflow trigger that is called just before a Shadow Attribute is saved. I think this can be useful to design workflows to prevent the creation of proposal under certain conditions.

#### Questions

### DIFF
--- a/app/Model/ShadowAttribute.php
+++ b/app/Model/ShadowAttribute.php
@@ -196,6 +196,23 @@ class ShadowAttribute extends AppModel
 
         // convert into utc and micro sec
         $this->data = $this->Attribute->ISODatetimeToUTC($this->data, $this->alias);
+
+        $trigger_id = 'shadow-attribute-before-save';
+        $isTriggerCallable = $this->isTriggerCallable($trigger_id);
+        if ($isTriggerCallable) {
+            $triggerData = $this->data;
+            $workflowErrors = [];
+            $logging = [
+                'model' => 'ShadowAttribute',
+                'action' => 'add',
+                'id' => $this->data['ShadowAttribute']['id'],
+                'message' => __('The workflow `%s` prevented the saving of this proposal.', $trigger_id)
+            ];
+            $workflowSuccess = $this->executeTrigger($trigger_id, $triggerData, $workflowErrors, $logging);
+            if (!$workflowSuccess) {
+                return false;
+            }
+        }
         return true;
     }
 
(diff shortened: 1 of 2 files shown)


app/Model/WorkflowModules/trigger/Module_shadow_attribute_before_save.php
```diff
@@ -0,0 +1,45 @@
+<?php
+include_once APP . 'Model/WorkflowModules/WorkflowBaseModule.php';
+
+class Module_shadow_attribute_before_save extends WorkflowBaseTriggerModule
+{
+    public $id = 'shadow-attribute-before-save';
+    public $scope = 'attribute';
```
mokaddem: change this to either `proposal` or `shadow_attribute`

app/Model/WorkflowModules/trigger/Module_shadow_attribute_before_save.php
```diff
@@ -0,0 +1,45 @@
...
+    public $name = 'Shadow Attribute Before Save';
+    public $description = 'This trigger is called just before a Shadow Attribute is saved in the database';
+    public $icon = 'cube';
```
mokaddem: Change `cube` to `comment`

app/Model/ShadowAttribute.php
```diff
@@ -196,6 +196,23 @@ public function beforeSave($options = array())
...
+
+        $trigger_id = 'shadow-attribute-before-save';
+        $isTriggerCallable = $this->isTriggerCallable($trigger_id);
+        if ($isTriggerCallable) {
+            $triggerData = $this->data;
+            $workflowErrors = [];
+            $logging = [
+                'model' => 'ShadowAttribute',
+                'action' => 'add',
+                'id' => $this->data['ShadowAttribute']['id'],
```
mokaddem: This might throw an exception if we're adding a new proposal since it won't get an ID yet.

mokaddem (commented):
Thanks for the good PR! I also thing it would be useful to have a trigger for proposals.
Please find some feedback. It's mostly related to visuals.
Cheers!

vincenzocaputo:
Hi! I think to have fixed these points. Many thanks!

mokaddem:
Hey there!
Thanks you very much for your contribution. It has been merged into `develop` (448b5dbdf0a0e8ee821ef5fb0d5d2a5fbef64827). It won't take long until it gets released on `2.4`.
Cheers!

PR #9602, opened 2024-03-01
Update Event.php

karenyousefi:
fix error Undefined offset: 0 in [/var/www/MISP/app/Model/Event.php, line 3682]

`Notice: Undefined offset: 0 in [/var/www/MISP/app/Model/Event.php, line 3682]
Trace:
ErrorHandler::handleError() - APP/Lib/cakephp/lib/Cake/Error/ErrorHandler.php, line 230
Event::updatedLockedFieldForAllAnalystData() - APP/Model/Event.php, line 3682
...`

As a user, I encountered an issue where an undefined offset notice was being logged, particularly when attempting to import an event via a JSON file into MISP. To address this, I implemented checks to ensure the validity of array indices before accessing them. The code now verifies if the 'Object' array is not empty before iterating over its elements. Additionally, it checks if each element within the 'Object' array exists before updating it using the updatedLockedFieldForAnalystData function. Furthermore, similar checks were added for the 'Attribute' array within each 'Object' element to prevent further errors.

These changes aim to eliminate the undefined offset notice error, especially during event imports via JSON files, and improve the robustness of the code by handling potential edge cases where array elements might not exist.

I tested the updated code locally to ensure it resolves the reported issue without introducing any regressions.

### DIFF
--- a/app/Model/Event.php
+++ b/app/Model/Event.php
@@ -3679,7 +3679,9 @@ class Event extends AppModel
         }
         if (!empty($event['Event']['Object'])) {
             for ($i=0; $i < count($event['Event']['Object']); $i++) { 
-                $event['Event']['Object'][$i] = $this->updatedLockedFieldForAnalystData($event['Event']['Object'][$i]);
+                 if (isset($event['Event']['Object'][$i])) {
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks a lot, well spotted!

PR #9615, opened 2024-03-13
fix: Attachments deletion when accepting a delegation request

vincenzocaputo:
#### What does it do?

This pull request fixes an issue that causes the deletion of attachments when accepting an Event delegation request.

#### Questions


adulau:
Thank you! Good catch.

PR #9636, opened 2024-03-23
fix: [doc:rhel-installer] Correct conditional addition of httpd Liste…

Wachizungu:
…n 443 line

#### What does it do?

Hope these are the correct files to edit to be included in the installer.
Fixes an issue with rhel installer which has been there for a while. Installation added a line 'a Listen 443', so the httpd service didn't start properly, giving an error.

This should still keep the addition conditional. 

#### Questions


adulau:
Awesome! Thank you very much.

PR #9641, opened 2024-03-25
chg: [docs:new-background-workers] add rhel specific steps to migrati…

Wachizungu:
…on guide

#### What does it do?

Adds RHEL specific steps to the new background workers migration guide.

@righel

#### Questions


adulau:
Thank you!

PR #9664, opened 2024-04-05
OIDC: customize usernameProperty, add default role, use default_org, 

ftoppi:
#### What does it do?

Add OIDC config setting `username_property` to define which claim to use. Defaults to `email` to keep the same behavior.
Add OIDC config setting `default_role` in case IdP does not send the claim `roles_property` AND IdP userInfo cannot be queried. Defaults to `null` to keep the same behavior.
Use OIDC config setting `default_org` in case IdP does not send the claim `organization_property`.
Does not fail if IdP userInfo cannot be queried.

#### Questions

- [ ] Does it require a DB change? NO.
- [ ] Are you using it in production? YES.
- [ ] Does it require a change in the API (PyMISP for example)? NO.

### DIFF
--- a/app/Plugin/OidcAuth/Lib/Oidc.php
+++ b/app/Plugin/OidcAuth/Lib/Oidc.php
@@ -50,7 +50,7 @@ class Oidc
         }
 
         $organisationProperty = $this->getConfig('organisation_property', 'organization');
-        $organisationName = $claims->{$organisationProperty} ?? null;
+        $organisationName = $claims->{$organisationProperty} ?? $this->getConfig('default_org');
 
         $organisationUuidProperty = $this->getConfig('organisation_uuid_property', 'organization_uuid');
         $organisationUuid = $claims->{$organisationUuidProperty} ?? null;
--- a/app/Plugin/OidcAuth/Lib/Oidc.php
+++ b/app/Plugin/OidcAuth/Lib/Oidc.php
@@ -27,7 +27,8 @@ class Oidc
 
         $claims = $oidc->getVerifiedClaims();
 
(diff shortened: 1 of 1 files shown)


JakubOnderka:
Can you please elaborate why [b1fcdd7](https://github.com/MISP/MISP/pull/9664/commits/b1fcdd783b617e7efdfce2b853162fe718a37dcf) is needed?

ftoppi:
> Can you please elaborate why [b1fcdd7](https://github.com/MISP/MISP/pull/9664/commits/b1fcdd783b617e7efdfce2b853162fe718a37dcf) is needed?

If IdP doesn't send the org property, the connection is immediately rejected even if a default org is configured. IMHO, "default org" should be used if none is sent by IdP.

JakubOnderka:
Default organisation is set in https://github.com/MISP/MISP/blob/8934982ff2606b89399e646fdf88b6524a4dad30/app/Plugin/OidcAuth/Lib/Oidc.php#L59

So can you test it again?

ftoppi:
> Default organisation is set in
> 
> https://github.com/MISP/MISP/blob/8934982ff2606b89399e646fdf88b6524a4dad30/app/Plugin/OidcAuth/Lib/Oidc.php#L59
> 
> So can you test it again?

MISP rejects the connection if IdP does not send org name:

https://github.com/MISP/MISP/blob/8934982ff2606b89399e646fdf88b6524a4dad30/app/Plugin/OidcAuth/Lib/Oidc.php#L57

https://github.com/MISP/MISP/blob/8934982ff2606b89399e646fdf88b6524a4dad30/app/Plugin/OidcAuth/Lib/Oidc.php#L346

PR #9677, opened 2024-04-12
new: [feed:pullEvents] Added support of tag collection in feed configuration

mokaddem:
This allows to specify a tag collection for which all the tags will be applied on the pulled Events

### DIFF
--- a/app/Controller/FeedsController.php
+++ b/app/Controller/FeedsController.php
@@ -109,17 +109,14 @@ class FeedsController extends AppController
                 }
 
                 foreach ($feeds as &$feed) {
-                    if (!empty($feed['Feed']['tag_id'])) {
-                        if (substr($feed['Feed']['tag_id'], 0, 11) == 'collection_') {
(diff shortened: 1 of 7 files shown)


iglocska:
That change is a bit tricky, I'd rework it: instead of changing tag_id, could we add another field? There are two approaches I see working: a boolean that toggles whether we look at the tags table or the tag_collections or alternatively simply adding a tag_collection_id field would work too - and we attach whichever is set.

The change to varchars complicates things, makes lookups / searches more expensive and size wise it's going to be worse than the other options above (though since we're talking about the feeds table it's tiny either way). Still, if we could change to any of those it would be cleaner. 

My personal preference is in this order:

add tag_collections_id > add tag_source_is_collection > tag_id conversion to string 

Also, tag_collections_id wouldn't make any changes to existing APIs, it would rather just add an alternative.

mokaddem:
Ready for next round @iglocska

PR #9694, opened 2024-04-19
chg: [config] Set Oidc issuer

christianmg99:
#### What does it do?

It allows to specify an OIDC issuer other than the `provider_url`, it still defaults to `provider_url` if issuer is not specified, making the update transparent for current deployments.

#### Questions

### DIFF
--- a/app/Plugin/OidcAuth/Lib/Oidc.php
+++ b/app/Plugin/OidcAuth/Lib/Oidc.php
@@ -302,9 +302,10 @@ class Oidc
         $providerUrl = $this->getConfig('provider_url');
         $clientId = $this->getConfig('client_id');
         $clientSecret = $this->getConfig('client_secret');
+        $issuer = $this->getConfig('issuer', $providerUrl);
 
         if (class_exists("\JakubOnderka\OpenIDConnectClient")) {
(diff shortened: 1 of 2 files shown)


app/Plugin/OidcAuth/Lib/Oidc.php
```diff
@@ -302,9 +302,10 @@ private function prepareClient()
         $providerUrl = $this->getConfig('provider_url');
         $clientId = $this->getConfig('client_id');
         $clientSecret = $this->getConfig('client_secret');
+        $issuer = $this->getConfig('issuer', $providerUrl);
```
JakubOnderka, line 305: It is no need to use second argument `$providerUrl` is if `issuer` is not set, `$issuer` will be `null` as OpenIDConnectClient library expects.

christianmg99: I thought the same at first, but then I noticed that `getConfig()` would throw an exception even before reaching the library:
https://github.com/MISP/MISP/blob/471840ce33233a8d9db86766d007153f66727e8d/app/Plugin/OidcAuth/Lib/Oidc.php#L511

JakubOnderka: You are correct, merged.

PR #9697, opened 2024-04-21
fix: [galaxy_clusters] Add orgc filter option for index, set it as de…

Wachizungu:
…fault for galaxy view 'My Clusters'

#### What does it do?

As per discussion on gitter.
Note that I also updated the "My Clusters" button which is on the galaxy view page for example. This is also more aligned with events index (org events there means only events created by your org, not owned by).

#### Questions


iglocska:
Thanks a lot!

PR #9717, opened 2024-04-27
fix: [galaxyClusters:restSearch] filter on org_id and orgc_id if para…

Wachizungu:
…m set

#### What does it do?

allows filtering on org_id and orgc_id when using galaxy clusters restsearch (is currently broken because of wrong param names)

#### Questions


iglocska:
Thanks a lot!

Issue #9672, opened 2024-04-11
Feature Request: Event view optimizations for large number of objects


cyberjv:
### Is your feature request related to a problem? Please describe.

Some of my organization's use cases have events containing a large number of objects. With many different object templates being used repeatedly the event becomes a confusing mess of different mixed objects.
The event view needs some updates to simplify managing a complex event with a large number of objects.

### 1. Identifying field or name for each object
Currently in the object header the 'Object name' field is actually populated by the object template name. As a result all the objects with the same object template have the same name, even if the object has an attribute called name:
![image](https://github.com/MISP/MISP/assets/166504219/fc912af8-0e8d-4a88-b452-07dca3d9aa50)

_Proposed solution:_
Add a name/value to the object header that identifies the specific object instance compared to other objects of the same type. This could be one of the attributes of the object ('name' attribute on the above screenshot for example)
Rename the current 'Object name' in the header to 'Object type' or 'Object template'

### 2. Collapsible objects in the event view
Currently all the attributes of all the objects are visible at all times. This makes it difficult for the user to get an overview of all the objects and relationships. If objects can have specific names it would make sense to be able to collapse them to make it possible to have an overview of many events at once.
Being able to close objects also helps if the user has to go over each object for some reason, they could close the objects that have been processed/reviewed as they go, keeping track of the progress
_Proposed solution:_
Add a 'collapse object' button to the object header. Upon collapsing the header stays, but the style changes noticeably. (Solid background of the header turns to white box with boders for example)

### 3. Tags on objects
Add a button to add/remove a tag on all attributes of the object at once. If a tag is present for all attributes of an object at once, show it in the object header. 
This way 'views' of objects can easily be created to filter sets of objects effectively using the filtering tool

### 4. Object pinning and/or ordering
Currently the objects can be ordered based on the columns of the table the objects are in. With the default being 'order by latest changed attribute' the order of the objects changes all the time, adding to the confusion.
_Proposed solution:_
Add ability to pin objects to the top or bottom of the table.
Also add functionality to be able to set the order of objects manually (sample: Extra column with user editable values for ordering, hidden when empty)

### Describe alternatives you've considered

The event graph view is helpful with getting an overview of a lot of objects, it even tries to display identifying information next to the title of each object to make it possible to distinguish them. The weakness of that view is looking at specific details of objects and comparing them.

These would be some helpful updates to the Object Graph view:
[...]

PR #9750, opened 2024-05-22
new: [events:view] New UI feature allowing to collapse Attributes con…

mokaddem:
It comes with a MISP setting (`MISP.collapse_attribute_in_object`) to toggle a different behavior for the instance.
By default, Objects are not collapsed.
The Attribute included in the Object's row will always be the one having the highest `ui-priority` defined in its template. If there is none or if multiple Attributes have the same value, the display might be random.

**Collapsed view**
Screenshot from 2024-05-22 15-37-05

**Expanded view** (normal view)
Screenshot from 2024-05-22 15-37-27

### DIFF
--- a/app/Model/Server.php
+++ b/app/Model/Server.php
@@ -5930,6 +5930,14 @@ class Server extends AppModel
                     'type' => 'boolean',
                     'null' => true
                 ),
+                'collapse_attribute_in_object' => array(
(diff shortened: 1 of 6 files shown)


iglocska:
Thanks a lot!

adulau:
Related to #9672 - closing point number 2

PR #9758, opened 2024-05-30
Update ModulesController.php

Katz-Tal:
Hey, 
Please see this as well: https://github.com/MISP/misp-docker/pull/67

Fix issue with MISP modules (ModulesController.php file)

We had an error when using MISP modules (modules/queryEnrichment API endpoint)
After full investigation, this fixed our error.
I think it's very important to add it, since we would like to use the updated official version of MISP.

This is the full error, take from `error.log`:

```
Error: [TypeError] Argument 3 passed to Module::canUse() must be of the type array, string given, called in /var/www/MISP/app/Controller/ModulesController.php on line 25
Request URL: /modules/queryEnrichment
Stack Trace:
#0 /var/www/MISP/app/Controller/ModulesController.php(25): Module->canUse()
#1 [internal function]: ModulesController->queryEnrichment()
#2 /var/www/MISP/app/Lib/cakephp/lib/Cake/Controller/Controller.php(499): ReflectionMethod->invokeArgs()
#3 /var/www/MISP/app/Lib/cakephp/lib/Cake/Routing/Dispatcher.php(193): Controller->invokeAction()
#4 /var/www/MISP/app/Lib/cakephp/lib/Cake/Routing/Dispatcher.php(167): Dispatcher->_invoke()
#5 /var/www/MISP/app/webroot/index.php(101): Dispatcher->dispatch()
#6 {main}
After removing lines 25-27 on /var/www/MISP/app/Controller/ModulesController.php the issue was solved and we got MISP modules working again.
```

Thanks!

#### What does it do?

#### Questions

### DIFF
--- a/app/Controller/ModulesController.php
+++ b/app/Controller/ModulesController.php
@@ -22,9 +22,6 @@ class ModulesController extends AppController
         if (!Configure::read('Plugin.Enrichment_' . $modname . '_enabled')) {
             throw new MethodNotAllowedException('Module not found or not available.');
         }
-        if (!$this->Module->canUse($this->Auth->user(), 'Enrichment', $modname)) {
-            throw new MethodNotAllowedException('Module not found or not available.');
-        }
         $options = array();
 
         // Check module


ostefano:
CC @iglocska

iglocska:
Hmm, that doesn't seem like a valid solution, it simply removes the permission check, which isn't intended. I'll fix it in a few moments properly so that it correctly passes the module to the check.

PR #9764, opened 2024-06-05
chg: [diagnostics] add Database/MysqlObserverExtended to valid data s…

Wachizungu:
…ources list

#### What does it do?
Makes sure the datasource shows as okay on diagnostics page when using Database/MysqlObserverExtended

#### Questions


iglocska:
Thanks a lot!

PR #9771, opened 2024-06-10
fix: [doc] correct filenames in rhel background worker migration guid…

Wachizungu:
…e steps

#### What does it do?

Edit filenames so they match policy module name, otherwise running the related commands gives an error (my bad, thought to use a clearer policy name and then didn't copy everywhere).

#### Questions


adulau:
Thank you!

PR #9801, opened 2024-07-01
chg: [Attributes restSearch] added sort support for publish_timestamp

Benni0:
#### What does it do?

Querying attributes in a polling manner without loosing Attributes is difficult using the current API for the following reasons:
- Attribute timestamp cannot be used as restart filter, as sync mechanisms can import attributes with older timestamps
- Attribute id cannot be used, as this would loose all attribute updates.
- Event publish_timestamp could be a possible restart point, but it is currently not possible to sort after publish_timestamp.

This change adds the ability to sort queried attributes regarding the Event.publish_timestamp field and adds the Event.publish_timestamp field automatically to each attribute.
Due to this change it would be possible to query attributes since a specific publish_timestamp and sorted after this field in a paged manner. On the next interval, the polling can be restarted by the latest publish timestamp without loosing any attributes.

#### Questions

It adds the Event.publish_timestamp to each Attribute

#### Note:
I tried to avoid to modify the  `AppModel:findOrder` function first, but it would make sense to modify this function to make it more generic.

### DIFF
--- a/app/Model/AppModel.php
+++ b/app/Model/AppModel.php
@@ -4186,20 +4186,46 @@ class AppModel extends Model
 
     public function findOrder($order, $orderModel, $validOrderFields)
     {
-        if (!is_array($order)) {
-            $orderRules = explode(' ', strtolower($order));
-            $orderField = explode('.', $orderRules[0]);
-            $orderField = end($orderField);
-            if (in_array($orderField, $validOrderFields, true)) {
-                $direction = 'asc';
-                if (!empty($orderRules[1]) && trim($orderRules[1]) === 'desc') {
-                    $direction = 'desc';
+        if (is_string($order)) {
+            $orderRules = explode(',', $order);
+        } elseif (is_array($order)) {
+            $orderRules = $order; // to support multiple column order
+        }
+
+        $order = array();
+        foreach ($orderRules as $rule) {
+            if (!is_string($rule)) {
+                return null;
+            }
+            $ruleItems = explode(' ', trim($rule));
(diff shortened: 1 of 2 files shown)


Benni0:
@iglocska,
could you please have a look at this pull-request and tell me if I have a chance to get this merged?
This would solve some of my continuous indicator export problems.

Thank you!

ostefano:
+1, had to resort to `events/index` for my use case

Benni0:
@ostefano 
due to the latest changes in commit [1787811](https://github.com/MISP/MISP/commit/178781175e3301b0eac48059ccc4eb9765762a02) the order of attributes is fixed to `Attribute.id asc`.
It seems, the order parameter is completely ignored.

I would anyways recommend to fetch the events ordered by their publish timestamp, and then fetch the attributes for each event. During my tests it came out, this procedure is much faster than sorting attributes regarding the events publish timestamp.

iglocska:
> @ostefano due to the latest changes in commit [1787811](https://github.com/MISP/MISP/commit/178781175e3301b0eac48059ccc4eb9765762a02) the order of attributes is fixed to `Attribute.id asc`. It seems, the order parameter is completely ignored.
> 
> I would anyways recommend to fetch the events ordered by their publish timestamp, and then fetch the attributes for each event. During my tests it came out, this procedure is much faster than sorting attributes regarding the events publish timestamp.

The fixed attribute ID ordering has been reverted and it instead simply acts as a default. As thus it can be merged, I've done so now on develop. Thanks a lot!

PR #9841, opened 2024-07-25
new: Comment field for Bookmarks

janphiz:
This is my very first commit. I tried my best and hope it works.

#### What does it do?

I tried to add a comment field to the new Bookmark feature. Analysts could use this feature to add additional context on why they saved and shared a bookmark with their org. For example internal ticket ID, actions that have to be taken on the event, etc. 

#### Questions

   Maybe, since I added a new field for additional information.
   No.
   No, as far as I know is the bookmark feature not yet part of pyMISP. If so then yes.

### DIFF
--- a/app/View/Bookmarks/add.ctp
+++ b/app/View/Bookmarks/add.ctp
@@ -12,6 +12,11 @@ $fields = [
         'class' => 'input span6',
     ],
     [
+        'field' => 'Comment',
+        'type' => 'textarea',
+        'class' => 'input span6',
+    ],
+    [
         'field' => 'exposed_to_org',
         'type' => 'checkbox',
         'label' => __('Should this bookmark be exposed to all users from the organisation'),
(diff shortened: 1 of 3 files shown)


adulau:
It's a good point indeed but it seems your PR is just covering the UI and not the database part ;-)

janphiz:
Yes, that's what I meant with the the answer to DB change question. I haven't worked my way up to the DB part yet, but wanted to contribute the idea already :-)

adulau:
Thanks for the idea. @mokaddem will look at it and extend it if required.

mokaddem:
Hey there,
Thanks for the idea an the initial PR!
I added the missing parts and merged it into `develop` (https://github.com/MISP/MISP/commit/82718067ac3fb9c5fe4cd23c4c99047fc75daa03).
Cheers!

PR #9851, opened 2024-08-05
Default to env dependencies, and fallback to submodules'

ostefano:
#### What does it do?

* Fix requirements.txt and requirements-dev.txt 
* Make all scripts default to env python dependencies but fallback to submodules' in case they are not installed


ostefano:
@iglocska experimental, to be tested first

PR #9865, opened 2024-08-15
chg: [internal] Include in logged message subject and e-mail address …

JakubOnderka:
…when sending e-mail

#### Questions


iglocska:
I dig the idea, thanks!

PR #9876, opened 2024-08-21
Fixes performance of sightings restSearch when performing MISP sync

tomking2:
#### What does it do?

There is a bug in the current sightings `restSearch` function when pulling sightings on a per-event basis. 

Instead of pulling sightings just for each event sequentially in a loop, sightings for all events are pulled repeatedly, based on the number of events requested. This has a significant knock-on effect, as the amount of data to process was multiplying based on the number of events. Significant impact on both memory consumption and processing time. As simple as a variable change to change the code to its intended function.

For example when I request 20 events - lets say there are 1000 sightings in total - 20000 sightings are returned. When de-duplicated on the client side this drops down to the 1000 as intended.

#### Questions


iglocska:
Thanks, very nice catch!!

PR #9884, opened 2024-08-27
[Correlation] Generate correlations through multiple requests

tomking2:
#### What does it do?
This one might be up for debate, but I think it's a healthy addition that ensures consistent performance across all MySQL implementations (be that MariaDB, MySQL, AWS RDS or Azure MySQL etc.).

We were spotting that during ingestion of events with severe over-correlating values, we were seeing a major degradation of performance when generating correlation values.

The current code looks for the value in either `value1` or `value2` via an OR clause, leaving MySQL to find the best optimisation strategy for performing this query. We found that on our MySQL DB (which wasn't MariaDB), the optimisation strategy was not optimal for every request. It appeared that for over-correlating IOCs, instead of using a `sort_union` strategy on the `value1` and `value2` indexes, it was instead opting to use the `event_id` as the selected index. I couldn't find a consistent method to force the correct strategy.

What did this mean for our MISP instance? Instead of correlations taking <= 0.1s per attribute, it was taking around 5-7s per attribute. When trying to ingest an event with 4K objects and 80K attributes (every object had high correlations), it would've taken around 6 days to finish ingestion.

So what does this PR do? Instead of relying upon SQL to find a suitable optimisation strategy for the `value` conditions within the OR clause, separate DB calls are made for each value condition, enabling straightforward indexes to be used for the query. Correlation limits and bits like ACL are still adhered to during the lookups.

Potential impact - While this will ensure consistent performance when generating correlations, for MariaDB that appears to consistently create well-optimised queries (although I haven't done much validation here), a slight reduction in performance may be sighted as it now must make up to 3 DB calls instead of just the 1. However given many MISP instances will not be using MariaDB, I reckon this is a good tradeoff.

#### Questions

### DIFF
--- a/app/Model/Correlation.php
+++ b/app/Model/Correlation.php
@@ -289,13 +289,6 @@ class Correlation extends AppModel
         // stupid hack to allow statically retrieving the constants
         ClassRegistry::init('Attribute');
         $conditions = [
-            'OR' => [
-                'Attribute.value1' => $value,
-                'AND' => [
-                    'Attribute.value2' => $value,
-                    'NOT' => ['Attribute.type' => Attribute::PRIMARY_ONLY_CORRELATING_TYPES]
-                ]
-            ],
             'NOT' => [
                 'Attribute.type' => Attribute::NON_CORRELATING_TYPES,
             ],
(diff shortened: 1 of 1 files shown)


tomking2:
I've been chatting with someone on Gitter who is facing similar performance issues on calculation of correlations. They are using MariaDB.

Their query seems OK and creating the correct sort_union:

image

However the query is still extremely slow to return:
image

If we simulate splitting the request into multiple per field, we are seeing much better performance. For example this is one part of the query looking at `value1`.

image

Therefore I think this is a good change to make.

app/Model/Correlation.php
```diff
@@ -289,27 +289,45 @@ private function __getMatchingAttributes($value)
...
+            $correlatingAttributesPart = $this->Attribute->find('all', [
+                'conditions' => $conditionsPart,
+                'recursive' => -1,
+                'fields' => $this->getFieldRules(),
+                'contain' => $this->getContainRules(),
+                'order' => []
+            ]);
+
+            if (!empty($correlatingAttributesPart)) {
+                $correlatingAttributes = array_merge($correlatingAttributes, $correlatingAttributesPart);
```
JakubOnderka: Please use 
```php
array_push($correlatingAttributes, ...$correlatingAttributesPart);
```
This will not create new array, but instead will append `$correlatingAttributesPart` to existing array `$correlatingAttributes` – so this is faster and consumes less memory.

tomking2:
Thanks @JakubOnderka,

I've updated to use `array_push` as suggested.

tomking2:
@iglocska Shall I close this one off? I think it does the same thing as your recent change in v2.5.6?

Also curious what sparked your change here, were you seeing something similar to the performance issues above?

One thing we could pull from mine is accounting for `value2` and `extraConditions` in the over-correlating count, and perhaps using the result size to reduce limits

iglocska:
I should look more closely at PRs, totally missed that you had a solution for this. Yeah, it should be safe to close now. I was debugging with someone that had crazy low performance with the correlation generation due to the sort_union, mysql works in mysterious ways.

Also, to add an additional twist to it, it sadly needs 3 passes as we also have the advanced correlation rules. I so wish we would never have gone the value1|value2 way back in the day...

Issue #8598, opened 2022-09-20
Support: OpenID Connect Not Working/Docs Incomplete


jaimeatwork:
### Support Questions

I have attempted to configure OpenID connect but cannot seem to get it to work. 

I also had to enable Oidc auth in the bootstrap.php file by doing this

```
/**
 * Uncomment the following line to enable client SSL certificate authentication.
 * It's also necessary to configure the plugin �~@~T for more information, please read app/Plugin/CertAuth/reame.md
 */
CakePlugin::load('OidcAuth');
// CakePlugin::load('ShibbAuth');
// CakePlugin::load('LinOTPAuth');
```

My configuration, redacted:

```
....snip....
  'Security' =>
  array (
    'level' => 'medium',
....snip....
    'auth' =>
    array (
      0 => 'OidcAuth.Oidc',
    ),
....snip....
  array (
    'provider_url' => 'https://xxxx.xxxxxxxx.com', // sso provider
    'client_id' => 'client_id',
    'client_secret' => 'client_secret',
    'role_mapper' =>
    array (
      'misp-admin' => 1,
      'misp-user' => 3,
    ),
    'default_org' => 'MyOrg',
  ),
```

I consistently get the error `Error: invalid_client Description: The issuer of the request object is invalid.`. It does not appear any of the requests include the `iss` value in the URL or elsewhere I can see. 

Additionally, I tested from inside the deployment, I can curl to the wellknown OIDC endpoint and see the expected issuer key and value in the response. 

Do I need to take any specific steps to ensure the issuer is sent along? Thanks!

### MISP version

2.4.162

### Operating System

Ubuntu

### Operating System version

20.04

### PHP version

7.4.3

### Browser

Chrome

### Browser version

105

### Relevant log output

```shell
2022-09-20 20:38:03 Error: [JakubOnderka\OpenIDConnectClientException] Error: invalid_client Description: The issuer of the request object is invalid.
Request URL: /users/login?state=<redacted>&error=invalid_client&error_description=The+issuer+of+the+request+object+is+invalid.
Stack Trace:
#0 /var/www/MISP/app/Plugin/OidcAuth/Lib/Oidc.php(22): JakubOnderka\OpenIDConnectClient->authenticate()
#1 /var/www/MISP/app/Plugin/OidcAuth/Controller/Component/Auth/OidcAuthenticate.php(32): Oidc->authenticate()
[...]

PR #9905, opened 2024-09-09
Add scopes to OIDC client

JordanBarnartt:

#### What does it do?

This PR addresses an issue with OIDC functionality.  Currently, the OIDC client only checks for claims in the default scope of the authentication.  However, depending on the IdP, necessary attributes like email and roles are not included in this scope.

This change allows the user to configure in the `config.php` file which scopes they want to include, and adds each of them to the OIDC client object.

This partially addresses the issue in https://github.com/MISP/MISP/issues/8598.

#### Questions

- [ ] Does it require a DB change?  No
- [ ] Are you using it in production?  Yes
- [ ] Does it require a change in the API (PyMISP for example)?  No


JakubOnderka:
Thanks for your request, will be integrated by https://github.com/MISP/MISP/pull/9912 into develop branch.

PR #9913, opened 2024-09-12
chg: [internal] Raise exception if invalid data received from module

JakubOnderka:
#### Questions

### DIFF
--- a/app/Model/Event.php
+++ b/app/Model/Event.php
@@ -6295,7 +6295,7 @@ class Event extends AppModel
         }
     }
 
-    public function enrichment($params)
+    public function enrichment(array $params)
     {
         $option_fields = array('user', 'event_id', 'modules');
         foreach ($option_fields as $option_field) {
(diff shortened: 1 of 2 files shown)


app/Model/Event.php
```diff
@@ -6372,12 +6372,11 @@ public function enrichment($params)
                             throw new MethodNotAllowedException(h($module['name']) . ' service not reachable.');
                         } else if (!is_array($result)) {
                             continue 2;
+                        } else if (!isset($result['results'])) {
+                            throw new RuntimeException("Invalid response received from module {$module['name']}, response data do not contains results field.");
```
m5050, line 6376: The `Enrich Event` functionally now is breaking when exception is returned by this line.
Earlier, the `Enrich Event` ignores errors and complete on processing the reminder of attributes and objects.
When you look at number of the misp-modules, its expected to return`misperrors['error']` for attributes that were unable to be enriched (e.g. [revresedns.py](https://github.com/MISP/misp-modules/blob/main/misp_modules/modules/expansion/reversedns.py)).

PR #9924, opened 2024-09-23
fix: [expansion:popover] Pass correct ID type for attributes. Fix #9919

### LINKED ISSUE #9919: Bug: Popover hover enrichment is broken
### Actual behavior

Popup with hover enrichments doesn't appear.

I think the issue is related to this code:
https://github.com/MISP/MISP/blob/ef48a1c69ed70395eaaa7513a53fbccacba66cae/app/webroot/js/misp.js#L4065

element selector expects something in format Attribute_<attribute_id>_container , but is now getting attributes_<attribute_id>_container, because of below change:
https://github.com/MISP/MISP/commit/2b1d2cb3442895134c6eba065ec4d2b7fefc222c

The params for the function call are set here:
https://github.com/MISP/MISP/blob/ef48a1c69ed70395eaaa7513a53fbccacba66cae/app/webroot/js/misp.js#L4085

I would propose a fix but I'm not sure which other types showHoverEnrichmentPopover(type, id) can get and how to handle them.

### Expected behavior

When hovering over an attribute for which a hover enrichment is enabled / available, a popup with the hover enrichment result is shown.

### Steps to reproduce

Set up misp-modules on your instance, enable hashlookup expansion module and hover enrichments.
Create an event and add attribute md5 with value 8ed4b4ed952526d89899e723f3488de4 .

Hover over the attribute. Popup should appear (but doesn't)

### Version

2.4.198

### Operating System

Ubuntu

### Operating System version

20.04

### PHP version

7.4

### Browser

_No response_

### Browser version

_No response_

### Relevant log output

_No response_

### Extra attachments

_No response_

### Code of Conduct

Wachizungu:
#### What does it do?

fixes #9919

#### Questions

### DIFF
--- a/app/webroot/js/misp.js
+++ b/app/webroot/js/misp.js
@@ -4083,6 +4083,7 @@ $(document.body).on('mouseenter', '.eventViewAttributeHover', function () {
         currentPopover = '';
     }
     var type = $(this).attr('data-object-type');
+    if (type==='attributes') type = 'Attribute'; // Type translation to expected input for further processing
     var id = $(this).attr('data-object-id');
 
     if (type + "_" + id in ajaxResults["hover"]) {


iglocska:
Thanks a lot!

PR #9929, opened 2024-09-24
fix: [Export] Allow proposal blocking in Bro/Zeek exports

liviuvalsan:
#### What does it do?

It allows proposal blocking for Zeek exports. Without this one line fix the proposals to remove the IDS flag or delete attributes do not have any effect for the export to Zeek.

#### Questions


iglocska:
Thanks, makes sense!

PR #9941, opened 2024-10-04
Update UPGRADE.ubuntu2404.sh

cudeso:
With  `if [ $INSTALL_SSDEEP ]; then` the condition will always be True ('check for a non-empty string'); needs to be `if $INSTALL_SSDEEP; then` (check for the result of the 'command' INSTALL_SSDEEP)

#### Questions

- [ ] Are you using it in production? Almost


cudeso:
https://github.com/MISP/MISP/pull/9950

PR #9950, opened 2024-10-04
fix: [Upgrade] Fix conditionals in 2.5 upgrade script

tomking2:
#### What does it do?

I'm pretty sure two of the conditional statements in the upgrade script are slightly wrong, as they currently always resolve to 'true' as it's a string evaluation. This is the SSDEEP and supervisor checks.

This should correctly evaluate if the string is `true` or not.

#### Questions

### DIFF
--- a/INSTALL/UPGRADE.ubuntu2404.sh
+++ b/INSTALL/UPGRADE.ubuntu2404.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 # MISP 2.5 upgrade for MISP 2.4 installations on Ubuntu 24.04 LTS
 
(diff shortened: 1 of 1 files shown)


cudeso:
https://github.com/MISP/MISP/pull/9941

tomking2:
Ah sorry @cudeso, didn't see your PR; you beat me to it! 

There is a second conditional that needs fixing regarding the supervisor switch though. If you'd like to add this one to your PR I'll close this to save on duplicate PRs. Alternatively you can close down your PR if you'd prefer.

iglocska:
Thanks a lot! Sorry for all the delays.

Issue #9942, opened 2024-10-04
Bug: MISP sharing group "active" field via API


linozinho:
### Actual behavior

I want to change the status to "false"/"true" of the field "active" in the sharing group via api but it doesn't work

`POST: https://misp.example/sharing_groups/edit/1`
BODY: 
```
 {
     "SharingGroup": {
        "id": "1",
        "active": false
    }  
  } 
```
  
  RESPONSE:
```
"SharingGroup": {
        "id": "1",
        "name": "TEST",
        "releasability": "string",
        "description": "Sharing group TESTE",
        "uuid": "07f2c6c1-a93c-440b-88cf-28a6514e9b73",
        "organisation_uuid": "03076bd2-bdec-49c7-9cf2-8ddb0271d62c",
        "org_id": "2",
        "sync_user_id": "2",
        "active": true,
        "created": "2024-10-04 10:13:48",
        "modified": "2024-10-04 10:48:01",
        "local": false,
        "roaming": true
    },
```
   
![Screenshot from 2024-10-04 11-26-29](https://github.com/user-attachments/assets/eeced712-40d2-44c0-afa9-61459ef854e3)

    
I've tried several "HTTP BODY¨ and not one of them worked

[MISP API Documentation](https://www.misp-project.org/openapi/#tag/Sharing-Groups/operation/addSharingGroup)

### Expected behavior

"active" field should be false

### Steps to reproduce

1- POST: https://misp.example/sharing_groups/edit/1
2- Use this HTTP body

```
 {
     "SharingGroup": {
        "id": "1",
        "active": false
    }  
  } 
```
3- Send Request
  

### Version

v2.4.197

### Operating System

Ubuntu

### Operating System version

Ubuntu 22.04.4 LTS

### PHP version

PHP 7.4.33 

### Browser

Chrome

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

linozinho:
[Problem found](https://github.com/MISP/MISP/issues/9958)

PR #9954, opened 2024-10-07
chg: [SharingGroup] Enable editing 'active' flag from API

tomking2:
#### What does it do?

Resolves request #9942 by enabling users to update the `active` flag of a sharing group via the API

#### Questions


tomking2:
Close as covered by a9a6c15fb352b8d2ce4dfdf600e800b78df702bf

PR #9965, opened 2024-10-08
Allow site admin to see authkey from other admins

JSCU-CNI:
#### What does it do?

Improves the fix from 8577790e75b50d57b71a5c82d2e4611b130983f7 by explicitly checking if the current user is not a site admin before redacting the authkey.

#### Questions


iglocska:
Thanks, merged into develop, it will be in the next release. Good catch!

JSCU-CNI:
> Thanks, merged into develop, it will be in the next release. Good catch!

Good to hear! Should we close this PR? Unsure how the git flow works for this project.

iglocska:
Hmm yeah, good question. I wonder why this wasn't auto-closed. Let's wait until the next release, maybe it wasn't merged in time for the last one!

JSCU-CNI:
Can we close this PR @iglocska?

Issue #9958, opened 2024-10-07
Sharing group API "active" and  "local"  attributes missing 


linozinho:
https://github.com/MISP/MISP/blob/1399a8d2f00c3f2ec628716b8836566d41888735/app/Model/SharingGroup.php#L665C1-L666C1

Missing 
` $attributes = ['active', 'local'];`                
without these fields active and local cannot be changed via API

![Screenshot from 2024-10-07 14-01-06](https://github.com/user-attachments/assets/b3d8d158-68b3-41e1-9762-c7c343096759)

PR #9966, opened 2024-10-08
FIX:  change flag 'local' and 'active' sharing group via api

linozinho:
#### Pull Request per:
https://github.com/MISP/MISP/blob/1399a8d2f00c3f2ec628716b8836566d41888735/app/Model/SharingGroup.php#L665C1-L666C1

#### What does it do?

allow update a sharing groups "active" / "local" flag via API
Issue: #9958

#### Questions


linozinho:
@adulau Is the PR acceptable?

PR #9973, opened 2024-10-11
Improvement: Add Sharing Group "default_event_distribution" and allow editing of "active" and "local" sharing group flags via API

linozinho:
## Improvement:  Add Sharing Group "default_event_distribution" and allow editing of "active" and "local" sharing group flags via API

The purpose of this pull request is to add the sharing method by sharing group to "MISP.default_event_distribution" as this option is missing. 

Screenshot from 2024-10-11 11-08-02

Untitled design (2)

Organize sharing groups by id because in my opinion it is more organized in get via API and there are organization options in the graphical interface that the user can filter for themselves in their own way.

Untitled design

Allow editing of "active" and "local" sharing group flags via API

```
        "name": "testetetdtadsfasdfasdf",
        "releasability": "test",
        "description": "test",
        "active": true, //this
        "local": false, //this
        "roaming": false,
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ Information, news and updates are also regularly posted on the MISP project [Mas
 
 Installation
 -------------
-For test- og production installations we recommend you check out the possible options on [misp-project.org/download](https://www.misp-project.org/download/).
+For test- or production-installations we recommend you check out the possible options on [misp-project.org/download](https://www.misp-project.org/download/).
 
 Documentation
 -------------


linozinho:
@iglocska what do you think?

linozinho:
app/Model/SharingGroup.php

Without this the active flag does not work, is there any reason for this to be the case?

```diff
            $isSGOwner = !$user['Role']['perm_sync'] && $existingSG['org_id'] == $user['org_id'];
            if ($isUpdatableBySync || $isSGOwner || $user['Role']['perm_site_admin']) {
                $editedSG = $existingSG['SharingGroup'];
-                $attributes = ['name', 'releasability', 'description', 'created', 'modified', 'roaming'];
+                $attributes = ['name', 'releasability', 'description', 'created', 'modified', 'roaming', 'active', 'local'];
                foreach ($attributes as $a) {
                    if (isset($sg[$a])) {
                        $editedSG[$a] = $sg[$a];
```




app/Model/Server.php

without this, it is not possible to set the sharing group as distribution default

```diff
              ],
                'default_event_distribution' => array(
                    'level' => 0,
-                   'description' => __('The default distribution setting for events (0-3).'),
+                   'description' => __('The default distribution setting for events (0-4).'),
                    'value' => '',
                    'test' => 'testForEmpty',
                    'type' => 'string',
                    'options' => [
                        '0' => __('Your organisation only'),
                        '1' => __('This community only'),
                        '2' => __('Connected communities'),
-                       '3' => __('All communities')
+                      '3' => __('All communities'),
+                      '4' => __('Sharing Group')
                    ],
                ),
                'default_attribute_distribution' => array(
```

iglocska:
Looks good to me, thanks! Will cherry pick it onto 2.4-develop and it will make its way to 2.5 on the next release. Thanks!

linozinho:
Thank you very much 🙂

PR #9981, opened 2024-10-14
Update defaults.json

cudeso:
Enable delta merge on URLHaus feed to prevent the event from growing indefinite
#### Questions


adulau:
Thank you!

PR #9990, opened 2024-10-16
Fix bug: New users cannot use the platform when AAD is enabled

### LINKED ISSUE #9915: Bug: New users cannot use the platform when AAD is enabled
### Actual behavior

image
image

```bash
"MISP.disableUserSelfManagement" true
"MISP.disable_user_login_change" true
"MISP.disable_user_password_change" true
"Security.require_password_confirmation" false
```

When I create new users, they are unable to operate.

image

If I enable UserSelfManagement and password_change, they are required to set a new password.

### Expected behavior

New users should not be required to set a new password nor edit their profile on first login when AAD is enabled.

### Steps to reproduce

1. Enable AAD.
2. Disable user profile edit. This is done automatically by the docker image entrypoint [here] (https://github.com/MISP/misp-docker/blob/master/core/files/configure_misp.sh)
3. When a new user is created, the "needs to change password" cannot be disabled and thus passwordless users cannot use the platform.

### Version

2.4.197

### Operating System

Ubuntu

### Operating System version

22.04

### PHP version

7.2

### Browser

All browsers

### Browser version

All versions

### Relevant log output

_No response_

### Extra attachments

_No response_

### Code of Conduct

d-r-e:
Fixes #9915

Even when passwords are disabled, new users were forced to change their password when logging in through Azure Active Directory/ EntraID.

### DIFF
--- a/app/Controller/Component/ACLComponent.php
+++ b/app/Controller/Component/ACLComponent.php
@@ -977,6 +977,22 @@ class ACLComponent extends Component
             }
             return true;
         };
+        // Ensure new users with AAD enabled are not required to set a new password
+        $this->dynamicChecks['aad_password_change_disabled'] = function ($user) {
+            if (Configure::read('AadAuth.enabled') && $user['change_pw'] == 1) {
+                $user['change_pw'] = 0;
+                $this->User->save($user);
+            }
+            return true;
+        };
+        // Ensure new users with AAD enabled are not required to edit their profile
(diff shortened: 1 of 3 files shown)


d-r-e (commented):
Having issues when disabling passwords on AAD-only MISP instances, as users are required to change password anyway, never letting them use the platform. This fixes the issue.

Ment-LLB:
This issue is resolved in the latest 2.5.9 version.

PR #9992, opened 2024-10-17
Ref 9989

JakubOnderka:
#### Questions


iglocska:
Could you merge this into 2.4-develop too? Also, generally what we do now:

Develop for 2.4-develop -> merge into develop. We have a 6 month grace period with updates for 2.4 until we phase it out completely.

Issue #6846, opened 2021-01-11
Tags syncing issue with object/event already created and publish


jacksonfylle:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Support
| OS version (server)       | Ubuntu
| OS version (client)       | Arch
| MISP version / git hash   | v2.4.136 (99166f9f7b4b13ba4d7100c53bd0d3bf62346665) 
| Browser                   | Firefox

### Support Questions
I have noticed some strange behavior on tags when syncing to MISP :
+ Updating tags on already created object/event are not well synchronized
+ Tags attributes are well synchronized inside a new object/event when it is created for the first time. Attributes with tags added after are not properly sync.
+ Tags removed are not synchronized either.

#### Step to reproduce
MISP_A: Create event and create the first attribute with a global tag; Publish
MISP_A == MISP_B

MISP_A: Add new attribute with a tag; publish
---> The attribute is synchronized but the tag is missing.

Hope it is not a mistake from me. Thanks for your help.

david74552:
Is this linked to https://github.com/MISP/MISP/issues/2223, https://github.com/MISP/MISP/issues/5453, https://github.com/MISP/MISP/issues/7481 and https://github.com/MISP/MISP/pull/7493?
I understand they have the code, but chose to attempt to fix this issue using another approach (soft-delete).

Issue #5453, opened 2019-12-09
Sync issue with Tags on object attributes


andurin:
### Work environment

| Questions                 | Answers
|---------------------------|--------------------
| Type of issue             | Support
| OS version (server)       | Ubuntu
| OS version (client)       | Ubuntu
| PHP version               | 7.0
| MISP version / git hash   | 2.4.119
| Browser                   | Firefox

### Support Questions

I recently noticed that Tags on object attributes won't get synced to other misp instances. Normal attribute tags are getting synced - even if only additional.

Our use case:
We ingest new MISP Events from different sources to our "MISP-A" instance and doing a first review about malicious observables. To get a QA view on those first analysis our First Level Analysts will publish some interesting events to push that event to "MISP-B" where our Second Level Analysts live. Since we focus ourself on good or bad observables we use different tagging on attribute level. From time to time an Analyst on "MISP-B" recognize some false positive tags on object attributes and remove them or of course add additional tags. Those vetted and quality approved events should then be synced back to "MISP-A".

Current situation: 
Normal attributes tagged on MISP-B will be available on MISP-A after publishing. Removed tags from MISP-B will NOT be removed on MISP-A. So tags are getting synced only additional?

Object attributes tagged on MISP-B won't be available on MISP-A after publishing. Only a guess - removed object attribute tags won't get deleted also.

Questions for support:

- Are object attributes really handled in a different way while syncing?
- Why are Tags only beeing synced additional to potential existing tags?

andurin:
As discussed in chat, steps to reproduce:

Requirements: Two MISPs with sync from A-->B
As far as I can say it doesn't matter if push to B or pull from A or both.

1. Create an Event on "A". 
    Add a plain attribute + a object (f.e. url with fullurl, hostname, domain) to it. 
    Set an global tag also the plain attribute as also one of the object attributes. 
2. Let the event getting synced to "B"
Expectation: Event from A == Event on B

3. Now add an additional global Tag 
    to the plain attribute as also to the object attribute. 
    Publish the event again
Excpectation: Event from A == Event on B

Reality: 
Only the plain attribute get the newly added tag. The new tag on the object attribute get's lost.

mokaddem:
Hey @andurin,
We just pushed a fix https://github.com/MISP/MISP/commit/b83e825b7bf251a3e668e2717692ca3fa9100b01 that should resolve the issue.
If you still experience other sync issues, please come back to us.
Thanks a lot for the comprehensive report!

[...]

PR #9993, opened 2024-10-18
fix: [Sync] Remove deleted tags when cleanup tags server setting is enabled

### LINKED ISSUE #2223: Tag removal does not sync to other MISP instances
When a tag is deleted and the event re-published, the tag is not removed from the event on downstream MISP instances. This is counter intuitive and can lead to strange situations such as events that have two TLP colors.

tomking2:
#### What does it do?

Re-enables the functionality to delete missing tags on sync, fixing issues when tags are updated on once instance, but their removal does not propagate to the receiving MISP organisation. This only runs when the `remove_missing_tags` parameter is set for a server PULL. 

Tested on both 2.4 and 2.5.

Fixes #2223 and also related to #5453, #7481, #7493 and #6846

#### Questions

### DIFF
--- a/app/Model/Attribute.php
+++ b/app/Model/Attribute.php
@@ -2856,7 +2856,7 @@ class Attribute extends AppModel
         return $attribute;
     }
 
-    public function editAttributeBulk($attributes, $event, $user)
+    public function editAttributeBulk($attributes, $event, $user, $server = null)
     {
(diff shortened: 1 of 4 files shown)


iglocska:
Thanks a lot @tomking2, that flew under the radar for sure!

PR #10018, opened 2024-10-30
Update INSTALL.ubuntu2404.sh

sidxparab:
### What does it do?
Fixes some errors while running the latest Ubuntu 24.04 installer script.

### Modifications:

1. If a PECL installation or configuration command failed, the entire installer script exited out. Added `|| true` to ensure the script continues even when the command fails
2. The above listed reason is same for modifications to installation of 'SSDEEP'
3. Updated the Baseurl MISP setting to discard the deprecated method and use the latest one. Also, BaseURL needs to be specified using the protocol.
4. Updated the `Security.password_policy_complexity` parameter which was erroring out due to special characters not being escaped

### Errors:
Below are corresponding errors received for the above modifications

1. (Scripted exited out, as PECL was already installed on my EC2, leading to an error install failed)
2. (Script exited out due an error)
3. `Warning: This method is deprecated. Next time please use cake admin setSetting MISP.baseurl [baseurl]`
4. `Error: Method name must be a string`

### DIFF
--- a/INSTALL/INSTALL.ubuntu2404.sh
+++ b/INSTALL/INSTALL.ubuntu2404.sh
@@ -231,21 +231,23 @@ print_ok "PHP and MySQL configured..."
 
 print_status "Installing PECL extensions..."
 
-sudo pecl channel-update pecl.php.net &>> $logfile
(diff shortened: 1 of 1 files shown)


iglocska:
Nice fixes, thanks a lot!

Issue #10020, opened 2024-11-01
Bug: Warning on wrong remote version in compatibility check for server


crym:
### Actual behavior

Application log (warning): `Warning: Sync to Server ('8') initiated, but version 2.4.111 is required in order to be able to pull proposals from the remote side.`

Connection test on server:
```
Local version: 2.5.1
Remote version: 2.5.1
Status: OK
Compatibility: Proposal pull disabled (remote version < v2.4.111)
POST test: Received sent package
``` 

### Expected behavior

No errors or warnings as both servers are the same version.

### Steps to reproduce

Have two servers with version 2.5.0 or both with version 2.5.1. Enable pull/push between these.  
Check warnings in logs on `{misp_server}/logs/index/filter:errors`  
Go to  `{misp_server}/servers` and click `Run` on the server connection test.  

### Version

2.5.1

### Operating System

almalinux

### Operating System version

8

### PHP version

8.2.13

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

crym:
Looks to be related to this check https://github.com/MISP/MISP/blob/c9fe5692969c08b755d9ee64fa0bdb603703a22d/app/Controller/ServersController.php#L1788
Will submit a pull request

PR #10021, opened 2024-11-01
fix(connection test): check not working for minor version upgrade

crym:
#### What does it do?

Quick fix for #10020 

#### Questions


iglocska:
Thanks a lot!

Issue #10044, opened 2024-11-14
Bug: OidcAuth does not implement cache-control or security headers


capile:
### Actual behavior

When using OidcAuth the plugin controller is used, instead of the application controller, so when the redirection to the authentication/authorization endpoints is done (on `OpenIDConnectClient.redirectToEndpoint‎`), none of the configured headers are sent.

This breaks cache-control and session handling, causing the authentication to fail (for missing session parameters) on any server that uses a caching reverse proxy or CDN. 

### Expected behavior

apply the configured response headers to every request

### Steps to reproduce

on a server configured with OidcAuth, try: `curl -isL https://<hostname>/users/login` and look for Cache-Control headers, for example. Other headers should also be missing (like CSP).

### Version

2.5.1

### Operating System

misp-docker on kubernetes

### Operating System version

22.04?

### PHP version

8.2

### Browser

any

### Browser version

any

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #10045, opened 2024-11-14
Send registered headers before passing to plugin

capile:
The JakubOnderka\OpenID-Connect-PHP (OidcAuth plugin) will not use the controller when sending redirects, causing any cache or security-related headers to be omitted. Refers to: https://github.com/MISP/misp-docker/issues/185

#### What does it do?

If it fixes  `#10044`

#### Questions

no
almost
no


iglocska:
Thanks, merged into 2.4-develop, it will percolate to 2.5 on the next release!

adulau:
Merged in https://github.com/MISP/MISP/commit/40fe311827154e3381e2a27d7e1098ee6118ebe3

Issue #10024, opened 2024-11-05
Feature Request: add analyst_data/indexMinimal endpoint to OpenApi doc


Wachizungu:
### Is your feature request related to a problem? Please describe.

analyst_data/indexMinimal endpoint OpenApi doc is missing

### Describe the solution you'd like

Added documentation

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #10051, opened 2024-11-19
[openapi] document analystData/indexMinimal endpoint

righel:
#### What does it do?

closes https://github.com/MISP/MISP/issues/10024

#### Questions


Wachizungu:
Thank you!

PR #10075, opened 2024-12-03
Catch errors during opening of GeoIP database

JSCU-CNI:
#### What does it do?

When the GeoIp2 database cannot be opened, e.g. due to a filesystem error, an InvalidArgumentException may be thrown during creation of the Reader object. Since this is not caught, users cannot be authenticated when this issue appears, throwing a HTTP 500 error in the process.

Simply adding the creation of the Reader to the already existing try/catch-block fixes it.

#### Questions


iglocska:

adulau:
Merged by @iglocska  in https://github.com/MISP/MISP/commit/9f24c76cedcd99817975b97ab0ea58ffc02603bb

Issue #10077, opened 2024-12-04
Bug: Unable to install MISP on fresh Ubuntu 22.04 image


d0t1q:
### Actual behavior

Web page returns 500 response. 

[Wed Dec 04 02:05:13.254372 2024] [ssl:error] [pid 30906] AH02604: Unable to configure certificate misp.local:443:0 for stapling
[Wed Dec 04 02:11:50.664550 2024] [php7:warn] [pid 47677] [client 192.168.50.150:3014] PHP Warning:  require_once(/var/www/MISP/app/Vendor/autoload.php): failed to open stream: No such file or directory in /var/www/MISP/app/Config/core.php on line 298
[Wed Dec 04 02:11:50.664586 2024] [php7:error] [pid 47677] [client 192.168.50.150:3014] PHP Fatal error:  require_once(): Failed opening required '/var/www/MISP/app/Vendor/autoload.php' (include_path='.:/usr/share/php') in /var/www/MISP/app/Config/core.php on line 298

### Expected behavior

200 response

### Steps to reproduce

On a fresh install on Ubuntu 22.04 (attempted multiple times on fresh installs). 

The installer encounters multiple errors when being ran with the following command

taylor@szeth:~$wget --no-cache -O /tmp/INSTALL.sh https://raw.githubusercontent.com/MISP/MISP/2.4/INSTALL/INSTALL.sh
bash /tmp/INSTALL.sh -A

I've included the installer output log, post install the system is non-functional.

[installer_output.txt](https://github.com/user-attachments/files/18001812/installer_output.txt)

The errors appear to start around line 4469 when it gets to the cake install and it appears to be an issue with PHP 

```
Next step: Installing CakePHP
No composer.lock file present. Updating dependencies to latest instead of installing from lock file. See https://getcomposer.org/install for more information.
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires php >=8.1.0,<9.0.0 but your php version (7.4.33) does not satisfy that requirement.
  Problem 2
    - Root composer.json requires matthiasmullie/scrapbook >=1.5.3 -> satisfiable by matthiasmullie/scrapbook[1.5.3, 9999999-dev].
    - matthiasmullie/scrapbook 1.5.3 requires php >=8.0.0 -> your php version (7.4.33) does not satisfy that requirement.
  Problem 3
[...]

PR #10080, opened 2024-12-05
fix: 2.4 install.sh script git clone reference 

d0t1q:
Fixes: #10077
Install script for 2.4 references wrong branch when cloning repository


iglocska:
Woah, nice catch, thanks a lot!

PR #10082, opened 2024-12-05
fix: [collections] internal server error thrown on view if not site a…

Wachizungu:
…dmin

#### What does it do?

Fixes typo, which results in internal server error being thrown on viewing a collection if the user is not a site admin.

Please note this PR is on develop! I'm sorry I don't have 2.4 dev instance atm.

#### Questions

### DIFF
--- a/app/Model/Collection.php
+++ b/app/Model/Collection.php
@@ -115,7 +115,7 @@ class Collection extends AppModel
         if ($user['Role']['perm_site_admin']) {
             return true;
         }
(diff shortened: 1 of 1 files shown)


iglocska:
@JakubOnderka - please be aware that any such fixes need to make their way into 2.4-develop too. The workflow we use now is: 

Fixes/new features that are NOT php8 related go into 2.4-develop. Before each release (or also inbetween when we need it) 2.4-develop is merged into develop. This allows both 2.4 and 2.5 to stay current until the grace period for 2.4 support ends.

PR #10084, opened 2024-12-06
fix: [internal] Fix "Integrity constraint violation: 1052 Column 'id' in where clause is ambiguous" when fetching galaxy matrix

JakubOnderka:
#### Questions


iglocska:
Please merge it into 2.4-develop first ;)

PR #10086, opened 2024-12-09
fix: [UI] wrong closing html tag on hard/soft delete form for a button

Wachizungu:
#### What does it do?

Fixes wrong html closing tag (no impact on users, as it looks like browsers autocorrect this kind of thing).

#### Questions


iglocska:
Thanks a lot!

PR #10088, opened 2024-12-09
fix: typos in warning messages triggered when you go to openapi, but …

Wachizungu:
…don't have API key or access.

#### What does it do?
Fixes a couple of typos

#### Questions


iglocska:
Thanks a lot!

PR #10089, opened 2024-12-09
chg: [OTP] mention authenticator app in enter OTP label

Wachizungu:
#### What does it do?

Clarifies the label a bit. Points out to people they will not get the OTP via e-mail, as people keep getting confused.

#### Questions


iglocska:
Thanks, let's hope it's not needed too frequently ;)

PR #10093, opened 2024-12-10
fix: typo in serverSetting related error message

Wachizungu:
#### What does it do?

Burns some of the witches (there are some hiding in misp-modules too)!

#### Questions


iglocska:
Typo or freudian slip? Anyway, thank you :)

PR #10094, opened 2024-12-10
fix: [internal] 'Call to a member function query() on null' when fetching cache

JakubOnderka:
#### Questions


iglocska:
Thanks a lot!

PR #10095, opened 2024-12-10
chg: [internal] Include SQL query in case of failed execution

JakubOnderka:
#### Questions


iglocska:
Could you keep the changes mirrored to MySQLObserverExtended?

PR #10122, opened 2024-12-22
fix: [Templates] use correct encoding in upload_file view element

Wachizungu:
#### What does it do?

using json_encode here breaks the functionality. Change to use h() instead.

#### Questions


iglocska:
Thanks a lot

PR #10127, opened 2024-12-22
fix: [Templates:index] correctly display template names

Wachizungu:
#### What does it do?

Fixes issue with template index view. Org names were shown in the 'name' columns instead of the template names.

Also modified column name Share to Shared.

#### Questions


iglocska:
Thanks a lot!

PR #10130, opened 2024-12-24
fix: [stix2misp] _from_misp -> _is_stix2_from_misp

JakubOnderka:
It looks like _from_misp method was renamed to _is_stix2_from_misp

#### Questions


iglocska:
Ping @chrisr3d

PR #10132, opened 2024-12-29
chg: Add link to security advisories to SECURITY.md

Wachizungu:
#### What does it do?

Just adds the link to the advisories.

#### Questions


adulau:
Thank you!

PR #10134, opened 2024-12-30
new: [test] Include misp_stix_converter in stixtest.py

JakubOnderka:
#### Questions


JakubOnderka:
@chrisr3d I have to revert your commit 9dd3c26faf0649a94b5209499fa544995e916c6a, as this version changes some methods that are used by `stix2misp.py` scripts and this change was not reflected in MISP repo. Also no new version of `misp-stix` was released.

PR #10135, opened 2024-12-30
Fixes the response on three rest API requests: GalaxyClusterRelation (Add/Edit) and  SightingsDB (Add)

tomking2:
#### What does it do?

I noticed that three requests using the $this->RestReponse->saveSuccessResponse were incorrectly skipping the $id parameter of the function. This meant that the `format` (e.g. "application/json") was being provided as an ID, not the output format.

This was working OK until mid this year, but as a byproduct of this commit https://github.com/MISP/MISP/commit/f08a2eaec25f0212c22b225c0b654bd60d089ef9#diff-de60e7431e287347ac0b63a74341482e0fe06916b7e891391e56c11d8a7e54ae any invalid formats resulted the response being encoded with HTML entities. This spits out bad JSON and broke PyMISP parsing of the JSON output as result.

This pull request does two things:
1. Places the $format in the correct positional parameter for the `saveSuccessResponse`
2. Passes the ID in the omitted `$id` parameter, 'cause we have an ID available for all 3 replacements so why not.

Before fix:
```
'{&quot;saved&quot;:true,&quot;success&quot;:true,&quot;name&quot;:&quot;GalaxyClusterRelation added&quot;,&quot;message&quot;:&quot;GalaxyClusterRelation added&quot;,&quot;url&quot;:&quot;/galaxy_cluster_relations/add/application/json&quot;,&quot;id&quot;:&quot;application/json&quot;}'
```

After fix:
```json
{"saved":true,"success":true,"name":"Relationship added.","message":"Relationship added.","url":"/galaxy_cluster_relations/add/72493","id":"72493"}
```

#### Questions

### DIFF
--- a/app/Controller/GalaxyClusterRelationsController.php
+++ b/app/Controller/GalaxyClusterRelationsController.php
@@ -131,7 +131,7 @@ class GalaxyClusterRelationsController extends AppController
             if (isset($clusterSource['authorized']) && !$clusterSource['authorized']) {
                 $errors = array($clusterSource['error']);
             }
-            
+
             if (!empty($relation['GalaxyClusterRelation']['tags'])) {
                 $tags = explode(',', $relation['GalaxyClusterRelation']['tags']);
                 $tags = array_map('trim', $tags);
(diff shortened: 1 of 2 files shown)


righel:
Good catch. Thanks!

PR #10141, opened 2025-01-03
fix: [authkeys] add accepts the user_id via URL params and posted JSO…

Wachizungu:
…N body

#### What does it do?

Fixes likely copy pasta error, sets correct variable name to accept the user_id via URL params and posted JSON body.

#### Questions


iglocska:
Nice catch!

PR #10144, opened 2025-01-03
chg: [AuditLog] add SharingGroupOrg and SharingGroupBlueprint filteri…

Wachizungu:
…ng options

#### What does it do?

Allows you to filter on these model types using the auditlog query builder.

#### Questions


iglocska:
Thanks a lot!

PR #10149, opened 2025-01-05
chg: [AuditLog] add valid models like notes and opinions, to possible…

Wachizungu:
… filters list

#### What does it do?

Adds some more valid models to auditlogscontroller, so they show up when building a filter. I checked that these actually get logged, and that filtering works after.

I was also wondering whether "Task" ever gets logged in audit logs? It doesn't have this $actsAs "AuditLog", but it is in this list for some reason.

#### Questions


adulau:
Thank you!

PR #10151, opened 2025-01-06
[new] Features for sync. and fixes

mokaddem:
# New synchronisation features

- Added new permission flag to allow internal pull: `perm_sync_internal`
    - Before, pull would downgrade the distribution regardless of the state of the `internal instance` flag
    - Now, if the remote user has the permission `perm_sync_internal`, the pulling instance will not
downgrade the distribution anymore
    - This way, it's always up to the remote to decide if the downgrade happens or not

- Added new permission flag to allow tag mirroring: `perm_sync_authoritative`
    - If a user is given this permission flag, tag saved locally will strictly match the data received, which may delete tags
    - The "generic" name of the permission could allow to implement similar behavior in the future

- The synchronisation option `remove_missing_tags` now also remove missing tags at Event and Event-Report level

# Changes & Bug fixes
- Added number of Events to be synchronized along with the number of Event that have been blocked to be synchronized due to sync. filtering rules
- Galaxy synchronisation now correctly handles Orgs and Sharing-Groups
- Tags on event-report are correctly synced, taking into account sync filtering rules


app/Controller/EventReportsController.php
```diff
@@ -233,7 +233,7 @@ public function addTag($id = false, $tag_id = false)
                             }
                         }
                     } else {
-                        $tagId = $this->EventReports->EventReportTag->Tag->lookupTagIdForUser($this->Auth->user(), trim($tag_id));
```
iglocska, line 236: This sounds like a bug fix that is unrelated.

app/Controller/EventReportsController.php
```diff
@@ -322,10 +322,8 @@ public function removeTag($id = false, $tag_id = false, $galaxy = false)
                 'recursive' => -1,
                 'fields' => array('Tag.name')
             ));
-            if ($this->EventReport->EventReportTag->delete($eventReportTag['EventReportTag']['id'])) {
-                if (empty($eventReportTag['EventReportTag']['local'])) {
-                    $this->EventReport->Event->unpublishEvent($event);
-                }
+            $detachSuccess = $this->EventReport->detachTag($eventReportTag['EventReportTag']['id'], $id, !empty($eventReportTag['EventReportTag']['local']));
```
iglocska, line 325: This will no longer unpublish if the tag wasn't a local tag - is that intentional?

app/Model/Attribute.php
```diff
@@ -2911,7 +2911,10 @@ public function editAttributePostProcessing($attributes, $event, $user, $server)
                     A solution to still keep the behavior for previous instance could be to not soft-delete the Tag if the remote instance
                     has a version below x
                 */
-                if (isset($server) && isset($server['Server']['remove_missing_tags']) && $server['Server']['remove_missing_tags']) {
+                if (
+                    (isset($server) && isset($server['Server']['remove_missing_tags']) && $server['Server']['remove_missing_tags']) ||
```
iglocska, line 2915: Even though it's not what we disgussed, I just realised a use case where this could be improved: For authoritative pulls ONLY remove / update global tags.

app/Model/Event.php
```diff
@@ -4378,7 +4428,21 @@ public function _edit(array &$data, array $user, $id = null, $jobId = null, $pas
...
             }
+            // Also remove existing tags not present in the incoming data if user is sync_user and has authoritative flag set
             if (isset($data['Event']['Tag']) && $user['Role']['perm_tagger']) {
+                if (
+                    (isset($server) && isset($server['Server']['remove_missing_tags']) && $server['Server']['remove_missing_tags']) ||
+                    ($user['Role']['perm_sync'] && !empty($user['Role']['perm_sync_authoritative']))
+                ) {
+                    $existingTags = $this->EventTag->find('all', [
+                        'recursive' => -1,
+                        'conditions' => ['event_id' => $this->id],
```
iglocska: same comment as above, I'd add 'local' => 0 here as a condition.

app/Model/EventReport.php
```diff
@@ -315,7 +356,7 @@ public function buildACLConditions(array $user)
                         'OR' => array(
                             'Event.org_id' => $user['org_id'],
                             'EventReport.distribution' => array('1', '2', '3', '5'),
-                            'AND '=> array(
+                            'AND ' => array(
```
iglocska: Just to make sure, did this work as intended? Seems like the old version had a mistake in this `'AND '` should be `'AND'`.

app/Model/EventReport.php
```diff
@@ -350,7 +391,7 @@ public function attachReportCountsToEvents(array $user, $events)
             if (!$user['Role']['perm_site_admin'] && $event['Event']['org_id'] != $user['org_id']) {
                 $conditions['AND'][] = [
                     'EventReport.distribution' => [1, 2, 3, 5],
-                    'AND '=> [
+                    'AND ' => [
```
iglocska: `'AND '` => `'AND'`

app/Model/EventReport.php
```diff
@@ -488,6 +529,47 @@ public function reArrangeReport(array $report)
...
+            return $saveResult;
+        }
+        return $saveResult;
+    }
+
+    public function detachTag($eventReportTagID, $eventReportID, $local = false): bool
+    {
+        $success = $this->EventReportTag->delete($eventReportTagID);
+        if ($success) {
+            if (empty($local)) {
```
iglocska, line 568: ok, this might catch the previous issues I believed I saw. To be confirmed?

app/Controller/EventReportsController.php
mokaddem: The new function `detachTag` takes care of updating the report's timestamp and potentially unpublishes the event.

app/Model/Event.php
mokaddem: Indeed, the local conditions was done in `AttributeTag->pruneOutdatedAttributeTagsFromSync` where local tag were skipped for removal. It's more efficient to filter it here instead.

app/Model/Attribute.php
mokaddem: Indeed, the local conditions was done in AttributeTag->pruneOutdatedAttributeTagsFromSync where local tag were skipped for removal. It's more efficient to filter it here instead.

app/Model/EventReport.php
mokaddem: Nice catch!

PR #10158, opened 2025-01-08
chg: replace deprecated harvesthq/chosen by community fork

marjatech:
#### What does it do?

Replaces Chosen (JS Lib) from [harvesthq](https://github.com/harvesthq/chosen) by a Fork from [JJJ](https://github.com/JJJ/chosen)
Old v1.8.7
New v2.2.0

while not highly active under development its at least not deprecated

#### Questions


mokaddem:
Hey there,
Thanks for the report and taking the time to actually do a PR.
Since this library is unmaintained, and more importantly, we're planning of phasing out many old JS dependencies (including this one), I'm not keen on updating it - unless there is a security issue.
Did you find a security concern worth updating to an unmaintained fork?
Thanks again!

marjatech:
No this was more like a try to find something "more" maintained than the original one, phasing it out sounds like a much better option here :)

mokaddem:
Cool! Thanks for the report and the PR anyway, it's really appreciated :smile:

PR #10165, opened 2025-01-10
Create INSTALL.debian12.sh

rubentroncon:
- Replaced strings 'Ubuntu' -> 'Debian'
- Edited os_version_check() to check for Debian 12 instead of Ubuntu
- Set PHP package versions to 8.2 (because 8.4 was not available in default apt repositories)
- Fixed ssdeep install (apt install libfuzzy-dev and create symlink to /usr/local/lib, otherwise install fails)

#### What does it do?

* It adds an install script for MISP on Debian to the repository
* It fixes an issue when installing ssdeep (Install fails because libfuzzy.so is not found)

#### Questions

### DIFF
--- a/INSTALL/INSTALL.debian12.sh
+++ b/INSTALL/INSTALL.debian12.sh
@@ -1,0 +1,681 @@
+#!/bin/bash
+# MISP 2.5 installation for Debian 12
+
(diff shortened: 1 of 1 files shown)


iglocska:
Absolute hero, thanks a lot!

PR #10208, opened 2025-01-31
update gridstack to v1.2.1

marjatech:
#### What does it do?

Upgrades gridstack Javascript library from **v1.0.0** to **v1.2.1**

While v1.2.1 is not the newest version of gridstack it is a minor upgrade without any code changes required in MISP.

#### Questions


mokaddem:
Hey again,
Same comment as for issue https://github.com/MISP/MISP/pull/10158#issuecomment-2647592305 It would have been great to migrate to the latest version, but since we were also planning of phasing this one out, I don't really see the point.
Your thoughts?

marjatech:
Sure, for this one I do have a point. The gridstack.js v1.0.0 actually embeds jQuery in v3.4.1 which does have known vulnerabilities.
While v1.2.1 is not latest, it embeds jQuery in v3.5.1 which is not known vulnerable as of today. So that's the intention here.

mokaddem:
That's a nice catch. Thanks!

PR #10228, opened 2025-02-11
Update README.md

cudeso:
Add SSL verification error/solution for AadAuth plugin

#### Questions


adulau:
Thank you!

PR #10232, opened 2025-02-14
chg: [doc:OidcAuth] fix typo and clarify some configuration steps

Wachizungu:
#### What does it do?

Clarifies some doc. Would appreciate specifically checking the part about Security.auth_enforced true . I am not 100% sure about that / what it does.

#### Questions


iglocska:
Thank you!

PR #10237, opened 2025-02-19
Attribute tag removal: fix timestamp touch

rkollataj:
#### What does it do?

Adds missing field (i.e. Attribute.type) when fetching an attribute to make timestamp of an attribute updated after global tag removal.

#### Questions


iglocska:
Thanks a lot, merged in develop!

Issue #10231, opened 2025-02-13
Bug: Argument list too long in [/var/www/MISP/app/Lib/Tools/BackgroundJobs/BackgroundJob.php, line 78


AliceGrey:
### Actual behavior

When a large event is created and triggers the event-after-save-new workflow, the workflow fails to run with the error "Argument list too long in [/var/www/MISP/app/Lib/Tools/BackgroundJobs/BackgroundJob.php, line 78".

### Expected behavior

Workflow runs

### Steps to reproduce

Create a simple event-after-save-new workflow with the following blocks:
Event After Save New -> IF::Organization(check for a specific org) -> IF::Tag (If not tagged with any for a specific tag) -> If true :: Tag Operation (Add global Tag) -> Publish Event

(Ignore in the image that the module is disabled - this screenshot was taken in a test environment)
<img width="1790" alt="Image" src="https://github.com/user-attachments/assets/dc8dc8d7-cb59-443e-bc73-542998b36aef" />

Create a new event with many tags & attributes and with the specific org and without the tags.

<img width="2672" alt="Image" src="https://github.com/user-attachments/assets/319d40b0-5cce-4421-b663-c05121169ef4" />

Check jobs

<img width="3152" alt="Image" src="https://github.com/user-attachments/assets/f1ab13f4-2c8f-4e58-8de5-6d1ffbb2a429" />

### Version

2.5.6

### Operating System

AlmaLinux in Docker Container

### Operating System version

9.5

### PHP version

8.2.25

### Relevant log output

```shell
2025-02-13 19:29:55 Warning: proc_open(): Exec failed: Argument list too long in [/var/www/MISP/app/Lib/Tools/BackgroundJobs/BackgroundJob.php, line 78]
2025-02-13 19:29:55 Error: [JOB ID: UUID REDACTED] - failed with error code 127 after 0.005 seconds. STDERR: Warning Error: proc_open(): Exec failed: Argument list too long in [/var/www/MISP/app/Lib/Tools/BackgroundJobs/BackgroundJob.php, line 78]
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

vincenzocaputo:
I'm experiencing the same issue.

vincenzocaputo:
[...]

PR #10247, opened 2025-03-02
fix: [Workflow] posix_spawn(): Argument list too long

vincenzocaputo:
#### What does it do?

With this pull request I propose a possible fix for the issue #10231.

The error occurs when workflows are executed on large MISP events that contain many tags and/or Galaxies. I believe the issue may also arise when events have a high number of attributes and attachments.

The problem is caused by executing the 'workflow' command through the proc_open function, where the entire MISP event is passed as a JSON string argument, exceeding the command-line length limit.

My solution involves saving the MISP event data to a temporary file (using FileAccessTool) and passing the file path as a command argument instead of the entire MISP event.

I have tested this solution in the following use cases:
- Workflow with the Concurrent Task logic module
- Event-after-save-new workflows (as mentioned in the issue)

#### Questions


JakubOnderka:
@vincenzocaputo You should use `BackgroundJobsTool::enqueueDataFile` to insert data and then `BackgroundJobsTool ::fetchDataFile` to fetch data file by given path.

vincenzocaputo:
> @vincenzocaputo You should use `BackgroundJobsTool::enqueueDataFile` to insert data and then `BackgroundJobsTool ::fetchDataFile` to fetch data file by given path.

@JakubOnderka  thanks for the hint! I've applied the changes and it seems to work correctly.

mokaddem:
Hey there,
Thanks for keeping maintaining this branch. It's now merged.
Cheers!

Issue #9936, opened 2024-09-26
Bug: Requirement to change password on first login prevents smooth user setup with external authentication (Entra-ID)


cudeso:
### Actual behavior

When new users are created and they login via external authentication (AAD / Entra-ID) for the first time they are required to change their password. If you have `disable_user_password_change`set to False they can't change their password on first login. Also, as an Admin you can not uncheck the box "User must change password" as it's grayed out.

Linked to https://github.com/MISP/MISP/issues/9806

### Expected behavior

When `disable_user_password_change` is False, do not have required password change on first login. Or, have an option on user create to disable force password change. @

### Steps to reproduce

- Have these settings:
  - MISP.disableUserSelfManagement => true
  - MISP.disable_user_login_change => true
  - MISP.disable_user_password_change => true
- Create a user with user role
- Do not set a password
- Uncheck "send credentials automatically"

- Login as the user via AAD
- You get the error message that users cannot do SelfManagement
   -  If you set disableUserSelfManagement to False and login, you then get the messages users cannot change password
- It becomes impossible for new users to login

- As admin you cannot change  "User must change password" as it's grayed out. You can only change it if `disable_user_password_change` is True

### Version

2.4.198

### Operating System

Ubuntu

### Operating System version

22

### PHP version

7.4

### Extra attachments

If you create the user via PyMISP then there's no problem because you can set the **change_pw** to False

```
org_id = 1 # MISP organisation ID
change_pw = 0
autoalert = 0
contactalert = 0
disabled = 0

role_id = 3

user = MISPUser()
user.email = username
user.change_pw = change_pw
user.org_id = org_id
user.role_id = role_id
user.user_invited_by = "Via scripting"
user.autoalert = autoalert
user.contactalert = contactalert
user.disabled = disabled

result = misp.add_user(user, pythonify=True)
print(result)
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #10255, opened 2025-03-08
Fix first login change password bug when the setting disable_user_password_change is set to true

vincenzocaputo:
#### What does it do?

This pull request should resolve the bug preventing the first login of a user. The issue occurs when the setting disable_user_password_change is set to true.

As stated in the setting's description, it should be enabled only if user passwords are managed through an external system. So, in this case the password change on first login should be skipped to avoid the following error:
image

I think this issue is related to #9936.

#### Questions


iglocska:
thanks a lot!

PR #10260, opened 2025-03-13
fix: [validation] use is_numeric and filter_var to validate integer a…

### LINKED ISSUE #10259: Bug: AttributeValidationTool incorrectly checks string input with is_int
### Actual behavior

The integer Attribute type cannot be validated because the AttributeValidationTool receives the value as a string type. In the screenshot, I've added a simple `var_dump($value);` statement to the `'integer'` check within [AttributeValidationTool.php](https://github.com/MISP/MISP/blob/2.5/app/Lib/Tools/AttributeValidationTool.php#L569-L573), which shows that $value is testing a string. Not a PHP coder, but it seems this check will always fail. I've tested this with PyMISP as well and cannot add a MISPAttribute with type set to 'integer'. This is also true for MISPObject templates (e.g. diamond-event).
Image

### Expected behavior

Adding an attribute with `integer` type and value of `123` (or any other valid integer) will validate.

### Steps to reproduce

Clone misp-docker, create an event, attempt to add an attribute of type 'integer' and value of an integer.

I've tested the with misp-docker and NUKIB/misp docker. The var_dump was added to NUKIB/misp by modifying the running container.

### Version

2.5.7

### Operating System

Docker

### Operating System version

latest

### PHP version

latest

### Browser

_No response_

### Browser version

_No response_

### Relevant log output

```shell

```

### Extra attachments

_No response_

### Code of Conduct

kfinny:
…ttribute type

Fixes #10259 

Please test. I'm not properly setup for MISP development or automated testing. I've done some basic testing by modifying the source code of a docker instance and testing with various inputs. Also did basic testing with PyMISP.

#### Questions

- [ ] Are you using it in production? Tested on docker instance

### DIFF
--- a/app/Lib/Tools/AttributeValidationTool.php
+++ b/app/Lib/Tools/AttributeValidationTool.php
@@ -567,8 +567,10 @@ class AttributeValidationTool
                   }
                 return true;*/
             case 'integer':
-                if (is_int($value)) {
-                    return true;
+                if (is_numeric($value)) {
+                    if (filter_var($value, FILTER_VALIDATE_INT)) {
+                        return true;
+                    }
                 }
                 return __('The value has to be an integer value.');
             case 'iban':


iglocska:
Thanks, that is indeed a good catch. Not sure why this wasn't flagged before. Thanks!

PR #10263, opened 2025-03-17
fix: [logging] warning on remote server version during sync

crym:
#### What does it do?
The following log line is created in application logs:
`Warning: Sync to Server ('2') initiated, but version 2.4.111 is required in order to be able to pull proposals from the remote side.`
Both local and remote is running v2.5.6
Related to #10021 

#### Questions


iglocska:
Thanks a lot!

Issue #10264, opened 2025-03-18
Bug: Event Publishing time usage increases when the user which initates the publish task is associated with many events.


Benni0:
### Actual behavior

The time it takes to publish an event increases if the user who initiates the publishing action ~~belongs to a group that has many events~~. If you try to edit the user in the users menu, the user also loads slow.
**There is an relation between slow publishing and the user, but it's not the organisation.**

~~In [EventShell.php](https://github.com/MISP/MISP/blob/ca5da8622931d01dcb480dd8b968e7d3a121af9d/app/Console/Command/EventShell.php#L420C19-L420C34), the UserId is derived from the job and the `User->getAuthUser()` function is used to get the User instance. The `getAuthUserByConditions` function requests the user including the organisation from the database and the organisation seems to include all events of the organisation.~~

~~The same behavior can be observed, when the user is edited on the "List Users" page.~~

### Expected behavior

The user should have no influence on the time required for the publish action.

### Steps to reproduce

Publish an unpublished event using an user which has many associated events.
~~View an user, which belongs to an organisation which has much events, on the "List Users" page.~~ -> that might be another issue.

### Version

2.5.7

### Operating System

CentOS

### Operating System version

9

### PHP version

8.2.25

### Relevant log output

```shell

```

### Extra attachments

As a workaround, I've modified the `publish()` function in [EventShell.php](https://github.com/MISP/MISP/blob/ca5da8622931d01dcb480dd8b968e7d3a121af9d/app/Console/Command/EventShell.php#L420), to use a fixed userId of an user which belongs to an organisation without any events. Due to this, publishing takes about two seconds instead of about two minutes. But due to this modification, the wrong user is shown in audit log.

I would suggest to implement a function for this, where only the user without the organisation is fetched from the database.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

iglocska:
Hmm, I can't seem to be able to reproduce this. Dumping the contents of the retrieved user object only contains the role, organisation and the associated server (if set), Events shouldn't be loaded.

```
        $user = $this->find('first', [
            'conditions' => $conditions,
            'fields' => $full ? [] : $this->describeAuthFields(),
            'recursive' => -1,
            'contain' => [
                'Organisation',
                'Role',
                'Server',
            ],
        ]);
```

This code block requests the user object, with an explicit list of relations (recursive -1 and the list of contain objects ensures this).

[...]

PR #10267, opened 2025-03-20
fix: Do not fetch users associated objects recursive for audit logging

Benni0:
#### What does it do?
Currently users are fetched with all associated events for audit logging, but only `org_id` is used.
If the user is associated with many events, this might take much time.
If only the user instance, without associated objects is fetched, audit logging is much faster.

If it fixes an existing issue, please use github syntax: #10264

#### Questions


iglocska:
Thanks a lot!

PR #10282, opened 2025-03-28
Load AuthPlugins based on what is specified in the configuration file

ostefano:
@iglocska @righel as discussed


iglocska:
Thanks a lot

PR #10292, opened 2025-04-03
Update LdapAuthenticate.php to escape characters when searching for member groups

### LINKED ISSUE #10291: Bug: LDAP group lookup does not properly escape special characters causing ldapDefaultRoleId mapping to fail
TLDR: This line does not properly escape special characters in AD and causes group lookup to fail when integrating with our Microsoft AD. Additionally the space after the 'member=' doesn't seem to be valid syntax for a search filter.

https://github.com/MISP/MISP/blob/2ee749dd939c3206af8458de02749b4c7e3a0dfe/app/Plugin/LdapAuth/Controller/Component/Auth/LdapAuthenticate.php#L116

In my ad, this is what my distinguished name looks like:

```
CN=Userton\, User (12345),OU=TEST,DC=example,DC=com
```
Since there are special characters, the search for member groups fails with a bad filter error.
When I updated line 116 to escape the special characters, the member group lookup works.
```php
$filter = '(member=' . ldap_escape($ldapUserData[0]['dn'], "", LDAP_ESCAPE_FILTER) . ')';
```
My ldap test config:
```php
  'LdapAuth' =>
  array (
    'debug' => true,
    'ldapServer' => 'ldaps://ldaps.example.com:636',
    'ldapDn' => 'DC=example,DC=com',
    'ldapReaderUser' => 'CN=ldapbindaccount,OU=Process-Accts,OU=TEST,DC=example,DC=com',
    'ldapReaderPassword' => 'bindpassword',
    'mixedAuth' => true,
    'ldapTlsRequireCert' => 'LDAP_OPT_X_TLS_NEVER',
    'ldapTlsCrlCheck' => 'LDAP_OPT_X_TLS_CRL_NONE',
    'ldapTlsProtocolMin' => 'LDAP_OPT_X_TLS_PROTOCOL_SSL3',
    'updateUser' => true,
    'ldapDefaultRoleId' =>
    array (
      'MISP_Admin' => 1,
      'MISP_User' => 3,
      'MISP_ReadOnly' => 6,
    ),
    'ldapDefaultOrg' => 1,
  ),
```
I think updating that single line to escape special characters should resolve the issue. I've never really used the ldap php stuff before, so not entirely sure if this change will break anything else.

ppatryaps:
Resolves #10291

### DIFF
--- a/app/Plugin/LdapAuth/Controller/Component/Auth/LdapAuthenticate.php
+++ b/app/Plugin/LdapAuth/Controller/Component/Auth/LdapAuthenticate.php
@@ -113,7 +113,7 @@ class LdapAuthenticate extends BaseAuthenticate
     private function getUserMemberships($ldapconn, $ldapUserData)
     {
         $groups = [];
-        $filter = '(member= ' . $ldapUserData[0]['dn'] . ')';
+        $filter = '(member=' . ldap_escape($ldapUserData[0]['dn'], "", LDAP_ESCAPE_FILTER) . ')';
         $ldapUserMemberships = ldap_search($ldapconn, self::$conf['ldapDn'], $filter, ['cn']);
 
         if ($ldapUserMemberships) {


righel:
implemented by https://github.com/MISP/MISP/pull/10293

PR #10300, opened 2025-04-08
fix: [CorrelationRules:edit] UI fixes, correctly populate selectors f…

Wachizungu:
…ield, and don't try weird ajax redirect magic.

Developed during 2025 hackathon.lu

#### What does it do?

Fixes some UI issues

#### Questions


iglocska:
Thanks a lot!

PR #10304, opened 2025-04-09
chg: [db] make event info collation case insensitive

Wachizungu:
- Developed during 2025 hackathon.lu

#### What does it do?

Changes event info column collation, so comparisons are case insensitive. This will change restSearch results when filtering events based on event info.

#### Questions


iglocska:
Thanks a lot!

Issue #10219, opened 2025-02-04
Feature Request: Display extended event in the index or not


adulau:
### Is your feature request related to a problem? Please describe.

Display extended event in the index or not

### Describe the solution you'd like

Display extended event in the index or not

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

mokaddem:
Also include filtering options in the API

PR #10305, opened 2025-04-09
Add new feature: display extended event in the index or not (reupload)

ThomasLcr:
**What does it do?**

Added a feature that indicates in the Info column if an event extends another one.By default, we can consider the extended event option as unchecked, which only puts an icon, indicating the extension, at the end of the event information. If the box is checked then it displays the full description of the extended event.

Modification of the dynamic size of the Info column in the event index to make it easier to read.

Screenshot from 2025-04-07 15-37-58

Screenshot from 2025-04-07 15-39-29

Related to https://github.com/MISP/MISP/issues/10219


app/View/Elements/Events/eventIndexTable.ctp
```diff
@@ -165,8 +165,40 @@
...
+                                aria-label="<?= __('See extended event') ?>">
+                                    <?= h($extends_id)?>
+                                </a>
+                                : <?= h($extends_info) ?>
+                            </p>
+                        </span>
+                    </div>
+                <?php else: ?>
+                    <a href="<?= h($baseurl) ?>/events/view/<?= h($extends_id) ?>" 
+                    title="Extends event <?= h($extends_id) ?>"
```
mokaddem: Could we have i18n in this code block? e.g.
```php
title="Extends event <?= h($extends_id) ?>"
```
=>
```php
title="<?= __('Extends event %s', h($extends_id) ?>"
```

app/Controller/EventsController.php
```diff
@@ -970,6 +980,8 @@ private function __indexColumns()
             $possibleColumns[] = 'owner_org';
         }
 
+        $possibleColumns[] = 'extending';
```
mokaddem, line 986: I'd prefer having the column hidden by default. I think the icon is useful enough for most use cases.
Adding it to `DEFAULT_HIDDEN_INDEX_COLUMNS` would make sense in my opinion

mokaddem (changes requested):
Hey,
Thanks for the good work!
If you could review the comments above and update this branch to the latest version of 2.5 / 2.5-develop, that would be great.
Thanks!

app/View/Elements/Events/eventIndexTable.ctp
ThomasLcr: Done in [3dee3f1](https://github.com/MISP/MISP/pull/10305/commits/3dee3f1c985d33f1e76409ab23f4ecc662399164)

app/Controller/EventsController.php
ThomasLcr: Done in [3dee3f1](https://github.com/MISP/MISP/pull/10305/commits/3dee3f1c985d33f1e76409ab23f4ecc662399164)

mokaddem:
Already merged. Thanks!

PR #10306, opened 2025-04-09
Split Known Identifier from False Positive Warnings in the event view

ThomasLcr:
**What does it do**

When an attribute is linked to an activated  warninglist, it adapts the message according to the warning category (knowledge identifier or false positive).

Fixed the filtering option from the hyperlinks in the warnings widgets.

Screenshot from 2025-04-09 15-29-49

### DIFF
--- a/app/files/feed-metadata/defaults.json
+++ b/app/files/feed-metadata/defaults.json
@@ -297,7 +297,7 @@
     "Feed": {
       "name": "OpenPhish url list",
       "provider": "openphish.com",
-      "url": "https://openphish.com/feed.txt",
+      "url": "https://raw.githubusercontent.com/openphish/public_feed/refs/heads/main/feed.txt",
       "rules": "",
       "enabled": false,
       "distribution": "3",


app/webroot/css/main.css
```diff
@@ -2104,10 +2103,22 @@ a.discrete {
...
+
+.false_positive h4 {
     color: red;
 }
 
+.known_identifier {
+    border: 1px solid orange;
+}
+
+.known_identifier h4 {
```
mokaddem: Do not forget to bump the value in AppController.
https://github.com/MISP/MISP/blob/c533290d574095eca7581dd8037c3c46991363d7/app/Controller/AppController.php#L36

To force reload web assets.

app/Model/Event.php
```diff
@@ -5260,10 +5260,16 @@ private function __prepareAttributeForView(
...
             }
 
-            if ($filterType['warning'] == 0) { // `both`
-                // pass, do not consider as `both` is selected
-            } else if (!empty($attribute['warnings']) || !empty($attribute['validationIssue'])) { // `include only`
-                $include = $include && ($filterType['warning'] == 1);
+            if ($filterType['warning'] == 0) { // `all`
+                // pass, do not consider as `all` is selected
+            } else if (!empty($attribute['warnings']) || !empty($attribute['validationIssue'])) { // Include warning
+                if ($filterType['warning'] == 3 && $attribute['warnings'][0]['warninglist_category'] == "false_positive") { // False positive
```
mokaddem: Looking at the first entry in the list of warnings `[0]` will lead to inconsistent result if the Attribute matches on the 2 list type.

app/Model/Warninglist.php
```diff
@@ -180,7 +180,11 @@ public function attachWarninglistToAttributes(array &$attributes)
                             'warninglist_name' => $warninglists[$warninglistId]['name'],
                             'warninglist_category' => $warninglists[$warninglistId]['category'],
                         ];
-                        $eventWarnings[$warninglistId] = $warninglists[$warninglistId]['name'];
+                        if ($attributes[$redisResultToAttributePos[$pos]]['warnings'][0]['warninglist_category'] == "false_positive") {
```
mokaddem: Same comment than above with the `[0]`. It places the warning list in the wrong bucket.

app/View/Elements/genericElements/SidePanels/Templates/eventWarnings.ctp
```diff
@@ -1,27 +1,50 @@
...
-        $baseurl,
-        (int)$id,
-        __('View warninglist %s', h($name)),
-        __('View warninglist')
+$linksByCategory = [];
+
+foreach ($event['warnings'] as $category => $warnings) {
+    foreach ($warnings as $id => $name) {
+        $linksByCategory[$category][] = sprintf(
+            '<a href="#attributeList" title="%s" onclick="toggleWarningFilter(\'warninglistId:%s\')">%s</a> <a href="%s/warninglists/view/%s" class="black fa fa-search" title="%s" aria-label="%s"></a>',
```
mokaddem: Just a question, why not use `setAttributeFilter` as it was used before?

mokaddem (changes requested):
Cool PR! If you could fix the bug with the `[0]` as commented above, that would be great.
Thanks!

app/View/Elements/genericElements/SidePanels/Templates/eventWarnings.ctp
ThomasLcr: I used toggleWarningFilter because if you click on the name of the warning list, it filters the attributes to show only those that belong to that warning list, as with setAttributeFilter, but if I click again on the name of the warning list, it unapplies the filter whereas setAttributeFilter does not.

app/Model/Event.php
ThomasLcr: Do you mean that an attribute can be in 2 different warning lists, where one can be False Positive and the other one Known identifier?

app/Model/Warninglist.php
ThomasLcr: Done in [a29dd67](https://github.com/MISP/MISP/pull/10306/commits/a29dd6705bc1a972c8eb7f203df84e255d68fb41)

app/Model/Event.php
ThomasLcr: Done in [a29dd67](https://github.com/MISP/MISP/pull/10306/commits/a29dd6705bc1a972c8eb7f203df84e255d68fb41)

Issue #10288, opened 2025-04-03
[Feature Request] Add Warninglist in /attributes/search


mokaddem:
### Is your feature request related to a problem? Please describe.

When you search for attributes in the UI, you don’t see the warning that the value is in a warninglist, (nor can you filter them out.)

### Describe the solution you'd like

- Allow showing the warnings in the search result
- (potentially if not too costly) Allow filtering based on warninglist hits

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Wachizungu:
It looks like this is implemented / can be closed?

PR #10314, opened 2025-04-16
new: [Attributes] Add Warninglist in /attributes/search

ThomasLcr:
**What does it do?**

Added functionality to allow the user to exclude IOCs on a warning list when performing a filter search on attributes.
Added a small indicator in /attributes/index to show whether an attribute is in a warning list.

Screenshot from 2025-04-16 11-14-51

Screenshot from 2025-04-16 11-18-30

Related to https://github.com/MISP/MISP/issues/10288

### DIFF
--- a/app/Controller/AttributesController.php
+++ b/app/Controller/AttributesController.php
@@ -60,6 +60,9 @@ class AttributesController extends AppController
             if ($key === 'to_ids' && $value === '0') {
                 unset($filters[$key]);
             }
+            if ($key === 'warnings' && $value === '0') {
+                unset($filters[$key]);
+            }
             if (is_array($value)) {
                 $filters[$key] = $this->cleanDefaultFormValues($value);
             } elseif ($value === '') {
@@ -82,7 +85,7 @@ class AttributesController extends AppController
         }
         $params['conditions']['AND'][] = $this->Attribute->buildConditions($user);
         $paramArray = [
-            'value' , 'type', 'category', 'org_id', 'tags', 'to_ids', 'first_seen', 'last_seen', 'limit', 'page', 'sort', 'direction'
+            'value' , 'type', 'category', 'org_id', 'tags', 'to_ids', 'first_seen', 'last_seen', 'limit', 'page', 'sort', 'direction' 
         ];
         $filterData = array(
             'request' => $this->request,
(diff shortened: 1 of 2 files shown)


app/Controller/AttributesController.php
```diff
@@ -131,11 +138,21 @@ public function index()
                 $params['limit'] = $filters['limit'];
             }
             $attributes = $this->Attribute->fetchAttributes($user, $params);
+            if (isset($params_warnings)) {
```
mokaddem: Maybe you could rely on the `enforceWarninglist` param, since you're already using `includeWarninglistHits`.

app/Controller/AttributesController.php
```diff
@@ -131,11 +138,21 @@ public function index()
...
+                $attributes = array_filter($attributes, function($attribute) {
+                    return !isset($attribute['Attribute']['warnings']);
+                });;
+            }
         } else {
             $params['page'] = !empty($filters['page']) ? $filters['page'] : 1;
             $params['limit'] = !empty($filters['limit']) ? $filters['limit'] : 60;
             $this->paginate['conditions'] = $conditions;
             $attributes = $this->Attribute->fetchAttributes($user, $params);
+            if (isset($params_warnings)) {
```
mokaddem: Same comment as above

mokaddem (commented):
Another cool PR!
Have a look at the comments and let me know what you think!

ThomasLcr:
> Another cool PR! Have a look at the comments and let me know what you think!

Thanks for the tip !

app/Controller/AttributesController.php
ThomasLcr: Done in [047cae2](https://github.com/MISP/MISP/pull/10314/commits/047cae26e5dfbd22363292dbbfdb59a6afa744c7)

app/Controller/AttributesController.php
ThomasLcr: Done in [047cae2](https://github.com/MISP/MISP/pull/10314/commits/047cae26e5dfbd22363292dbbfdb59a6afa744c7)

app/Controller/AttributesController.php
```diff
@@ -108,7 +111,9 @@ public function index()
         }
         $this->set('params', $params);
         $conditions = $this->Attribute->buildFilterConditions($user, $filters, false);
-        $params = [];
+
+        $params = isset($params['enforceWarninglist']) ? ['enforceWarninglist' => 1] : [];
```
mokaddem: Are you trying to do something like this instead of resetting `$params`?
```php
if (isset($params['enforceWarninglist']) {
    $params['enforceWarninglist'] = 1;
}
```

mokaddem (changes requested):

app/Controller/AttributesController.php
ThomasLcr: Thanks for the correction !

PR #10318, opened 2025-04-18
new: [Event|API] Extended flag on /events/restseach and /events/index

ThomasLcr:
**What does it do?**

Added 2 boolean flag : 
  - extending, which indicates events that extend another event
  - extended, which shows events that are extended by other events

Related to https://github.com/MISP/MISP/issues/10219


app/Model/Event.php
```diff
@@ -2943,6 +2951,86 @@ public function set_filter_published(&$params, $conditions, $options)
         return $conditions;
     }
 
+
+    public function set_filter_extended(&$params, $conditions, $options)
+    {   
+        xdebug_break();
```
mokaddem: Remove this. :wink:

app/Model/Event.php
```diff
@@ -2943,6 +2951,86 @@ public function set_filter_published(&$params, $conditions, $options)
...
+        ));
+
+        //If there is no event with extends_uuid(extending), there is basically no event extended
+        if (empty($events)) {
+            $conditions['AND'][] = array('Event.id' => -1);
+            return $conditions;
+        }
+
+        // Extract the UUIDs of the events that are extended and remove duplicates
+        $targetUuids = array_unique(Hash::extract($events, '{n}.Event.extends_uuid'));
```
mokaddem: This and its associated find query can be simplified by using `$this->find('column', [])`

mokaddem (changes requested):
Thanks for the PR!
It would be handy to also expose these two filters in the /events/index "Filter Event Index" modal.
https://github.com/MISP/MISP/blob/de1007a632e8b43e69f3d92bc2801765ff71c674/app/Controller/EventsController.php#L1109-L1128
What do you think?

app/Model/Event.php
ThomasLcr: I hunted them all, but this one seems to have escaped me :angry:

app/Model/Event.php
ThomasLcr: Done in commit c2b6728

ThomasLcr:
> Thanks for the PR! It would be handy to also expose these two filters in the /events/index "Filter Event Index" modal.
> 
> https://github.com/MISP/MISP/blob/de1007a632e8b43e69f3d92bc2801765ff71c674/app/Controller/EventsController.php#L1109-L1128
> 
> What do you think?

I think this could be a useful feature. For the filtering function on extending/extended, do you think we can only apply a boolean YES/NO filter or would it be better to use a parameter Event ID, like for example, only include events that extend this ID?

ThomasLcr:
> Thanks for the PR! It would be handy to also expose these two filters in the /events/index "Filter Event Index" modal.
> 
> https://github.com/MISP/MISP/blob/de1007a632e8b43e69f3d92bc2801765ff71c674/app/Controller/EventsController.php#L1109-L1128
> 
> What do you think?

Thanks for the idea. It's done in [0b7f94f](https://github.com/MISP/MISP/pull/10318/commits/0b7f94f28040a5bd880a3b7f2168b8b1b20e8131)

mokaddem:
> > Thanks for the PR! It would be handy to also expose these two filters in the /events/index "Filter Event Index" modal.
> > https://github.com/MISP/MISP/blob/de1007a632e8b43e69f3d92bc2801765ff71c674/app/Controller/EventsController.php#L1109-L1128
> > 
> > What do you think?
> 
> Thanks for the idea. It's done in [0b7f94f](https://github.com/MISP/MISP/pull/10318/commits/0b7f94f28040a5bd880a3b7f2168b8b1b20e8131)

A boolean like you did is perfect. Thanks!

Issue #10320, opened 2025-04-21
Feature Request: Create a feature to set default rest queries by default if not set (like for `limit`)


adulau:
### Is your feature request related to a problem? Please describe.

As you know we have the issue if there is no `limit` set, a request can kill a instance to get all data. To not impact existing MISP user, we propose to create a setting to set the default limit (maybe some other parameters) to avoid dangerous queries if the parameter is not set.

### Describe the solution you'd like

It should be a parameter in the settings to allow user or site admin to have a default limit in the MISP instance. 

As it's just a setting, it won't impact the existing users. 

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

github-germ:
Good idea.

Might be helpful, since the server cannot kill long running backend database queries, if there was a documented suggestion for folks to consider configuring a method for the database to kill these processes ( g.g. creating a [MySQL watchdog event](https://dev.mysql.com/doc/refman/8.4/en/create-event.html) -- example below)

We see cases where the client initiating an extremely long running request unknowing of the impact, who may even no longer be there (disconnected for example due to max_execution_time in PHP or httpd), can trigger an unintentional MISP DoS.

Suggertion from ChatGPT:
```
CREATE EVENT kill_long_queries
ON SCHEDULE EVERY 10 MINUTE -- Runs every 10 minutes
DO
BEGIN
    DECLARE done INT DEFAULT FALSE;
    DECLARE process_id INT;
    DECLARE cur CURSOR FOR 
        SELECT ID FROM information_schema.PROCESSLIST 
        WHERE TIME > 21600 AND USER <> 'root'; -- 21600 seconds = 6 hours
    
    DECLARE CONTINUE HANDLER FOR NOT FOUND SET done = TRUE;
 
    OPEN cur;
    read_loop: LOOP
        FETCH cur INTO process_id;
        IF done THEN
            LEAVE read_loop;
        END IF;
        SET @query = CONCAT('KILL ', process_id);
        PREPARE stmt FROM @query;
        EXECUTE stmt;
        DEALLOCATE PREPARE stmt;
    END LOOP;
 
    CLOSE cur;
END;
```

cudeso:
Could this be implemented as part of the role permissions? Similar as how you now set memory limits (`Role.memory_limit`)?

Byhird:
+1 this, we're struggling operating an ISAC and trying to get everyone who hits our APIs to enforce limits on their requests is an impossible task.

[...]

PR #10325, opened 2025-04-22
new: [Roles|API] New attribute in Roles and server setting for number of result by request

ThomasLcr:
**What does it do?**

Added a new server parameters allow you to define a default limit for the number of results for a request. The value 0 allows an unlimited number of results.
A new ‘result_limit_count’ attribute has also been added to allow you to set a specific limit on the number of results for the role. If it is not defined, it takes the default limit defined in the server parameters (unlimited for site admin role).

When making a request, any user can therefore set a limit indicator equal to or lower to the one of their role.
Works with the following endpoints:
- /attributes/restSearch
- /events/restSearch
- ~~/events/index~~
- /objects/restSearch

New UI admin setting :
Screenshot from 2025-04-22 13-52-02

New role attribute:
Screenshot from 2025-04-25 11-03-18
Screenshot from 2025-04-25 11-03-54

Related to https://github.com/MISP/MISP/issues/10320


app/Controller/AppController.php
```diff
@@ -1436,6 +1436,23 @@ public function restSearch()
...
+        $user = $this->Auth->user();
+        $roleId = $user['role_id'];
+        $this->loadModel('Role');
+        $role = $this->Role->find('first', [
+            'conditions' => ['Role.id' => $roleId],
+            'recursive' => -1,
+            'fields' => ['Role.result_limit_count']
+        ]);
+        $roleLimit = isset($role['Role']['result_limit_count']) ? (int)$role['Role']['result_limit_count'] : 0;
+
```
mokaddem: https://github.com/MISP/MISP/blob/5c64f76da90472abee9205c5fedb9a4501c03722/app/Controller/AppController.php#L1439-L1449

mokaddem: Since this might be used multiple time in the future, make it a function in `userModel.php`

app/Controller/EventsController.php
```diff
@@ -816,33 +815,51 @@ private function __indexRestResponse(array $passedArgs)
...
+        $this->loadModel('Role');
+        $role = $this->Role->find('first', [
+            'conditions' => ['Role.id' => $roleId],
+            'recursive' => -1,
+            'fields' => ['Role.result_limit_count']
+        ]);
+        $roleLimit = isset($role['Role']['result_limit_count']) ? (int)$role['Role']['result_limit_count'] : 0;
 
         if (empty($rules['limit'])) {
+            $rules['limit'] = $roleLimit !== 0 ? $roleLimit : 20000;
```
mokaddem: Let's only keep the limit count for rest-search queries only. Please, revert that part that touches to the /events/index.

app/Model/Role.php
```diff
@@ -127,6 +127,9 @@ public function beforeSave($options = array())
         if (empty($this->data['Role']['rate_limit_count'])) {
             $this->data['Role']['rate_limit_count'] = 0;
         }
+        if (!isset($this->data['Role']['result_limit_count']) || $this->data['Role']['result_limit_count'] === '') {
```
mokaddem: Since the server setting is called `default_restsearch_limit` and we only want to apply it for rest-search queries, wouldn't it make sense to make it explicit in the name? Something like `restsearch_max_limit` or similar.

app/Model/AppModel.php
```diff
@@ -2290,6 +2290,10 @@ public function updateDatabase($command)
             case 138:
                 $sqlArray[] = "ALTER TABLE `events` MODIFY info text CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL;";
                 break;
+            case 139:
+                $sqlArray[] = "ALTER TABLE `roles` ADD `limit_search_results` tinyint(1) NOT NULL DEFAULT 0;";
```
mokaddem: Where is `limit_search_results` it used? I see the `add` and `edit` form but not where it's used elsewhere.

mokaddem (changes requested):

app/Model/Role.php
ThomasLcr: I think this name makes sense because if the admistrator creates a new role without giving an explicit limit, it will take this one. Moreover, he may give a higher limit than the server limit so it might be odd to call it ‘max’.

app/Controller/AppController.php
ThomasLcr: Here is my proposition in [40cda5c](https://github.com/MISP/MISP/pull/10325/commits/40cda5ca7c5411dc591b49476686ce0d6290d113)

Issue #10321, opened 2025-04-21
Feature Request: If a return format is not specified in the API, the Accept header should be respected and take precedence.


adulau:
### Is your feature request related to a problem? Please describe.

If a return format is not specified in the API, the Accept header should be respected and take precedence.

### Describe the solution you'd like

If a return format is not mentioned in the API, the accept header should work and take the precedent.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #10335, opened 2025-04-30
new: [API] Accept header can define format

ThomasLcr:
#### What does it do?
When performing a restSearch, if the returnFormat flag is not set, it searches the request header for an explicit format for the response. If the format is valid, the response will be in this format.

Related to https://github.com/MISP/MISP/issues/10321


app/Controller/AppController.php
```diff
@@ -1399,11 +1399,17 @@ public function restSearch()
         );
         $exception = false;
         $filters = $this->_harvestParameters($filterData, $exception, $this->_legacyParams);
+        $acceptHeader = $this->request->header('Accept');
         if (empty($filters) && $this->request->is('get')) {
             throw new BadRequestException(__('Restsearch queries using GET and no parameters are not allowed. If you have passed parameters via a JSON body, make sure you use POST requests.'));
         }
         if (empty($filters['returnFormat'])) {
-            $filters['returnFormat'] = 'json';
+            if (preg_match('#application/([a-zA-Z0-9\-\+]+)#', $acceptHeader, $matches)) {
```
mokaddem, line 1409: This wouldn't work for `Accept: text/csv` which is used a lot.

mokaddem (changes requested):

app/Controller/AppController.php
ThomasLcr: Thanks for the feedback ! I added the support of type text as well as quality parameters in my last commit [38612af](https://github.com/MISP/MISP/pull/10335/commits/38612aff7a8c46aaa8d0193a396f838badffb42f)

PR #10342, opened 2025-05-06
Update OpenPhish feed URL to new GitHub location

drego85:
#### What does it do?

This PR updates the OpenPhish feed URL to point to their new GitHub-hosted location.

The original feed URL occasionally becomes unavailable, while the GitHub-based feed is now officially listed as a reliable alternative.

Switching to the GitHub feed ensures better reliability and availability of threat intelligence data.

#### Questions


adulau:
It makes sense. Thank you very much!

PR #10343, opened 2025-05-07
[OTP/OIDC] Allow otp multiple auth

righel:
#### What does it do?

Fixes local auth not working when `OidcAuth.mixedAuth=true` and `Security.otp_required=true`

#### Questions


cvandeplas (approved):
is that all to support multiple auth? nice catch !!  thanks !

righel:
> is that all to support multiple auth? nice catch !! thanks !

There were some minor changes done before on the Oidc plugin but overall was straightforward!

PR #10345, opened 2025-05-08
add: [Warninglists] Split Known Identifier from False Positive in the…

ThomasLcr:
(re-upload due to a merge conflict with the new release)
**What does it do**

When an attribute is linked to an activated  warninglist, it adapts the message according to the warning category (knowledge identifier or false positive).

Fixed the filtering option from the hyperlinks in the warnings widgets.

Screenshot from 2025-04-09 15-29-49


mokaddem:
Already merge. Thanks!

PR #10353, opened 2025-05-21
Fix #2737: Improve password requirement indication.

### LINKED ISSUE #2737: Display password requirements on UI
### Expected behavior

Indication for the passwords requirements whenever a password change is needed. 

### Actual behavior

No indications at all.

Just get to [misp]/users/pwd and try to change your password.

misp_issue

monjiapawne:
#### What does it do?

Fixing issue: Display password requirements on UI #2737
Provides more verbose indication of failed password change. If the complexity regex isn't the default, it falls back to displaying the raw regex.

#### Questions


iglocska:
Cherry picked onto 2.4-develop, thanks a lot!

iglocska:
Reverted, this actually completely breaks new installations, as the password complexity validation has been moved to the beforefilter. 

The issue with that: We enroll an initial user where we disable the complexity check on installation - in order to do that (with hard coded simple values) we disable the validation temporarily by unloading the validation scripts. This reintroduces it and blocks the initial user from ever being created.

Additionally, it displays the validation rules on the login page, not just on user creation/modification/password reset. Definitely not something we would like instances to disclose to unauthenticated users.

PR #10372, opened 2025-06-03
update Cerebrate UI components and respect `skip_proxy` flag 

DocArmoryTech:
- The `skip_proxy` setting is now fully supported in the Cerebrate UI, including:
  - View: Displays current proxy status.
  - Edit: Allows users to update the setting.
  - Index: Shows the proxy configuration per Cerebrate entry.

This also ensures the `skip_proxy` setting is properly respected in HTTP requests made via `SyncTool::setupHttpSocket()`. Previously, the method defaulted to using the 'Server' model key, which caused it to overlook `skip_proxy` and other configuration values when used by models like 'Cerebrate'.

To resolve this:
- `AppModel::setupHttpSocket()` was extended to accept an optional `$model` parameter. If provided, it is passed through to `SyncTool::setupHttpSocket()` to ensure the correct settings are used.
- Backward compatibility is maintained for other subclasses of `AppModel` that do not yet pass a model identifier.

Tested and confirmed working in production with `skip_proxy` enabled.

### DIFF
--- a/app/Model/AppModel.php
+++ b/app/Model/AppModel.php
@@ -3510,16 +3510,21 @@ class AppModel extends Model
         return true;
     }
 
-    public function setupHttpSocket($server, $HttpSocket = null, $timeout = false)
+    public function setupHttpSocket($server, $HttpSocket = null, $timeout = false, $model = null)
     {
         if (empty($HttpSocket)) {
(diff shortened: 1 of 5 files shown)


iglocska:
Thanks a lot!

PR #10377, opened 2025-06-05
new: [Bookmarks] allow site admins to create global bookmarks

goodlandsecurity:
#### What does it do?

This PR allows site admins to create global bookmarks (exposed to all users on an instance). This feature will be especially beneficial in community instances, allowing site admins to share bookmarks for search queries to view intel specific to an industry, threat actor, threat type, etc. while not requiring the end user to need to create their own.

#### Questions

### DIFF
--- a/app/Controller/BookmarksController.php
+++ b/app/Controller/BookmarksController.php
@@ -37,9 +37,14 @@ class BookmarksController extends AppController
                 $conditions = [
                     'OR' => [
                         'Bookmark.user_id' => $this->Auth->user()['id'],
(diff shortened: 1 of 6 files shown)


iglocska:
That would indeed be handy, but it is a bit dangerous as it is - could you modify it so that only site admins can add global bookmarks? Also, could you add a new db upgrade script to accomodate for the new field? Have a look at for example update 138, just search AppModel.php for 138 to see what was added to make it work. 

Thanks for the PR and sorry that it took this long to respond.

PR #10386, opened 2025-06-12
fix: allow editing protected event flag on sync and consider it a sig…

Wachizungu:
…nificant change

#### What does it do?

Fixes an issue with unprotected events to protected changes not syncing properly.

e.g.:
1. Create event -> sync it (pull)
2. Turn the event into protected mode -> sync it (pull) -> crypto keys got added but the event wasn't flagged as protected

To be verified together with a to be done fix to protected event pull sync (which is currently broken).

Also:
Sorry for missing the meeting today! I won't make it a habit!

#### Questions


iglocska:
Merged into develop, thanks a lot!

PR #10398, opened 2025-06-19
Add feed from phish.co.za (Phishing.Database) - 2.4

cudeso:
#### What does it do?

Add feed from phish.co.za (Phishing.Database) - 2.4

#### Questions


adulau:
Thank you!

PR #10400, opened 2025-06-20
Migrate GitHub Actions workflow to Forgejo Actions (reupload)

ThomasLcr:
This PR allows to run current MISP workflow on a forgejo instance.


adulau:
Awesome thank you!

PR #10401, opened 2025-06-20
fix: [sync:Event] pass fingerprint to _add if it exists, to avoid bre…

Wachizungu:
…aking issue

#### What does it do?

Dumb temporary fix for a breaking issue in pull sync for protected event.

_**Fairly sure this should be improved / refactored!!!!**_

#### Questions

### DIFF
--- a/app/Model/Server.php
+++ b/app/Model/Server.php
@@ -504,7 +504,14 @@ class Server extends AppModel
                     return false;
                 }
             }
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks a lot!

adulau:
Fixed in 4b2e6c557c453e8aabdcc240f2d072c16ed65ea1 and already released.

PR #10409, opened 2025-06-30
fix: [EventReport] allow rendering of mermaid.js diagrams with arrows

Wachizungu:
#### What does it do?

Fixes the mermaid rendering of diagrams using arrows (e.g. --> or <-->) in event reports.

The XSS for which this partialEscapeHtml was introduced, is only in the error message rendering I believe. The rest should already be fine (famous last words), as mermaid uses DOMPurify.

Of course, review by you true pros is appreciated.

#### Questions


iglocska:
Thanks a lot!

Issue #10426, opened 2025-07-19
Bug: ssdeep failure on pristine Ubuntu 22.04 server


SteveClement:
### Actual behavior

```
checking for ssdeep... configure: error: "Could not find 'libfuzzy.so'. Try specifying the path to the ssdeep build directory."
[ERROR] Jakub Onderka's PHP8 SSDEEP extension compilation and installation failed. Please check /var/log/misp_install.log for more details.
```

### Expected behavior

It should just install.

### Steps to reproduce

Vanilla install with official 24.04 installer script.

Potential fix: sudo apt install libfuzzy-dev

### Version

2.5

### Operating System

Ubuntu

### Operating System version

22..04

### PHP version

8.3

### Relevant log output

```shell
checking for ssdeep... configure: error: "Could not find 'libfuzzy.so'. Try specifying the path to the ssdeep build directory."
[ERROR] Jakub Onderka's PHP8 SSDEEP extension compilation and installation failed. Please check /var/log/misp_install.log for more details.
```

### Code of Conduct

- [ ] I agree to follow this project's Code of Conduct

SteveClement:
Installing libfuzzy-dev does not solve the issue.

SteveClement:
Resuming installer after flipping back to no ssdeep works to get a fully functional MISP install.

SteveClement:
Relevant lines in installer: https://github.com/MISP/MISP/blob/caf8524f8a98cf2058d607c5091ac267999e25f3/INSTALL/INSTALL.ubuntu2404.sh#L251C1-L260C3

SteveClement:
Quick fix would be: ln -s /usr/lib/x86_64-linux-gnu/libfuzzy.so /usr/lib/libfuzzy.so

SteveClement:
> Quick fix would be: ln -s /usr/lib/x86_64-linux-gnu/libfuzzy.so /usr/lib/libfuzzy.so

This has been implemented in the debian installer. Closing this issue as it is fixed for me.

PR #10428, opened 2025-07-19
fix: [ssdeep] Fixes issue #10426 of pecl ssdeep not compiling

SteveClement:
Fixes issue #10426 now ssdeep pecl will now compile and install.


adulau:
Thank you!

PR #10431, opened 2025-07-22
fix: [TagCollections] correct permission check in removeTag()

Frisb7:
#### What does it do?

It fixes an existing issue #8706

Fixes the logic error in `TagCollectionsController::removeTag()` that threw
a *ForbiddenException* when the user **had** permission to modify the tag
collection.  Adds the missing negation so only unauthorised users are blocked.

#### Questions


iglocska:
Thanks a lot!

PR #10460, opened 2025-08-20
Feature/simpler shibbauth behavior

guillomovitch:
#### What does it do?

This PR makes configuration of Shibboleth (actually: SAML) authentication a bit more easier, by making handling of configuration directives more consistent and more logical.

Before:
- DefaultOrg is only used if non-documented useDefaultOrg is true
- DefaultRole is not used if dynamic role assignment fails (login fails immediatly), and overrides dynamic assignement if set

After:
- dynamic organisation assignement is attempted if OrgTag is defined, DefaultOrg if used as a fallback if set
- dynamic role assignement is attempted if GroupTag and GroupRoleMatching are defined, DefaultRole is used as a fallback if set

Documentation has been updated accordingly, with more accurate wording (at least, for someone understanding SAML), and actual examples.

#### Questions

- it does not require a DB change
- it does not requires a change in the API
- we don't use it in production (yet)


adulau:
Thanks for the PR. What will be the impact of the changes for existing Shibbauth configuration in MISP? Will it break the current ones?

guillomovitch:
This is indeed a behavior change:
- undocumented useDefaultOrg is now ignored, meaning DefaultOrg will be used, if defined, as a fallback
- DefaultRole will be used, if defined, as a fallback, whereas login previously failed for lack of role if dynamic assignment failed

As a consequence, some users could have their organization changed, and some users could now login, whereas they were previously rejected. Considering those changes as breaking or not is a matter of opinion.

PR #10464, opened 2025-08-21
remove quotes, as they trigger a javascript syntax error

guillomovitch:
#### What does it do?

This commit fixes an issue preventing the definition of a remote server when using french language

#### Questions

- it does not require a DB change
- it does not require a change in the API (PyMISP for example)?
- we are using it in production


adulau:
Merci/Thank you!

Issue #10453, opened 2025-08-12
Bug: API and Download as .. ignore proposale no ids since 2.5.16


HiS3:
### Actual behavior

- setting: `MISP.proposals_block_attributes true` 
- Attributes have a proposal for ids_no
- API-request or _download as..._ Attributes from event with only ids=true

**BUT**  here I also get the attributes where the ids proposal is set to _false_ or _no_

I see this Bug since `v.2.5.16` and also in `v.2.5.18`

My guess is that the API query ignores the set proposal despite the setting

Since this is the only way to handle the FP rate in IDS well, it prevents the current updates of the instances.

### Expected behavior

- setting: `MISP.proposals_block_attributes true` 
- Attributes have a proposal for ids_no
- API-request or _download as..._ Attributes from event with only ids=true

Got _**only**_ attributes with ids=true (direct or via proposals)

### Steps to reproduce

- set  setting: `MISP.proposals_block_attributes true` 
- go in a Event
- set on one or more Attributes a proposal for _no ids_
- export the attributes via API-request or _download as..._  with `to_ids:1`

### Version

2.5.18

### Operating System

MISP as Docker (official Repo)

### Operating System version

22.04

### PHP version

8.2.29

### Browser

Firefox, RestApi, PyMISP

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

HiS3:
As an addition.
also the explicit setting of 
```json
{
  "to_ids": 1,
  "allow_proposal_blocking": true
}
```
 in the api requests does not lead to the desired result

marjatech:
This time the desired behaviour still works when using: 
`"includeProposals": 1` 

However that should not be necessary. See also former issue #8076 fixed in #8288

@iglocska: The issue seems to be related to the refactoring in 245fdc4774f9a6846185e95daabb2e226a9be74c

PR #10465, opened 2025-08-21
fix: bind ShadowAttribute Model also to allow proposals to block attributes

marjatech:
#### What does it do?

Possible Fix for #10453 
Option `proposals_block_attributes` is not working right now. 

#### Questions


iglocska:
Thanks a lot!

Issue #10468, opened 2025-08-26
Bug: Broken link to "Bound Server" in User view


DocArmoryTech:
### Actual behavior

When viewing a user configured as a "Sync user for" a remote server, the User's details include a broken link to the "Bound Server" e.g. (`https://misp.local/servers/view/17`)

The current Cake template `/app/View/User/view.ctp` generates a link pointing to an incorrect/non-existent server action:
```  
        'html' => $this->Html->link($user['Server']['name'], ['controller' => 'servers', 'action' => 'view', $user['Server']['id']]),
```

### Expected behavior

When viewing a User the "Bound Server" link should point to the relevant server's index preview.

### Steps to reproduce

1. configure a local MISP user as "Sync user for" a remote server
2. View the MISP user's profile
3. Following link to "Bound Server"

### Version

2.5.18

### Operating System

Ubuntu

### Operating System version

22.04

### PHP version

8.1

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #10469, opened 2025-08-26
fix(users-view): correct action name in server link

DocArmoryTech:
#### What does it do?

Changed the link action from 'view' to 'previewIndex' in app/View/Users/view.ctp to fix navigation to the correct "Bound Server" see #10468


iglocska:
Thanks a lot!

PR #10475, opened 2025-08-29
Solving the non-synchronization of local tags when pulling between internal instances

ThomasLcr:
Deletion of content brought by https://github.com/MISP/MISP/pull/7100
This PR prevented the propagation of local tags with internal and non-internal instances.
Moreover a flag _excludeLocalTags_ has since been introduced to do this work.

(solved also the issue for pulling local galaxy cluster)


iglocska:
Thanks a lot, nice catch!

Issue #5095, opened 2019-09-03
Activity heat-map statistic might be incorrect


mokaddem:
Activity heat-map statistic might not show relevant data (way to much items displayed compared to the action done).
-> Just check if everything is working as intended

Isaaciyo:
Hello, I'm a student from the Codepath AI301 class specializing in Python, and I would love to work on this issue. Is this issue still active?

iglocska:
@Isaaciyo - no it's been stale for almost 7 years, consider MISP has undergone many an iteration since then.

PR #10489, opened 2025-09-05
Fix nids duplicate emails

musharafmaqbool:
## What does it do?
This PR fixes duplicate email rules appearing in NIDS export functionality, ensuring clean and accurate export output.

**Changes:**
- Remove duplicate emailDstRule call for generic 'email' case
- Prevents duplicate entries in NIDS export output  
- Generic email attributes now default to source rules only
- Users can still use 'email-dst' for destination-specific rules

**Fixes:** #5095

## Testing
- [ ] NIDS export no longer shows duplicate email rules
- [ ] Generic email attributes export correctly
- [ ] email-dst still works for destination-specific rules

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -3417,7 +3417,6 @@ class EventsController extends AppController
         if ($this->_isSiteAdmin()) {
             $this->Flash->info(__('Warning, you are logged in as a site admin, any export that you generate will contain the FULL UNRESTRICTED data-set. If you would like to generate an export for your own organisation, please log in with a different user.'));
         }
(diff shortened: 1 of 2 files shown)


musharafmaqbool:
Hi @adulau,

Thank you for the feedback. You're absolutely right - I understand now that each fix should be in its own separate branch with its own PR. 

I've just created separate branches for each fix:
1. **DB Export Malware Warning** - This branch only contains the warning message for database exports
2. **NIDS Email Duplicate Fix** - I'm creating this as a separate branch that only addresses the duplicate email rules issue

I apologize for initially combining both changes in one PR. I'll make sure to follow the one-fix-per-PR approach going forward. Each branch now focuses on solving one specific issue with its own commit history.



Thank you for your patience and guidance on the proper contribution process.

Issue #3101, opened 2018-03-27
Add a warning note in DB export that we don't include the malware sample (which are on the fs)


adulau:
Add a warning note in DB export that we don't include the malware sample (which are on the fs)

stbam:
Hey I'm a student of codepath and i'm looking to work on this issue. I know its old but if it still needs hands feel free to assign it to me!

adulau:
This was already resolved some years. By the way, you can do PR directly and avoid asking for permission in the issue. Thank you.

PR #10490, opened 2025-09-05
Add db export malware warning

musharafmaqbool:
## What does it do?
Adds a warning message to the database export functionality to inform users that malware samples are not included in database exports since they are stored on the filesystem rather than in the database.

**Changes:**
- Added warning message to inform users about malware sample exclusion
- Warning appears when users access the export function
- Helps set proper expectations about export contents

**Fixes:** #3101

## Testing
- [ ] Warning message appears when accessing DB export
- [ ] Message clearly explains malware samples are not included
- [ ] Export functionality still works normally

## Questions
- [x] Are you using it in production? This is a user-facing warning improvement

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -3417,6 +3417,8 @@ class EventsController extends AppController
         if ($this->_isSiteAdmin()) {
             $this->Flash->info(__('Warning, you are logged in as a site admin, any export that you generate will contain the FULL UNRESTRICTED data-set. If you would like to generate an export for your own organisation, please log in with a different user.'));
         }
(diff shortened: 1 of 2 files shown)


musharafmaqbool:
Hi @adulau,

Thank you for the feedback. You're absolutely right - I understand now that each fix should be in its own separate branch with its own PR. 

I've just created separate branches for each fix:
1. **DB Export Malware Warning** - This branch only contains the warning message for database exports
2. **NIDS Email Duplicate Fix** - I'm creating this as a separate branch that only addresses the duplicate email rules issue

I apologize for initially combining both changes in one PR. I'll make sure to follow the one-fix-per-PR approach going forward. Each branch now focuses on solving one specific issue with its own commit history.


Thank you for your patience and guidance on the proper contribution process.

PR #10498, opened 2025-09-17
Forgejo pipeline for synchronisation

ThomasLcr:
# Synchronisation Pipeline

This PR introduces a workflow for Forgejo that automatically deploys multiple MISP instances within the same run.
The workflow is designed to perform Python synchronisation tests between these instances.


adulau:
Thank you very much!

PR #10499, opened 2025-09-18
fix: [Event] Switch Correlation and Warninglist filtering option

ThomasLcr:
In the Event view, the attribute filtering options on correlations and warninglists have been reversed.
This PR resolves this issue.


iglocska:
Merged into 2.5's develop branch, thanks a lot!

PR #10511, opened 2025-10-10
chg: enable test pull rules button with rules set

righel:
#### What does it do?
 enable test pull rules button with rules set, change wording to make it more intuitive 

image

#### Questions


iglocska:
Much appreciated!

PR #10515, opened 2025-10-13
Spelling 1

jsoref:

#### What does it do?

Fixes a subset of the errors identified in https://github.com/check-spelling-sandbox/MISP/actions/runs/18449400245/attempts/1#summary-52560834094

#### Questions

- [x] Does it require a DB change? Not sure
- [x] Are you using it in production? No
- [x] Does it require a change in the API (PyMISP for example)? Not sure


app/Config/config.default.php
```diff
@@ -228,7 +228,7 @@
         'ldapDN'             => 'dc=example,dc=com',
         'ldapSearchFilter'   => '', // Search filter to limit results from ldapsearh fx to specific group. FX
         //'ldapSearchFilter'   => '(objectclass=InetOrgPerson)(!(nsaccountlock=True))(memberOf=cn=misp,cn=groups,cn=accounts,dc=example,dc=com)',
-        'ldapSearchAttribut' => 'uid',          // filter for search
```
iglocska, line 231: Whilst valid (that's indeed an annoying spelling error), this change is actually dangerous as users will already have custom, untracked config.php files with ldapSearchAttribut set - meaning their instances would become unreachable after an update. I'd remove all instances of ldapSearchAttribut -> ldapSearchAttribute changes altogether.

iglocska:
Thanks a lot, merged into develop with the change to ldapSearchAttribut => ldapSearchAttribute reverted in a follow up commit, as these changes would cause issues with existing installations.

PR #10523, opened 2025-10-18
Action modules to add tags based on MMDB and vulnerability information

cudeso:
#### What does it do?

Two workflow action modules.
- Module_tag_country_asn_from_enrichment
  -  Add the **country**, **country code**, **ASN** number and **ASN organisation** based on the enrichment module. A variant of the existing `Module_assign_country_from_enrichment`, but with ASN information and it uses tags
- Module_tag_cve_from_enrichment
  - Add the CVSS score, risk level and vulnerability summery (as comment) to an attribute. This module does not use the cve enrichment module, but queries `cvepremium.circl.lu` directly. The cve enrichment module did not return all results, hence the direct calls.

#### Questions


mokaddem (commented):
Few unstructured notes. Will come back with more details later on:
### 1.
Whenever you modify data, you have to update the roaming data as well, so that later modules see the modified data. Check this for tags: https://github.com/MISP/MISP/blob/ef7d7dde1074e74f3e43b1b513aa015a581896b4/app/Model/WorkflowModules/action/Module_tag_operation.php#L140-L147 and this for modifying attributes: https://github.com/MISP/MISP/blob/ef7d7dde1074e74f3e43b1b513aa015a581896b4/app/Model/WorkflowModules/action/Module_attribute_ids_flag_operation.php#L4

### 2.
Not sure how long cvepremium will be maintain. Using vulnerability-lookup instead is preferable. Anyway, that's for the future, we could merge it.

### 3.
TODO for myself (could be a bug in the `TagOperation` module), doing modification on local tags should not "touch" the attributes. Only global tag operations should bump the timestamp.

### 4.
https://github.com/MISP/MISP/pull/10523/files#diff-a56abc4d290a53cedca5d5abc4dcec045e908d9c58538cf6a96a9f3b93d677efR92
Maybe use the `JsonTool:decode` function for error catching?

cudeso:
Thanks. Will look into 1/ & 4/ soon and update the PR.

cudeso:
@mokaddem  Thank you for the suggestions to update the PR. 

I addressed 1 and 2, please let me know if anything should be coded differently.

For the tags I had to execute `$this->_buildFastLookupForRoamingData($roamingData->getData())` prior and post the update. Otherwise I triggered an error where it could not find the ID of the attribute in the `_addTag` call.
Also copied the private function `genTagObjectsFromTagNames` from Module_tag_operation.

adulau:
Thanks a lot for the contribution!

PR #10524, opened 2025-10-21
Correlation - fix ipv4-mapped ipv6 address handling

sid1033224:
#### What does it do?

Correlations break when encountering an ipv4-mapped ipv6 address of the form `::ffff::a.b.c.d/ZZ` since the test simply checks for the presence or absence of `.` characters, leading to an ipv6 address with an embedded ipv4 address being mishandled as an ipv4 address.  

This culminates in an exception when trying to shift left by a negative number (since these CIDRs typically have prefix lengths of 96 or greater, and the netmask handling code assumes 32 or less.)  Ultimately, this behaviour leads to the inability to add `ip-addr` attributes to any event.

#### Questions


iglocska:
Thanks a lot, sorry for missing this earlier!

PR #10534, opened 2025-11-03
Improved the accessibility of the server settings screens

obert01:
On each setting row, I have added a pure ARIA button (invisible), that switch the cell to edit mode when activating. This avoid having to move the mouse cursor and double-click, which is challenging for screen reader users. 

The behavior for "normal" users should remain unchanged.


iglocska:
As always, much appreciated @obert01 !

PR #10550, opened 2025-11-25
fix: [UI] Return and use updated last_login time, opposed to 2nd last_login

### LINKED ISSUE #10487: Bug: Last login shows wrong date/time in user interface
### Actual behavior

The last login in the user interface when logging in shows the next to last login, rather than the actual last login.

### Expected behavior

The last login in the user interface when logging should show the actual last login date and time.

### Steps to reproduce

Login, note the time and date. Logout, then login again and note the date and time from the last login.

### Version

2.5.6 and 2.5.18

### Operating System

Debian, RHEL

### Operating System version

bookworm

### PHP version

8.x

### Browser

Firefox

### Browser version

Latest

### Relevant log output

```shell

```

### Extra attachments

_No response_

### Code of Conduct

tomking2:
#### What does it do?

Fixes #10487. Code was not actually using the last_login updated by `updateLoginTimes` (the old current_login). Instead, it was using the previous value, meaning the last_login was 1 behind the true value.

Simply leveraging the return from updateLoginTimes to update the user to show the correct date.

#### Questions

### DIFF
--- a/app/Controller/UsersController.php
+++ b/app/Controller/UsersController.php
@@ -1290,7 +1290,7 @@ class UsersController extends AppController
         if (empty($authUser['disabled'])) {
             $this->User->extralog($authUser, "login");
         }
-        
+
         $this->User->Behaviors->disable('SysLogLogable.SysLogLogable');
         $user = $this->User->find('first', array(
             'conditions' => array(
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks a lot!

PR #10568, opened 2025-12-10
fix: [doc:OpenAPI] align sharing group blueprint add sharing_group_id…

Wachizungu:
… definition

#### What does it do?

As proposed in dev chat.

#### Questions


iglocska:
Thank you, sir!

Issue #10561, opened 2025-12-04
Bug: Jobs failed when pushing the TI to TAXII server with push_taxii


ba1dh3ad:
### Actual behavior

I'm setting up the platform with MISP [official docker project](https://github.com/MISP/misp-docker).

When try to push TI to remote TAXII2 server, an error was raised:

```
STDERR: 'taxii_push [CRITICAL] An error occurred!
Traceback (most recent call last):
File "/var/www/MISP/app/files/scripts/taxii/taxii_push.py", line 385, in convert_misp_dir
stix_bundle = convert_misp_file(event_file)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/var/www/MISP/app/files/scripts/taxii/taxii_push.py", line 355, in convert_misp_file
converter.parse_json_content(str(misp_file))
File "/usr/local/lib/python3.12/site-packages/misp_stix_converter/misp2stix/exportparser.py", line 57, in parse_json_content
json_content = json.loads(json_content)
^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.12/json/init.py", line 346, in loads
return _default_decoder.decode(s)
^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.12/json/decoder.py", line 338, in decode
obj, end = self.raw_decode(s, idx=_w(s, 0).end())
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.12/json/decoder.py", line 356, in raw_decode
raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
File "/var/www/MISP/app/files/scripts/taxii/taxii_push.py", line 444, in main
push_content(args.dir, args.baseurl, args.api_root, args.collection, args.key)
File "/var/www/MISP/app/files/scripts/taxii/taxii_push.py", line 431, in push_content
for taxii_envelope_bytes in make_taxii_envelopes(
^^^^^^^^^^^^^^^^^^^^^
File "/var/www/MISP/app/files/scripts/taxii/taxii_push.py", line 277, in make_taxii_envelopes
stix_object = next(stix_objects, None)
^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #10569, opened 2025-12-11
Suggested fix for issue #10561

mdhirt:
Suggested fix for issue #10561 Bug: Jobs failed when pushing the TI to TAXII server with push_taxii

- [N] Are you using it in production?


adulau:
@chrisr3d @iglocska Can you confirm this fix?

PR #10575, opened 2025-12-15
Somes fixes for the Regexp page

ThomasLcr:
## Fix in View

Apply a fix on top of the existing PR
[https://github.com/MISP/MISP/pull/10572](https://github.com/MISP/MISP/pull/10572)

---

## Fix in Controller

Prevent accidental deletion of `Regexp` entries when their type is set to **`All`** during edition.

### Previous behavior

1. Create a new `Regexp` with the type set to **All**
2. Edit this `Regexp` without changing any field
3. Save the form
4. The `Regexp` entry is deleted from the database

This happened because the cleanup logic removed all similar entries after the new `All` entry was created, including the newly created one.

---

### Current behavior

1. Create a new `Regexp` with the type set to **All**
2. Edit this `Regexp` without changing any field
3. Save the form
4. The `Regexp` entry remains correctly stored in the database

NB: The edit operation still follows the existing *delete + recreate* logic, and this behavior has not been modified.

### DIFF
--- a/app/Controller/RegexpController.php
+++ b/app/Controller/RegexpController.php
@@ -72,10 +72,13 @@ class RegexpController extends AppController
             unset($this->request->data['Regexp']['id']);
             // If 'all' is set, it overrides all other type settings. Create an attribute with the "all" setting and save it. Also, delete the original(s)
             if ($this->request->data['Regexp']['all'] == 1) {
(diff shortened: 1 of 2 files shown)


iglocska:
Thanks a lot!

PR #10576, opened 2025-12-15
Migration for Object Template

ThomasLcr:
Other migrations using the index table factory.


iglocska:
Lovely, thanks a lot!

Issue #10580, opened 2025-12-16
Enhancement: Add Object -> open Search directly (skip the dropdown menu(s))


pettai:
### Is your feature request related to a problem? Please describe.

Since you're pushing for add Object over Attributes, the workflow for Objects needs to be improved!

Add Object defaults the user to the less helpful 2x of drop-down Object scope menu(s), a workflow which is cumbersome and slow (compared to add an Attribute workflow, for instance).

### Describe the solution you'd like

One fix could be that the Add Object jumps to the "search bar" directly in the All Object scope.
This could be enable by a configuration knob if the old behavior can't be changed by default)

### Describe alternatives you've considered

Bonus 1: Display the 5 most commonly used objects as fast-dial buttons

### Additional context

Bonus 2: this behavior applies to other "adds", like Add Tags, so it could be adopted there too

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

pettai:
Hey @adulau, @mokaddem do you mind looking at my minimalistic solution to this, PR here #10585

Wachizungu:
Hello, I see the associated PR was merged. Does that mean this issue can be closed?

pettai:
Yes, thanks for notifying

PR #10585, opened 2025-12-18
Add Object faster workflow

pettai:
#### What does it do?

Addresses Issue https://github.com/MISP/MISP/issues/10580 

I makes the Add Object jump into the search bar directly, so the user don't have to click several times, first to select ("meta")category of templates (or All Object, which I think most users uses anyway) and another click to open the search for template to select. 

#### Questions


pettai:
I now see that commit https://github.com/MISP/MISP/pull/10585/commits/69e8c704d52cf3914ce38b64a5127e9ec616e3e5 isn't optimal, since it actually allowed for multiple choices, which is wrong in this case.

iglocska:
Thanks a lot!

Issue #10583, opened 2025-12-18
Bug: Workflow Logic Module “IF :: Threat Level”: select default must be option key (ID), not label, as the label default breaks UI preselection + runtime type consistency


mzeeb:
### Actual behavior

In the workflow logic module “IF :: Threat Level”, the Threat Level parameter is a select built from the threat level mapping:
	•	1 = High
	•	2 = Medium
	•	3 = Low
	•	4 = Undefined

If the parameter default is set to the label (e.g. "Low"), the builder and runtime behave inconsistently:
	1.	The workflow builder dropdown preselects the wrong entry (shows High).
	2.	If the dropdown is not touched, runtime receives "Low" (string) instead of 3 (int).
	3.	If the user changes the dropdown, runtime then receives the correct numeric key.

### **Default is label (bad)**

```php
        $this->params = [
            [
                'id' => 'condition',
                'label' => 'Condition',
                'type' => 'select',
                'default' => 'equals',
                'options' => $this->operators,
            ],
            [
                'id' => 'threatlevel',
                'label' => 'Threat Level',
                'type' => 'select',
		'default' => 'Low', // <---- Here is the wrong value
                'options' => $this->threatlevels_mapping,
                'placeholder' => __('Pick a threat level'),
            ],
        ];
```

- UI preselects **High**
    
- runtime value is "Low" (string) if untouched
    
- runtime value becomes 3 (int) only after changing dropdown

### Expected behavior

For type: select, the parameter default should be the **option key**.

The UI should preselect the corresponding label, and runtime should always receive the key.

If the default is set to the numeric key (3), both UI preselection and runtime behavior are correct and consistent.
  

### **Default key**

```php
        $this->params = [
            [
                'id' => 'condition',
                'label' => 'Condition',
                'type' => 'select',
                'default' => 'equals',
                'options' => $this->operators,
            ],
            [
                'id' => 'threatlevel',
                'label' => 'Threat Level',
                'type' => 'select',
		'default' => 3, // <---- Here is the fixed value
                'options' => $this->threatlevels_mapping,
                'placeholder' => __('Pick a threat level'),
            ],
        ];
```

- UI preselects **Low**
    
- default runtime value is 3 (int), independent of UI interaction

### Steps to reproduce

Reproduction steps
	1.	Add IF :: Threat Level node to a workflow
	2.	Open workflow builder → dropdown preselects High
	3.	Save without changing dropdown and execute → runtime receives "Low" (string)
[...]

PR #10587, opened 2025-12-18
fixed workflow logic

Adeosunomoshalewa:
#10583
Fixed  Workflow Logic Module for the  “IF :: Threat Level
#10583


adulau:
Should be fixed in develop branch. Thanks again for the reported issue.

Issue #10578, opened 2025-12-16
Bug: Not simple to find the correct object template 


pettai:
### Actual behavior

After attending many MISP trainings, this question always comes up:

"How do report an ActiveDirectory Username?"

There is no attribute for AD account name, nor an easy way of finding a fitting object for this...

### Expected behavior

We'd like to see an "ad-username" or "(ms-)username" (or similarly named) attribute, so it's easy for the MISP users to find.
This new attribute could also be included in an object with more metadata about the account.

### Steps to reproduce

Add Attribute -> Search for "user" yields user-agent, target-user, github-user

### Version

2.5.x

### Operating System

Ubuntu

### Operating System version

22.04

### PHP version

8.x

### Browser

Firefox

### Browser version

latest

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

adulau:
The object for AD is usually https://www.misp-project.org/objects.html#_user_account and where there is a set of `account-type` . If you see something to update in the object, let us know.

pettai:
Thanks for looking at this issue 👍 

Since you point at an object, and you what users of MISP to find/use that object conveniently, then I suggest this:
https://github.com/MISP/MISP/issues/10580.

Currently, Attributes are much simpler to add since it's current workflow is simpler much faster.

pettai:
Btw. This PR is also touching the subject of how unhelpful it is to find the correct object template:
https://github.com/MISP/MISP/issues/9821 (although, it's about coin-address vs cryptocurrency)

pettai:
> The object for AD is usually https://www.misp-project.org/objects.html#_user_account and where there is a set of `account-type` . If you see something to update in the object, let us know.

The only thing that I don't like about it (makes it harder to find) is the category in which it's placed "ftm". 
It's not optimal to search for a sysadmin object in the "follow the money" category ...

adulau:
It's in `misc` category https://github.com/MISP/misp-objects/blob/main/objects/user-account/definition.json

I can change it if you want.

pettai:
No need to change the category, I don't think it would help.

An issue in the GUI is the sorting/displaying of the search results, and it's also a bit confusing that the objects have the same name except for letters in caps and a dash

[...]

PR #10588, opened 2025-12-19
Enhanced sorting in generic_picker.ctp

pettai:
Adds better sorting, so the most relevant item ends up at the top of the list.

#### What does it do?

Sorts based on the search string and puts items that starts with the search string on top (instead of sorting the items in alphabetical order)

Example: using Add Object, and searching for the an object template. Typing in "user" in the search field, will sort and highlight "user-account" at the top 
The current behavior puts "ftm-UserAccount" at the top, then "github-user", etc. 

Fixes Issue described in https://github.com/MISP/MISP/issues/10578

#### Questions


iglocska:
That's a good idea, thanks!

PR #10590, opened 2025-12-19
Server views migration

ThomasLcr:
## Migration of all Server views

- Index [done] 
- Preview index [done]

Screenshot from 2025-12-22 14-51-47


iglocska:
I've just had a check and it produces a bunch of notice errors, make sure you have debug enabled when testing it (also seeing it in org fields):

image



Besides that the pagination/sorting seems to be broken:

image

ThomasLcr:
Thanks for the feedback. Indeed, I forgot to test the pagination and debug mode.

iglocska:
Confirmed, much appreciated!

Issue #10563, opened 2025-12-05
JakubOnderka/OpenID-Connect-PHP not "really" supported


maxime-escourbiac:
Hello Team, thanks again for the Hack.lu event this year it was awesome !

We would like to report recurring issues with the OpenID authentication module in MISP, which currently depends on the external library OpenID-Connect-PHP (https://github.com/JakubOnderka/OpenID-Connect-PHP).

In its current form, the MISP integration does not fully support several essential behaviours required by modern IdPs. We encountered issues that were only resolved after applying the fixes introduced in this upstream pull request: https://github.com/JakubOnderka/OpenID-Connect-PHP/pull/28

Some Other MISP (using Okta) users have faced problems (https://github.com/JakubOnderka/OpenID-Connect-PHP/issues/26)

Given these repeated difficulties and the limited maintainership of the upstream project, we would like to propose an alternative approach: Would it be possible to directly integrate the four OpenID-related files into the MISP codebase instead of relying on the external dependency?

We would be happy to assist and integrate some fixes based on the issue reported on Jakub's project.

Thank you for considering this proposal!

adulau:
It's indeed a very good proposal. Don't hesitate to make PR too.  Thanks a lot.

maxime-escourbiac:
Hello @adulau, I saw you assigned someone, just to sure and avoid duplicate work. 

Yes, I can open a PR for that I think I will have some time next week to move the code from Jakub's project..

Thanks for the quick response.

+++

iglocska:
Please do, I'll keep an eye out then for your fix instead of touching it :) Thanks a lot!

maxime-escourbiac:
Hey @iglocska , I began to think about the integration an I figured out that the task will be more complex than expected specially for keeping the unit test developed.

The original CI/CD did not work so I made some changes on our forked project to fix that. In addition I bump the minimal php version required for the library.

My question is, will you consider to integrate our "OpenID-Connect-PHP" library instead of the original one and we will ensure the support for MISP usage ? 

https://github.com/certmichelin/OpenID-Connect-PHP

iglocska:
Hi @maxime-escourbiac-bib sure, though I would suggest to merge it upstream into @JakubOnderka's repo long term (will make maintenance easier) - but until then we can definitely go with your fork! 

Could you create a PR for the replacement?

maxime-escourbiac:
Hello @iglocska,

I've just submit the PR for this issue #10599.

[...]

PR #10599, opened 2026-01-06
[Oidc] Move from JakubOnderka to CertMichelin openid-connect-php

maxime-escourbiac:
This PR integrate a change for the library managing OIDC authentication, as mentioned in #10563.

We had some issues when integrating with Forgerock using the Request Object approach (for example, duplicated scope values between the JWT and POST parameters).

To address this, we added an optional flag that allows enforcing the “legacy” / standard parameter-based flow, even when request_parameter_supported in the .well-known/openid-configuration is set to true.

It introduces an optional configuration attribute (disable_request_object) allowing MISP to disable the Request Object flow and always use standard authorization parameters when needed.

The default behavior remains unchanged, ensuring full backward compatibility. Once this PR will be validated, I will submit a change on Docker integration project.

In addition, this PR improves the CI/CD pipeline and ensures better compatibility with the PHP versions currently supported by MISP.

Thanks in advance for your review !

+++

### DIFF
--- a/app/Plugin/OidcAuth/Lib/Oidc.php
+++ b/app/Plugin/OidcAuth/Lib/Oidc.php
@@ -205,7 +205,7 @@ class Oidc
 
         try {
             $oidc->refreshToken($userInfo['refresh_token']);
-        } catch (CertMichelin\ErrorResponse $e) {
+        } catch (CertMichelin\ErrorResponse | JakubOnderka\ErrorResponse $e) {
             if ($e->getError() === 'invalid_grant') {
                 $this->log($user['email'], "Refreshing token is not possible because of `{$e->getMessage()}`, considering user is not valid");
                 return false;
@@ -290,7 +290,7 @@ class Oidc
     }
     
     /**
-     * @return \CertMichelin\OpenIDConnectClient
+     * @return OpenIDConnectClient CertMichelin if available, otherwise JakubOnderka for keeping backward compatibility.
      * @throws Exception
      */
     private function prepareClient()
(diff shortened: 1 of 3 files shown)


app/Plugin/OidcAuth/Lib/Oidc.php
```diff
@@ -205,7 +205,7 @@ public function isUserValid(array $user, $ignoreValidityTime = false, $update =
 
         try {
             $oidc->refreshToken($userInfo['refresh_token']);
-        } catch (JakubOnderka\ErrorResponse $e) {
```
iglocska, line 208: Could you duplicate and wrap this try/catch into two if checks checking for the existence of CertMichelin\OpenIDConnectClient and JakubOnderka\OpenIDConnectClient? Otherwise if an admin didn't update the dependencies (extremely common) the catch will no longer execute for them.

iglocska:
I just had a look at the fork and there are some things removed that are puzzling (such as README.md and CHANGELOG.md content) could you re-add those? And additionally, could you do a PR on Jakub's library with your changes?

adulau:
Maybe keeping the whole history of the fork is also interesting (as the code was also a fork from another repository).

maxime-escourbiac:
Hello @iglocska & @adulau, thanks for the first review.

I’ve adapted the code paths (both exception handling and creation logic) to be compliant with Jakub’s dependency when CertMichelin is not present, ensuring there is no runtime crash when users apply the update. I also reintegrated and fixed the changelog, and pushed a 1.3.0 release to align the version number with the forked project. Regarding the README.md, I did not reintroduce the removed section, as Jakub’s project is a global OpenID client, whereas our fork is intended solely for use within MISP and does not support usage outside that integration.

+++

PR #10601, opened 2026-01-11
Added the "edge-extension-id" as Attribute, and Regex validator to "edge-extension-id" and "chrome-extension-id"

eCrimeLabs:
#### What does it do?

This adds the "edge-extension-id" as an attribute option to support data similar to the "chrome-extension-id"

#### Questions


iglocska:
It looks like I never replied to this, it is merged on develop, thanks a lot.

PR #10607, opened 2026-01-16
fix: [taxonomies] update getTagsForNamespace call to prevent OOM error

### LINKED ISSUE #10606: Bug: Taxonomies addTag causes OOM on large instance
### Actual behavior

When adding a single tag via the taxonomy UI endpoint `/taxonomies/addTag/taxonomy_id:<id>/name:<tag>`, the request fails with a PHP fatal out-of-memory error. In my case PHP exhausted 6 GB (`memory_limit=6442450944`).

While the request is running, MariaDB shows CakePHP executing a query against `attribute_tags` with a large `IN (...)` list of tag IDs, returning 9,398,861 rows. CakePHP attempts to hydrate the full result set into PHP arrays, which causes the OOM.

This occurs even though the action is conceptually “add one tag to taxonomy”.

### Expected behavior

Adding a single tag via `/taxonomies/addTag/...` should not load tag connector tables (`AttributeTag` / `EventTag`) for all tags in the taxonomy namespace, and should complete successfully regardless of the size of `attribute_tags`.

At most it should check whether the tag exists and then add or update the tag record.

### Steps to reproduce

1. Run MISP on an instance with a large `attribute_tags` table (in my case, `attribute_tags` contains millions of rows).
2. Navigate to Taxonomies and attempt to add a tag via the UI or directly call:
   `/taxonomies/addTag/taxonomy_id:314/name:atr%3Avulnerability-asset-type%3D%26quot%3Bapplication-software%26quot%3B`
3. Observe the request eventually fails with PHP OOM.

Optional confirmation:

* During the request, run `SHOW FULL PROCESSLIST;` on MariaDB and observe a query similar to:
  `SELECT AttributeTag.id, AttributeTag.attribute_id, ... FROM attribute_tags AS AttributeTag WHERE AttributeTag.tag_id IN (...)`
* Running an equivalent count query shows the result set size is very large:
  `SELECT COUNT(*) FROM attribute_tags WHERE tag_id IN (...);` returns `9,398,861`.

### Version

2.5.25

### Operating System

Ubuntu

### Operating System version

Debian 12 (misp-docker)

### PHP version

8.2.29

### Browser

_No response_

### Browser version

_No response_

### Relevant log output

```shell
2026-01-16 13:28:10 Error: Fatal Error (1): Allowed memory size of 6442450944 bytes exhausted (tried to allocate 20480 bytes) in [/var/www/MISP/app/Lib/cakephp/lib/Cake/Model/Datasource/Database/Mysql.php, line 285]
2026-01-16 13:28:10 Error: [InternalErrorException] Internal Server Error
Request URL: /taxonomies/addTag/taxonomy_id:<id>/name:<name>
Error in: /var/www/MISP/app/Lib/cakephp/lib/Cake/Error/ErrorHandler.php, Line: 292
Stack Trace:
#0 /var/www/MISP/app/Lib/cakephp/lib/Cake/Error/ErrorHandler.php(243): ErrorHandler::handleFatalError()
#1 /var/www/MISP/app/Lib/cakephp/lib/Cake/Core/App.php(970): ErrorHandler::handleError()
#2 /var/www/MISP/app/Lib/cakephp/lib/Cake/Core/App.php(943): App::_checkFatalError()
#3 [internal function]: App::shutdown()
#4 {main}
```

### Extra attachments

_No response_

### Code of Conduct

jsman:
#### What does it do?

Fixes #10606.

Prevents PHP out-of-memory when adding a tag via Taxonomies by avoiding eager loading of tag connector relations for an entire taxonomy namespace.

#### Questions

### DIFF
--- a/app/Model/Taxonomy.php
+++ b/app/Model/Taxonomy.php
@@ -432,7 +432,7 @@ class Taxonomy extends AppModel
         if (empty($taxonomy)) {
             return false;
         }
-        $tags = $this->Tag->getTagsForNamespace($taxonomy['Taxonomy']['namespace']);
+        $tags = $this->Tag->getTagsForNamespace($taxonomy['Taxonomy']['namespace'], false);
         $colours = $paletteTool->generatePaletteFromString($taxonomy['Taxonomy']['namespace'], count($taxonomy['entries']));
         foreach ($taxonomy['entries'] as $k => $entry) {
             $colour = $colours[$k];


iglocska:
Yikes, nice catch. Thanks a lot!

PR #10615, opened 2026-01-27
 Apply security best practices, add dependency review and OpenSSF sco…

pethers:
#### What does it do?

Created partly with StepSecurity app

* Enable comment summary in PR for dependency review

* OpenSSF Scorecard

#### Questions

- [NO ] Does it require a DB change?
- [ NO] Are you using it in production?
- [ NO] Does it require a change in the API (PyMISP for example)?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,3 +15,13 @@ updates:
     schedule:
       # Check for updates to GitHub Actions every weekday
       interval: "daily"
+
+  - package-ecosystem: pip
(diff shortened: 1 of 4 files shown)


pethers:
What I require for all Hack23 open source https://github.com/Hack23/ISMS-PUBLIC/blob/main/Open_Source_Policy.md#-policy-requirements 

image

Would be nice if a tool used by Swedish government organsations, at least had a better baseline.

PR #10618, opened 2026-01-29
Fixed ARIA attributes for read-only settings

obert01:
Only shows the "edit" ARIA button if the given setting is editable (nnot CLI-only).


iglocska:
Much appreciated, merged into develop!

PR #10619, opened 2026-01-30
chg: [internal] Reduce memory usage when processing events

JakubOnderka:
#### Questions


iglocska:
Thanks a lot, that's an awesome catch!

PR #10635, opened 2026-02-13
fix: [schema] remove incorrect requirement for sharing_group

Wachizungu:
#### What does it do?

Fixes duplicate / incorrect UUID requirement in the sharing_group part of the schema definition.

I wasn't sure if the same should be done for 2.4, as I suppose that one is obsolete.

#### Questions


iglocska:
Thank you, sir!

PR #10638, opened 2026-02-13
fix: [analyst_data:blocklists] use correct controller name for redire…

Wachizungu:
#### What does it do?

Just aligns with line 91, this should be the correct controller name I think.

I am not sure this code is callable in any way using the GUI at the moment (so I couldn't test), but perhaps @mokaddem can confirm.

#### Questions


iglocska:
Cheers!

PR #10640, opened 2026-02-14
fix: [SharingGroups:removeServer] use correct SharingGroupServer ID f…

Wachizungu:
…or deletion

#### What does it do?

Fixes a logic error. The function was:

1. Searching for a SharingGroupServer in the SharingGroup which had a matching server ID to the one provided as param
2. Then deleting the SharingGroupServer with ID equal to the server ID provided as param (not the matching SharingGroupServer ID)

Basically: the wrong SharingGroupServer was getting deleted, if there happened to be a coincidental match.

#### Questions


iglocska:
Oh la la thanks!

PR #10641, opened 2026-02-14
fix: [Servers:eventBlockRule] use success response for success, and fail r…

Wachizungu:
…esponse for fail

#### What does it do?
Fixes issue with wrong response type being used

#### Questions


Wachizungu:
This one is wrong / incorrect. Doesn't error but still.
Better version in https://github.com/MISP/MISP/pull/10668/changes

PR #10642, opened 2026-02-14
fix: [Cerebrate] correct error message in captureSg

Wachizungu:
#### What does it do?

Corrects a likely copy-pasta issue. Wrong error message.

#### Questions


iglocska:
Thanks, another

image

PR #10643, opened 2026-02-14
fix: [TaxiiServers] correct saveFailResponse action name for getColle…

Wachizungu:
…ctions

#### What does it do?

Corrects a copy pasta error. Sets the correct action name for the getCollections saveFailResponse

#### Questions


iglocska:
Thank you!

PR #10644, opened 2026-02-14
fix: [TaxiiServer] use correct casing for 'Basic' authorization header

Wachizungu:
#### What does it do?

Upercases the B in Basic, as per https://datatracker.ietf.org/doc/html/rfc7617#section-1.1

#### Questions


iglocska:
Thanks a lot!

PR #10645, opened 2026-02-14
fix: [Servers:saveCert] correct file size check

Wachizungu:
#### What does it do?

Adds () around the comparison, so the ! is correctly applied

#### Questions


iglocska:
wow, nice

PR #10646, opened 2026-02-14
fix: [Galaxy:detachCluster] set correct orgc id for attributes

Wachizungu:
#### What does it do?

Fixes a likely typo.
I don't think it has an impact at the moment (because it is not used anywhere in a way that could lead to abuse), but it could lead to ACL issue later on, so probably better to fix.

#### Questions


iglocska:
Thanks!

image

PR #10647, opened 2026-02-14
fix: [SharingGroupBlueprint:execute] set correct conditional 'id' val…

Wachizungu:
…ue to send to view

#### What does it do?

swaps the conditional assignment values. When $id is empty -> set to 'all', otherwise to $id.
I don't believe this value is actually used in the view though....

#### Questions


iglocska:
Nice, thanks!

PR #10648, opened 2026-02-14
fix: [GalaxyElements:index] correctly set canModify, to display delet…

Wachizungu:
…e button for those who have the correct privileges

#### What does it do?

Makes the delete button show up correctly (currently it is never shown).

image

#### Questions


iglocska:
Nice find, thanks!

PR #10649, opened 2026-02-14
fix: [UI:logs] don't try to do weird ajax redirect magic on search, d…

Wachizungu:
…isplay result properly

#### What does it do?

Fixes messed up UI on search on /logs/index

Without this fix you get something like this, instead of the index page updating properly:
image

#### Questions


iglocska:
Ooops, thank you!

PR #10653, opened 2026-02-15
fix: [SharingGroup] correctly allow user without sync perm to update …

Wachizungu:
…own sharing group, using REST api

#### What does it do?

fixes a bug that made it so users with perm_sharing_group but without perm_sync, weren't able to update sharing groups belonging to their org via REST api.

##### Before the fix
**Result of query**
owner-cant-edit-sharing-group-without-sync-perm-1

**User**
owner-cant-edit-sharing-group-without-sync-perm-2

**Org**
image

**Role**
owner-cant-edit-sharing-group-without-sync-perm-3

##### After the fix
owner-cant-edit-sharing-group-without-sync-perm-4-after-fix

#### Questions


iglocska:
Nice logic fix!

PR #10654, opened 2026-02-15
fix: [Servers] use 'validationErrors' instead of 'validationError'

Wachizungu:
#### What does it do?

Fixes likely typo.

Didn't test after this change so please check, or let me know if I need to!

#### Questions


iglocska:
Another mistake that Claude wouldn't have made... Thank you for finding it ;)

PR #10655, opened 2026-02-15
chg: [Feeds] remove redundant tags query in edit

Wachizungu:
#### What does it do?

Removes a redundant db query / assignment.
Not sure which one of the 2 should be kept, so leaving the one that ended up being used at the moment.

#### Questions


iglocska:
Every drop of performance helps, thanks!

PR #10656, opened 2026-02-15
fix: [ShadowAttribute] correct typo in afterSaveCorrelation db query …

Wachizungu:
…params

#### What does it do?

fixes a typo

#### Questions


iglocska:
Bloody hell. Nice catch!

PR #10657, opened 2026-02-15
fix: [Sightings] return saved 'false' on failures

Wachizungu:
#### What does it do?

Sets saved false for failure cases.

#### Questions


iglocska:
Thanks a lot!

Issue #3404, opened 2018-06-28
misp-modules timeout too low by default (on VM)


SteveClement:
Update packer build script to take a longer time out into consideration.

Potentially update this default to larger then 10 seconds.

WARNING:misp-modules:Timeout on ocr

SteveClement:
updated to 300 seconds.

Issue #3403, opened 2018-06-28
Tornado in misp-modules timing out when submit large PDF


SteveClement:
When submitting a large-ish file and script execution takes some time.
Tornado times out.

Investigate: how to make tornado wait :)

Todo: Update VM and INSTALL materials to take change into account.

ERROR:tornado.application:Exception in Future <Future at 0x7fb7580d81d0 state=finished raised WandError> after timeout
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/tornado/gen.py", line 936, in error_callback
    future.result()
  File "/usr/lib/python3.6/concurrent/futures/_base.py", line 425, in result
    return self.__get_result()
  File "/usr/lib/python3.6/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
  File "/usr/lib/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/local/lib/python3.6/dist-packages/misp_modules/__init__.py", line 185, in run_request
    response = mhandlers[x['module']].handler(q=jsonpayload)
  File "/usr/local/lib/python3.6/dist-packages/misp_modules/modules/import_mod/ocr.py", line 54, in handler
    image = img.composite(pdf.sequence[p], top=pdf.height * p, left=0)
  File "/usr/local/lib/python3.6/dist-packages/wand/image.py", line 543, in wrapped
    result = function(self, *args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/wand/image.py", line 2196, in composite
    self.raise_exception()
  File "/usr/local/lib/python3.6/dist-packages/wand/resource.py", line 222, in raise_exception
    raise e
wand.exceptions.WandError: wand contains no images `MagickWand-28' @ error/magick-image.c/MagickCompositeImageChannel/2043

SteveClement:
Fixed in upcoming misp-modules PR

PR #10661, opened 2026-02-16
fix: propagate Plugin.<family>_timeout to misp-modules /query payload

Frisb7:
#### Problem
* MISP reads `Plugin.<family>_timeout` (e.g., `Plugin.Action_timeout`) and applies it as the **HTTP client/socket timeout** when calling the module service `/query`.
* However, `misp-modules` enforces the **actual execution timeout** in Tornado using the JSON payload field `timeout` (defaulting to `DEFAULT_TIMEOUT = 300s` when missing).
* Since MISP does not include `timeout` in the `/query` JSON body, long-running Action modules still time out around ~5 minutes even when `Plugin.Action_timeout` is increased.

#### What does it do?
Inject `timeout` into the `/query` JSON payload before calling `sendRequest('/query', ...)`:
This aligns the configured `Plugin.<family>_timeout` with the **Tornado `with_timeout()` enforcement** in `misp-modules`, preventing the default 300s cutoff.

#### Related issues
* #3403 (Tornado timeout on long processing)
* #3404 (misp-modules timeout too low / “Timeout on ocr”)

#### Questions
* [ ] Does it require a DB change?
* [x] Are you using it in production?
* [ ] Does it require a change in the API (PyMISP for example)?

### DIFF
--- a/app/Model/Module.php
+++ b/app/Model/Module.php
@@ -283,6 +283,9 @@ class Module extends AppModel
         } else {
             $timeout = Configure::read('Plugin.' . $moduleFamily . '_timeout') ?: 10;
         }
+        if (!isset($postData['timeout'])) {
+            $postData['timeout'] = (int) $timeout;
+        }
         try {
             return $this->sendRequest('/query', $timeout, $postData, $moduleFamily);
         } catch (Exception $e) {


Frisb7:
needs some fix

iglocska:
Cheers, good catch!

PR #10662, opened 2026-02-16
fix: [TagCollections] correctly set a specific tag when submitting a …

Wachizungu:
…string like 'tag_name'

#### What does it do?

Submitting something like below currently just adds a tag at random, which is odd. Still not a perfect solution (what's the point of a LIKE to add 1 tag?), but okay.

This function has a bunch of problems, just starting with this.

image

#### Questions


iglocska:
Thank you, that's indeed wonky.

PR #10663, opened 2026-02-16
fix: [TagCollections] make it possible to add multiple tags at once v…

Wachizungu:
#### What does it do?

Fixes broken code, allows submitting something like this:
image

Also removes duplicate field in 'fields'

#### Questions


iglocska:
cheers!

PR #10664, opened 2026-02-16
fix: [TagCollections:removeTag] allow removing tags by setting tag_co…

Wachizungu:
…llection id in the POST body

#### What does it do?

Allows doing this:
image

without the change, the post body tag_collection id is not used

#### Questions


iglocska:
Nice, thanks!

PR #10666, opened 2026-02-17
fix: [ObjectTemplates] fix likely typo in pagination sort param

Wachizungu:
#### Questions


iglocska:
Thanks a lot!

PR #10668, opened 2026-02-18
fix: [ServersController:eventBlockRule] use success response for succ…

Wachizungu:
…ess, and fail response for fail

#### Questions


iglocska:
Mon dieu, thanks!

PR #10669, opened 2026-02-18
fix: [Tasks] use 'validationErrors' instead of 'validationError' in c…

Wachizungu:
…ontroller

#### What does it do?

Fixes likely typo.

Didn't test after this change so please check, or let me know if I need to!

#### Questions


iglocska:
Ooops. Nice catch.

PR #10670, opened 2026-02-19
CI: Fix PHP version in GitHub Actions matrix

Frisb7:
## What does it do?

Fixes a typo in the GitHub Actions workflow where `setup-php` receives an invalid PHP version (`${{ matrix.php }}a`).

#### Change
- Use the matrix value directly:
  - `php-version: ${{ matrix.php }}`

#### Why
The extra trailing character makes the version string invalid (e.g., `8.1a`, `8.3a`) and can cause CI setup to fail or behave unexpectedly.

#### Notes
No product/runtime code changes; workflow-only fix.

## Questions

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,7 +49,7 @@ jobs:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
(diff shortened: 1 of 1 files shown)


iglocska:
Nice catch, thanks!

PR #10671, opened 2026-02-19
fix: [Collections] make edits work for collections with sharing group…

Wachizungu:
… distribution

#### What does it do?

Make sure you don't get error on line 91 (because Event is not loaded at the right spot):
`$canSGBeUsed = $this->Event->SharingGroup->checkIfCanBeUsed($this->Auth->user(), $this->_isRest(), $data, 'Collection');`

To test:
Create a collection with distribution set to a sharing group.
Try editing the name of the collection afterwards -> internal server error
With this fix -> no error

#### Questions


iglocska:
Gracias!

PR #10672, opened 2026-02-20
fix: [Model:AnalystData] use correct variable name for empty check, b…

Wachizungu:
…efore filtering fields for notes in fetchChildNotesAndOpinions

#### What does it do?

fixes likely copy paste error.

I didn't verify functionality after, so please check extra.

#### Questions


iglocska:
Thank you sir, the copy pasta monster hit again!

PR #10680, opened 2026-02-24
Fix: [Decaying Models] Check for undefined array key decay_score

miaowdelivery:
Check for undefined array key "decay_score".

**What does it do?**
Addresses the error below:
```
Undefined array key "decay_score" in [app/Model/MispAttribute.php, line 2223
Warning: Warning (2): foreach() argument must be of type array|object, null given in [app/Model/MispAttribute.php
```

**Questions**


WallieGizmo:
The mentioned errors are triggered, when selecting events, using the option "excludeDecayed": true.
Result is an empty set.
On version 2.5.15, no error was triggered, and a complete(?) set of events was provided.

iglocska:
Thanks a lot!

PR #10698, opened 2026-03-12
Add `-y` to libfuzzy-dev install in `INSTALL.ubuntu2404.sh`

ComfortCurve2000:
#### What does it do?

If SSDEEP is selected for install then the script requires interaction. This change keeps everything non-interactive.
Add `-y` to libfuzzy-dev install in `INSTALL.ubuntu2404.sh` line 256.

#### Questions


adulau:
Good point. Thank you.

PR #10710, opened 2026-03-20
Iimprove install script

claudex:
#### What does it do?

Improvement in install script

#### Questions

- [x] Are you using it in production? No

Do no merge it yet, I need to run some more tests, but feel free to comment.

That's a lot of changes but I've split them in multiple commits to allow review by step and to document the changes for reference. I thinks it better to keep the multiple commits instead of squash them to avoid losing the reason of the change.


iglocska:
Awesome, thanks a lot, that's thorough as hell!

PR #10711, opened 2026-03-22
Remove Danish MISP Community entry

eCrimeLabs:
#### What does it do?

Removed Danish MISP Community metadata from defaults.json.

#### Questions


adulau:
Do you know one community which could replace this one in DK?

eCrimeLabs:
Currently not it is in the process of being re-thought :)

adulau:
Thank you and keep us informed if this changes in the future.

PR #10733, opened 2026-04-08
Fix baseurl in few more places

vpiserchia:
#### What does it do?

It fixes the baseurl in few places

#### Questions


iglocska:

PR #10737, opened 2026-04-12
fix: event delete and restSearchExport fail when MISP.use_uuids_in_urls = true

lekanjava:
## Summary

Fixes #10631 — Event deletion (and rest search export) from the UI fails with `405 / Invalid JSON input` when `MISP.use_uuids_in_urls` is set to `true`.

## Root Cause

When `MISP.use_uuids_in_urls` is enabled, the UI sends UUIDs instead of numeric IDs in URL paths. The JavaScript calls `GET /events/delete/<uuid>` to load the delete confirmation popup.

In `EventsController::delete()`, the GET branch (which renders the confirmation form) had:

```php
$eventList = is_numeric($id) ? [$id] : $this->_jsonDecode($id);

PR #10738, opened 2026-04-13
Show anonymized (AAAA****ZZZZ) API key in remote server view (#10731)

### LINKED ISSUE #10731: Feature Request: Show anonymized (AAAA****ZZZZ) API key in remote server view.

To ease debugging around what key is used. If i want to verify what key is used for a remote server i now hav to check in the DB using SQL.

In /servers/index, show anonymized (AAAA****ZZZZ) API key for remote server in the table view. To ease debugging around what key is used. Optionally show it anonymized when the "Sync user" "View" button is used.

_No response_

### Additional context

_No response_

### Code of Conduct

lekanjava:
Show anonymized (AAAA****ZZZZ) API key in remote server view (#10731)

What does it do?
It implements a feature to display an anonymized version of the API key used for remote server connections in both the /servers/index table view and the "Sync user" view button results.

Debugging remote server connections often requires verifying that the correct API key is being used. Previously, this required direct SQL access to the database. This change allows Site Administrators to see the masked key (format: AAAA****ZZZZ) directly in the UI, simplifying the debugging process while maintaining security.

Key Implementation Details:

Model Logic: Added Server::anonymizeAuthkey() which handles both plain and encrypted keys.
Access Control: The anonymized key is only fetched and displayed if the user is a Site Administrator. It is unset for all other users to ensure no leakage via UI or REST API.
API Support: Added the masked authkey field to the Server index API response (restricted to site admins).
Fixes #10731

Questions
 Are you using it in production? No (Development/Testing)
 Does it require a change in the API (PyMISP for example)? Yes (Added masked authkey field to server index response for site admins)

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -3709,7 +3709,7 @@ class EventsController extends AppController
                 $this->redirect(array('action' => 'index'));
             }
         } else {
-            $eventList = is_numeric($id) ? [$id] : $this->_jsonDecode($id);
+            $eventList = (is_numeric($id) || Validation::uuid($id)) ? [$id] : $this->_jsonDecode($id);
             $this->request->data['Event']['id'] = json_encode($eventList);
             $this->set('idArray', $eventList);
             $this->layout = false;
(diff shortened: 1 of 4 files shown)


elhoim:
This PR carries #10737's change as well as its own. `app/Controller/EventsController.php` here is byte-identical to #10737 — the same `(is_numeric($id) || Validation::uuid($id))` edit in `delete()` and `restSearchExport()` — and nothing in the description mentions a dependency, so a reviewer reading this in isolation would take that hunk to be part of the anonymised-API-key feature.

The anonymised-key work itself is the other four files (`ServersController.php`, `Server.php`, `Servers/index.ctp`).

Either would resolve it: note in the description that this is stacked on #10737 and should merge after it, or rebase onto `2.5` so the diff shows only the API-key change. The second is easier to review.

PR #10739, opened 2026-04-14
Force site admin to change org name from ADMIN (#1070)

### LINKED ISSUE #1070: Force site admin to change org name from ADMIN
Right now after installation the default organisation generated by MISP is called "ADMIN". 

We should force admins to change the organisation name on first login, similarly to how we force a password change / terms to be accepted.

lekanjava:
What does it do?
It implements a forced check for Site Administrators who belong to the default "ADMIN" organisation. Following a fresh installation, MISP creates a default organisation named "ADMIN". To ensure proper configuration and security, this PR forces the administrator to rename this organisation upon login.

The implementation is consistent with other "lockdown" features in MISP (like forced password changes):

Site Admins in the "ADMIN" org are redirected to the organisation edit page.
Direct navigation to other parts of the platform is blocked until the name is changed.
The user session is automatically refreshed upon a successful rename to clear the lock.
Fixes #1070

Questions
 Are you using it in production? No (Development/Testing)
 Does it require a change in the API (PyMISP for example)? No

### DIFF
--- a/app/Controller/AppController.php
+++ b/app/Controller/AppController.php
@@ -762,6 +762,20 @@ class AppController extends Controller
             }
         }
 
+        // Check if the host organisation is still named "ADMIN" and force a name change
+        if (
+            $this->_isSiteAdmin() &&
+            $user['Organisation']['name'] === 'ADMIN' &&
+            !$this->_isControllerAction([
+                'organisations' => ['admin_edit'],
+                'users' => ['logout', 'login']
+            ])
+        ) {
+            $this->Flash->info(__('Please change the default organisation name ("ADMIN") before proceeding.'));
+            $this->redirect(['controller' => 'organisations', 'action' => 'edit', 'admin' => true, $user['org_id']]);
+            return false;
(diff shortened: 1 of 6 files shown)


elhoim:
This PR appears to sit at the top of a three-branch stack and carries both of the others' changes.

- `app/Controller/EventsController.php` is byte-identical to #10737 (the `(is_numeric($id) || Validation::uuid($id))` edit).
- `ServersController.php`, `Server.php` and `Servers/index.ctp` are the anonymised-API-key feature from #10738.

The ADMIN-org lockdown this PR is actually about is `AppController.php` (+14) and `OrganisationsController.php` (+3) — 17 lines out of the 70 shown.

Nothing in the description mentions the dependency, so the diff reads as one feature bundling three unrelated changes. Rebasing onto `2.5` once #10737 and #10738 land would reduce this to the two files it is about; alternatively, noting the stack and the merge order in the description would let a reviewer skip the inherited hunks.

PR #10756, opened 2026-04-29
fix: [ACLComponent] make canModifyWarninglist return true for roles w…

Wachizungu:
…ith perm_warninglist

#### What does it do?

Without this change, the edit button doesn't show up on /warnignlists/view/<id> page with overmind theme, for people with perm_warninglist

#### Questions


iglocska:
Thanks a lot!

PR #10759, opened 2026-05-04
Initial work for supporting HA

vpiserchia:
#### What does it do?

Add HA capabilities to MISP, this is an initial work in order to get attention and help from the community

#### Questions

### DIFF
--- a/app/Config/config.default.php
+++ b/app/Config/config.default.php
@@ -94,6 +94,7 @@ $config = array(
         'event_view_filter_fields'       => 'id, uuid, value, comment, type, category, Tag.name',
         'live'                           => true,
         'uuid'                           => '',
+        'primary_uuid'                   => '',
         'delegation'                     => true,
         'max_correlations_per_event'     => 5000,
         'disable_auto_logout'            => false,
--- a/app/Controller/AppController.php
+++ b/app/Controller/AppController.php
@@ -1096,6 +1096,16 @@ class AppController extends Controller
         return $this->userRole['perm_site_admin'];
     }
 
+    // checks if HA is enabled and this instance is the Primary one
+    protected function _isHAPrimary()
+    {
+        if (!empty(Configure::read('MISP.uuid')) && !empty(Configure::read('MISP.primary_uuid'))) {
+            return (Configure::read('MISP.uuid') === Configure::read('MISP.primary_uuid'));
+        }
+        // back compatibility
+        return true;
+    }
+
     private function __captureParam($data, $param, $value)
     {
         if ($this->{$this->modelClass}->checkParam($param)) {
--- a/app/Controller/TasksController.php
+++ b/app/Controller/TasksController.php
@@ -57,7 +57,7 @@ class TasksController extends AppController
             throw new MethodNotAllowedException('You are not authorised to do that.');
         }
         if (!$this->_isHAPrimary()) {
-           throw new MethodNotAllowedException('The is not the primary instance in a HA deployment.');
+           throw new MethodNotAllowedException('Not a Primary instance in a HA deployment.');
         }
 
         $task = $this->Task->find('first', array(
@@ -90,7 +90,7 @@ class TasksController extends AppController
             throw new MethodNotAllowedException('You are not authorised to do that.');
         }
         if (!$this->_isHAPrimary()) {
-           throw new MethodNotAllowedException('The is not the primary instance in a HA deployment.');
+           throw new MethodNotAllowedException('Not a Primary instance in a HA deployment.');
         }
 
         $this->set('dropdownData', $this->__getDropdownData());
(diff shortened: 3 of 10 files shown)


vpiserchia:
In the previous PR  I mistakenly merged submodule changes. The new commit fixes the things

claudex:
Could you explained the scenario how this could be used. Because, if I understand correctly, you'll have 2 instances with the same database, in which you'll want to store the settings to synchronize them. This means that if the uuid is in the database one way or another, all instances will be primary 

Another issue I see is that you'll have externally configure the failover from one instance to another, wouldn't be simpler to just start the background jobs in that case on the new primary instance ?

In any case, it will be better to have a function to check the primary to avoid the same copy paste value everywhere which will be a pain to edit later.

vpiserchia:
> Could you explained the scenario how this could be used. Because, if I understand correctly, you'll have 2 instances with the same database, in which you'll want to store the settings to synchronize them. This means that if the uuid is in the database one way or another, all instances will be primary
> 

My usecase is a DR scenario. where I have two sites where one instance is active/primary and the other passive/secondary. A clustered db in between would keep the data synchronized.
As far as I know (please correct me) the MISP.uuid is not stored anywhere in the DB and it belongs to the app/Config/configure.php code. This file and the entire MISP code is not shared nor synchronized; it is private to the "site".
This would allow to differentiate the instances. (two or more)

> Another issue I see is that you'll have externally configure the failover from one instance to another, wouldn't be simpler to just start the background jobs in that case on the new primary instance ?

You are right, the failover must be triggered manually (by switching the primary_uuid setting). A background job on the new primary might work.

The issue is how can I prevent the previous primary node to run tasks in the case it starts over again. This is indeed unsolved with the current code; that's way I was thinking about to a different approach with two improvements:

1. store the primary_uuid in the DB (AdminSettings)
2. wrap the DB changes (C-UD) with a lock to prevent an instance to change the DB unless it holds the lock
 
> 
> In any case, it will be better to have a function to check the primary to avoid the same copy paste value everywhere which will be a pain to edit later.

The point 2. above would solve this as well

vpiserchia:
Thinking twice, the trigger of the failover should be allowed only to a Site Admin after a warning; independently from where it is activated. that is failover should be site transparent. otherwise a failing site will not be able to change the instance mode (primary/secondary)

claudex:
> As far as I know (please correct me) the MISP.uuid is not stored anywhere in the DB and it belongs to the app/Config/configure.php code.

If `system_setting_db` is enabled, it could be read from the database (with a fallback on the config file) like any other setting,

vpiserchia:
> > As far as I know (please correct me) the MISP.uuid is not stored anywhere in the DB and it belongs to the app/Config/configure.php code.
> 
> If `system_setting_db` is enabled, it could be read from the database (with a fallback on the config file) like any other setting,


Ok, this helps for keeping the db synchronized. HA deployment MUST require this param to be set; but then also the MISP.uuid is synchronized and it will not be possible to discern  the primary from the secondary

vpiserchia:
the `system_setting_db` is the opposite of what I want to achieve. I cannot afford a multi-master situation. Duplicated tasks will run and DB will be bloated

righel:
I don't believe infrastructure deployment choices should be handled in business logic code, the amount of places you should call `_isHAPrimary` is extremely large (probably hundreds if not thousands) and sometimes not very intuitive as many operations that look like read also write data, only missing one and the secondary will write to the db. 
On top of that, the application will be very broken when in secondary mode with all that unexpected `return false` statements. The PR doesnt solve the failover, if the primary fails who flips the flag?
The web application is not really the weak point for a HA MISP deployment, it's usually the db so proper replica, clustering should solve this, you can also add a reverse proxy to manage the load/traffic MISP receives.

claudex:
After some thinking. If you want to have a DR instance ready but not doing anything, you can just stop the workers by disabling supervisord and start it when you do the switch. On the previous primary instance, you just stop supervisord and prevent it to start and it won't do anything. Would that cover your use case ?

vpiserchia:
> I don't believe infrastructure deployment choices should be handled in business logic code, the amount of places you should call `_isHAPrimary` is extremely large (probably hundreds if not thousands) and sometimes not very intuitive as many operations that look like read also write data, only missing one and the secondary will write to the db. 
totally right, this approach is way too much intrusive and error prone. Fixing the DR and HA should be done elsewhere; but that said, I was looking to some way to put the MISP in read only mode.

>On top of that, the application will be very broken when in secondary mode with all that unexpected `return false` statements. The PR doesnt solve the failover, if the primary fails who flips the flag?
At this point it was not implemented, this discussion was a way to start discussion and collect the ideas

> The web application is not really the weak point for a HA MISP deployment, it's usually the db so proper replica, clustering should solve this, you can also add a reverse proxy to manage the load/traffic MISP receives.

The idea was to have a single settings that would stop the task execution. maybe the suggestion from @claudex  is a simple  way. more on his comment


thanks

PR #10764, opened 2026-05-06
fix: [CollectionElement] correct typo in mayView, properly access arr…

Wachizungu:
…ay element

#### What does it do?

Fixes a typo that would lead to an error if the function was used. No impact as this function isn't used currently (Collection-mayView is used instead, it has this version of the code).

#### Questions


adulau:
Thank you. I'm wondering if we similar typo in the code base with `(` instead of `[` but I bet not.

Wachizungu:
There are a few $array( instead of array(

PR #10765, opened 2026-05-06
fix: [Collections] correct typo preventing users belonging to the cre…

Wachizungu:
…ating/syncing user's org, from modifying data

#### What does it do?

These typos prevented users from the same organization as the creator, that were not the creator, from editing data.
Something similar for sync users, but that part I didn't test/verify. I don't think collections even sync yet though.

#### Questions


adulau:
Good catch!

PR #10766, opened 2026-05-07
fix: [server:sync] Internal synchronisation also propagates local tag…

### LINKED ISSUE #10110: Bug: Local, exportable tags not exported between internal instances after being updated in source instance
### Actual behavior

Events with attributes containing local tags already synced between internal instances do not get updated local tags on attributes on sync/update/fetch when the event itself was already synced prior to new tags being added.

Adding a global/public tag to the attribute and republishing the event leads to the updated local tag also being synced over. Simply unpublishing the event and republishing it does not help.

### Expected behavior

Events already synced from A to B should have their local tags updated in B if these have been changed after previous sync

### Steps to reproduce

Two instances: A and B, where B is set ut to pull from A.

0. Create local, exportable tags "local:tag:test" and "local:tag:test2" in instance A
1. Create event in A with an internal reference attribute with any content. Apply tag "local:tag:test" to this attribute and sync event over to instance B.
2. Verify event in instance B contains attribute correctly tagged "local:tag:test"
3. Add tag "local:tag:test2" to the attribute in instance A
4. Attempt to pull updated event to instance B via both /servers/pull/1/full and /servers/pull/1/update, as well as "Fetch this event" on /servers/previewEvent/1/<guid>
5. Attribute in instance B never receives updated tag "local:tag:test2".

### Version

2.5.2

### Operating System

Debian (official docker image)

### Operating System version

12

### PHP version

8

### Browser

All

### Browser version

_No response_

### Relevant log output

_No response_

### Extra attachments

_No response_

### Code of Conduct

mokaddem:
Internal synchronization also propagates local tags if the event was already synced.

- Fix #10110

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -847,7 +847,7 @@ class EventsController extends AppController
             $rules = [
                 'recursive' => -1,
                 'fields' => array('id', 'timestamp', 'sighting_timestamp', 'published', 'uuid', 'protected'),
-                'contain' => array('Orgc.uuid'),
+                'contain' => array('Orgc.uuid', 'EventTag'),
             ];
         } else {
             // Remove user ID from fetched fields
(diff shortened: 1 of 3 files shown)


mokaddem:
Mr. @iglocska, since this is touching the sync, a review would be appreciated :smiley:

iglocska:
Looks good to me, makes sense. Thanks a lot!

PR #10767, opened 2026-05-08
chg: [Bookmark] log changes to audit logs

Wachizungu:
#### What does it do?
Makes it so bookmark modifications are logged to audit logs

#### Questions


iglocska:
Thanks a lot!

PR #10768, opened 2026-05-08
fix: [Attributes:addTag] allow adding multiple tags at once with list…

Wachizungu:
… of tag strings

#### What does it do?

I suppose this was merged from a fix originally done on 2.4, and didn't have the MispAttribute update

#### Questions


iglocska:
Thanks, indeed, looks like a 2.4 merge fail.

PR #10769, opened 2026-05-08
fix: [MispAttribute:advancedAddMalwareSample] set correct sharing gro…

Wachizungu:
…up id

#### What does it do?

Fixes a likely copy paste error.

Note: I did NOT validate this change

#### Questions


iglocska:
Yikes, nice find, thanks!

PR #10770, opened 2026-05-09
fix: [Warninglist:enableWarninglist] return 'Warninglist disabled' in…

Wachizungu:
… flash message when warninglist is disabled by passing 0 as url param

#### What does it do?

/warninglists/enableWarninglist/<id>/0 was correctly disabling a warninglist, but returning "Warninglist enabled" in the response.

Note: it is a bit odd that calling /warninglists/enableWarninglist/<id> actually disables a warninglist (enable param is false by default).

#### Questions


iglocska:
Thanks a lot!

PR #10771, opened 2026-05-11
Create a blocklist via the MISP workflows

cudeso:
#### What does it do?

Update a blocklist file with event attributes. Supports removal-by-tag, atomic writes with one-generation backup, and optional rolling filesystem snapshots.

image

#### Questions


iglocska:
Thanks a lot!

PR #10773, opened 2026-05-14
fix: [AttributeValidationTool] add start anchor to uuid validation regex

Wachizungu:
The UUID validation regex was missing a leading `^` anchor, allowing strings with arbitrary prefixes to pass validation. Added `^` to ensure the full value is matched from the start.

#### Questions


iglocska:
Thanks a lot!

PR #10774, opened 2026-05-14
fix: [UI] correctly show add custom galaxy button in side menu, for u…

Wachizungu:
…sers with access to the function

#### What does it do?

Makes sure you get this button / link:
image
#### Questions


iglocska:
Thanks a lot!

PR #10775, opened 2026-05-15
Update TAXII2 related code to conform to TAXII2 spec

microblag:
# What does it do?
The existing Taxii 2 config makes some assumptions about the shape of the taxii server that don't match the spec. This PR attempts to correct these deviations and cleanly updating existing configs.

## Questions
[Yes] Does it require a DB change?
[No] Are you using it in production?
[No] Does it require a change in the API (PyMISP for example)?


iglocska:
Thanks a lot!

PR #10776, opened 2026-05-15
fix: [Model:ObjectReference] correct 'assignment instead of compariso…

Wachizungu:
…n' bug in smartSave()

#### What does it do?
Not 100% sure about impact of the bug and the fix, but does look wrong hence the PR.

#### Questions


iglocska:
Good eye!

PR #10777, opened 2026-05-15
fix: [Bookmarks] use mayView for view() ACL check

Wachizungu:
#### What does it do?

This has pretty much 0 impact, as org users could still see these properly on index / menu, that's why I'm not flagging it as security fix / reporting that way.

#### Questions


iglocska:
Gracias!

PR #10779, opened 2026-05-16
fix: [ShadowAttributesController] harden shadow attribute batch creation workflow

skraft9:
#### What does it do?

This PR hardens the batch import logic in `ShadowAttributesController` to ensure data integrity during bulk operations.

Currently, the batch import loop relies on the model instance state to determine if a record is a new entry. This change explicitly clears the `id` field from the request data within the loop. This ensures that the framework consistently treats every item in a batch import as a new proposal (INSERT) rather than accidentally allowing an update (UPDATE) to existing records.

**Changes:**
- Added `unset($this->request->data['ShadowAttribute']['id'])` within the batch processing loop in `ShadowAttributesController::add()`.

#### Questions


iglocska:
Thanks a lot, very nice catch.

PR #10780, opened 2026-05-17
[doc] Fix default release to 2.5

frbayart:
#### What does it do?

Fix the default release to 2.5

#### Questions


adulau:
Thank you!

PR #10783, opened 2026-05-19
fix: [Feed] correct typos in __filterEventsIndex condition

Wachizungu:
#### What does it do?

Untested fix.
Suppose it should make the NOT filtering based on creator org work.

#### Questions


iglocska:
Oh ffs, nice catch.

PR #10788, opened 2026-05-22
[Security] Enforce strict URL schemes in TaxiiServersController

skraft9:
#### What does it do?

This PR improves input validation and connection handling for the `discovery_url` parameter in `TaxiiServersController::getRoot`. 

Previously, the endpoint lacked strict protocol validation, which could allow malformed URIs to be passed to `HttpSocket`. Additionally, raw socket exceptions were returned directly in the UI, leading to verbose error states when connections failed.

**Changes:**
- Added regex enforcement to ensure only `http://` and `https://` schemes are processed.
- Implemented network boundaries to prevent the socket from attempting routing to loopback (`127.0.0.0/8`) and cloud metadata endpoints (`169.254.169.254`).
- Standardized the error response block to log the detailed network error internally while returning a clean, generic message to the user, preventing verbose socket states from leaking into the UI.

#### Questions

### DIFF
--- a/app/Controller/TaxiiServersController.php
+++ b/app/Controller/TaxiiServersController.php
@@ -5,7 +5,6 @@ class TaxiiServersController extends AppController
 {
     public $components = array('Session', 'RequestHandler');
 
-
     public function beforeFilter()
     {
         // No need for CSRF tokens for a search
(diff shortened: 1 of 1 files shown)


skraft9:
- Verified patch will return a consistent error message, regardless if the destination host/service is unreachable. 
- Verified restrictions are effective when attempting to connect to localhost.

Screenshot 2026-05-22 133225

iglocska:
Thanks a lot, I dig it!

PR #10789, opened 2026-05-24
Fix Taxii Server view Collections UI element

lilithlarose:
#### What does it do?

Currently there is a bug in Taxii Server view, selecting collections triggers a malformed request i.e. https://[baseurl]https//[baseurl]/taxii_servers/collectionsIndex/[id]

This PR changes this request to return in the correct format i.e. https://baseurl]/taxii_servers/collectionsIndex/[id] in this view by removing a baseurl prepend in the component definition as this is currently handled in the base Accordion component.

#### Questions

### DIFF
--- a/app/View/TaxiiServers/view.ctp
+++ b/app/View/TaxiiServers/view.ctp
@@ -61,13 +61,13 @@ echo $this->element(
         ],
         'children' => [
             [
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks a lot, merged into develop!

PR #10792, opened 2026-05-26
[Security] Prevent deletion 500 error caused by 4-byte UTF-8 characters

skraft9:
#### What does it do?

This PR improves the stability of the event deletion process by ensuring better handling of multi-byte character encoding between different database tables.

**Technical Details:**
In some environments, a mismatch can occur between the character encoding supported by the primary `events` table and the `event_blocklists` metadata table. Specifically, when deleting an event that contains 4-byte UTF-8 characters (such as certain emojis or mathematical symbols), the transition to the blocklist logic can trigger a database-level exception if the target table is using a 3-byte configuration.

Because this exception occurs during the `beforeDelete` lifecycle hook, it interrupts the transaction, resulting in a 500 error and preventing the event from being removed through the standard interface.

**The Fix:**
Added a sanitization step to the `beforeDelete` function in `app/Model/Event.php` to filter character sequences that exceed the 3-byte limit before metadata is processed for blocklisting. This ensures that the deletion process remains functional and robust across various database configurations and legacy environments.

#### Questions

* [ ] Does it require a DB change? **No**
* [ ] Are you using it in production? **Verified in a containerized development environment.**
* [ ] Does it require a change in the API (PyMISP for example)? **No**

### DIFF
--- a/app/Model/Event.php
+++ b/app/Model/Event.php
@@ -370,15 +370,22 @@ class Event extends AppModel
         if (Configure::read('MISP.enableEventBlocklisting') !== false && empty($this->skipBlocklist)) {
             $this->EventBlocklist = ClassRegistry::init('EventBlocklist');
             $orgc = $this->Orgc->find('first', array('conditions' => array('Orgc.id' => $this->data['Event']['orgc_id']), 'recursive' => -1, 'fields' => array('Orgc.name')));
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks a lot, makes sense!

Issue #10716, opened 2026-03-24
Bug: Curiosity edgecase, Freetext import seems to be unable to deal with U+2003 "Em Space" characters


samitainio:
### Actual behavior

This one is a bit difficult to describe but it seems that whitespace character "U+2003" a.k.a  "Em Space" seems to be not handled correctly and it concatenates lines to filename attributes in some cases (see screenshot I don't know how to describe this one :D ) while in singular attributes before or after the attribute it just fails to detect e.g. file hashes.

### Expected behavior

All whitespaces are handled correctly and I can be my usual lazy analyst copy pasting iocs into freetext import 😃 

### Steps to reproduce

Case example ioc list from: https://cert.gov.ua/article/6287707

Copied the iocs as-is from the article into freetext import. Iocs type detection is filename if em spaces present and it detects a filename on the same line but in some tests it just failed to detect anything e.g. single hash with em space surrounding it without filename present.

### Version

Currently installed version… v2.5.35 (f8a38ba2bc56ca304b3411b23cf853312afe11d8)

### Operating System

Ubuntu 

### Operating System version

24.04.4 LTS

### PHP version

8.3.6

### Relevant log output

```shell
Issue #10715 seen happening at the same time but not necessarily related to this issue.
```

### Extra attachments

<img width="1975" height="571" alt="Image" src="https://github.com/user-attachments/assets/39421b91-128e-412f-8b9f-c4a085cf37cb" />

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #10794, opened 2026-05-27
Fix: Handle Em Space (U+2003) and Unicode separators in FreeText import

skraft9:
## What does it do?
This PR resolves issue #10716. 

The FreeText import parser was previously using `\s` for whitespace detection, which is limited to ASCII characters. This caused it to fail when encountering Unicode whitespace characters like the Em Space (U+2003), often found when copying IOCs from threat intelligence reports. This resulted in filenames and hashes being incorrectly concatenated together.

**Changes:**
- Updated the regex in `ComplexTypeTool::parseFreetext` to include `\p{Z}+` (the Unicode separator character class).
- Added the `/u` (Unicode) modifier to the `preg_split` pattern to ensure the string is treated as UTF-8.
- This ensures that various Unicode spaces are correctly treated as delimiters, significantly improving the robustness of the FreeText import for real-world analyst data.

## Questions
- [ ] Does it require a DB change? *No*
- [ ] Are you using it in production? *Verified in local containerized MISP environment with reproduction test strings.*
- [ ] Does it require a change in the API (PyMISP for example)? *No*

### DIFF
--- a/app/Lib/Tools/ComplexTypeTool.php
+++ b/app/Lib/Tools/ComplexTypeTool.php
@@ -310,8 +310,7 @@ class ComplexTypeTool
     {
         // convert non breaking space to normal space and all unicode chars from "other" category
         $input = preg_replace("/\p{C}+|\xc2\xa0/u", ' ', $input);
-        $iocArray = preg_split("/\r\n|\n|\r|\s|\s+|,|<|>|;/", $input);
-
+        $iocArray = preg_split("/\r\n|\n|\r|\s|\p{Z}+|,|<|>|;/u", $input);
         preg_match_all('/\"([^\"]+)\"/', $input, $matches);
         array_push($iocArray, ...$matches[1]);
 


skraft9:
## Pre-Patch Check
```bash
sudo docker exec -it misp-docker-misp-core-1 php -r '
    $input = "filename.exe" . "\u{2003}" . "hash.md5";
    // Put the parser logic here and echo the results
    $result = preg_split("/[\s]+/", $input);
    print_r($result);
'
Array
(
    [0] => filename.exe hash.md5
)
```

## Simulate Post-Patch
```bash
sudo docker exec -it misp-docker-misp-core-1 php -r '
    $input = "filename.exe" . "\u{2003}" . "hash.md5";
    // Using \p{Z} with the /u modifier
    $result = preg_split("/\r\n|\n|\r|\s|\p{Z}+|,|<|>|;/u", $input);
    print_r($result);
'
Array
(
    [0] => filename.exe
    [1] => hash.md5
)
```

The pre-patch regex ignored the Em Space because it only recognized ASCII whitespace, whereas the simulated fix enables Unicode awareness with the `/u` flag and `\p{Z}` property, allowing the parser to correctly split on all types of Unicode spaces.

---

## Verified Fix

FreeText Import Sample:
```
malware.exe d41d8cd98f00b204e9800998ecf8427e
evil_script.ps1 192.168.1.1
readme.txt deadbeefdeadbeefdeadbeefdeadbeef
```

image

---

Before the patch: The parser would treat `malware.exe[EM_SPACE]d41d8cd98f00b204e9800998ecf8427e` as a single, invalid filename, or fail to recognize the hash entirely.

After the patch: MISP should successfully split the line into two distinct attributes:
`filename: malware.exe`
`md5: d41d8cd98f00b204e9800998ecf8427e`

iglocska:
Nice catch, thanks! Bloody unicode issues...

Issue #10636, opened 2026-02-13
Bug: "star button" to set home page is broken, internal server error is returned


Wachizungu:
### Actual behavior

Stack Trace:
#0 /var/www/MISP/app/webroot/index.php(107): Dispatcher->dispatch()
#1 {main}
2026-02-13 14:16:25 Error: [TypeError] json_decode(): Argument #1 ($json) must be of type string, array given
Request URL: /userSettings/setHomePage
<redacted> (User #1) @ <redacted>
Error in: /var/www/MISP/app/Model/UserSetting.php, Line: 120
Stack Trace:
#0 /var/www/MISP/app/Model/UserSetting.php(120): json_decode()
#1 /var/www/MISP/app/Model/UserSetting.php(248): UserSetting::validate_homepage()
#2 /var/www/MISP/app/Model/UserSetting.php(514): UserSetting->checkSettingValidation()
#3 /var/www/MISP/app/Controller/UserSettingsController.php(371): UserSetting->setSetting()
#4 [internal function]: UserSettingsController->setHomePage()
#5 /var/www/MISP/app/Lib/cakephp/lib/Cake/Controller/Controller.php(500): ReflectionMethod->invokeArgs()
#6 /var/www/MISP/app/Lib/cakephp/lib/Cake/Routing/Dispatcher.php(193): Controller->invokeAction()
#7 /var/www/MISP/app/Lib/cakephp/lib/Cake/Routing/Dispatcher.php(167): Dispatcher->_invoke()
#8 /var/www/MISP/app/webroot/index.php(107): Dispatcher->dispatch()
#9 {main}

### Expected behavior

"Home page" is set to the current page on clicking the star button.

### Steps to reproduce

Ideally open dev tools > network tab before

1. Go to event index
2. Click the set homepage button (the star button in the global menu)
3. Nothing happens -> 500 returned can be seen in dev tools network tab

### Version

2.5.32

### Operating System

Ubuntu

### Operating System version

24.04

### PHP version

8.3

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

elhoim:
This looks resolved by #10795 (merged into `develop`, now on `2.5` and released in v2.5.39).

[...]

PR #10795, opened 2026-05-27
Fix: Handle array payload in UserSetting::validate_homepage to prevent PHP 8 TypeError #10636

skraft9:
## What does it do?

This PR resolves issue #10636.

The `setHomePage` functionality was returning a 500 Internal Server Error on PHP 8+. This occurred because the request payload was already being parsed into an array by the framework before reaching `UserSetting::validate_homepage()`. In PHP 8+, passing an array to `json_decode()` throws a fatal `TypeError` (Argument `#1` must be of type string, array given), which crashed the request.

Changes:
* Updated `UserSetting::validate_homepage()` to check if the incoming `$value` is a string before attempting to decode it.
* If `$value` is already an array, it bypasses `json_decode()`, preventing the fatal type error and allowing the homepage to be saved successfully.

## Questions

- [ ] Does it require a DB change? *No*
- [x] Are you using it in production? *Verified in local containerized MISP environment.*
- [ ] Does it require a change in the API (PyMISP for example)? *No*

### DIFF
--- a/app/Console/Command/AdminShell.php
+++ b/app/Console/Command/AdminShell.php
@@ -690,7 +690,7 @@ class AdminShell extends AppShell
                 $this->AdminSetting->delete($lock['AdminSetting']['id']);
             }
             $processId = empty($this->args[0]) ? false : $this->args[0];
-            $this->Server->runUpdates(true, false, $processId);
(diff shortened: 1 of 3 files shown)


iglocska:
Much appreciated!

PR #10796, opened 2026-05-28
fix: [SharingGroupOrg] correctly log removed org id, when removing an…

Wachizungu:
… org from a sharing group

#### What does it do?

Makes sure the add and remove logs are aligned:
image

#### Questions


iglocska:
Thanks a lot!

PR #10797, opened 2026-05-28
fix: [Users] correct some spelling errors in user related messages

Wachizungu:
#### Questions


iglocska:
Thanks, I can't English.

PR #10798, opened 2026-05-28
fix: [User] instantiate CryptoGraphicKey correctly for users:getGpgPu…

Wachizungu:
…blicKey

#### What does it do?

Function was giving an internal server error due to a typo in class name.

#### Questions


iglocska:
Thanks a lot!

PR #10803, opened 2026-05-31
fix: [CRUDComponent:edit] correctly override values if 'override' par…

Wachizungu:
…am is set

#### What does it do?
**'override' wasn't use anywhere with edit -> so no impact of the bug yet.
Aligned with 'add'.  Values are in $input[$modelName][$field]

#### Questions


iglocska:
Thanks a lot!

PR #10822, opened 2026-06-05
Fix: Correct galaxy distribution schema and v151 migration

### LINKED ISSUE #10815: Bug: v2.5.39: SQL query for `151` issue
### Actual behavior

When upgrading to v2.5.39 we noticed an error with the SQL for db version 151.
```The executed SQL query was: ALTER TABLE `galaxies` ADD `distribution` tinyint(4) NOT NULL DEFAULT 0; The returned error is: SQLSTATE[42S21]: Column already exists: 1060 Duplicate column name 'distribution' However, as this error is allowed, the update went through.```
Even though the update went through the definition of the distribution column in the galaxies table remains to have a default value of NULL and not 0 as is implied by the ALTER TABLE statement above.
```ALTER TABLE `galaxies` MODIFY `distribution` tinyint(4) NOT NULL DEFAULT 0;```
The above sql fixes this.

### Expected behavior

Use a modify statement if the column exists.

### Steps to reproduce

Upgrade from a version prior to 2.5.39 to that version.
Check the update progress or the update_database logs.

### Version

2.5.39

### Operating System

Ubuntu

### Operating System version

24.04

### PHP version

8.3

### Browser

_No response_

### Browser version

_No response_

### Relevant log output

```shell

```

### Extra attachments

_No response_

### Code of Conduct

skraft9:
#### What does it do?

Fixes #10815

During the v151 database migration, upgrading an older instance that already contained a `distribution` column in the `galaxies` table (where it allowed `NULL`) resulted in a `1060 Duplicate Column` error. Because this error is intentionally suppressed during upgrades, the `ADD` command was skipped, leaving the existing column with a default of `NULL` instead of the required `0`.

This PR makes two adjustments to resolve the desync:
1. Updates the v151 update script in `AppModel.php` to use `MODIFY` instead of `ADD` for the `galaxies` table's `distribution` column, ensuring existing columns correctly inherit `NOT NULL DEFAULT 0`.
2. Updates `db_schema.json` so the blueprint correctly defines the distribution default as `"0"` instead of `null`, establishing the proper source of truth for the ORM.

#### Questions

- [x] Does it require a DB change? (Yes)

### DIFF
--- a/app/Model/AppModel.php
+++ b/app/Model/AppModel.php
@@ -2692,7 +2692,7 @@ class AppModel extends Model
                 $sqlArray[] = "ALTER TABLE `galaxy_clusters` MODIFY `description` text CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;";
                 break;
             case 151:
-                $sqlArray[] = "ALTER TABLE `galaxies` ADD `distribution` tinyint(4) NOT NULL DEFAULT 0;";
+                $sqlArray[] = "ALTER TABLE `galaxies` MODIFY `distribution` tinyint(4) NOT NULL DEFAULT 0;";
                 break;
             case 'fixNonEmptySharingGroupID':
                 $sqlArray[] = 'UPDATE `events` SET `sharing_group_id` = 0 WHERE `distribution` != 4;';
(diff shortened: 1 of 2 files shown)


iglocska:
Thanks a lot, this explains some issues we saw with the upgrade on some instances!

PR #10826, opened 2026-06-05
Fix: Category mismatch when using bulk Type change in Freetext Import

### LINKED ISSUE #10799: Bug: Category not matched when Type changed for all in Freetext Import
### Actual behavior

When using the Freetext import to populate attributes, if the bulk type change option is used to change type of all attributes to a one falling into different than the pre-popupated category, the category is not updated respectively. This results in inconsistency that seemingly causes no attributes to be added to an event.

### Expected behavior

Category should be updated to the one under which the Type selected exists.

### Steps to reproduce

1. Create an event
2. Select "Populate from..."
3. Select "Freetext Import"
4. Paste multiple lines
5. On the "Freetext Import Results" page, use the "Change all" button to change the Type of all attributes to one that falls under different category than the originally populated type.
6. Observe that Type is changed to the desired one for all attributes, while Category remains.

### Version

2.5.38

### Operating System

Ubuntu

### Operating System version

24.04

### PHP version

8.3

### Browser

Chrome

### Browser version

148.0.7778.179

### Relevant log output

```shell

```

### Extra attachments

Image

### Code of Conduct

skraft9:
## What does it do?

Fixes #10799.

When using the Freetext Import bulk "Change all" functionality, the Type dropdowns were updated programmatically using `.val()`, which failed to trigger the necessary JavaScript event listeners. As a result, the Category field remained static and did not re-map to the new Type.

This PR adds `.trigger('change')` to the bulk update loop within `changeFreetextImportExecute()`. This forces the UI to re-evaluate the category-type relationship, ensuring the Category field correctly synchronizes with the newly selected Type.

## Questions

### DIFF
--- a/app/webroot/js/misp.js
+++ b/app/webroot/js/misp.js
@@ -2857,7 +2857,9 @@ function changeFreetextImportExecute() {
     $('.typeToggle').each(function() {
         if ($(this).val() === from) {
             if (selectContainsOption("#" + $(this).attr('id'), to)) {
-                $(this).val(to);
+                // Update the value, then manually trigger the 'change' event
+                // so the Category dropdowns update automatically.
+                $(this).val(to).trigger('change');
             }
         }
     });


skraft9:
## Patch Testing

Added sample data in Freetext Import
Screenshot 2026-06-05 165013

Used the "Change All" button and verified the Category changes.
Screenshot 2026-06-05 165028

iglocska:
Cheers, makes sense!

PR #10827, opened 2026-06-05
Fix: Prevent OverCorrelatingValue::truncate() crash on null input

### LINKED ISSUE #10763: Bug: MalwareBazaar MISP feed fails with OverCorrelatingValue::truncate() null value error
### Actual behavior

Fetching the MalwareBazaar MISP-format feed fails consistently.

The background job starts correctly and imports part of the events, but eventually terminates with:

OverCorrelatingValue::truncate():
Argument #1 ($value) must be of type string, null given

The fetch process stops completely and no active fetchFeed process remains running.

Disabling correlation on the feed does not prevent the error.

### Expected behavior

The MalwareBazaar MISP-format feed should import successfully without crashing.

If malformed or null attributes are encountered, they should be skipped or handled gracefully without terminating the entire fetch job.

Disabling feed correlation should also prevent correlation-related processing from causing the fetch to fail.

### Steps to reproduce

1. Install MISP 2.5.37
2. Enable the MalwareBazaar MISP-format feed from abuse.ch
3. Enable feed fetching
4. Start the feed fetch from the GUI or using:
   sudo -u www-data /var/www/MISP/app/Console/cake Server fetchFeed 1 64
5. Wait for the import process to start
6. The job progresses partially, then terminates with:

   OverCorrelatingValue::truncate():
   Argument #1 ($value) must be of type string, null given

7. The fetchFeed process disappears from the process list

Additional note:
- Disabling feed correlation does not prevent the issue

### Version

2.5.37

### Operating System

Ubuntu

### Operating System version

24.04

### PHP version

8.3.6

### Browser

_No response_

### Browser version

_No response_

### Relevant log output

```shell
Error: OverCorrelatingValue::truncate(): Argument #1 ($value) must be of type string, null given, called in /var/www/MISP/app/Model/OverCorrelatingValue.php on line 30

#0 /var/www/MISP/app/Model/OverCorrelatingValue.php(30): OverCorrelatingValue::truncate()
#1 /var/www/MISP/app/Model/OverCorrelatingValue.php(46): OverCorrelatingValue->isBlocked()
#2 /var/www/MISP/app/Model/Event.php(5212): OverCorrelatingValue->block()
#3 /var/www/MISP/app/Model/Event.php(5046): Event->dedupAttributesForCorrelation()
#4 /var/www/MISP/app/Model/Feed.php(1194): Event->__add()
#5 /var/www/MISP/app/Model/Feed.php(765): Feed->__addEventFromFeed()
#6 /var/www/MISP/app/Model/Feed.php(1282): Feed->downloadFromFeed()

[JOB ID: ...] - failed with error code 1
```

### Extra attachments

_No response_

### Code of Conduct

skraft9:
## What does it do?

Fixes #10763.

The MalwareBazaar MISP-format feed occasionally contains malformed attribute values that result in a `null` value being passed to `OverCorrelatingValue::truncate()`. In PHP 8.3, passing `null` where a `string` is expected causes a fatal `TypeError`, which terminates the entire background feed fetch process.

This PR updates `OverCorrelatingValue::truncate()` to accept a nullable string (`?string`) and adds a defensive guard to return an empty string if the input is `null`. This ensures the deduplication pipeline gracefully handles malformed attributes without crashing the import job.

## Patch Testing
The fix was verified locally inside the MISP core container by isolating the method and simulating a `null` input, confirming the fatal error is successfully bypassed:

```bash
# Executed within the misp-core container
php -r 'class App { public static function uses($c, $t){} } class AppModel{} require "/var/www/MISP/app/Model/OverCorrelatingValue.php"; echo "Result: [" . OverCorrelatingValue::truncate(null) . "]\n";'

# Output confirms null is caught and an empty string is returned:
Result: []
```

## Questions

### DIFF
--- a/app/Model/OverCorrelatingValue.php
+++ b/app/Model/OverCorrelatingValue.php
@@ -8,12 +8,14 @@ class OverCorrelatingValue extends AppModel
     /** @var array */
     private $blockedValues = [];
 
-    public static function truncate(string $value): string
+    public static function truncate(?string $value): string
     {
+        if ($value === null) {
+            return '';
+        }
         $value = mb_strtolower($value);
         return mb_substr($value, 0, 191);
     }
-
     public static function truncateValues(array $values): array
     {
         return array_map(function(string $value) {


iglocska:
Thanks a lot, that's a nice catch!

PR #10829, opened 2026-06-07
new: [feeds] add Rectifyq MISP Feeds

rectifyq:
What does it do?
Added Rectifyq MISP Feeds (ICS-OT related feed, Malaysian context related feed)
Refer: https://rectifyq.com/resources

Questions
 Are you using it in production? Yes
 Does it require a change in the API (PyMISP for example)? Not sure


iglocska:
Thanks a lot!

PR #10833, opened 2026-06-09
fix: [UI] screenshot_box popup display issue when opening images atta…

AxeR44:
#### What does it do?

Fixed an issue causing the popup to appear behind the gray overlay when opening an image attached to an event

#### Questions


iglocska:
Thanks a lot!

PR #10837, opened 2026-06-11
feat: Bulk Add/Remove Tags on the Events index

### LINKED ISSUE #10831: Feature Request: Tagging multiple events at the same time

We rely heavily on labels to organize data in our MISP instance. Often when another instance pushes a large amount of new events, we want to assign them the same tag. This is a lot of work, because we can only tag one event at a time.

When selecting multiple events in the event overview, we would like an option that allows us to add one or more labels to all these events at the same time.

Ideally, you would be able to select events that are not on the current page, and apply labels to these as well. This way you can add labels to all events missing a specific filter.

It's possible to solve this using the API. However this is not very user friendly. 

### Additional context

_No response_

### Code of Conduct

skraft9:
## What does it do?

Fixes #10831

Creates bulk **Add Tags** and **Remove Tags** buttons to the Events index toolbar, letting a user tag or untag multiple events in one action instead of opening each event individually.

Selecting one or more events reveals the two new buttons next to the Delete/Export buttons.

---

## How it works

The feature is implemented on top of existing, authenticated MISP endpoints. 

- The picker is opened with the native `getPopup(id, 'tags', 'selectTaxonomy')` flow.
- Tag application reuses the native global `quickSubmitTagForm()` path, which fetches a fresh form (and CSRF token) per event via `fetchFormDataAjax()` before posting to `/events/addTag/<id>`. This avoids stale-token issues and keeps payload formatting identical to single-event tagging. Because `addTag` already accepts a JSON array of tag ids, applying multiple tags / tag collections works without extra handling.
- Tag removal uses the native confirm-form flow: `GET /events/removeTag/<id>/<tagId>` (which returns a form with a fresh token) followed by a `POST`, per event.
- Only events the user is permitted to modify (`data-can-modify`) are acted on. All ACL and taxonomy-exclusivity checks remain server-side in the existing actions.

---

## Implementation notes

- New file: `app/webroot/js/mass-event-tagging.js`, loaded only on the full-page Events index render via a single `$this->Html->script(...)` include (guarded by `!$ajax`).
- The script is self-contained in an IIFE and is a no-op if the native helpers are unavailable, so it cannot break unrelated JS.
- Buttons match the native toolbar styling (`btn btn-small btn-inverse`, icon-only with tooltips) produced by the `ListTopBar` element.
- Requests are issued sequentially with a per-target success/failure count and a final summary message; the page reloads once the batch completes.

---

## Questions

- [ ] Are you using it in production? Local docker / test instance

---

## Testing

Verified on a local MISP `v2.5.40` instance:
- Adding a tag to multiple selected events.
- Removing a tag from multiple selected events.
- Buttons appear only when at least one modifiable event is selected, and match the native toolbar visually.

---
New button to Add Tag for multiple events.
Screenshot 2026-06-11 171831

---
New button to Remove Tag for multiple events.
Screenshot 2026-06-11 171847

---
Select Tag for bulk add.
image

---
Select Tag for bulk add.
image

---
Successfully applied tag in bulk to two events at once.
image

---

### DIFF
--- a/app/View/Events/index.ctp
+++ b/app/View/Events/index.ctp
@@ -215,3 +215,6 @@ if (!$ajax
 if (!$ajax) {
     echo $this->element('/genericElements/SideMenu/side_menu', array('menuList' => 'event-collection', 'menuItem' => 'index'));
 }
+if (!$ajax) {
+    echo $this->Html->script('mass-event-tagging');
+}
--- a/app/webroot/js/mass-event-tagging.js
+++ b/app/webroot/js/mass-event-tagging.js
@@ -1,0 +1,191 @@
+/**
+ * Bulk Add / Remove Tags on the Events index.
+ *
+ * Loaded only on /events/index (see the include in app/View/Events/index.ctp).
+ * Reuses MISP's native tag picker and submit path rather than intercepting raw
+ * AJAX, so CSRF tokens, payload format, and ACL checks are all handled by core.
+ *
+ * Mechanism (verified against MISP 2.5.40 source):
+ *   - The native picker (getPopup(id,'tags','selectTaxonomy')) calls the global
+ *     quickSubmitTagForm(selectedTagIds, addData) when the user picks a tag.
+ *   - We wrap that function once. While "mass mode" is armed we ignore the
+ *     single picker event and instead loop over every selected event, fetching
+ *     a fresh CSRF-bearing form per request via the native helpers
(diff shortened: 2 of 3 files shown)


DaanWillems:
Amazing! Thank you. I hope it's released soon :)

iglocska:
Apologies, saw this PR too late and implemented something very similar in functionality on develop. It however lacks the ability to mass remove tags. Sorry for not checking before jumping on the issue!

PR #10838, opened 2026-06-11
feat: Reveal hidden setting behaviours (CLI only / file only / redacted) in server settings

### LINKED ISSUE #10812: Feature Request: More verbose setting posture information

Some of the server settings have some hidden behaviours, such as file only settings (for security), CLI only, etc. Users are unaware of which setting has which of these flags set.

Expose them in the server settings.

_No response_

### Additional context

_No response_

### Code of Conduct

skraft9:
## What does it do?

Fixes #10812

This PR reveals hidden behaviour server settings as small badges next to the setting name in **Administration → Server Settings & Maintenance**

New badges shown:

- **CLI only** this setting can only be changed from the command line (`cli_only`).
- **File only** for security reasons this value is always stored in the config file, never in the database (`SystemSetting::isSensitive()`: passwords, API keys, secrets, and the salt / encryption key).
- **Redacted** this value is hidden in the UI (`redacted`).
- **Read only** this setting cannot be edited from the web interface (`editable => false`).

## How it works

The behavioural flags already travel with each setting definition and reach the view; they were just never rendered. Therefore, this change is almost entirely presentational.

- `Server::__sortFinalSettings()` and `Server::serverSettingReadSingle()` now annotate each setting row with a `file_only` boolean derived from `SystemSetting::isSensitive()`, alongside the existing `setting` name. (`SystemSetting` is already imported in `Server.php`; `isSensitive()` is public static.)
- `app/View/Elements/healthElements/settings_row.ctp` builds a badge per behaviour present on the row, using MISP's existing `label label-*` styling with explanatory tooltips, and renders them next to the setting name.
- The previous inline `[CLI only]` text in the Description column is removed, since it is now represented consistently as a badge with the others.

## Implementation notes

- No new model logic and no controller changes the data was already present on each row.
- No database changes, no API changes.
- 2 files modified: `app/Model/Server.php` (annotation) and `app/View/Elements/healthElements/settings_row.ctp` (rendering).

## Questions

- [ ] Does it require a DB change? **No**
- [ ] Are you using it in production? **Local docker / test instance**
- [ ] Does it require a change in the API (PyMISP for example)? **No**

## Testing

Verified on a local MISP `v2.5.40` instance: badges render correctly across the MISP and Security tabs, including `CLI only` on CLI-restricted settings, `File only` on password / secret / key settings, and all three composed on `Security.encryption_key`. 

No layout regressions and no impact on the inline edit behaviour.

---

image

---

image

---

image

---


iglocska:
That's lovely, thanks a lot!

PR #10843, opened 2026-06-14
fix: [event] add redirects non-Overmind themes to missing view2 (#10839)

### LINKED ISSUE #10839: Bug: Creating new event redirects to /events/view2/{eventid}
### Actual behavior

When creating a new event, the user then gets redirected to /events/view2/{eventid} (Where {eventid} is a number).
However, view2 doesnt exist so this results in:
"Error: [MissingViewException] View file "Events/view2.ctp" is missing."

The event gets created tho, but the redirect seems broken. 

Is this theme-related maybe?
I'm using the default theme and not Overmind.

### Expected behavior

I expect it to create the event and then redirect me to the event. 

### Steps to reproduce

1. Run latest MISP-container (2.5.40)
2. Have MISP.enable_themes = false
3. MISP.default_theme = Default (or None/empty)
4. Create an Event.

### Version

2.5.40

### Operating System

Alma

### Operating System version

9.8

### PHP version

8.3.31

### Browser

Firefox, Chrome

### Browser version

_No response_

### Relevant log output

```shell
error.log says


Error: [MissingViewException] View file "Events/view2.ctp" is missing.
Exception Attributes: array (
'file' => 'Events/view2.ctp',
)
Request URL: /events/view2/{REDACTED_EVENT_ID}
Error in: /var/www/MISP/app/Lib/cakephp/lib/Cake/View/View.php, Line: 1031
Stack Trace:
#0 /var/www/MISP/app/Lib/cakephp/lib/Cake/View/View.php(470): View->_getViewFileName()
#1 /var/www/MISP/app/Lib/cakephp/lib/Cake/Controller/Controller.php(973): View->render()
#2 /var/www/MISP/app/Lib/cakephp/lib/Cake/Routing/Dispatcher.php(200): Controller->render()
#3 /var/www/MISP/app/Lib/cakephp/lib/Cake/Routing/Dispatcher.php(167): Dispatcher->_invoke()
#4 /var/www/MISP/app/webroot/index.php(107): Dispatcher->dispatch()
#5 {main}
```

### Extra attachments

_No response_

### Code of Conduct

skraft9:
## What does it do?

Fixes #10839.

Previously, submitting the Add Event form always redirected users to `/events/view2/{id}`. The `view2` action only ships a themed template (Overmind, EventTest) and lacks a default template at `app/View/Events/view2.ctp`. On the default theme, this redirect lands on a view that does not exist, causing CakePHP to throw a `MissingViewException` (*"View file Events/view2.ctp is missing"*). While the event is successfully created in the database, the user is presented with a 500 error page instead of the new event.

This PR guards the `view2` redirect behind the Overmind theme check and falls back to the standard `view` action for all other themes.

## How it works

The change introduces a single conditional statement within `EventsController::add()`. It mirrors the pattern already utilized in `AttributesController` and elsewhere in `EventsController`, where `view2` redirects are explicitly gated on `$this->theme === "Overmind"` with a fallback to `view`. Overmind users retain the new event view layout, while all other users are safely routed back to the working `view` action.

**Regression Source:** Introduced in commit `a6a06665f` (*"new: [Overmind] Start the migration of Event View"*), which swapped this post-create redirect from `view` to `view2` without providing a baseline default template.

## Questions

* [ ] Does it require a DB change? No
* [ ] Are you using it in production? Verified in local containerized MISP environment / test instance.
* [ ] Does it require a change in the API (PyMISP for example)? No

## Testing

Verified on a local MISP v2.5.40 instance with `MISP.default_theme = Default` and `MISP.enable_themes = false` (matching the exact conditions from the issue report):

* **Before the fix:** Creating an event redirects to `/events/view2/{id}`, which returns an HTTP 500 error due to `MissingViewException` (Verified on `.210` via redirect to `/events/view2/32`).
* **After the fix:** Creating an event redirects to `/events/view/{id}`, which returns a clean HTTP 200 OK and properly renders the new event (Verified on `.210` via redirect to `/events/view/33`).
* **Behavior Breakdown:** `/events/view/{id}` functions flawlessly for the default theme; `/events/view2/{id}` is now exclusively hit under the Overmind theme, which provides its own template files.ent-add-view2-redirect`

### DIFF
--- a/app/Controller/EventsController.php
+++ b/app/Controller/EventsController.php
@@ -3336,7 +3336,11 @@ class EventsController extends AppController
                     } else {
                         // redirect to the view of the newly created event
                         $this->Flash->success(__('The event has been saved'));
-                        $this->redirect(array('action' => 'view2', $this->Event->getID()));
+                        if ($this->theme === "Overmind") {
+                            $this->redirect(array('action' => 'view2', $this->Event->getID()));
+                        } else {
+                            $this->redirect(array('action' => 'view', $this->Event->getID()));
+                        }
                     }
                 } else {
                     if ($this->_isRest()) { // TODO return error if REST


iglocska:
Cheers, we discussed this internally the other day, the whole view2 thing wasn't that clever. I'll merge it for now, but in the longer run we should have something more abstracted that auto redirects to view2 on the appropriate themes internally, but with fixes like these it's not that urgent.

Thanks a lot!

PR #10850, opened 2026-06-15
fix: [doc] correct typo in OidcAuth documentation

Wachizungu:
#### Questions


righel:
Good catch, thanks!

PR #10854, opened 2026-06-16
Handle Redis BLPOP socket timeouts in mispzmq

cudeso:
#### What does it do?

- Handle Redis BLPOP socket timeouts in mispzmq
- Reconnect on Redis connection errors and
  continue publishing instead of exiting on timeouts

#### Questions


iglocska:
Thank you sir!

PR #10859, opened 2026-06-18
Scheduled taxii push

microblag:
- add TAXII as a scheduled task type with single-server and all-enabled push modes
  - enqueue TAXII pushes through the scheduler worker and server shell
  - add an enabled flag for TAXII servers, defaulting enabled for upgrades and new records
  - block manual and scheduled pushes for disabled TAXII servers
  - expose TAXII enabled state in default and Overmind add/index/view UI
  - filter scheduled TAXII task selection to enabled servers

#### Questions

- [Yes] Does it require a DB change?
- [No?] Does it require a change in the API (PyMISP for example)?


iglocska:
Thanks a lot!

PR #10860, opened 2026-06-18
Proxy support for taxii push

microblag:
Give the taxii push script proxy support and pass proxy settings to it.

#### Questions


iglocska:
Thanks a lot! (I keep forgetting about your username change, sorry for the delay...)

PR #10865, opened 2026-06-19
fix: [event:graph] honour extended view in event graph

tsattolo:
One line regression from d61dc04c5

#### What does it do?

Shows objects and attributes from extending events in an event's extended view.

#### Questions


Wachizungu:
Issue still present in MISP 2.5.46. Manually tested this patch and looks good to me.

PR #10867, opened 2026-06-20
fix: [Dashboards:EventStreamWidget] Force default columns if only inv…

Wachizungu:
…alid 'fields' are requested

#### Questions


iglocska:
Thanks a bunch!

PR #10870, opened 2026-06-22
Add option to disable OIDC PAR

BytesUnlimited-net:
#### What does it do?

Adds a new OIDC configuration option:

`OidcAuth.disable_pushed_authorization_request`

The option defaults to `false` and disables OAuth 2.0 Pushed Authorization Requests (PAR) when enabled.

This was added for compatibility with older Keycloak deployments using Kerberos authentication. These versions advertise a `pushed_authorization_request_endpoint` in their OIDC discovery document, but using PAR can lead to authentication failures.

A related upstream PR adds a dedicated method for disabling PAR:

https://github.com/certmichelin/OpenID-Connect-PHP/pull/2

This PR uses that method when available and falls back to the existing provider configuration mechanism otherwise.

#### Questions

* [ ] Does it require a DB change?
* [x] Are you using it in production?
* [ ] Does it require a change in the API (PyMISP for example)?


righel (approved):
Look good to me.
I also see you also added: https://github.com/MISP/misp-docker/pull/427

Thanks.

iglocska:
Thanks a lot!

PR #10877, opened 2026-06-23
chg: [Servers] add remote user's org name and uuid to 'view sync user'

### LINKED ISSUE #10876: Feature Request: Have the option to show more details (remote org name/uuid) about the remote user for "sync"

Which organization the remote user is linked to, has a huge impact on ACL, especially if you use sharing groups.

If people "mess up" the creation of the local org and user, this can create a lot of issues, which are sometimes difficult to debug. Being able to quickly check the org UUID of the remote user, would save a lot of headaches.

On the servers/index page:
When you click on "View" sync user, also include the remote org uuid, and perhaps other details like org name.

Image

I know you can call /users/view/me via an API call, but it's not always straightforward, depending on the environment you are in.

### Additional context

_No response_

### Code of Conduct

Wachizungu:
#### What does it do?

Fixes #10876 

Example result:
image

#### Questions

### DIFF
--- a/app/Model/Server.php
+++ b/app/Model/Server.php
@@ -5341,6 +5341,8 @@ class Server extends AppModel
 
             $results = [
                 __('User') => $user['User']['email'],
+                __('Organisation name') => $user['Organisation']['name'],
+                __('Organisation UUID') => $user['Organisation']['uuid'],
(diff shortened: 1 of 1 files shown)


iglocska:
Good idea, cheers!

PR #10878, opened 2026-06-24
fix: [Server] Correct typos in default value and description for MISP…

Wachizungu:
….log_errors_ndjson_path

### What does it do
Make it this, which I think is what it was supposed to be:
image

#### Questions


iglocska:
Thanks a lot!

PR #10883, opened 2026-06-27
fix: [freetext import] keep valid category on type change (#10873)

### LINKED ISSUE #10873: Bug: In Populate from function in freeTextImport, Network activity change to Payload delivery
### Actual behavior

Since you have injected IP addresses and want to change all value ip-dst to ip-src, the category is also changed from network activity to payload delivery

### Expected behavior

When you want to use the massive fonction to convert ip-src to ip-src, only this parameter had to be changed

### Steps to reproduce

From an event, click on populate from, submit a list of ip adresses, try to make a massive change ip-src to ip-dst and you will otice that the category is also changed. It was not like this before update.

### Version

2.4.42

### Operating System

Ubuntu 5.15.0-181-generic

### Operating System version

11

### PHP version

8.4.22

### Browser

Firefox

### Browser version

152.0.1 

### Relevant log output

```shell

```

### Extra attachments

Image

Image

### Code of Conduct

skraft9:
Fixes #10873. 

On type change in Freetext Import, keep the current category if it's still valid for the new type instead of always resetting to the default.

  - DB change? No
  - Used in production? No
  - API change? No

### DIFF
--- a/app/View/Events/resolved_attributes.ctp
+++ b/app/View/Events/resolved_attributes.ctp
@@ -282,9 +282,15 @@
                 }
                 */
                 var $categorySelect = $('#' + currentId);
+                var previousCategory = $categorySelect.val();
                 $categorySelect.empty();
                 for (var category in currentOptions) {
                     $categorySelect.append(new Option(category, category));
+                }
+                // Keep the user's current category if it is still valid for the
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks a lot!

Issue #10817, opened 2026-06-04
Feature Request: more sources for country identification for the dashboard map


iglocska:
### Is your feature request related to a problem? Please describe.

- country-code object relationship
- geolocations

### Describe the solution you'd like

Add them

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

skraft9:
Released with 2.5.43 - should be good to close

PR #10884, opened 2026-06-27
feat: [dashboard] geomap country sources: geolocation + country-code objects

skraft9:
#### What does it do?

Delivers on feature request #10817. 

Adds two country sources to the dashboard geolocation map widget: geolocation objects (`countrycode` field) and the country-code object relation. Both feed the existing ISO-alpha2 tally and are individually toggleable. 

Tested on a live 2.5.42 instance with seeded objects:
* Both sources tally to the correct countries and blend with the existing sources
* Input normalized (fr -> FR), country names and alpha-3 codes rejected
* Recency window and per-source toggles honored
* Full render path returns the WorldMap renderer with correct data
* Cache stays isolated per config; default config picks up the new sources
* Verified visually: 12 seeded countries all shaded on the dashboard map

#### Questions

- [X] Does it require a DB change? (No)
- [X] Does it require a change in the API (PyMISP for example)? (No)

### DIFF
--- a/app/Lib/Dashboard/AttributeGeoMapWidget.php
+++ b/app/Lib/Dashboard/AttributeGeoMapWidget.php
@@ -13,7 +13,7 @@
  * mapping is computed transiently per render and never persisted, so
  * there is nothing stable to link a region to.)
  *
- * It blends up to four independently-selectable sources (config key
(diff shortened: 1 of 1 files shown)


iglocska:
Awesome, thanks a lot!

Issue #10818, opened 2026-06-04
Feature Request: sharing group improvements


iglocska:
### Is your feature request related to a problem? Please describe.

- confidential sharing groups (can't see other orgs, required for NIS2)
- multiple sharing groups

### Describe the solution you'd like

- confidential sharing groups (can't see other orgs, required for NIS2)
- multiple sharing groups

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

cyberdashy:
Not exactly what you are after, but I have seen organisations codify the external organisation added to sharing groups. Recipients syncing the sharing group wont be able to identify the members, but it does require external documentation to manage the aliases.

skraft9:
@iglocska 2 PRs ready for review:

- #10885 
- #10886

PR #10885, opened 2026-06-27
feat: [sharing groups] confidential sharing groups: hide member org roster

skraft9:
#### What does it do?

Delivers part 1 of 2 feature request #10818 (the confidential sharing groups part; scoping the multiple sharing groups is next).

- Adds a per-sharing-group `confidential` flag. 
- When set, the member organisation roster is hidden from everyone except the owning organisation and site admins, on this instance and on synchronised peers. 
- Required for NIS2 style sharing where the participant list itself is sensitive. 
- Membership is still stored and used for access control, only its display is hidden, so distribution and sync are unaffected.

Tested on a live `2.5.42` instance with two orgs:
* Roster hidden from member orgs on a confidential SG, shown to owner org and site admins, across the SG view, SG index, shared event/attribute fetch, and the distribution graph
* Roster still shown normally on non-confidential SGs (no regression)
* Member org keeps full access to events shared via a confidential SG (confidentiality hides the roster, not the data)
* Flag persists on create and edit (round-tripped `0` to `1` to `0`) and is included in the sync attribute set
* Migration `154` applied cleanly, schema diagnostics green
* Two leaks caught and fixed during testing: the SG index and the distribution graph were not gating per SG

#### Questions

- [X] Does it require a DB change? (Yes, migration `154` adds `sharing_groups.confidential`)
- [X] Does it require a change in the API (PyMISP for example)? (No)

### DIFF
--- a/INSTALL/MYSQL.sql
+++ b/INSTALL/MYSQL.sql
@@ -1341,6 +1341,7 @@ CREATE TABLE `sharing_groups` (
   `modified` datetime NOT NULL,
   `local` tinyint(1) NOT NULL,
   `roaming` tinyint(1) NOT NULL DEFAULT 0,
+  `confidential` tinyint(1) NOT NULL DEFAULT 0,
   PRIMARY KEY (`id`),
   UNIQUE KEY `uuid` (`uuid`),
   UNIQUE KEY `name` (`name`),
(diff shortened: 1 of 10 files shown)


iglocska:
Let's lump this in with the upcoming release, this and the other SG changes are pretty hefty. Thanks a lot, it will come in super handy!

PR #10886, opened 2026-06-27
feat: [sharing groups] multiple sharing groups per event for read access

skraft9:
#### What does it do?

Delivers the multiple sharing groups part of #10818

* **Core Change:** Adds an `event_sharing_groups` link table to grant secondary read access to an event and its event-level attributes.
* **New Endpoints:** Adds `addSharingGroup` and `removeSharingGroup`.
* **Security:** Purely additive `OR` logic. It never loosens existing checks. Attribute visibility stays fail-closed, so org-only attributes remain hidden from secondary group members.

**Scope & Notes:**
* Sync, UI visibility, and attribute-level inheritance are excluded from this initial slice.
* Uses migration 154. If the first half of #10818 merges first, this will need renumbering.
* I read "multiple sharing groups" as one event shared with several groups at once. Please let me know if that needs to be reworked.

**Tested on a live 2.5.42 instance:**
* Link grants read access; removal revokes it immediately.
* Org-only attributes stay hidden from additional-group members.
* Behaviour is consistent across event view, index, and attribute `restSearch`.
* Write path successfully blocks attaching groups the user cannot access.

#### Questions

- [X] Does it require a DB change? (Yes, migration 154)
- [X] Does it require a change in the API (PyMISP for example)? (Yes, new endpoints)

### DIFF
--- a/INSTALL/MYSQL.sql
+++ b/INSTALL/MYSQL.sql
@@ -471,6 +471,20 @@ CREATE TABLE `event_locks` (
   KEY `timestamp` (`timestamp`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb3 COLLATE=utf8mb3_general_ci;
 /*!40101 SET character_set_client = @saved_cs_client */;
+
+--
+-- Table structure for table `event_sharing_groups`
+--
+
+CREATE TABLE `event_sharing_groups` (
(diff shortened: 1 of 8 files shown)


iglocska:
Thanks a lot for providing this, that's a super requested feature. I'll not merge it for this version yet as it's a pretty critical change and I want to spend more time with it, let's push it for the next version though. Massive thanks!

PR #10892, opened 2026-07-06
fix: add zeromq supervisor setting and status check

microblag:
## What does it do?
Adds the missing `Plugin.ZeroMQ_supervisor_managed` server setting as a CLI-only boolean setting, disabled by default.

When ZeroMQ is supervisor-managed, disabling Plugin.ZeroMQ_enable no longer attempts to kill the ZeroMQ process from the web process. This avoids failures in deployments where the ZeroMQ process is owned by an external supervisor; disabling the setting now only stops MISP from publishing messages.

When this setting is enabled, ZeroMQ diagnostics no longer require the local Python ZeroMQ library check from the web process. This supports deployments where web and supervisor processes are not on the same system.

ZeroMQ diagnostics now verify liveness using `PubSubTool::statusCheck()` instead of the PID-file based `checkIfRunning()` check, so diagnostics confirm that the service is actually responding.

### Questions

### DIFF
--- a/app/Model/Server.php
+++ b/app/Model/Server.php
@@ -2591,6 +2591,9 @@ class Server extends AppModel
         // If we are trying to change the enable setting to false, we don't need to test anything, just kill the server and return true.
         if ($setting === 'Plugin.ZeroMQ_enable') {
             if ($value == false || $value == 0) {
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks a lot!

PR #10895, opened 2026-07-08
fix: [ServersController:updateJSON] correctly return update results w…

Wachizungu:
…hen not using async update

#### What does it do?

Fixes a typo that caused "results" to be null for all updates

#### Questions


iglocska:
Cna't tpye. Thanks!

PR #10897, opened 2026-07-08
fix: [OrgEventsWidget] Fix target date computation

elena19m:
Computation of the target month and year failed to account for the case where the start date was before January of the previous year. The code considered the case when the target month was below 1, but trying to go back further than (12 + the current month) months resulted in a negative value for the target month. Changing the `if` to a `while` compensates for multiple years.

#### Questions


iglocska:
Thanks a lot!

PR #10900, opened 2026-07-09
fix: [Export] Remove stray leading newlines from module export files

svenlepamets:
#### What does it do?
Fixes an issue introduced in 2.5.40 - when exporting a MISP event through a MISP module, the file would always begin with two line feeds. This would corrupt, for example, an XML file.

The fix has been tested locally.

#### Questions

### DIFF
--- a/app/Model/EventReport.php
+++ b/app/Model/EventReport.php
@@ -1,5 +1,3 @@
-
-
 <?php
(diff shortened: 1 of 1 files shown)


iglocska:
Nice catch, thanks a lot!

PR #10907, opened 2026-07-13
fix: [export:cache] fix event scope export and allow specifying hash …

Wachizungu:
…algorithm

#### Questions


iglocska:
Much appreciated!

PR #10922, opened 2026-07-27
fix: [installer] correct cake Admin quoting in INSTALL.ubuntu2404.sh (#10919)

### LINKED ISSUE #10919: Bug: INSTALL.ubuntu2404.sh silently skips `Admin runUpdates` and `MISP.osuser` due to a quoting bug (`.../cake Admin` executed as a single path)
### Actual behavior

`INSTALL/INSTALL.ubuntu2404.sh` on the `2.5` branch quotes `"${MISP_PATH}/app/Console/cake Admin"` as a **single token** on two lines (467-468 as of commit `f63f851`):

https://github.com/MISP/MISP/blob/f63f85106b9235b7109aeaba9393dc30f86a4e5a/INSTALL/INSTALL.ubuntu2404.sh#L467-L468

```bash
sudo -u "${APACHE_USER}" "${MISP_PATH}/app/Console/cake Admin" setSetting "MISP.osuser" "${APACHE_USER}" &>>$logfile
sudo -u "${APACHE_USER}" "${MISP_PATH}/app/Console/cake Admin" runUpdates &>>$logfile
```

sudo therefore tries to execute a file literally named `cake Admin` (with a space), which does not exist, so **both commands fail on every fresh install**. Because both invocations redirect all output to `$logfile` and are not followed by `error_check`, the failure is completely silent: the installer finishes with a success message, but `MISP.osuser` was never set and the initial `Admin runUpdates` pass never ran.

These are the only two occurrences (`grep -n '/cake Admin"' INSTALL.ubuntu2404.sh`); every other `cake` invocation in the script uses the correct form `"${MISP_PATH}/app/Console/cake" Admin ...`. `UPGRADE.ubuntu2404.sh` is not affected.

### Expected behavior

`Admin setSetting MISP.osuser` and `Admin runUpdates` should actually execute during installation, using the same quoting as the rest of the script:

```diff
-sudo -u "${APACHE_USER}" "${MISP_PATH}/app/Console/cake Admin" setSetting "MISP.osuser" "${APACHE_USER}" &>>$logfile
-sudo -u "${APACHE_USER}" "${MISP_PATH}/app/Console/cake Admin" runUpdates &>>$logfile
+sudo -u "${APACHE_USER}" "${MISP_PATH}/app/Console/cake" Admin setSetting "MISP.osuser" "${APACHE_USER}" &>>$logfile
+sudo -u "${APACHE_USER}" "${MISP_PATH}/app/Console/cake" Admin runUpdates &>>$logfile
```

Optionally, adding `error_check` after each of these two calls would prevent this class of silent failure in the future. After applying this two-line fix locally, both steps complete normally on a fresh install. Happy to open a PR if that helps.

### Steps to reproduce

1. Fetch the installer from the `2.5` branch (commit `f63f851` at the time of writing):
   `wget -O INSTALL.ubuntu2404.sh https://raw.githubusercontent.com/MISP/MISP/refs/heads/2.5/INSTALL/INSTALL.ubuntu2404.sh`
2. `grep -n '/cake Admin"' INSTALL.ubuntu2404.sh` → shows the two affected lines (467-468).
3. Run the installer on a pristine Ubuntu 24.04 server → it completes and reports success.
4. Reproduce the failing invocation exactly as the script runs it:
   `sudo -u www-data "/var/www/MISP/app/Console/cake Admin" runUpdates; echo "exit=$?"` → `command not found`, exit 1 (see log output below).

### Version

2.5.44

### Operating System

Ubuntu Server

### Operating System version

24.04.4 LTS (pristine VM)

### PHP version

8.3.6

### Browser

N/A — server-side installer bug

### Browser version

_No response_

### Relevant log output

```shell
$ sudo -u www-data "/var/www/MISP/app/Console/cake Admin" runUpdates; echo "exit=$?"
sudo: /var/www/MISP/app/Console/cake Admin: command not found
exit=1
```

### Extra attachments

_No response_

### Code of Conduct

skraft9:
Fixes #10919.

Diagnosis and diff are @jjfz123's from the issue report - this PR is their proposed change, verified end to end.

### Problem

- `INSTALL/INSTALL.ubuntu2404.sh` lines 467-468 quote the binary and the shell name as one token: `"${MISP_PATH}/app/Console/cake Admin"`.
- sudo therefore looks for a file literally named `cake Admin`, which does not exist, and both commands fail with `command not found` on every fresh install.
- Both calls redirect to `$logfile` and neither is followed by `error_check`, so the failure is silent: the installer reports success while `MISP.osuser` was never set and the initial `Admin runUpdates` never ran.
- These are the only two occurrences in the script. Every other cake invocation, including `User init` and `User change_pw` on the very next lines, already uses the correct form. `UPGRADE.ubuntu2404.sh` is unaffected.

### Fix

- Quote only the binary path, so the subcommand is passed as its own argument.
- Two lines, matching the form used everywhere else in the script.

### Testing

Reproduced and verified against a 2.5.44 install, running the invocations exactly as the script does:

- Before, broken form: `sudo: /var/www/MISP/app/Console/cake Admin: command not found`, exit 1, nothing written to the logfile.
- After, fixed form: both commands exit 0, and the logfile contains `Setting "MISP.osuser" changed to "www-data"` and `All updates completed.`
- `bash -n` on the modified script passes, and no occurrence of the bad pattern remains.

### Questions

- Does it require a DB change? No
- Are you using it in production? No, verified on a 2.5.44 test instance
- Does it require a change in the API (PyMISP for example)? No

### DIFF
--- a/INSTALL/INSTALL.ubuntu2404.sh
+++ b/INSTALL/INSTALL.ubuntu2404.sh
@@ -464,8 +464,8 @@ error_check "Ensure redis is running"
 
 print_status "Running MISP updates"
 
-sudo -u "${APACHE_USER}" "${MISP_PATH}/app/Console/cake Admin" setSetting "MISP.osuser" "${APACHE_USER}" &>>$logfile
-sudo -u "${APACHE_USER}" "${MISP_PATH}/app/Console/cake Admin" runUpdates &>>$logfile
+sudo -u "${APACHE_USER}" "${MISP_PATH}/app/Console/cake" Admin setSetting "MISP.osuser" "${APACHE_USER}" &>>$logfile
+sudo -u "${APACHE_USER}" "${MISP_PATH}/app/Console/cake" Admin runUpdates &>>$logfile
 MISP_USER_KEY_FILE="$(mktemp)"
 sudo -u "${APACHE_USER}" "${MISP_PATH}/app/Console/cake" User init >"${MISP_USER_KEY_FILE}"
 MISP_USER_KEY="$(tr -d '\n' <"${MISP_USER_KEY_FILE}")"


iglocska:
Oops, thanks a lot!

PR #10955, opened 2026-08-15
Update INSTALL.ubuntu2404.sh

TheyCallMeZamani:
omit the quotation after the "ADMIN" word. it cause an error to run cake ! the quotations should embrace the cake path only.


iglocska:
Thanks a lot, woah, how did that go unnoticed?

PR #10957, opened 2026-08-18
Add ThreatCluster feeds to default feed metadata

Jam0k:
This PR adds two feeds from ThreatCluster (https://threatcluster.io), a threat-intelligence platform that clusters reporting from 20,000+ public sources and extracts validated indicators.

**1. ThreatCluster OSINT Feed** — `misp` source format
- Base URL: https://threatcluster.io/misp (serves `manifest.json`, `hashes.csv`, and per-event JSON per the MISP feed spec)
- Daily events with deduplicated attributes (network indicators + file hashes), correlated threat context (actor / malware / CVE galaxy tags where known), `tlp:clear` tagged
- No authentication required

**2. ThreatCluster IOC Blocklist (network indicators)** — `freetext` source format
- URL: https://threatcluster.io/api/iocs/public/feed.txt
- High-confidence malicious domains and IPs from the last 30 days, one per line with `#` comment header
- Indicators pass an automated validation pipeline (allowlist/anti-false-positive checks) before publication; false-positive contact is in the feed header

Both feeds are free for any use, served over HTTPS with no rate limits that would affect normal MISP fetch intervals. Several MISP instances already consume the feed by manual configuration; this adds it to the defaults for discoverability.

Happy to adjust entry fields (distribution, tags, defaults) to whatever the maintainers prefer.

### DIFF
--- a/app/files/feed-metadata/defaults.json
+++ b/app/files/feed-metadata/defaults.json
@@ -2505,5 +2505,51 @@
       "exportable": true,
       "hide_tag": false
     }
+  },
+  {
+    "Feed": {
+      "name": "ThreatCluster OSINT Feed",
+      "provider": "ThreatCluster",
+      "url": "https://threatcluster.io/misp",
+      "rules": "{\"tags\":{\"OR\":[],\"NOT\":[]},\"orgs\":{\"OR\":[],\"NOT\":[]}}",
+      "enabled": true,
+      "distribution": "3",
+      "default": false,
+      "source_format": "misp",
+      "fixed_event": false,
+      "delta_merge": false,
+      "publish": false,
+      "override_ids": false,
+      "settings": "{\"csv\":{\"value\":\"\",\"delimiter\":\"\"},\"common\":{\"excluderegex\":\"\"}}",
+      "input_source": "network",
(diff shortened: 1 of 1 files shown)


adulau:
Thanks for the pr. https://threatcluster.io/misp is not working and giving a JSON error message where an Authentification Bearer is required.

Jam0k:
Thanks for taking a look and apologies, that was a bad bit of presentation on our side rather than an auth requirement.

The feed is fully public. no key, no token, at any point. What you hit was a helper JSON at the base URL that described our separate subscriber tier, and it included an "authentication": "Required - Bearer token" line for that tier. I've reworded it now so the base URL now states "authentication": "none — this feed is fully public, no key or token required" up front.

The feed itself is unchanged and works as submitted. I've verified it end-to-end against a clean MISP 2.5 instance added as source_format: misp with no auth configured, Server fetchFeed pulled 8 events / 370 attributes with no errors:

https://threatcluster.io/misp/manifest.json (8 events)
https://threatcluster.io/misp/<uuid>.json (200, no auth)
https://threatcluster.io/misp/hashes.csv  (200, no auth)

Happy to adjust anything else.

PR #10966, opened 2026-08-27
[low] fix: [Collections] delegate CollectionElement::mayView()/mayModify() to Collection

### LINKED ISSUE #10937: CollectionElement::mayView()/mayModify() are an unreachable copy of Collection's and would error if called
Reporting rather than opening a PR, because the fix is a maintainer judgement call (delete vs. delegate) rather than a mechanical correction.

## What

`CollectionElement` defines its own `mayView()` and `mayModify()` (app/Model/CollectionElement.php:133-186). Both query for a `Collection` row **against the `collection_elements` table**:

```php
$collection = $this->find('first', [
    'recursive' => -1,
    'conditions' => ['Collection.id' => $collection_id]
]);
...
$collection['Collection']['org_id']   // read afterwards
```

`$this` is `CollectionElement`, so the `FROM` clause is `collection_elements`. With `recursive => -1` the `belongsTo Collection` association is not joined, so the alias `Collection` is not in scope and `Collection.id` is not a resolvable column. The query cannot succeed, and `$collection['Collection']` would be unpopulated regardless.

## Currently unreachable

Every real call site goes through the association to `Collection`'s own (schema-correct) implementations:

```
app/Controller/CollectionElementsController.php:45,77,103,114,156
    $this->CollectionElement->Collection->mayView(...)
    $this->CollectionElement->Collection->mayModify(...)
```

A grep for direct `CollectionElement->mayView(` / `->mayModify(` calls returns nothing. So nothing is broken in production today.

## Why it is worth raising anyway

The dead copy is being **actively maintained as though it were live**. Commit `5488f020d` ("fix: [Collections] correct typo preventing users belonging to the creating/syncing user's org, from modifying data") applied the same `Orgc_id`/`Org_id` → `orgc_id`/`org_id` typo fix to *both* `Collection.php` and `CollectionElement.php` in a single commit — i.e. someone spent effort fixing a method nothing calls, and will do so again.

There is also a related latent hazard: `Collection.php:150` used a strict `=== 4` on the `distribution` column, which is returned as a string because `app/Config/database.default.php` sets `PDO::ATTR_STRINGIFY_FETCHES => true`. I have a PR open for the live copy (#10927). `CollectionElement.php:176` carries the identical comparison, but I deliberately left it out of that PR — patching a comparison inside a method that would throw an SQL error before reaching it would imply the method works.

## Options, none of which I want to pick unilaterally

1. **Delete both methods.** Cleanest if they are genuinely vestigial; removes the duplicate-maintenance burden the git history shows already cost a wasted fix.
2. **Replace the bodies with delegation** — `return $this->Collection->mayView($user_id, $collection_id);` — if some future or external consumer is expected to call them on this model.
3. **Fix them properly** (query `$this->Collection`, correct the `=== 4`) if they are meant to grow independent element-level semantics later.

Happy to submit a PR for whichever you prefer.

Found during a read-only review of the `2.5` tree at `5d150ae89`.

elhoim:
**BLUF —** `CollectionElement::mayView()` and `mayModify()` are dead copies of `Collection`'s that query the wrong table.

- **Problem** — With `recursive => -1` on `CollectionElement` the `Collection` alias is never joined, so the conditions cannot resolve, and `mayView()` also dereferences a `SharingGroup` association the model does not have. Nothing calls them today — `CollectionElementsController` goes through the association to `Collection`'s own correct methods.
- **Fix** — Replaces both bodies with delegation to `Collection`.
- **Effect** — No runtime behaviour change; it removes a second copy that has to be fixed twice — it has already attracted a duplicated typo fix, and still carries a stale strict distribution comparison corrected upstream.

## Problem

`CollectionElement::mayModify()` and `CollectionElement::mayView()` (`app/Model/CollectionElement.php:133-186` before this change) are a copy of `Collection`'s implementations, but they run the lookup against the wrong table:

```php
$collection = $this->find('first', [
    'recursive' => -1,
    'conditions' => ['Collection.id' => $collection_id]
]);
...
$collection['Collection']['org_id']   // read afterwards
```

`$this` is `CollectionElement`, so the `FROM` clause is `collection_elements`. With `recursive => -1` the `belongsTo Collection` association (`app/Model/CollectionElement.php:13-18`) is not joined, so the alias `Collection` is not in scope — `Collection.id` is not a resolvable column and the query cannot succeed. Even if it did, `$collection['Collection']` would never be populated. `mayView()` additionally dereferences `$this->SharingGroup`, which `CollectionElement` does not have.

Nothing calls these today — every real call site goes through the association to `Collection`'s own, schema-correct implementations:

```
app/Controller/CollectionElementsController.php:45,77,103,114,156
    $this->CollectionElement->Collection->mayModify(...)
    $this->CollectionElement->Collection->mayView(...)
```

The reason this is still worth fixing is that the dead copy is being maintained as though it were live: commit `5488f020d` applied the same `Orgc_id`/`Org_id` → `orgc_id`/`org_id` typo fix to *both* `Collection.php` and `CollectionElement.php`, i.e. effort was spent fixing a method nothing calls. `CollectionElement.php` also still carries the strict `=== 4` distribution comparison that was corrected in `Collection.php` by #10927 — a fix that has to be remembered twice.

## Fix

Option 2 from the issue: replace both bodies with delegation to `Collection`.

```php
public function mayModify(int $user_id, int $collection_id)
{
    return $this->Collection->mayModify($user_id, $collection_id);
}
```

Delegation rather than deletion because it preserves the public API — these are public model methods and MISP modules/scripts could conceivably call them — while ending the duplicate-maintenance burden and inheriting the `(int)` cast from #10927 for free. The typed signatures are kept as-is. If maintainers would rather delete the methods outright (option 1 in the issue), I am happy to change the PR to that.

`$this->Collection` is available via the existing `belongsTo` association and is already used the same way by `CollectionElement::touchCollection()` (`app/Model/CollectionElement.php:127`).

## Verified

- `php -l app/Model/CollectionElement.php` — no syntax errors (PHP 8.5).
- Re-grepped the tree: no caller invokes `mayView`/`mayModify` directly on `CollectionElement`, so this change is unreachable-by-construction from existing code paths and cannot regress behaviour.
- Confirmed `Collection::mayModify()`/`mayView()` (`app/Model/Collection.php:107,134`) implement exactly the logic the removed copies intended, against the correct table.

## Not verified

- No runtime execution, no functional or unit test run — this is a static change only.

Fixes #10937

### DIFF
--- a/app/Model/CollectionElement.php
+++ b/app/Model/CollectionElement.php
@@ -132,57 +132,12 @@ class CollectionElement extends AppModel
 
     public function mayModify(int $user_id, int $collection_id)
     {
-        $user = $this->User->getAuthUser($user_id);
-        $collection = $this->find('first', [
-            'recursive' => -1,
-            'conditions' => ['Collection.id' => $collection_id]
-        ]);
-        if ($user['Role']['perm_site_admin']) {
-            return true;
-        }
-        if (empty($user['Role']['perm_modify'])) {
-            return false;
-        }
-        if (!empty($user['Role']['perm_modify_org'])) {
-            if ($user['org_id'] == $collection['Collection']['orgc_id']) {
-                return true;
-            }
-            if ($user['Role']['perm_sync'] && $user['org_id'] == $collection['Collection']['org_id']) {
-                return true;
-            }            
-        }
-        if (!empty($user['Role']['perm_modify']) && $user['id'] === $collection['Collection']['user_id']) {
-            return true;
-        }
-        return false;
+        return $this->Collection->mayModify($user_id, $collection_id);
     }
 
     public function mayView(int $user_id, int $collection_id)
     {
-        $user = $this->User->getAuthUser($user_id);
-        $collection = $this->find('first', [
-            'recursive' => -1,
-            'conditions' => ['Collection.id' => $collection_id]
-        ]);
-        if ($user['Role']['perm_site_admin']) {
-            return true;
-        }
-        if ($collection['Collection']['org_id'] == $user['org_id']) {
-            return true;
-        }
-        if (in_array($collection['Collection']['distribution'], [1,2,3])) {
-            return true;
-        }
-        if ($collection['Collection']['distribution'] === 4) {
-            $SharingGroup = ClassRegistry::init('SharingGroup');
-            $sgs = $this->SharingGroup->fetchAllAuthorised($user, 'uuid');
-            if (isset($sgs[$collection['Collection']['sharing_group_id']])) {
-                return true;
-            } else {
-                return false;
-            }
-        }
-        return false;
+        return $this->Collection->mayView($user_id, $collection_id);
     }
 
     public function deduceType(string $uuid)


elhoim:

410e8ef35ab6 (`chg: [collections] Drop dead CollectionElement::mayView/mayModify duplicates`) removes the duplicated `mayView`/`mayModify` bodies from `CollectionElement.php` outright. That is a better fix than this branch, which only delegated them to `Collection` — deleting the dead copies removes the divergence risk entirely rather than papering over it. The code this PR edits no longer exists, which is also why the branch shows as conflicting.

Thanks for taking care of it.

PR #10968, opened 2026-08-27
[critical] fix: do not apply regexp replacements to pdb attribute values

### LINKED ISSUE #1420: Do not apply regexp for path to pdb attribute type
### Work environment

| Questions | Answers |
| --- | --- |
| Type of issue | Bug |
| MISP version / git hash | 2.4.49 (misppriv as of 08 AUG) |
### Expected behavior
- PDB path attributes are not generified using regexps.
### Actual behavior
- PDB path attributes are generified using regexps (**C:\Users\mr.tekide\Documents\** becomes **%USERPROFILE%\Documents\**).
- Add a PDB path attribute to a stock install of MISP.
  - For example: _C:\Users\mr.tekide\Documents\Visual Studio 2013\Projects\paccrypt11njratmalii\paccryptalipnahzade\obj\Debug\LManager.pdb_ from [here](https://citizenlab.org/2016/08/group5-syria/).

The whole point of PDB paths is to preserve unique/identifying informations such as the user-names.

### LINKED ISSUE #2823: PDB paths getting generified via regexp
PDB paths sometimes contain directories that can be (and are) generified via the regexps;
for example: 
`c:\Users\USA\Documents\Visual Studio 2008\Projects\New folder (2)\kasper\Release\kasper.pdb` (from [here](https://www.threatconnect.com/blog/kasperagent-malware-campaign/) ) is generified to:
`%USERPROFILE%\Documents\Visual Studio 2008\Projects\New folder (2)\kasper\Release\kasper.pdb `
because of the regexps; while values to be useful for rules are to be exact and un-modified.

Food for thought:
- The easiest way would be to not apply the regexps to attributes of type PDB.
- If there are more attribute types where regexp should not be applied to, then it might be more future-proof to add a flag per attribute types to disable application of regexps to those.

elhoim:
**BLUF —** Default regexps rewrite `pdb` debug-symbol paths to `%USERPROFILE%`, destroying the original value.

- **Problem** — `RegexpBehavior::runRegexp()` applies every type-ALL regexp to incoming attribute values, and the default dataset shipped in `INSTALL/MYSQL.sql` rewrites Windows user paths to `%USERPROFILE%`. `pdb` values are debug-symbol paths lifted verbatim from a binary, so the build machine's user directory — often the most identifying part of the artefact — is silently stripped before storage, with the original never recorded, breaking correlations and rules that depend on the exact string.
- **Fix** — Adds `pdb` to the existing `EXCLUDED_TYPES` constant alongside `sigma`, `size-in-bytes`, `counter` and `float`.
- **Effect** — `pdb` values are stored exactly as submitted on every instance, including ones with their own path-normalising regexps.

## Root cause

`app/Model/Behavior/RegexpBehavior.php:12`

```php
const EXCLUDED_TYPES = ['sigma', 'size-in-bytes', 'counter', 'float'];
```

`runRegexp()` applies every regexp whose `type` is `ALL` (or matches the attribute type) to the incoming value. A fresh install ships default regexps of type `ALL` that rewrite Windows user paths — `INSTALL/MYSQL.sql:1845-1847` (and the same rows in `INSTALL/POSTGRESQL-data-initial.sql:291-293`):

```sql
(16, '/.:.Users\\\\(.*?)\\\\/i', '%USERPROFILE%\\\\', 'ALL'),
(17, '/.:.DOCUME~1.\\\\(.*?)\\\\/i', '%USERPROFILE%\\\\', 'ALL'),
(18, '/.:.Documents and Settings\\\\(.*?)\\\\/i', '%USERPROFILE%\\\\', 'ALL'),
```

`pdb` attributes are debug-symbol paths lifted verbatim out of a binary. They legitimately contain the build machine's user directory, and that directory name is often the most identifying part of the artefact. Under the default dataset:

```
c:\Users\USA\Documents\Visual Studio 2008\Projects\New folder (2)\kasper\Release\kasper.pdb
```

is silently stored as

```
%USERPROFILE%\Documents\Visual Studio 2008\Projects\New folder (2)\kasper\Release\kasper.pdb
```

destroying the forensic value and breaking any correlation or rule that depends on the exact string. This is #1420 and its duplicate #2823.

## The change

One line — add `pdb` to the existing exclusion list:

```php
const EXCLUDED_TYPES = ['sigma', 'size-in-bytes', 'counter', 'float', 'pdb'];
```

### Why a code-level exclusion rather than retuning the default regexp dataset

In #2823 @iglocska noted the default regexps are optional and the dataset could simply be revised. That is true but it does not solve the reported bug: the guarantee that a `pdb` value survives intact has to hold on every instance, including instances that have added their own path-normalising regexps of type `ALL` — which is exactly what an analyst deployment tends to do. Excluding the type is also the precedent already set in this file: `sigma`, `size-in-bytes`, `counter` and `float` are excluded because rewriting them corrupts a structurally exact value, and `pdb` is the same class of value. The shipped SQL datasets are left untouched, so path normalisation continues to work for every other type.

### Other types considered

`filename`, `filename|md5` and friends were considered and deliberately left alone — path generification is frequently *wanted* there, and there is no reported breakage. Hash and numeric types are already implicitly safe because the default regexps do not match them. Only `pdb` has a demonstrated, reported failure, so only `pdb` is changed.

### Known trade-off

`EXCLUDED_TYPES` short-circuits before the loop, so an excluded type also bypasses blocklist regexps (the `return false` branch, empty `replacement`). This is pre-existing behaviour shared by the four existing exclusions rather than something introduced here, and it is stated explicitly so it is not mistaken for an oversight. If blocklisting should apply to excluded types, that is a separate change to the behaviour's structure.

## Verification

Done:
- Confirmed the constant and the `runRegexp()` flow against `2.5` (`app/Model/Behavior/RegexpBehavior.php`, read in full).
- Confirmed the shipping default regexps of type `ALL` in both `INSTALL/MYSQL.sql` and `INSTALL/POSTGRESQL-data-initial.sql`.
- `php -l app/Model/Behavior/RegexpBehavior.php` — no syntax errors (PHP 8.5).

Not done (no environment available to me):
- No live MISP instance was used; the end-to-end path (attribute add → `runRegexp` → stored value) was not exercised at runtime.
- The test suite was not run.

Fixes #1420
Fixes #2823

### DIFF
--- a/app/Model/Behavior/RegexpBehavior.php
+++ b/app/Model/Behavior/RegexpBehavior.php
@@ -9,7 +9,7 @@ class RegexpBehavior extends ModelBehavior
 {
     private $__allRegexp = null;
 
-    const EXCLUDED_TYPES = ['sigma', 'size-in-bytes', 'counter', 'float'];
+    const EXCLUDED_TYPES = ['sigma', 'size-in-bytes', 'counter', 'float', 'pdb'];
 
     /**
      * replace the current value according to the regexp rules, or block blocklisted regular expressions


Wachizungu:
I can confirm the issue, and I do agree this isn't nice. My advice is to merge this.

Just wondering about the wider implications... As a similar issue arises with significant usernames / home folders (file attributes, text attribute for directory object path...)

elhoim:
Thanks — and that is the right question to ask, because I think the answer is that `pdb` is genuinely different from the other path-bearing types rather than the first of a long list.

The distinction is whether the username in the value is **incidental** or **load-bearing**.

For `filename`, `text`, or a directory path in an object, a local username is incidental. It is an artefact of where the sample happened to be executed, and stripping it before sharing is exactly what the Regexp replacement feature was built for. Those types should stay in scope.

`pdb` is the opposite case. Per `MispAttribute.php:4271` it is *"Microsoft Program database (PDB) path information"*, `to_ids => 0`, category *Artifacts dropped* — it is not an indicator to be matched, it is a fingerprint of the **build machine**. The developer's directory layout and account name are the identifying content: that is what makes `C:\Users\<handle>\Documents\Visual Studio\...\Release\foo.pdb` useful for clustering samples by actor. Redacting the username there does not sanitise an indicator, it destroys the evidence, and silently — the stored value no longer matches what the analyst submitted or what anyone else in the community holds.

So I would keep `EXCLUDED_TYPES` narrow and reserve it for types where a substitution makes the value *semantically wrong* rather than merely redacted. That is what the existing entries have in common: `sigma` (a rewritten rule is a broken rule), `size-in-bytes`, `counter`, `float` (a substitution yields a non-number). `pdb` fits that test; `filename` and `text` do not.

For the cases you raise, the mechanism already exists and is finer-grained than the exclusion list: `Regexp` rows carry a `type` column, and `RegexpBehavior::runRegexp()` only applies a rule when `$regexp['type'] === 'ALL' || $regexp['type'] === $type`. An operator who wants to scrub usernames from filenames but not from PDB paths can scope the rule to `filename` instead of `ALL`. The pain you are describing really comes from `ALL` rules being the path of least resistance.

If you would like, I can follow up separately with a docs note on that — scoping replacements by type rather than `ALL` — since that seems like the general answer, whereas this PR is just the one type where no operator choice can make a replacement correct.

Issue #1221, opened 2016-06-06
Duplicate attributes in same event


elhoim:
When bulk adding attributes to events from the ajax add attribute, i sometimes get duplicate attributes.
I am adding thousands of attributes to 3 or 4 different events, one browser tab per event and not adding attributes concurrently to the same event.

The duplicate attributes have both the to_ids flag set.

Not sure how to diagnose or reproduce this.

elhoim:
I have unique UUIDs and unique IDs for attributes, and they are not deleted but they have duplicate value, attribute type and attribute category.

iglocska:
Hmph. That definitely sounds like a bug. Got any idea how I can reproduce it?

elhoim:
My useage scenario was that I was adding domain attributes in 4 events from browser using the ajax "add attributes". Each event was in the thousands of domains.
I can send you files privately but artificial ones should work too.

Browser used was: Firefox 38.8 ESR on Windows 7

iglocska:
Hmm, so the domains had a few duplicates I guess. I'll check sounds weird -
maybe there is some weird concurrency issue that allows it to skip the
validation? No idea how this can happen. I'll look into it.

wilson-santos-pf:
Hi,

We are also facing the same concurrent problem.
Our data set sometimes has the same hostname, but with different ports. That data is exported as a "Network activity | hostname" attribute. 
We are making concurrent requests to MISP's API using PyMISP. This causes attributes to be duplicated. Even the tags on those attributes get duplicated. Then when publishing the event we get an exception like this _"Failed validating 'uniqueItems' in schema"_.
On sequential requests MISP rejects this, and simply says that a similar attribute already exists.
We are using MISP **2.4.80**.

iglocska:
This should now be fixed. Bulk imports now pre-validate and deduplicate before saving the data.

elhoim:
Revisiting this one: the closing rationale (bulk imports pre-validate and deduplicate) is the `value|type|category` hash dedup in `Event::_add()` (app/Model/Event.php:5410), which only covers duplicates *within a single submitted batch on event creation*. Two things are still open:

[...]

PR #10972, opened 2026-08-27
[high] fix: restore temporarily removed attribute validation rules (Refs #1221)

elhoim:
**BLUF —** A leaked validation-rule removal disables the duplicate-attribute check for the rest of a sync run.

- **Problem** — Three call sites unset the `value.uniqueValue` attribute validation rule and never put it back, and because CakePHP 2 models are request-wide singletons the removal leaks into every later attribute save in the same process. `Event::_add()` with `$fromXml=true` runs in a loop from `Server::__pullEvent()` and `Feed`, so after the first new event is captured the check is off for the rest of the pull.
- **Fix** — Adds `MispAttribute::removeValidationRules()` and `restoreValidationRules()`, and restores the rules in a `finally` block at all three sites.
- **Effect** — Sync and feed runs stop silently accepting duplicate type/category/value attributes.

## What

Attribute validation rules that are removed *temporarily* are never put back. Since CakePHP 2 models are registry singletons, `$this->validate` is shared for the whole request / worker run, so the removal silently leaks into every subsequent attribute save handled by the same process.

The rule that leaks is `value.uniqueValue` (`valueIsUnique()`), i.e. the only thing that stops the same `type` / `category` / `value` triplet being stored twice in one event. That is the check behind #1221.

Three sites, all `unset()` without restore:

* `Event::_add()` (`app/Model/Event.php:5318-5319`, `$fromXml` branch)
* `MispAttribute::captureAttribute()` (`app/Model/MispAttribute.php:3161`, `breakOnDuplicate=false`)
* `MispAttribute::validateAttribute()` (`app/Model/MispAttribute.php:2831-2833`, `$context=false`)

The `Event::_add()` one is the one that bites in practice. `_add(..., $fromXml=true)` is called in a loop from `Server::__pullEvent()` (`app/Model/Server.php:489`/`491`) and `Feed` (`app/Model/Feed.php:1024`/`1194`). As soon as one *new* event is captured during a pull, `value.uniqueValue` is gone for the rest of that job — so every `_edit()` that follows in the same run saves attributes with the uniqueness rule disabled. `Event::_edit()` has no intra-batch dedup of its own (unlike `_add()`, which dedups by `value|type|category` hash at `Event.php:5410`), and `editAttributeBulk()` culls bad records purely on the validation pass (`MispAttribute.php:3327-3345`) — with the rule gone, nothing culls duplicates.

## Change

Adds `MispAttribute::removeValidationRules()` / `restoreValidationRules()` and uses them at the three sites, restoring in a `finally` so early returns and exceptions cannot leak the state. Behaviour inside each scope is unchanged; in `Event::_add()` the removal is also moved down next to the capture loop it exists for, so the two early returns above it (`$fromPull` / existing-uuid) no longer leak either.

## Behavioural note (please read)

For a pull run, this restores `value.uniqueValue` for the `_edit()` calls that follow an `_add()`. That is the behaviour a pull already has today when the run happens to start with an edit, so this makes the two orderings consistent rather than introducing a new rule — but it does mean incoming duplicates that previously slipped through a mixed add+edit pull will now be dropped and logged as validation errors, as they already are elsewhere. Flagging it explicitly in case you want it scoped differently.

## Not addressed (deliberately)

* **No DB-level constraint.** `attributes` has unique keys on `id` and `uuid` only (`INSTALL/MYSQL.sql:116-128`, `db_schema.json`). `valueIsUnique()` is a check-then-act `find()`-then-`save()` with no transaction, row lock or upsert, so two concurrent API calls adding the same attribute can still both pass the check and both insert — exactly what @wilson-santos-pf described in #1221. Only a unique index on `(event_id, type, category, value1, value2, deleted, object_id)` (or an equivalent hash column) really closes that, and that is a schema migration + a data-cleanup decision for maintainers, so it is not in this PR.
* **`fast_update`.** `valueIsUnique()` returns `true` unconditionally when `fast_update` is set (`MispAttribute.php:977`); the comment there says this is a deliberate tradeoff, so it is left alone.
* `Event::_edit()` (`Event.php:5628-5629`) also unsets `Attribute.validate['event_id']` without restoring, and the line next to it unsets `validate['value']['unique']` — a rule name that does not exist (the rule is `uniqueValue`), so it is a no-op. Left as-is to keep this PR to one concern; happy to follow up.

## Testing

`php -l` clean on both files. I have no instance to run the suite against, and the concurrency analysis above is reasoning from the code, not an observed race.

Refs #1221

### DIFF
--- a/app/Model/Event.php
+++ b/app/Model/Event.php
@@ -5313,10 +5313,6 @@ class Event extends AppModel
         if ($fromXml) {
             // Workaround for different structure in XML/array than what CakePHP expects
             $data = $this->cleanupEventArrayFromXML($data);
-            // the event_id field is not set (normal) so make sure no validation errors are thrown
-            // LATER do this with   $this->validator()->remove('event_id');
-            unset($this->Attribute->validate['event_id']); // otherwise gives bugs because event_id is not set
-            unset($this->Attribute->validate['value']['uniqueValue']); // unset this - we are saving a new event, there are no values to compare against and event_id is not set in the attributes
         }
         unset($data['Event']['id']);
         if (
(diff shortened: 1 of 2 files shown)


elhoim:
One clarification on my own wording above, to avoid overclaiming: restoring `value.uniqueValue` culls duplicates *of attributes that already exist in the DB*. It does not cull duplicates that are siblings inside the same incoming batch — `editAttributeBulk()` validates the whole batch with `'validate' => 'only'` and then saves with `'validate' => false`, so two new same-value attributes in one edit push both pass validation (neither is in the DB yet) and both insert. `Event::_add()` has the `value|type|category` hash dedup for exactly this; `Event::_edit()` has no equivalent. Mirroring that dedup on the edit path would close it, but it changes sync semantics, so I would rather leave that call to you than fold it into this PR.

PR #11004, opened 2026-08-27
ci: code-coverage job measuring the unit and live suites

elhoim:
**Branch:** `elhoim/MISP:ci/coverage-harness` → `MISP/MISP:2.5`
**10 files, +7153** — of which **6327 lines are a generated lock file**. The reviewable surface is 826 lines across 9 files.

This is the PR in the series most likely to be refused, and it should stand on its own merits. It adds a CI job, and CI jobs cost minutes forever.

## What the job does

`.github/workflows/coverage.yml`, on push and PR to `2.5`:

1. Installs PHP 8.3 with **pcov** and the pinned dependency set.
2. Installs and configures a MISP instance (`build/coverage/ci_setup_misp.sh`), mirroring what `main.yml` does inline.
3. Runs `phpunit -c phpunit.xml --coverage-clover`.
4. Runs PyMISP's live suite and the `tests/testlive_*` suites under instrumentation (`build/coverage/ci_run_live_suite.sh`).
5. Merges, reports, and **fails if the union of the two suites covers fewer than `MIN_UNION_LINES` distinct statements**.
6. Prints the report on the run summary page and uploads the raw data.

### The part that does not exist elsewhere

MISP's Python suites drive a running instance over HTTP. PHPUnit is not in that process and cannot attribute a single line of what they exercise — which is most of `Controller/` and `Model/`. Any coverage number counting only `app/Test/` therefore understates MISP badly and points refactoring effort at the wrong places.

`build/coverage/covcollect.php` is hooked in as `auto_prepend_file` for **both** the Apache and the CLI SAPI. It starts pcov on every request and console invocation and writes one capture per process on shutdown. `merge_coverage.py` merges the captures; `report.py` intersects them with PHPUnit's clover report — which supplies the denominator, every executable statement in the tree — and prints unit, live, union, overlap and a per-subsystem breakdown.

Collection begins only once `$MISP_COV_DIR/ENABLED` exists, which is how instance setup (`runUpdates`, `updateJSON`, the schema import) is kept out of the numbers: it is not test coverage.

## What it costs

One additional job per push and per PR to `2.5`. It performs the same install `main.yml`'s `misp` job performs and runs a superset of that job's Python suites, so expect it in the same range as that job, plus a couple of minutes for pcov and the merge. **It has not been timed on MISP/MISP's runners** — the workflow only triggers on PRs whose base is `2.5`, so no run of it exists outside a fork.

If that cost is unacceptable on every PR, changing `on:` to `push` only, or to a weekly `schedule`, keeps the number and drops most of the minutes. The ratchet is less useful that way — it stops being a gate and becomes a report — but it is a one-line change and a reasonable place to compromise.

## Why the gate is on lines, not on a percentage

A percentage floor was tried first and failed on its own terms: a merge that added roughly 2300 statements of new, untested code took union coverage from **20.18% to 19.82%** — while the number of covered lines went **up by 35**. Nothing had stopped working. The denominator is the size of the tree, which nobody controls per-PR, so a gate hung off it eventually demands that somebody lower the gate — and a ratchet that gets lowered is decoration.

Covered lines fall only when the tests stop executing code, which is the thing worth gating on. The percentage is still reported; it just does not gate.

## The floor is deliberately slack, and asks to be raised

`MIN_UNION_LINES` ships at **15000**, which is well below anything measured. That is intentional and the workflow says so in a comment.

Everything measured so far was measured with a much larger test suite than this PR proposes — the fork's development branch, whose other tranches are the other PRs in this series. Those numbers do not transfer:

```
23791  unit + live, at the point coverage was first measured (20.18% of the
       tree at the time)
23826  after merging upstream 2.5 (19.82% of a bigger tree)
24152  with a PHP integration suite added and counted   (+326 lines)
30081  with PyMISP's live suite un-broken               (25.02%)
```

The right value for *this* configuration is unknown until it runs once on MISP/MISP. **Raise `MIN_UNION_LINES` to the `union` line of the first green run, less about 1.5% of headroom**, rather than shipping a red gate on day one. The headroom matters because some live tests fail rather than error and which ones can vary between runs.

## The known gap: PyMISP's live suite does not fully complete

CI measures less coverage than a developer machine does — about 25.0% versus about 24.8% locally in the best measured configuration, and considerably worse before this was addressed. The cause is ordering: PyMISP's `setUpClass` and `tests/testlive_comprehensive_local.py`'s both create an organisation literally named `Test Org`. PyMISP's tears its own down; the local one does not. Run the local one first and `add_organisation` returns `{'errors': ...}`, `cls.test_org` becomes a dict, and every one of PyMISP's tests errors in `setUpClass` in about seven seconds — contributing nothing at all, silently.

`ci_run_live_suite.sh` therefore runs PyMISP's suite **first**, exactly as `main.yml` orders it, and says why at the point where it matters. Some PyMISP tests still fail rather than error on a clean runner; those cost a little coverage and are why the floor wants headroom. **What it means for the number: treat it as a lower bound on what MISP's tests reach, not as a measurement of MISP.**

## Two settings that differ from a plain mirror of `main.yml`

Both are in `ci_setup_misp.sh` and both are load-bearing:

- `MISP.background_jobs 0` — this job runs no workers, so with it on `Event::publishRouter` enqueues to a queue nothing drains and events never reach `published=1`. `main.yml` gets away with it because it starts supervisor. With it off, `publishRouter` falls through to `Event::publish()` and sets `published=1` inline.
- The six `Plugin.ZeroMQ_*` settings `main.yml` sets — without them PyMISP's `test_zmq` fails here and passes there.

`tests/curl_tests_GH.sh` is a separate step placed before the Python suites, for the same `AUTO_INCREMENT`/event-id-1 reason documented in the `integration-harness` PR of this series.

## Why the lock file

`app/composer.lock` is gitignored, so `composer install` in CI is really `composer update`: it re-resolves from the network every run, and the same commit can go from green to red because a dependency published. A coverage *ratchet* measured against a moving dependency set is measuring the weather.

It lives at `build/composer-ci.lock`, not `app/composer.lock`, on purpose: it was resolved under PHP 8.3, and `symfony/console`, `symfony/filesystem`, `symfony/string` and `nette/utils` in it require PHP >= 8.2 — dropping it into `app/` would make the `build (ubuntu-22.04, 8.1)` job in `main.yml` unresolvable. Pinning is scoped to the one job with a gate to protect.

It is pinned against **#10964's exact `app/composer.json`**. If #10964 is amended, the lock goes stale and `composer validate` fails loudly (exit 2) rather than silently drifting; regenerating it is two commands, given in the workflow comment.

If the lock is judged not worth 6327 lines in the tree, deleting it and the two lines that copy it is a self-contained change — the job then works, just noisily.

## Depends on / merge order

Requires **#10964**: `coverage.yml` runs `phpunit -c phpunit.xml`, and the lock is resolved against that PR's `app/composer.json`. Nothing else in this series is required. It touches no file any other PR in the series touches, so it can merge at any point after #10964.

## Running it locally

Full instructions in `docs/coverage.md`. In short, against an instance you already have:

```bash
cd app && ./Vendor/bin/phpunit -c phpunit.xml --coverage-clover ../clover.xml

# php.ini, for BOTH the web SAPI and CLI:
#   pcov.enabled=1
#   pcov.directory=/var/www/MISP/app
#   auto_prepend_file=/var/www/MISP/build/coverage/covcollect.php
mkdir -p /tmp/cov && chmod 777 /tmp/cov
cat > build/coverage/covconfig.php <<'EOF'
<?php
return ['cov_dir' => '/tmp/cov', 'app_root' => '/var/www/MISP/app/'];
EOF
touch /tmp/cov/ENABLED          # nothing is recorded before this exists

# ...run the live suites, then:
python3 build/coverage/merge_coverage.py /tmp/cov merged.json /var/www/MISP/app/
python3 build/coverage/report.py clover.xml merged.json --min-union-lines 15000
```

`build/coverage/test_report.py` covers the reporter itself — in particular that the union counts a line once however many suites covered it. Run it with `python3 -m pytest build/coverage/test_report.py`. It is deliberately not wired into the workflow, to avoid installing pytest for one file.

## What this does not do

- **Does not add or change a single test.** It measures what MISP already has.
- Does not run the PHP integration suite — that suite is a separate PR in this series and is not required here. Once it lands, adding it to this job is four lines and was worth roughly **+326 covered lines** when measured.
- Does not change `.github/workflows/main.yml`, or how, when or in what order the existing suites run there.
- Does not gate on a coverage percentage, and does not fail a PR for adding untested code — only for making the existing tests execute less.
- Does not publish to Codecov or any third-party service; the report is a build artifact and a line on the run summary.
- Does not touch application code.

## What was not verified

- **The job has never run on MISP/MISP.** The workflow triggers only on PRs based on `2.5`, which no fork branch is.
- **shellcheck was not available** on the machine this was prepared on, nor in the container available to it. The two shell scripts were checked with `bash -n` only.
- The `MIN_UNION_LINES` value shipped here is a deliberately slack placeholder, not a measurement of this configuration. See above.

---

### Where this sits in the series

This is one of seven small PRs that together add a test and coverage capability to MISP. They were deliberately split so each can be reviewed, accepted or rejected on its own.

| PR | What it adds | Depends on |
|---|---|---|
| #10964 | PHPUnit 9.6 + `app/phpunit.xml` so coverage can run on PHP 8 | — |
| #10999 | Unit conformance suites (widgets, workflow modules, exports, tools) | #10964 |
| #11000 | Integration harness — model tests against a real database | #10964, #10999 |
| #11001 | Characterization of `Event`'s read and write paths | #11000 |
| #11002 | Live behavioural suites (dashboards, workflows, exports, console) | — |
| #11003 | Golden-snapshot contract for the restSearch API | #11002 |
| #11004 | Coverage job measuring the unit and live suites | #10964 |
| #11005 | Documentation of the three test layers | merge last |

**Suggested merge order:** #10964 → #10999 → #11000 → #11001, and independently #11002 → #11003. #11004 can land any time after #10964. #11005 last, since it documents whatever ends up merged.

#11002 and #11003 touch no PHP at all and are independent of the rest.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,211 @@
+# Measures how much of MISP the existing test suites actually execute.
+#
+# The PHPUnit suite is measured directly with pcov. The Python live suite
+# drives MISP over HTTP, so PHPUnit cannot attribute its coverage: pcov is
+# hooked into both mod_php and the CLI through auto_prepend_file
+# (build/coverage/covcollect.php), every request writes a capture, and the
+# captures are merged with the PHPUnit clover report into one union number.
+#
+# The ratchet exists because coverage that is not enforced silently decays.
+#
+# Requires app/phpunit.xml and the PHPUnit 9.6 bump from MISP/MISP#10964.
+name: coverage
+
+on:
+  push:
+    branches: [ '2.5' ]
+  pull_request:
+    branches: [ '2.5' ]
+
+env:
+  PHP_VERSION: '8.3'
+
+  # The ratchet is on covered LINES, not on a coverage percentage.
+  #
+  # A percentage floor was tried first and failed on its own terms: a merge
+  # that added ~2300 statements of new, untested code took the percentage
+  # DOWN while the number of covered lines went UP. The denominator is the
+  # size of the tree, which nobody controls per-PR, so a gate hung off it
+  # eventually demands that someone lower the gate - and a ratchet that gets
+  # lowered is decoration. Covered lines fall only when the tests stop
+  # executing code, which is the thing worth gating on. The percentage is
+  # still reported, it just does not gate.
+  #
+  # This starting value is deliberately slack. The number this job will
+  # report has never been measured on MISP/MISP's own runners, and it is
+  # sensitive to how much of the live suite completes. Rather than guess a
+  # tight floor and hand the project a red gate on day one, the floor starts
+  # below anything plausible and asks to be raised:
+  #
+  #   RAISE THIS to the `union` line of the first green run on 2.5, minus a
+  #   little headroom (~1.5% is enough; some live tests fail rather than
+  #   error, and which ones can vary between runs), and raise it again as
+  #   each tranche of tests lands.
+  MIN_UNION_LINES: '15000'
+
+jobs:
+  coverage:
+    runs-on: ubuntu-24.04
+
+    services:
+      mariadb:
+        image: mariadb:10.11
+        env:
+          MARIADB_ROOT_PASSWORD: bar
+          MARIADB_DATABASE: misp
+          MARIADB_USER: misp
+          MARIADB_PASSWORD: blah
+        ports: [ '3306:3306' ]
+        options: >-
+          --health-cmd="mariadb-admin ping -h 127.0.0.1 -uroot -pbar"
+          --health-interval=5s --health-timeout=3s --health-retries=30
+      redis:
+        image: redis:7
+        ports: [ '6379:6379' ]
+        options: >-
+          --health-cmd="redis-cli ping"
+          --health-interval=5s --health-timeout=3s --health-retries=30
+
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          submodules: 'recursive'
+
+      - name: Setup PHP with pcov
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ env.PHP_VERSION }}
+          extensions: mysql, mbstring, xml, opcache, readline, redis, gd, apcu, intl, zip, pcntl, pcov
+          ini-values: memory_limit=2048M, pcov.enabled=1
+
+      # The `zip` BINARY is required, not just ext-zip: MISP shells out to it
+      # when building attachment archives, and its absence surfaces as
+      # `proc_open(): posix_spawn() failed: No such file or directory`.
+      - name: Install system deps
+        run: |
+          sudo apt-get -y update --allow-releaseinfo-change
+          sudo apt-get -y install curl zip unzip gnupg python3 python3-pip \
+                                  python3-virtualenv apache2 libapache2-mod-php${{ env.PHP_VERSION }}
+
+      # app/composer.lock is gitignored, so `composer install` here is really
+      # `composer update`: it re-resolves from the network every run, and the
+      # same commit can go from green to red because a dependency published.
+      # A coverage RATCHET measured against a moving dependency set is
+      # measuring the weather. build/composer-ci.lock pins it.
+      #
+      # The lock lives here rather than at app/composer.lock on purpose: it
+      # was resolved under PHP 8.3, and symfony/console, symfony/filesystem,
+      # symfony/string and nette/utils in it require PHP >= 8.2 - dropping it
+      # into app/ would make the `build (ubuntu-22.04, 8.1)` job in main.yml
+      # unresolvable. Pinning is therefore scoped to this job, which is the
+      # one with a gate to protect. `composer validate` fails loudly (exit 2)
+      # if a change to app/composer.json leaves the lock stale, at which point
+      # regenerate it with:
+      #   cd app && composer update --ignore-platform-reqs \
+      #     && mv composer.lock ../build/composer-ci.lock
+      - name: Install PHP dependencies
+        run: |
+          cp build/composer-ci.lock app/composer.lock
+          cd app
+          composer validate --no-check-all --no-check-publish
+          composer install --no-progress --no-interaction
+
+      # -- instance --------------------------------------------------------
+      # Mirrors the install main.yml performs inline, kept in a script so the
+      # two cannot drift.
+      - name: Install and configure MISP
+        run: bash build/coverage/ci_setup_misp.sh
+
+      # -- PHPUnit suite ---------------------------------------------------
+      # After the install, as in main.yml, and deliberately BEFORE the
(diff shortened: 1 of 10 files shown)


elhoim:
Pushed a follow-up commit correcting a defect I found while running the job against a live instance.

`cake Admin setSetting MISP.background_jobs 0` is refused without `--force` — *"Setting change rejected ... It is highly recommended that this setting is enabled"* — and **the refusal does not exit non-zero**, so passing it through the settings loop left it at its default of `true` with no visible failure.

The consequence was that the inline-publishing setup did nothing: with background jobs still on, `Event::publishRouter` enqueues to a queue this job starts no worker to drain, so events never reach `published=1` and the export / publish-timestamp searches fail — the exact symptom the setting was there to prevent.

For what it is worth, the run that surfaced this also confirmed the rest of the setup works: `PyMISP/tests/testlive_comprehensive.py` executed all 85 tests (69 passed, 14 failed, 2 skipped) with no `setUpClass` error, where it had previously errored all 85 in about seven seconds.

iglocska:
Closing this for now, we run a fair bit of the tests locally but wouldn't want to have them for every commit, we're already struggling with the CI length. I have plans to change how we run the CI and then we can-re add some of the additional tests.

The PHP tests are mostly stale or auto generated and of low value currently, so leaving them out of the CI for now is on purpose. The first step will be to come up with a sane set of mock data for the JSON ingested structures (galaxies, objects, etc) and avoid actually ingesting the sub modules, as that takes up 40%+ of the CI run. Once we fix that we can look into hooking up the additional .py tests (from pymisp + MISP/tests/ at least).

elhoim:
A correction to my comment above, for the record, since this PR is closed but the claim is wrong and searchable.

I wrote that `cake Admin setSetting MISP.background_jobs 0` is refused without `--force` and that **"the refusal does not exit non-zero"**. The second half is false. `AdminShell::setSetting()` calls `Shell::error()`, which ends in `_stop(self::CODE_ERROR)` with `CODE_ERROR = 1`, and `CakeObject::_stop()` is `exit($status)`. It exits 1.

The behaviour I actually observed was real — the setting stayed at its default and nothing appeared to fail — but the cause was the exit status being swallowed by the surrounding settings loop, not the command returning 0. `--force` was still the right fix; my explanation of why it was needed was wrong.

Also, for whatever it is worth, the figure I quoted from you here is on the conservative side of what I measured: on run 32854863121 the `Update JSON` step is 225s of an 865s job (26.0%) on 8.1 and 216s of 798s (27.1%) on 8.3, with the recursive submodule checkout another ~9%. And within that ingestion, galaxies alone are 80.4% of the cost.

PR #11006, opened 2026-08-27
fix: [Tags:index] Fixed the attachment of a taxonomy to a tag

ThomasLcr:
Before the fix, all tags prefixed with the name of a taxonomy were automatically associated with that taxonomy (see images). 
Screenshot from 2026-08-27 17-20-11

Now, if the tag is not part of the taxonomy, then it is not associated with it.
Screenshot from 2026-08-27 17-20-04


iglocska:
Thanks a lot.

PR #11009, opened 2026-08-28
[cosmetic] fix: [navbar] gate Sharing Group Blueprints listing entry on index, not add

elhoim:
**BLUF —** The Sharing Group Blueprints listing menu entry is gated on `add` rather than `index`.

- **Problem** — In `NavbarHelper::buildSyncMenu()`, the List Sharing Group Blueprints entry links to `SharingGroupBlueprints/index` but gates its visibility on `canAccess('SharingGroupBlueprints', 'add')`.
- **Fix** — Changes the requirement check to `'index'` so the entry's visibility matches the action it links to.
- **Effect** — Users authorised to view the listing but not to create blueprints see the menu entry instead of having to type the URL by hand.

## Summary

The "List Sharing Group Blueprints" entry in the Sync navbar menu links to `SharingGroupBlueprints/index`, but its visibility was gated on `canAccess('SharingGroupBlueprints', 'add')` instead of `'index'`. This fixes the mismatch so the entry's visibility tracks the action it actually links to.

## Current code (before fix)

`app/View/Helper/NavbarHelper.php`, in `buildSyncMenu()`, around line 513-520:

```php
[
    'label' => __('List Sharing Group Blueprints'),
    'url' => $baseurl . '/SharingGroupBlueprints/index',
    'controller' => 'SharingGroupBlueprints',
    'action' => 'index',
    'requirement' => $this->Acl->canAccess('SharingGroupBlueprints', 'add'),
    'icon' => 'fas fa-drafting-compass'
]
```

The `url` and `action` fields both correctly target `SharingGroupBlueprints/index` (a real, existing controller action), but `requirement` checks the `add` action instead.

## Mechanism

`requirement` is the gate `NavbarHelper::build()` uses to decide whether to render this menu entry at all. Because it checks a different action (`add`) than the one the entry links to (`index`), the entry's visibility is decoupled from whether the viewing user can actually reach the page:

- A user with `index` rights but not `add` rights on `SharingGroupBlueprints` is fully authorized to view the listing page, but the menu entry never appears — they'd have to know the URL and type it manually.
- A user with `add` rights but not `index` rights sees a working-looking link whose target action their role isn't actually checked against.

## Reproduction performed by the validating agent

> 1) Read `app/View/Helper/NavbarHelper.php:513-519` — the 'List Sharing Group Blueprints' entry has `url => $baseurl.'/SharingGroupBlueprints/index'`, `action => 'index'`, but `requirement => $this->Acl->canAccess('SharingGroupBlueprints', 'add')` (mismatched action string). 2) Ran the pinning test directly against the running mispapp container: `podman exec mispapp bash -c "cd /var/www/MISP/app && Vendor/bin/phpunit --filter testSharingGroupBlueprintsEntryIsGatedOnTheAddAction Test/Unit/View"` -> `OK (1 test, 1 assertion)`. That test grants only `['SharingGroupBlueprints','add']` permission and asserts the `/SharingGroupBlueprints/index`-linking entry is absent from the rendered navbar's 'sync' menu — confirming a user who can reach the index page (has 'index' rights) but lacks 'add' rights never sees the menu entry. 3) Confirmed both target actions are real, valid controller endpoints: `grep -n 'public function index\|public function add' app/Controller/SharingGroupBlueprintsController.php` shows both `index()` (line 18) and `add()` (line 46) exist. 4) Sanity-checked the running instance: `curl http://127.0.0.1/SharingGroupBlueprints/index` and `.../add` both return 302 to `/users/login` (auth required, route resolves, no routing error).

This PR's author additionally re-verified the fix's effect directly, outside the pinning test (which only exercises the pre-fix scenario and is left unmodified — see below): a standalone harness loaded the patched `NavbarHelper` and drove `build()` with two ACL doubles —

- **index granted, add denied**: entry is now **present** (was absent before the fix — this is the case the bug hid).
- **index denied, add granted**: entry is now **absent** (correctly hidden, since the user can't reach `index`).

## User-visible consequence

A user whose role has `index` rights on `SharingGroupBlueprints` (the intended permission for listing them) but not `add` rights never sees the "List Sharing Group Blueprints" menu entry, even though they are fully authorized to view that page and must navigate to it by typing the URL directly.

## Fix

Change the `requirement` on that entry to `$this->Acl->canAccess('SharingGroupBlueprints', 'index')`, matching the entry's own `url`/`action` fields (the same pattern already used correctly by every other entry in this menu, e.g. the "List Sharing Groups" entry directly above it).

### Alternatives considered and rejected

- **Change `action` to `'add'` instead of fixing `requirement`**: rejected — the entry's `url` unambiguously points at `/index`, and `add` is a different, unrelated page (a create form). Changing `action` to `add` would make the entry claim to link to a create action while actually linking to the listing, which is worse, not better.
- **Leave it alone since a workaround (typing the URL) exists**: rejected — the whole point of the navbar is discoverability; a correctly-permissioned user shouldn't need to already know the URL.

## Scope note

Out of scope: the commented-out "Add Sharing Group Blueprint" entry a few lines above (lines 505-512) has the *same* mismatch in the opposite direction — `action => 'add'` but `requirement => canAccess(..., 'index')`. It is dead code (fully commented out) and was not part of the validated defect group for this PR, so it was left untouched. Flagging it here for visibility only.

## Behaviour change flag

**BEHAVIOUR CHANGE**: This changes which users see the "List Sharing Group Blueprints" navbar entry. Specifically, a role granted `index` on `SharingGroupBlueprints` but not `add` will now see the entry (previously hidden); a role granted `add` but not `index` will now have the entry hidden (previously shown). This is a visibility-only change — no routing, permission enforcement, or controller behaviour is altered; the underlying `SharingGroupBlueprintsController::index()` action's own access control (outside this helper) is unchanged.

## Pinning test

This is pinned by `app/Test/Unit/View/NavbarHelperTest.php::testSharingGroupBlueprintsEntryIsGatedOnTheAddAction`, which currently documents this as a `KNOWN-DEFECT` and asserts today's (buggy) behaviour. That test file was **not** modified by this PR per the reviewing session's instructions — it lives on another branch, and the maintaining session will update its annotation once this fix merges. Its assertion (only-`add`-granted → entry absent) still holds after this fix, since only-`add`-granted still lacks `index`; the case that actually flips (`index`-granted, `add`-denied → entry now present) is not covered by that specific test and was verified independently as described above.

---

---

*Priority revised medium -> cosmetic: my own correction, not a maintainer request. As I noted on the PR, the original body overstated the effect: the menu entry is not hidden from an authorised user, so the user-visible impact is smaller than 'medium' implies.*

### DIFF
--- a/app/View/Helper/NavbarHelper.php
+++ b/app/View/Helper/NavbarHelper.php
@@ -515,7 +515,7 @@ class NavbarHelper extends AppHelper {
                         'url' => $baseurl . '/SharingGroupBlueprints/index',
                         'controller' => 'SharingGroupBlueprints',
                         'action' => 'index',
-                        'requirement' => $this->Acl->canAccess('SharingGroupBlueprints', 'add'),
+                        'requirement' => $this->Acl->canAccess('SharingGroupBlueprints', 'index'),
                         'icon' => 'fas fa-drafting-compass'
                     ]
                 ]


elhoim:
Correction to this PR: the code change is right, but the body overstates its effect and should be read with this in mind.

The body claims a user "never sees the menu entry, even though they are fully authorized". **No assignable role can reach that state.** `ACLComponent::ACL_LIST[sharingGroupBlueprints]` gives:

```
add   => [perm_sharing_group]   (ACLComponent.php:798)
index => [perm_sharing_group]   (ACLComponent.php:806)
```

Identical requirements, and `AclHelper::canAccess()` consults exactly that table, so a user holding one necessarily holds the other. The verification in the original body used hand-built ACL doubles (index granted, add denied) that correspond to no real permission set.

What this change actually is: the menu entry links to `index` and declares `action => index`, but gated its visibility on `add`. Aligning the gate with the link is correct on its own terms and makes the entry robust if those two ACL entries ever diverge — but it is **not** a live user-visible fix today, and there is no behaviour change under stock MISP.

Apologies for the overstatement; the code stands, the claim did not.

PR #11013, opened 2026-08-28
[cosmetic] fix: [galaxy] AclHelper::canPublishGalaxyCluster() ignores perm_publish

elhoim:
**BLUF — priority: cosmetic** (revised down from *high* after maintainer review: the impact is a button that is
shown but does not work, not an actual privilege escalation — the server-side ACL still refuses the action).
The Publish Cluster control is shown to users lacking `perm_publish`; this PR fixes the check.

- **Problem** — `AclHelper::canPublishGalaxyCluster()` delegates to `ACL->canModifyGalaxyCluster()` instead of the existing `ACL->canPublishGalaxyCluster()`, so it ignores `perm_publish` and returns true for any `perm_galaxy_editor`.
- **Fix** — Points the helper at the correct ACL method and updates both `side_menu.ctp` templates accordingly.
- **Effect** — The Publish/Unpublish cluster control only appears for users who actually hold `perm_publish`.

## Summary

`AclHelper::canPublishGalaxyCluster()` (the helper the side menu uses to decide whether to render the Publish/Unpublish control) checked the wrong ACL primitive, so it ignored `perm_publish` entirely.

### Current code (before this PR)

```php
// app/View/Helper/AclHelper.php
public function canPublishGalaxyCluster(array $cluster)
{
    return $this->ACL->canModifyGalaxyCluster($this->me, $cluster);
}
```

```php
// app/View/Elements/genericElements/SideMenu/side_menu.ctp (both the toggle
// at ~line 1636 and the publish-only link at ~line 1717 call this helper)
if ($this->Acl->canPublishGalaxyCluster($cluster)) {
    ...
    'text' => $cluster['GalaxyCluster']['published'] ? __('Unpublish Cluster') : __('Publish Cluster')
    ...
}
```

Meanwhile `ACLComponent` already has a correct, dedicated method that nobody was calling:

```php
// app/Controller/Component/ACLComponent.php
public function canPublishGalaxyCluster(array $user, array $cluster)
{
    if (!$this->canModifyGalaxyCluster($user, $cluster)) {
        return false;
    }
    return (bool)$user['Role']['perm_publish'];
}
```

## Mechanism

`AclHelper::canPublishGalaxyCluster()` delegated to `ACL->canModifyGalaxyCluster()` instead of `ACL->canPublishGalaxyCluster()`, so the helper returned `true` for any `perm_galaxy_editor` regardless of `perm_publish`.

## Reproduction (quoted from the validating agent, and re-verified here)

> Ran the pinning test inside the mispapp container against the untouched main clone (branch `test-coverage-foundation`, unmodified):
> `podman exec mispapp bash -c "cd /var/www/MISP/app && ./Vendor/phpunit/phpunit/phpunit --configuration phpunit.xml --filter testCanPublishGalaxyClusterIgnoresThePublishPermissionOnTheHelper Test/Unit/View/SmallHelpersTest.php"` -> `OK (1 test, 2 assertions)`. The test asserts `ACLComponent::canPublishGalaxyCluster($me,$cluster)` is `false` for a `perm_galaxy_editor=true`/`perm_publish=false` user, while `AclHelper::canPublishGalaxyCluster($cluster)` is `true` for the same user/cluster — both assertions pass, confirming the divergence.
>
> To determine whether this is UI-only or a real bypass, traced the actual server-side gate: `AppController::beforeFilter` (line 436) calls `$this->ACL->checkAccess($user,$controller,$action)` on every request BEFORE any controller code runs; `ACLComponent::ACL_LIST['galaxyClusters']['publish'] = ['AND' => ['perm_galaxy_editor','perm_publish']]`. Verified this route-level gate empirically with a real `ACLComponent::checkAccess()` call run in a disposable, untracked `app/tmp/` scratch test inside the container (deleted immediately after) -> `OK (1 test, 3 assertions)`, confirming `canUserAccess(user_with_editor_no_publish, 'galaxyClusters','publish') === false`, `canUserAccess(user_with_both) === true`, `canUserAccess(user_with_publish_only_no_editor) === false`.

For this PR itself (not the pinning test, which lives on another branch and is out of scope to edit — see below), I re-verified the fix in isolation: I extracted the two changed `AclHelper` methods via a disposable scratch harness (`app/tmp/`, deleted after) with a fake `$this->ACL` that records which method was invoked, and confirmed `canPublishGalaxyCluster()` now calls `ACL->canPublishGalaxyCluster()` (not `canModifyGalaxyCluster()`), and the new `canUnpublishGalaxyCluster()` calls `ACL->canModifyGalaxyCluster()`. I also lint-checked both changed files with `podman exec mispapp php -l` (copied into the container's `app/tmp/`, not committed).

## User-visible impact (before this fix)

A galaxy-editor without `perm_publish` saw the "Publish Cluster" (and, via the same shared conditional, "Unpublish Cluster") control rendered in the side menu, but clicking it produced a 403 `ForbiddenException` — the real security boundary (`AppController::beforeFilter` → `ACLComponent::checkAccess` with `AND(perm_galaxy_editor, perm_publish)` on the `galaxyClusters/publish` route) was intact and enforced correctly, independent of this helper. **This was a misleading-affordance / UX bug, not an authorization bypass.**

## Fix

- `AclHelper::canPublishGalaxyCluster()` now delegates to the existing, already-correct `ACL->canPublishGalaxyCluster()`.
- The naive one-line swap was rejected: `side_menu.ctp` reuses the *same* helper call to gate a single toggle whose text/action flips between "Publish Cluster" and "Unpublish Cluster" depending on `cluster.published`. But `ACL_LIST['galaxyClusters']['unpublish']` only requires `perm_galaxy_editor` (no `perm_publish`) — a plain swap would have hidden the legitimate "Unpublish" control from editors who are actually allowed to unpublish.
- Added a sibling `AclHelper::canUnpublishGalaxyCluster()` (delegating to `ACL->canModifyGalaxyCluster()`, matching the ACL_LIST asymmetry) and changed the toggle in `side_menu.ctp` to branch on `cluster.published`: use `canUnpublishGalaxyCluster()` when the cluster is published (offering "Unpublish"), `canPublishGalaxyCluster()` otherwise (offering "Publish").
- The publish-only link at `side_menu.ctp` ~line 1717 already called `canPublishGalaxyCluster()` and needed no change — it becomes correct automatically once the helper is fixed.

### Pinning test

`app/Test/Unit/View/SmallHelpersTest.php::testCanPublishGalaxyClusterIgnoresThePublishPermissionOnTheHelper` pins the buggy behaviour this PR fixes. That file lives on another (internal, not-yet-merged) branch and is **not** modified by this PR — the test's annotation will be updated separately once this fix merges, per the maintainers' own process for this test.

## Behaviour change flag

**BEHAVIOUR CHANGE (UI only, no new errors introduced):** a `perm_galaxy_editor` user without `perm_publish` will no longer see a "Publish Cluster" control that always 403'd when clicked. They will still correctly see "Unpublish Cluster" on already-published clusters they're allowed to unpublish. No previously-silent failure starts reporting a new error here — if anything, a control that always errored is now hidden. No controller, model, or route-level ACL logic is touched.

## Out of scope

A secondary, unrelated observation was made while tracing the controller-side re-check: `GalaxyCluster::fetchIfAuthorized()`'s `'publish'` branch (`app/Model/GalaxyCluster.php`) has a condition (`orgc_id != user_org_id && perm_publish`) that looks backwards relative to its own error message. This was investigated and found **UNCONFIRMED as exploitable** — every route that reaches this branch is already gated by the route-level ACL (`AND(perm_galaxy_editor, perm_publish)` for publish; `perm_galaxy_editor` alone for unpublish), so no reachable-with-effect bypass was found or constructed. It is not touched by this PR.

## Why this analysis might be wrong

- I did not check for another `side_menu.ctp` variant (e.g. under a `Themed/` directory) in case one exists in a different branch/version and also calls `canPublishGalaxyCluster()` with different surrounding logic — only one `side_menu.ctp` calling this method was found in this checkout (`origin/2.5`).
- The fix was verified via a scratch harness that isolates the two changed methods with a fake `$this->ACL`, not via a full CakePHP `Helper` instantiation with a live `$this->me`/`$this->ACL` component — I did not run the actual pinning test against the fixed code (it lives on a branch not present here, and I was asked not to edit or import it), so I cannot show the pinning test going green in this PR; I can only show the method now calls the intended target.
- I did not check every other call site across the codebase (e.g. API/JSON responses, other views) for `canPublishGalaxyCluster`/`canModifyGalaxyCluster` usage on galaxy clusters beyond `side_menu.ctp` and `AclHelper.php` itself.

---

*Priority revised low -> cosmetic: maintainer assessment — Wachizungu: "rather cosmetic (low impact) issue, as the impact is that you see an unusable button".*

### DIFF
--- a/app/View/Elements/genericElements/SideMenu/side_menu.ctp
+++ b/app/View/Elements/genericElements/SideMenu/side_menu.ctp
@@ -1633,7 +1633,10 @@ $divider = '<li class="divider"></li>';
                                 'url' => $baseurl . '/galaxy_clusters/add/' . h($galaxy_id) . '/forkUuid:' . h($cluster['GalaxyCluster']['uuid']),
                                 'text' => __('Fork Cluster')
                             ));
-                            if ($this->Acl->canPublishGalaxyCluster($cluster)) {
+                            $canTogglePublishState = $cluster['GalaxyCluster']['published']
+                                ? $this->Acl->canUnpublishGalaxyCluster($cluster)
+                                : $this->Acl->canPublishGalaxyCluster($cluster);
+                            if ($canTogglePublishState) {
                                 echo $divider;
                                 echo $this->element('/genericElements/SideMenu/side_menu_link', array(
                                     'onClick' => array(
--- a/app/View/Helper/AclHelper.php
+++ b/app/View/Helper/AclHelper.php
@@ -122,6 +122,20 @@ class AclHelper extends Helper
      */
     public function canPublishGalaxyCluster(array $cluster)
     {
+        return $this->ACL->canPublishGalaxyCluster($this->me, $cluster);
+    }
+
+    /**
+     * Unlike publishing, unpublishing a galaxy cluster only requires
+     * perm_galaxy_editor (see ACLComponent::ACL_LIST['galaxyClusters']['unpublish']),
+     * not perm_publish. This mirrors that asymmetry so the side menu can offer the
+     * correct control depending on the cluster's current published state.
+     *
+     * @param array $cluster
+     * @return bool
+     */
+    public function canUnpublishGalaxyCluster(array $cluster)
+    {
         return $this->ACL->canModifyGalaxyCluster($this->me, $cluster);
     }
 
--- a/app/View/Themed/UiBeta/Elements/genericElements/SideMenu/side_menu.ctp
+++ b/app/View/Themed/UiBeta/Elements/genericElements/SideMenu/side_menu.ctp
@@ -1564,7 +1564,10 @@ $divider = '<li class="divider"></li>';
                                 'url' => $baseurl . '/galaxy_clusters/add/' . h($galaxy_id) . '/forkUuid:' . h($cluster['GalaxyCluster']['uuid']),
                                 'text' => __('Fork Cluster')
                             ));
-                            if ($this->Acl->canPublishGalaxyCluster($cluster)) {
+                            $canTogglePublishState = $cluster['GalaxyCluster']['published']
+                                ? $this->Acl->canUnpublishGalaxyCluster($cluster)
+                                : $this->Acl->canPublishGalaxyCluster($cluster);
+                            if ($canTogglePublishState) {
                                 echo $divider;
                                 echo $this->element('/genericElements/SideMenu/side_menu_link', array(
                                     'onClick' => array(


elhoim:
Correction to this PR, pushed as a second commit.

The body originally stated that only one `side_menu.ctp` calls this method in `origin/2.5`. **That is wrong.** There are four call sites:

```
app/View/Elements/genericElements/SideMenu/side_menu.ctp:1636          (patched in commit 1)
app/View/Elements/genericElements/SideMenu/side_menu.ctp:1717          (publish-only, !published-gated - no change needed)
app/View/Themed/UiBeta/Elements/genericElements/SideMenu/side_menu.ctp:1567   (MISSED - now patched)
app/View/Themed/UiBeta/Elements/genericElements/SideMenu/side_menu.ctp:1648   (publish-only, !published-gated - no change needed)
```

The UiBeta theme carries a byte-identical copy of the publish/unpublish toggle. With only the default theme patched, a `perm_galaxy_editor` user **without** `perm_publish` on the UiBeta theme would have lost the Unpublish control — which `ACL_LIST` grants on `perm_galaxy_editor` alone (`ACLComponent.php:528`). That would have traded a permission leak in one theme for a permission regression in the other.

The second commit applies the same `published ? canUnpublish : canPublish` branch at UiBeta:1567.

Wachizungu:
Would say this is a rather cosmetic (low impact) issue, as the impact is that you see an unusable button.

Curious about the reason for the asymmetry between publish and unpublish... I see events actually have similar logic in ACLComponent (perm_modify required for unpublish).

elhoim:
Agreed on the severity — retitled to `[low]` and revised the BLUF accordingly. The server-side ACL still refuses the action, so this is a button that does nothing rather than a privilege escalation.

On the asymmetry: it looks deliberate and it is consistent between galaxy clusters and events.

`ACLComponent.php`, galaxy clusters:

```php
'publish'   => array('AND' => ['perm_galaxy_editor', 'perm_publish']),   // :524
'unpublish' => array('perm_galaxy_editor'),                              // :528
```

and events, exactly as you noted:

```php
'publish'   => array('perm_publish'),   // :400
'unpublish' => array('perm_modify'),    // :423
```

In both cases publishing requires `perm_publish`, while unpublishing only requires the edit-level permission. That reads as intentional: publishing *increases* exposure and is the privileged act, whereas unpublishing is a retraction — you are undoing distribution, so gating it behind the lower "may edit this object" bar lets an editor pull something back without waiting for a publisher. The same reasoning is why `perm_modify` rather than `perm_publish` guards event unpublish.

The component method already encodes exactly that rule:

```php
public function canPublishGalaxyCluster(array $user, array $cluster)   // :1404
{
    if (!$this->canModifyGalaxyCluster($user, $cluster)) {
        return false;
    }
    return (bool)$user['Role']['perm_publish'];
}
```

so "can modify" **and** `perm_publish`. The defect is only in the view helper, which delegates `canPublishGalaxyCluster()` to `ACL->canModifyGalaxyCluster()` and therefore applies the *unpublish* rule to both buttons. This PR just points the helper at the method above, so the rendered control matches the rule the controller already enforces.

PR #11060, opened 2026-08-28
[low] chg: [CI] only check out the submodules the build job actually uses

elhoim:
**BLUF —** The build job checks out all eighteen submodules though only eight are used; this PR narrows it.

- **Problem** — The `build` job checks out all eighteen submodules recursively even though only eight are read by any step, and on run 32854863121 that checkout cost 77s of the 865s PHP 8.1 job and 67s of the 798s 8.3 job.
- **Fix** — Drops `submodules: 'recursive'` from `actions/checkout@v5` and replaces it with an explicit `git submodule update --init --recursive` limited to the eight paths actually used, keeping `--recursive` and no `--depth` so pinned commits stay reachable. The retained list deliberately includes `misp-decaying-models` and `noticelists`.
- **Effect** — Most of that per-job checkout time comes back on every build run.
- **Cost** — Submodules are now cloned anonymously over HTTPS, which works because all of them are public.

The `build` job checks out every submodule recursively, but only eight of the eighteen are ever read by a step in that job. This narrows the checkout to those eight.

### Measured cost

From GitHub Actions run 32854863121 on `2.5`, the `checkout` step with recursive submodules is **77s on the PHP 8.1 job (8.9% of the 865s job) and 67s on the 8.3 job (8.4% of 798s)**. Fetching ten fewer repositories should take most of that back.

### What changed

`submodules: 'recursive'` is dropped from `actions/checkout@v5` (which then defaults to not touching submodules at all) and replaced by an explicit `git submodule update --init --recursive -- <paths>` step listing the eight submodules the job uses. `--recursive` is kept so nested submodules of the retained paths behave exactly as before, and no `--depth` is used, so the pinned commits are always reachable.

One behavioural difference worth naming: with the `submodules:` input gone, `actions/checkout` no longer injects its auth config for submodule URLs, so these are cloned anonymously over HTTPS. All of them are public MISP/upstream repositories, so this works, but it is the only change beyond the path list.

### Submodules kept (8)

`app/Lib/cakephp`, `PyMISP`, `app/files/taxonomies`, `app/files/warninglists`, `app/files/misp-galaxy`, `app/files/misp-objects`, `app/files/noticelists`, `app/files/misp-decaying-models`.

The last one is easy to overlook: `PyMISP/tests/testlive_comprehensive.py::test_search_decay` calls `update_decaying_models()` and asserts `'NIDS Simple Decaying Model'` is present; `DecayingModel::__load_models` reads `app/files/misp-decaying-models/models` from disk and `update()` throws on an empty folder. Similarly `noticelists` is cheap to ingest but `Server::updateJSON()` iterates it unconditionally, so its absence would error the `Update JSON` step.

### Submodules excluded (10), with the reason each is unused by this job

**The six python STIX submodules** — `app/files/scripts/misp-stix`, `app/files/scripts/cti-python-stix2`, `app/files/scripts/python-stix`, `app/files/scripts/python-cybox`, `app/files/scripts/mixbox`, `app/files/scripts/python-maec`. The `MODULE_TO_DIRECTORY` block at the top of `stixtest.py`, `stix2misp.py`, `stix2/stix2misp.py` and `stix2/misp2stix2.py` calls `importlib.import_module(name)` first and only falls back to `sys.path.insert` of the local submodule directory **on ImportError**. `requirements.txt` pins `misp-stix>=2026.6.3`, whose own dependencies are `pymisp`, `stix`, `cybox`, `mixbox`, `maec`, `stix2` (and friends), so after `pip install -r requirements.txt` every import resolves from the venv and the local paths are never inserted. They remain the designed offline fallback for non-CI installs and stay in `.gitmodules` untouched — this only stops the CI job from fetching them.

**`app/files/scripts/misp-opendata`** — referenced only by `app/Lib/Export/OpendataExport.php` and by the diagnostics whitelist in `Server::isAcceptedSubmodule`. No step in this workflow performs an opendata export.

**`app/files/misp-event-templates`** — `EventTemplate::LIBRARY_DIR` is reached only via `EventTemplatesController::update` and `populateIfEmpty`. `tests/testlive_event_templates.py` exists in the repo but is not invoked by this workflow, and `app/Test/EventTemplate*.php` are pure unit tests with no filesystem reads.

**`app/files/misp-workflow-blueprints`** — `WorkflowBlueprint::REPOSITORY_PATH` is read only by `__readBlueprintsFromRepo` via `update()`, which is controller-triggered. No step hits it.

**`app/files/misp-iconify`** — the CSS actually served is a copy committed to the repo: `app/webroot/css/misp-iconify.css` is self-contained (base64 data-URI masked SVGs) and the layouts resolve the asset from webroot. (`app/webroot/css/misp-iconify-font.css` refers to webfonts that do not exist in the repo and is included by nothing, but that is a separate matter and not touched here.)

### Caveats

- Nothing was kept purely out of caution — each of the ten exclusions above is backed by a specific code path rather than an assumption. Conversely, `misp-decaying-models` and `noticelists` were *not* dropped despite looking droppable at first glance, for the reasons given above.
- **This has not been executed on MISP's runners.** The analysis is static, from reading the workflow and the code each step reaches; I have no way to run your CI. Please treat the first run on this branch as the actual verification.
- If anything does break, the revert is one line: delete the new step and put `with: submodules: 'recursive'` back on the checkout.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -79,8 +79,15 @@ jobs:
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v5
-        with:
-          submodules: 'recursive'
+
+      # Only the submodules this job actually reads are checked out. The other
+      # ten (the six python STIX ones, misp-opendata, misp-event-templates,
+      # misp-workflow-blueprints, misp-iconify) are never touched by any step
+      # here. See the PR that introduced this step for the per-submodule
+      # justification. To restore the old behaviour, drop this step and put
+      # `with: submodules: 'recursive'` back on the checkout above.
+      - name: Checkout the submodules used by this job
+        run: git submodule update --init --recursive -- app/Lib/cakephp PyMISP app/files/taxonomies app/files/warninglists app/files/misp-galaxy app/files/misp-objects app/files/noticelists app/files/misp-decaying-models
 
       - name: Stop default MySQL (if present)
         run: sudo service mysql stop || true
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -86,8 +86,15 @@ jobs:
       # here. See the PR that introduced this step for the per-submodule
       # justification. To restore the old behaviour, drop this step and put
       # `with: submodules: 'recursive'` back on the checkout above.
+      #
+      # /!\ --depth 1 is load-bearing. actions/checkout runs
+      # `git submodule update --init --force --depth=1 --recursive`, because its
+      # fetch-depth input defaults to 1. Doing the same by hand WITHOUT --depth
+      # clones the full history of every submodule instead, which is slower than
+      # the recursive checkout this step replaces - misp-warninglists alone is a
+      # 467 MB repository.
       - name: Checkout the submodules used by this job
-        run: git submodule update --init --recursive -- app/Lib/cakephp PyMISP app/files/taxonomies app/files/warninglists app/files/misp-galaxy app/files/misp-objects app/files/noticelists app/files/misp-decaying-models
+        run: git submodule update --init --recursive --depth 1 -- app/Lib/cakephp PyMISP app/files/taxonomies app/files/warninglists app/files/misp-galaxy app/files/misp-objects app/files/noticelists app/files/misp-decaying-models
 
       - name: Stop default MySQL (if present)
         run: sudo service mysql stop || true


elhoim:
Correcting a defect in this PR before anyone spends time on it.

As originally pushed, this change would have made the checkout **slower**, not faster, and the PR body presented the cause as a feature.

`actions/checkout` does not simply run `git submodule update --init --recursive`. It runs `git submodule update --init --force --depth=1 --recursive`, because its `fetch-depth` input defaults to `1` (`src/git-command-manager.ts`, `submoduleUpdate()`; `action.yml`, `fetch-depth`). The explicit step I added omitted `--depth`, which would have converted eight shallow submodule clones into eight full-history clones. `misp-warninglists` alone is a 467 MB repository, so that single omission would have cost more than the ~9% this PR sets out to save.

`c4f87a63a2` adds `--depth 1` and comments why it is load-bearing, so it does not get dropped again by someone tidying the command up.

The rest of the change stands: the ten excluded submodules are still unused by this job, and the revert is still one line. But the original body claimed the absence of `--depth` was deliberate ("no `--depth`, so pinned commits stay reachable") — that was wrong, and `actions/checkout` has always fetched these shallow.

Still not run on MISP’ s runners.

PR #11061, opened 2026-08-28
[low] fix: [updateJSONLite] implement the minimal JSON set the subcommand has always advertised

elhoim:
**BLUF —** `updateJSONLite` has always run a full ingest; this PR gives it the minimal set it advertises.

- **Problem** — `cake Admin updateJSONLite` advertises a minimal testing-only ingest, yet `Server::updateJSONLite()` has never existed; `AdminShell` now calls `Server->updateJSON(true)`, which takes no parameter, so the argument is discarded.
- **Fix** — Gives `Server::updateJSON()` an optional `$lite` flag that passes a base directory to the six `Model->update()` methods (`Galaxy`, `Noticelist`, `ObjectRelationship`, `ObjectTemplate`, `Taxonomy`, `Warninglist`), backed by a committed fixture tree under `app/files/lite/`.
- **Effect** — Tests get the minimal ingest the help text promises.
- **Cost** — Anyone relying on `updateJSONLite`'s accidental full ingest now gets the tiny fixture set instead; all new parameters are optional and trailing, so other callers are unaffected.

This is a first step towards what you outlined in #11004: "The first step will be to come up with a sane set of mock data for the JSON ingested structures (galaxies, objects, etc) and avoid actually ingesting the sub modules." It implements the subcommand that already promises exactly that.

## The history

`cake Admin updateJSONLite` has been registered in `AdminShell::getOptionParser()` with the help text `***TESTING ONLY*** Update the JSON definitions of MISP - but with a minimal set, only meant for testing.` — but the minimal set was never implemented:

- `Server::updateJSONLite()` has never existed in any commit. `git log -S 'function updateJSONLite' -- app/Model/Server.php` returns nothing across the whole history.
- `AdminShell::updateJSONLite()` therefore called an undefined method, i.e. it fatalled, until e68aaf9298 ("fix: [update json] fast mode fix") repointed it at `Server->updateJSON(true)`.
- `Server::updateJSON()` takes no parameter, so that `true` is silently discarded and "Lite" runs the identical full ingest.

So `updateJSONLite` today is a slower-to-say alias for `updateJSON`. This PR makes it do what its help text says.

**This is behaviour-changing for anyone who relies on today's accidental full-ingest behaviour of `updateJSONLite`.** After this PR it ingests a deliberately tiny fixture set. That is the documented contract of the subcommand, and `updateJSON` remains the way to ingest the real definitions.

## What the change does

`Server::updateJSON($lite = false)` now passes an optional base directory down to the six `Model->update()` methods. With `$lite`, that base is `app/files/lite/` — a small fixture tree committed to this repository — instead of the submodule-backed `app/files/`.

- Every new parameter is optional, trailing, and defaults to the exact path that is hardcoded today.
- The non-lite branch still calls `$model->update()` with **no arguments at all** (`$lite ? $model->update(...$arguments) : $model->update()`), so `cake Admin updateJSON`, the web-UI update buttons and `Server::updateSubmodule*` are untouched by construction.
- `AdminShell` needed no change — it already passes `true`.
- Because the fixtures are committed rather than shipped in submodules, `updateJSONLite` also works on a checkout made without `--recursive`.

`app/files/lite/README.md` documents each fixture and why it is there.

### Galaxies are addressed — they are the point

Galaxies are ~80% of the cold-ingest cost, so a change that left them alone would not be worth having. `Galaxy::update()` globs the galaxy definitions and the cluster packages from the same root, and skips any cluster package whose `type` was not loaded, so pointing both globs at the fixture tree means the 131 large cluster files are never even read, let alone decoded. The lite set ships two galaxies (`mitre-mobile-attack-attack-pattern`, `wiper`) with 25 + ~100 clusters.

### How the fixture set was chosen

Picked by reading the assertions in `PyMISP/tests/testlive_comprehensive.py`, `tests/testlive_comprehensive_local.py` and `tests/testlive_sync.py`, not by guessing:

- `wiper` sorts last alphabetically, so it becomes the highest-id galaxy that `galaxies()[0]` returns; it has clusters, and the name `Wiper` matches no other retained galaxy, which `test_search_galaxy`'s `len(ret) == 1` needs.
- Exactly three warninglists match `%dns resolv%` (`'3 warninglist(s) enabled'`). `public-dns-v4` and `public-dns-hostname` are trimmed to a handful of literal entries, keeping `8.8.8.8`, `9.9.9.9` and `1.1.1.1`, and deliberately not `8.8.8.9` or `1.11.71.4`, which must trip nothing. Their `version` is lowered to `1` so a later real `updateJSON` always supersedes them.
- `common-ioc-false-positive` keeps the empty-file MD5; `tlp` keeps its `red` predicate; `file`, `domain-ip` and `asn` object templates and the `gdpr` noticelist are looked up by name. `misp-objects/relationships/definition.json` is copied verbatim — trimming it gains nothing.

## Validation — please read this part

**I have not been able to execute the ingestion.** There was no database available in the environment this was written in, so nothing below is measured: no `cake` run, no test run, no timing. What was done is source reading, JSON validation of every fixture file, and `php -l` on all seven changed PHP files (clean; the pre-existing deprecation notices in `Server.php` and `Galaxy.php` are unrelated to this diff). Treat every claim about runtime behaviour as reasoning about the code, not as an observation.

To validate, on a fresh instance:

```
# non-lite must be unchanged
sudo -u www-data /var/www/MISP/app/Console/cake Admin updateJSON

# lite, ideally on a database that has never been ingested into
sudo -u www-data /var/www/MISP/app/Console/cake Admin updateJSONLite
```

Success for `updateJSONLite` looks like: it completes in a couple of seconds rather than minutes, reports success for all six structures, and leaves the database with **2 galaxies, 4 warninglists, 1 taxonomy (`tlp`), 3 object templates (`asn`, `domain-ip`, `file`), 1 noticelist (`gdpr`)** and the full relationship set. The stronger check is running the live suites against a lite-ingested instance and seeing `testlive_comprehensive.py` still pass — in particular `test_warninglists`, `test_search_galaxy`, `test_galaxy_cluster` and `test_freetext`, which carry the count assertions the fixture set was built around.

Failure would look like `'3 warninglist(s) enabled'` returning a different count, `search_galaxy` returning more than one result, or a missing named entry.

## Not covered by this PR

Being explicit, because this alone does not yet let CI drop the submodules:

- `testlive_comprehensive.py`'s `setUpClass` calls the per-controller `update_*` endpoints (`galaxies/update`, `warninglists/update`, …), which read the real `app/files/` paths. On a submodule-less checkout those become no-ops over empty directories rather than a re-ingest, but they do not honour the lite flag.
- `get_raw_object_template('domain-ip')` reads `app/files/misp-objects/objects/` from disk directly, not the database.
- The decaying-models submodule is needed by `testlive_comprehensive.py` independently of `updateJSON`.

So switching the workflow off `submodules: 'recursive'` is a follow-up, and I did not touch `.github/workflows/main.yml` here. Happy to do that in a separate PR, or to adjust any of the fixture choices if you would rather the set look different.


elhoim:
## Validation: this has now actually been run

The PR body says the change was never executed. It has been now, end to end, against a scratch database built from a structure-only dump of a live MISP schema. Reporting the results here, including the parts that are less tidy than the body implies.

### Lite mode works, and the "seconds, not minutes" claim holds

| structure | `updateJSONLite` | `updateJSON` (same code, same machine) |
|---|---|---|
| Galaxies | 0.92s | 461.30s |
| Noticelists | 0.07s | 0.06s |
| Warninglists | 0.17s | 79.19s |
| Taxonomies | 0.04s | 16.28s |
| ObjectTemplates | 0.18s | 22.33s |
| ObjectRelationships | 3.54s | 3.45s |
| **total** | **5.3s** | **9m43s** |

Exit status 0. Note that 3.5s of the 5.3s is `ObjectRelationships`, which the fixture set ships verbatim rather than trimmed.

### Resulting content matches what the PR claims

2 galaxies (`Mobile Attack - Attack Pattern`, `Wiper`) / 101 clusters / 676 elements / 15 cluster relations; 4 warninglists / 349 entries; 1 taxonomy (`tlp`, 8 predicates); 3 object templates (`asn`, `domain-ip`, `file`) / 57 elements; 1 noticelist (`gdpr`, 7 entries); and 313 object relationships — identical to the full run and to a live full-corpus instance, so "the full relationship set" is literally true.

### The `%dns resolv%` constraint: passes, but there is a collation detail worth recording

`PyMISP/tests/testlive_comprehensive.py` calls `toggle_warninglist(warninglist_name='%dns resolv%', force_enable=True)` and asserts the exact string `'3 warninglist(s) enabled'`, so the fixture set has to yield exactly three matches.

- A literal case-sensitive `name LIKE '%dns resolv%'` returns **0**, because `warninglists.name` is `utf8mb3_bin`.
- Case-insensitively it returns **exactly 3** (hostname, IPv4, IPv6 public DNS resolvers).
- A live full-corpus instance behaves **identically** — 0 case-sensitive, 3 case-insensitive.

So the fixture set reproduces production on exactly the terms the real corpus gives, which is the thing that matters. Flagging it because anyone verifying this by hand with a bare `LIKE` will get 0 and think the fixtures are broken.

Related: `8.8.8.8` is present in "List of known IPv4 public DNS resolvers" with `enabled=0`. Every warninglist lands `enabled=0` — in the lite run, in the full run, and on a live instance (0 of 125 enabled). PyMISP enables lists itself, so present-but-disabled is the correct production-matching state.

### The non-lite path is byte-for-byte unchanged

This is the claim that actually matters, so it was checked by content and not by reading the diff. Full `updateJSON` was run with the PR applied and the galaxy data fingerprinted:

```
clusters   = 55392
elements   = 367882
relations  = 63130
cluster_fp = 270805528385567381
elem_fp    = 826929597424496570
```

Identical to the pre-PR baseline. The fingerprint is `BIT_XOR(CONV(SUBSTRING(MD5(CONCAT_WS('|', ...)),1,15),16,10))`, which is order-independent so row ordering and auto-increment ids are free to differ; it is paired with the exact row counts because `BIT_XOR` cancels identical rows in pairs.

### Method, and its limits

The scratch database was rebuilt from a structure-only dump of the live schema rather than from `INSTALL/MYSQL.sql` — the latter is the pre-migration baseline and is missing columns that later migrations add, so ingestion fails against it with `Unknown column 'distribution'`.

One number that did not reconcile and is reported rather than explained: the full run's galaxy stage took 461s here, matching neither of the two figures measured earlier on this machine. The machine was loaded at the time. It was not chased, because the fingerprint is the correctness check and the wall clock is not — but it should not be quoted as a measurement of anything.

This was validated on a developer container, not on MISP's CI runners.

PR #11062, opened 2026-08-28
[low] ci: hook up the additional python suites in a separate job (depends on #11061)

elhoim:
**BLUF —** Committed live test suites never ran in CI; this PR adds a job that executes them.

- **Problem** — Several live suites already in the tree, including `tests/testlive_event_addtag.py`, `testlive_event_mass_actions.py`, `testlive_event_templates.py` and `testlive_collection_sync.py`, are never executed by CI, so regressions in them go unnoticed.
- **Fix** — Adds one purely additive `python-suites` job to `.github/workflows/main.yml`, run in parallel with `build`. It installs PyMISP's optional magic binding for `test_mimeType`, leaves `testlive_comprehensive` to `build`, and enables the `tlp` taxonomy for `test_search_index_by_tag`.
- **Effect** — Those four suites gain regression coverage on every run.
- **Cost** — Depends on #11061's cheap `cake Admin updateJSONLite` ingest step; merged first, the job only runs slower rather than failing.

This is step 2 of the plan sketched in #11004: *"The first step will be to come up with a sane set of mock data for the JSON ingested structures ... Once we fix that we can look into hooking up the additional .py tests (from pymisp + MISP/tests/ at least)."*

## Depends on #11061

**This PR should be merged after #11061** (`Admin updateJSONLite` + the `app/files/lite/` fixture tree). The new job's ingestion step is `app/Console/cake Admin updateJSONLite`, which is what keeps it affordable.

If it is merged first, nothing breaks: `AdminShell::updateJSONLite()` currently calls `Server->updateJSON(true)` and the argument is discarded, so the step performs a full ingest and the job is simply as slow as `build` (the measured `Update JSON` step is 216–225s of an 800–865s job). The failure mode is minutes, not a red run.

## What it adds

One new job, `python-suites`, running in parallel with `build`. Nothing is removed from `build` and no existing step is touched — the diff is purely additive.

Suites that exist in the tree and were never executed by CI:

| Suite | What it covers |
|---|---|
| `tests/testlive_event_addtag.py` | the single-event `/events/addTag/{id}` contract (tag id/name/JSON list, `collection_N` expansion, `local:1`, duplicate skip, `local_only`, unpublish side effect, permission refusals) |
| `tests/testlive_event_mass_actions.py` | `/events/addTag/selected` and `/galaxies/attachMultipleClusters/selected/event`, including the global→local downgrade matrix and its security boundary |
| `tests/testlive_event_templates.py` | the `/event_templates` REST surface — CRUD, duplicate, validate_definition, instantiate, import/export round-trip |
| `tests/testlive_collection_sync.py` | collection sync feature negotiation + push/pull round-trips, loopback mode (`HOST` + `AUTH` only) |
| PyMISP `tests/test_analyst_data.py`, `test_attributevalidationtool.py`, `test_fileobject.py`, `test_reportlab.py` | offline unit tests, no instance and no network, a couple of seconds in total |

The job also runs PyMISP's `testlive_comprehensive.py` and `tests/testlive_comprehensive_local.py`. That is deliberate duplication with `build`: the four new live suites need an instance anyway, and running the two big suites here as well is what makes this job the practical validation gate for the lite fixture set. Slimming `build` down once this is proven green is a follow-up, not this PR.

## Two settings that are easy to get wrong

**`MISP.background_jobs 0`, with `--force`.** This job deliberately starts no supervisor workers. Without the setting, `Event::publishRouter` enqueues to a queue nothing drains and published events never reach `published=1`. And `cake Admin setSetting MISP.background_jobs 0` is *refused* without `--force` ("It is highly recommended that this setting is enabled") — the refusal does **not** exit non-zero, so the setting silently stays at its default and the suites fail much later in ways that look unrelated. The `--force` and the reason for it are commented in the workflow.

**The six `Plugin.ZeroMQ_*` keys.** PyMISP's `test_zmq` subscribes to the publisher; without the redis coordinates and both `_enable` flags nothing is ever pushed and the test times out rather than failing loudly.

## Ordering constraint (please keep the comment)

PyMISP's `testlive_comprehensive.py` and `tests/testlive_comprehensive_local.py` both create an organisation named `Test Org`. PyMISP's tears its own down; the local suite's teardown is silently refused whenever one of its events still points at the org — `Organisation::beforeDelete` returns `false` and PyMISP surfaces that only as an error dict the teardown ignores — so `Test Org` survives it in practice. With the order reversed, all of PyMISP's tests error in `setUpClass` in about seven seconds while still appearing to "run". `build` already has this order right; here it is spelled out in a comment at the exact line so nobody reorders it later.

## What it costs

Baseline numbers are measured (GitHub Actions run 32854863121 on `2.5`); the new job's are estimates.

- `build` today: 865s (8.1) / 798s (8.3), of which `Update JSON` is 225s / 216s.
- `python-suites`: **one** PHP version (8.3 on noble), not a matrix — these suites exercise application behaviour, not PHP-version compatibility. Services are MariaDB + Redis only; no OpenLDAP, no supervisor.
- Estimated ~9–12 minutes of runner time per run: the same bootstrap as `build`, an ingest that lite mode turns from ~220s into seconds, then the comprehensive suites and the four new ones.
- **Wall clock of a PR should be unchanged** — the job runs in parallel with `build` and is expected to finish inside `build`'s duration. The cost is runner minutes, roughly +50% over today's two-job total.

If that is still too much, the natural dials are to drop the duplicated `testlive_comprehensive*` runs from this job once `build` is slimmed, or to gate the job on `pull_request` only.

## Deliberately excluded

- `tests/ui_event_mass_actions_check.py` — needs headless chromium over CDP plus `websocket-client`/`lxml`; a browser dependency in CI is a separate discussion.
- PyMISP `tests/testlive_sync.py` — needs two distinct instances.
- PyMISP `tests/test_neo4j.py` — needs a neo4j server.
- PyMISP `tests/test_emailobject.py` — `pymisp.tools.emailobject` imports `extract_msg`, `RTFDE` and `oletools`, none of which are in `requirements.txt`. `pymisp/tools/__init__.py` guards that import, but the test imports `EMailObject` directly, so it would be a pytest **collection error**, not a skip. Adding those three dependencies to CI is a separate call.
- `tests/testlive_collection_sync.py`'s two-instance write-branch assertions (loopback mode runs; a second instance does not).

Note also that `submodules: 'recursive'` is kept here on purpose: `testlive_comprehensive.py` re-ingests all five JSON corpora itself in `setUpClass` and `get_raw_object_template('domain-ip')` reads `app/files/misp-objects` from disk. Trimming the submodule checkout is a separate PR.

## Not executed

**None of this has been run on MISP's runners, or on any runner.** What was done: reading the suites and the existing workflow to establish what each one needs, verifying that the four PyMISP unit tests' imports are satisfied by the pinned `pymisp==2.5.34.1` wheel and `requirements.txt` (checked against the wheel's `pymisp/tools/__init__.py`), and validating that the resulting `main.yml` parses. The first real signal will be this PR's own CI run — which, until #11061 is in, will exercise the job with a full ingest.

One thing to watch on that first run: `build` runs the comprehensive suites *with* workers, this job runs them without. Any test that depends on a job actually being queued rather than executed synchronously would show up here and not there.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -713,8 +713,11 @@ jobs:
           #
           # /!\ --force is load-bearing: without it `setSetting` REFUSES this
           # value ("It is highly recommended that this setting is enabled") and
-          # still exits 0, so the setting would silently stay at its default and
-          # the suites would fail much later, in confusing ways.
+          # the setting stays at its default, so the suites would fail much
+          # later, in confusing ways. The refusal does exit non-zero
+          # (AdminShell::setSetting -> Shell::error -> _stop(CODE_ERROR)), so a
+          # step that runs this bare would fail loudly; it is a silent no-op
+          # only when the exit status is swallowed by a surrounding loop.
           app/Console/cake Admin setSetting --force "MISP.background_jobs" 0
           app/Console/cake Admin setSetting "SimpleBackgroundJobs.enabled" 0
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -789,20 +789,19 @@ jobs:
           sudo chmod -R g+ws `pwd`/app/tmp/logs
           . ./venv/bin/activate
 
-          # /!\ ORDER IS LOAD-BEARING -- DO NOT REORDER /!\
-          # PyMISP's testlive_comprehensive.py and tests/testlive_comprehensive_local.py
-          # both create an organisation literally named "Test Org". PyMISP's
-          # suite deletes its own in tearDownClass; the local one's teardown is
-          # silently refused whenever one of its events still points at the org
-          # (Organisation::beforeDelete returns false and PyMISP surfaces that
-          # only as an error dict that the teardown ignores), so "Test Org"
-          # survives the local suite in practice. Running the local suite first
-          # therefore makes every PyMISP test error in setUpClass -- in about
-          # seven seconds, while still looking like it "ran".
-          pushd PyMISP
-          cp tests/keys.py .
-          python -m pytest -v --durations=0 tests/testlive_comprehensive.py
-          popd
+          # PyMISP's own tests/testlive_comprehensive.py is deliberately NOT run
+          # here. It is already run by the `build` job, and this job is not a
+          # valid environment for it: background jobs are switched off above
+          # (MISP.background_jobs 0, SimpleBackgroundJobs.enabled 0) and the
+          # JSON corpora are the small updateJSONLite fixture set rather than
+          # the full submodules. Running it anyway failed four of its tests -
+          # test_feeds, test_freetext, test_roles_expanded and
+          # test_user_settings - purely because of those two differences; the
+          # same four pass in `build` on the same commit. It was only ever
+          # ordered before the local suite because both create an organisation
+          # named "Test Org" and the local suite's teardown leaves it behind,
+          # which would break PyMISP's setUpClass. With PyMISP's suite gone from
+          # this job that ordering constraint no longer applies here.
           python tests/testlive_comprehensive_local.py -v
 
           # Suites that live in tests/ and were never wired into CI. Each takes
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -758,6 +758,20 @@ jobs:
       # satisfied by requirements.txt are listed -- test_emailobject needs
       # extract_msg/RTFDE/oletools and test_neo4j needs a neo4j server, and an
       # unsatisfied import there is a collection error, not a skip.
+      # FileObject only emits its mimetype attribute when PyMISP's optional
+      # magic binding is importable; without it the attribute is simply absent
+      # and test_fileobject.py::test_mimeType fails with StopIteration from the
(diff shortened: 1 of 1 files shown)


elhoim:
Correcting a factual error I introduced in this PR, and in an earlier comment on #11004.

The body and the workflow comment both claimed that `cake Admin setSetting "MISP.background_jobs" 0` is refused without `--force` **and still exits 0**. That is wrong. `AdminShell::setSetting()` calls `Shell::error()`, which ends in `_stop(self::CODE_ERROR)` with `CODE_ERROR = 1`, and `CakeObject::_stop()` is `exit($status)`. The refusal exits **1**.

`--force` is still required — without it the setting is refused and stays at its default, which is exactly the failure this guards against — but the failure is loud, not silent. What I originally observed (the setting silently staying at its default with no visible error) was real; my explanation of it was not. The non-zero exit was being swallowed by the surrounding settings loop.

Fixed in the workflow comment by `1a83f5f`. Flagging it explicitly because a comment that tells the next reader "this command lies about its exit status" would send them chasing a bug that does not exist.

PR #11065, opened 2026-09-01
fix: [feed] guard missing disable_correlation setting in saveFreetextFeedData (#11007)

### LINKED ISSUE #11007: Bug: Undefined array key "disable_correlation" in [/var/www/MISP/app/Model/Feed.php, line 1358]
### Actual behavior

I imported the default feeds from https://github.com/MISP/MISP/blob/2.4/app/files/feed-metadata/defaults.json (linked on https://www.misp-project.org/feeds/ ). 
If I fetch the feed data, some feeds through an error.
It's in the function public function saveFreetextFeedData(array $feed, array $data, array $user, $jobId = false){}

In my opinion, it's reproduceable withe default feeds in 2.5 https://github.com/MISP/MISP/blob/2.5/app/files/feed-metadata/defaults.json.
Some freetext/csv feeds, don't have the "disable_correlation" Attribute.

### Expected behavior

The Variable $disableCorrelation  is set to false, right before but it seems, that an isset() for $feed['Feed']['settings']['disable_correlation'] is missing.
similar to Commit 19c4808 

### Steps to reproduce

Import freetext/csv feeds from the default feeds list like:
- https://feodotracker.abuse.ch/downloads/ipblocklist.csv 
- https://sslbl.abuse.ch/blacklist/sslblacklist.csv
- https://bazaar.abuse.ch/export/txt/md5/recent/

### Version

2.5.45

### Operating System

Ubuntu

### Operating System version

24.04

### PHP version

8.3

### Browser

_No response_

### Browser version

_No response_

### Relevant log output

```shell
Error: Warning Error: Undefined array key "disable_correlation" in [/var/www/MISP/app/Model/Feed.php, line 1358]
```

### Extra attachments

_No response_

### Code of Conduct

skraft9:
#### What does it do?

Fixes #11007.

`saveFreetextFeedData` builds the flag with `(bool) $feed['Feed']['settings']['disable_correlation'] ?? false`. The `(bool)` cast binds tighter than `??`, so the array key is read before the coalesce can guard it, and PHP 8 raises an "Undefined array key" warning for any feed whose `settings` array exists but does not carry `disable_correlation` (the case for several default freetext/CSV feeds). Wrapping the coalesce inside the cast resolves the key first, matching the guarded read a few lines up (Feed.php:1148-1149).

#### Questions

### DIFF
--- a/app/Model/Feed.php
+++ b/app/Model/Feed.php
@@ -1355,7 +1355,7 @@ class Feed extends AppModel
             }
             $disableCorrelation = false;
             if (!empty($feed['Feed']['settings'])) {
-                $disableCorrelation = (bool) $feed['Feed']['settings']['disable_correlation'] ?? false;
+                $disableCorrelation = (bool)($feed['Feed']['settings']['disable_correlation'] ?? false);
             }
             $event = array(
                 'info' => $feed['Feed']['name'] . ' feed',


iglocska:
Lovely thanks!

Issue #5500, opened 2020-01-10
414 Request-URI Too Long


RichieB2B:
When mass editing attributes, MISP does a GET request like:
```
https://misp.local/attributes/editSelected/28854/[2113775,2113833,2113834,2113835,2113830,2113831,2113832,2113827,2113828,2113829,2113824,2113825,2113826,2113821,2113822,2113823,2113818,2113819,2113820,2113815,2113816,2113817,2113812,2113813,2113814,2113791,2113792,2113793,2113809,2113810,2113811,2113806,2113807,2113808,2113803,2113804,2113805,2113800,2113801,2113802,2113797,2113798,2113799,2113794,2113795,2113796,2113785,2113786,2113787,2113788,2113789,2113790,2113782,2113783,2113784,2113779,2113780,2113781,2113776,2113777,2113778,2113774,2113773,2113772,2113764,2113771,2113770,2113769,2113768,2113767,2113766,2113765,2113763]?_=1578668575760
```
This can get quite long. Apache has a default [LimitRequestLine ](http://httpd.apache.org/docs/2.4/mod/core.html#limitrequestline)of 8190 bytes, but [CIS Apache Benchmark 1.10.1](https://www.cisecurity.org/benchmark/apache_http_server/) limits this to 512 bytes.

Perhaps MISP can use of POST for this?

iglocska:
Yeah that should be converted to a POST.

krayzpipes:
The same issue is present when merging objects representing x509 certificates.  It results in a GET request that is too long.

mokaddem:
Attribute mass edit and merging object are both no longer passing data via the URL. Thanks for the report!
(@krayzpipes it's cool to see that you are using that feature :smiley:)

PR #11066, opened 2026-09-01
fix: [sync] POST the UUID list when pulling analyst data to avoid 414 (#10958)

### LINKED ISSUE #10958: Server pull of Analyst Notes fails with large UUID batches (414 Request-URI Too Long)
Misp 2.5.39 : 
The pull of Analyst Notes fails when the batch contains around 100 UUIDs. Reducing the batch to ~40 UUIDs allows the pull to succeed. The request appears to be sent as a GET with the UUIDs included in the query string, which likely exceeds URL/header length limits (e.g. Apache LimitRequestLine, nginx large_client_header_buffers) on the receiving end.
This looks similar to a previously fixed issue on attributes/editSelected (#5500), where a GET request built with a long list of IDs also caused a 414 Request-URI Too Long, and the resolution was to convert the call to POST.

Steps to reproduce :
Configure a sync server with Analyst Data pull enabled.
Trigger a pull where the batch of Analyst Notes to fetch contains ~100 UUIDs.
Observe the pull failing (likely 414 Request-URI Too Long at the proxy/web server level).
Reduce the batch to ~40 UUIDs — the pull succeeds.
Suggested fix
Convert the relevant Analyst Notes pull request(s) from GET (UUIDs in query string) to POST (UUIDs in request body), consistent with the fix applied for #5500.

skraft9:
#### What does it do?

Fixes #10958.

`ServerSyncTool::fetchAnalystData` builds `/analyst_data/index/{type}` with every UUID appended as a named parameter and sends it as a GET. A large batch (around 100 UUIDs and up) pushes the request line past the web server limit (Apache `LimitRequestLine` / nginx `large_client_header_buffers`) and the pull fails with 414, which matches the ~40-ok / ~100-fail split in the report. This sends the UUID list in the POST body instead, the same shape `fetchIndexMinimal()` already uses; the `analyst_data/index` action harvests its filters (uuid included) from the POST body, so the receiving side is unchanged. Same GET-to-POST move as #5500.

Scope note: `fetchCollections()` carries the same GET-with-uuid shape and can hit 414 too, but its `collections/index` action reads the UUID list specifically from named params (there is a comment there on why a bare POST `uuid` filter would be ambiguous against the joined columns), so switching it needs a coordinated server-side change and is left out of this PR.

#### Questions

### DIFF
--- a/app/Lib/Tools/ServerSyncTool.php
+++ b/app/Lib/Tools/ServerSyncTool.php
@@ -265,14 +265,12 @@ class ServerSyncTool
             throw new RuntimeException("Remote server do not support analyst data");
         }
 
-        $params = [
-            'uuid' => $uuids,
-        ];
-
-        $url = '/analyst_data/index/' . $type;
-        $url .= $this->createParams($params);
-        $url .= '.json';
-        return $this->get($url);
+        // Send the UUID list in the POST body rather than the URL. As a GET
+        // every UUID rode in the request line as a named parameter, so a large
+        // batch overran the web server limit and the pull failed with 414. The
+        // analyst_data/index action already harvests its filters (uuid among
+        // them) from a POST body, the same shape fetchIndexMinimal() uses.
+        return $this->post('/analyst_data/index/' . $type, ['uuid' => $uuids]);
     }
 
     /**


iglocska:
Thanks, I have no idea why we didn't go for that in the first place, it was a terrible idea.

PR #11067, opened 2026-09-02
fix: [ui] focus URL lands on attributes outside the first page on large events (#10948)

### LINKED ISSUE #10948: Feature Request: Focus URL for large events

I daily generate URLs with the following format `https://<our MISP instance>/events/view/11/focus:9ae9bd81-10bc-4146-8feb-334667b29e67`. This brings up the event 11 and scrolls down to the attribute with the uuid requested.

This works really well for normal sized events.

But, for larger events where all attributes are not displayed on the first "page", this does not work at all.

Maybe this is known to someone but to me it is a problem. 

Is there a way to in the GET URL specify that all attributes should be expanded before applying the focus parameter?

_No response_

### Additional context

I run MISP 2.5.42

### Code of Conduct

skraft9:
## What does it do?

Fixes #10948.

On the event view, `focus:<attribute-uuid>` scrolls to the attribute once it is on the rendered page. On large events the target can sit on a later page and never render, so the scroll silently does nothing.

The pagination already jumps to the page of a focused top-level attribute, but it missed attributes nested inside an object, whose top-level list entry carries the object's uuid, not the attribute's. `rearrangeEventForView` now maps a focus that points at an object's attribute to that object's uuid, so the correct page renders and the existing client-side scroll lands on the attribute. It is a no-op when there is no focus or the focus is already a top-level entry, and it only reads the already ACL-filtered object list.

## Questions

### DIFF
--- a/app/Model/Event.php
+++ b/app/Model/Event.php
@@ -7551,6 +7551,24 @@ class Event extends AppModel
                 }
             }
         }
+        // The pagination tool jumps to the page holding the focused element by matching
+        // top-level uuids only. An attribute inside an object is not a top-level entry, so
+        // translate the focus to its containing object's uuid to land on the right page.
+        // The client-side scroll still targets the original attribute uuid.
+        if (!empty($passedArgs['focus'])) {
+            $focus = $passedArgs['focus'];
+            foreach ($objects as $object) {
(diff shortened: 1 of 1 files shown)


iglocska:
Thanks a lot!

PR #11072, opened 2026-09-02
fix: [UI] persist application log quick filter across pagination (#10687)

### LINKED ISSUE #10687: Bug: In the application logs - filter by something - go to next page - filter gone
### Actual behavior

In the /logs/index page you are able to use a filter. 
When you apply a filter, for example MISP update results, you basically use this link: 
`logs/index/filter:update`
If you click to go to the next page and use the next button at the bottom of the page, you always use this link: 
`logs/index/page:2 `
and your filter is removed. The bug is that you cannot go to the 2nd page of the filtered results. 

### Expected behavior

One would expect to see all the filtered results and be able to go to the 2nd page. 
Going to the next page should apply the same filter. 

### Steps to reproduce

Logs-> Application Logs
Apply one of the following filters, pick one that has many logs and will have a 2nd page
[Emails ](/logs/index/filter:email)
[Authentication issues ](/logs/index/filter:auth_issues)
[MISP Update results ](/logs/index/filter:update)
[Setting changes ](/logs/index/filter:settings)
[Warnings and errors](/logs/index/filter:errors)
Try to go to the  2nd page of the filtered results by clicking on the [next](/logs/index/page:2) page button

### Version

2.5.33

### Operating System

Ubuntu

### Operating System version

24.04

### PHP version

8.3

### Browser

Firefox

### Browser version

_No response_

### Relevant log output

```shell

```

### Extra attachments

_No response_

### Code of Conduct

skraft9:
#### What does it do?

Fixes #10687. The application-log quick filters (`filter:<name>`, e.g. `filter:email`) were applied to the query but not carried into the pagination and sort links, so paging to the next page fell back to unfiltered results. This adds `filter` to the Logs index table's `persistUrlParams`, so the active filter is preserved across page and sort navigation.

Verified on 2.5.44: with `filter:email` active, paging to page 2 and 3 keeps the filter and the results stay scoped to the filtered action.

#### Questions

### DIFF
--- a/app/View/Logs/index.ctp
+++ b/app/View/Logs/index.ctp
@@ -103,7 +103,9 @@
             ],
             'fields' => $fields,
             'title' => __('Application Logs'),
-            'persistUrlParams' => $paramArray
+            // Persist the active quick filter (filter:<name>) across pagination
+            // and sort links, otherwise the next page drops it (#10687).
+            'persistUrlParams' => array_merge($paramArray, ['filter'])
         ],
         'passedArgsArray' => $passedArgsArray,
     ]);


iglocska:
Thanks a lot, I keep forgetting about this!

PR #11073, opened 2026-09-02
fix: purgeByValue lost the leading-wildcard half of its OR

darkdi:
#### What does it do?

`purgeByValue()` builds its OR with two entries under the same array key, `CorrelationValue.value LIKE`, so PHP keeps only the second one and the query never looks for values that *end* with the searched string. Both correlation behaviours carry the identical block.

Wrapping each condition in its own array is the CakePHP way to OR two comparisons on one field, and it is the shape used elsewhere in the codebase, for example `$proposal_conditions['OR'][] = array(...)` in `Event.php`.

The visible effect is that purging a value leaves behind the correlation values that merely end with it, since only the `value%` half of the pattern ever reaches SQL.

#### Questions

Nothing open on my side. I did not touch the other duplicate-key spots I noticed while reading, `Server.php` at 1334 and 3207 and `EventReport.php` at 1763, since those want their own look and this is one fix per PR.


iglocska:
Nice catch, thanks!

PR #11076, opened 2026-09-02
[medium] perf: avoid Router::url route walk in Events/View/attribute_correlations.ctp link building

elhoim:
**BLUF —** Correlation links use `$baseurl` concatenation instead of a `Router::url` route walk per row.

- **Problem** — `app/View/Elements/Events/View/attribute_correlations.ctp` is rendered once per attribute row on the event view, and both anchors it emits were built with `$this->Html->link()` using an array URL, which sends every href through `Router::url` and its full route-table walk.
- **Fix** — Replaces those two call sites with plain `$baseurl` string concatenation, the same idiom already used for the identical targets in `row_attribute.ctp` and `PivotHelper.php`, leaving the emitted markup unchanged.
- **Effect** — The magnifier link at line 74 is ungated and renders for every attribute row whether or not it correlates, so events with many attributes drop a proportional amount of routing work per page render.

## What

`app/View/Elements/Events/View/attribute_correlations.ctp` is rendered once per attribute row on the event view. Both of the `<a>` tags it emits were built through `$this->Html->link()` with an **array** url, which routes every single href through `Router::url` and its full route-table walk. This PR replaces those two calls with `$baseurl` string concatenation — the idiom the codebase already uses for the *same* targets on the *same* page at `app/View/Elements/Events/View/row_attribute.ctp:104` / `:335` and `app/View/Helper/PivotHelper.php:54` / `:56`. The emitted markup is otherwise unchanged.

Two sites change:

- `:36` — the correlation link, once per correlated event (the `foreach` at `:16`), becomes `$baseurl . '/events/view/' . h($relevantId)`, plus `'/1/' . h($event['Event']['id'])` in the `$withPivot` case.
- `:74` — the ungated search magnifier, once per rendered attribute row, becomes `$baseurl . '/attributes/search/value:' . h(rawurlencode($object['value']))`.

## Why it is hot

Tier T1 **by granularity** (per attribute rendered), and I want to state the volume honestly rather than oversell it.

- `app/View/Elements/eventattribute.ctp:148` renders `'/Events/View/row_' . $object['objectType']` for every attribute — that is what `EventsController::viewEventAttributes` (`app/Controller/EventsController.php:1293`, `:1487`) and the embedded event view produce.
- `app/View/Elements/Events/View/row_attribute.ctp:210` then renders `Events/View/attribute_correlations` for **every** attribute, with no `'cache'` option.
- The magnifier at `:74` is **ungated** — it sits after the `if (!empty($event['Related'.$scope][$object['id']]))` block that closes at `:73`, so it renders for every attribute row whether or not the attribute correlates. The correlation link at `:36` is gated on the attribute actually correlating (`:2`).
- The element's other two callers are `row_proposal.ctp:108` and `genericElements/IndexTable/Fields/relatedEvents.ctp:15`.

**Gate, plainly:** no non-default setting is involved — this is default behaviour on the single most-used MISP page. But it is per *rendered* attribute, not per attribute in the DB: pagination caps a page at 60 rows (`app/Lib/Tools/CustomPaginationTool.php:10`, `:31`), unbounded only when `page == 0` sets `limit = count` (`:36-37`). So a page render costs on the order of 60–180 `Router::url` array calls, not millions, and this path never fires on API/export/sync. It is a cheap correct fix, not a page-level transformation.

## Mechanism

`HtmlHelper::link` (`app/Lib/cakephp/lib/Cake/View/Helper/HtmlHelper.php:349-352`) calls `$this->url()`, which MISP overrides at `app/View/Helper/AppHelper.php:35-41` only to inject `$url['admin'] = false` before delegating to `Helper::url` (`app/Lib/cakephp/lib/Cake/View/Helper.php:297-299`) = `h(Router::url($url))`.

`Router::url` (`app/Lib/cakephp/lib/Cake/Routing/Router.php:829-923`) takes its array branch: the prologue at `:860-899` rebuilds `$params` from the request stack, normalises the `'?'`, `'#'`, `'ext'` and prefix keys, unions in controller/plugin defaults — then loops **unconditionally** over `static::$routes` at `:903-914`, calling `CakeRoute::persistParams` (`CakeRoute.php:376-386`) and `CakeRoute::match` (`CakeRoute.php:398-478`) on each route until one matches, finally `_writeUrl` (`CakeRoute.php:489+`) on the winner.

`CakeRoute::match` is not cheap per route: `array_flip($this->keys)` + `array_intersect_key` + comparison (`:406-409`), `array_diff_key` on defaults (`:412-415`), `Router::namedConfig()` and `Router::prefixes()` (`:417-420`), a `foreach` over every key of `$url` (`:424-462`), then the options `preg_match` loop (`:468-476`).

The whole table is walked, because `/events/view/<id>` matches nothing until the very last route:

- the 8 admin paginator routes (`app/Config/routes.php:34-41`) die at `CakeRoute.php:413` — their defaults carry `'admin'`, which `AppHelper::url` has just unset;
- `mapResources`' `/events` dies at `:424-429` on `action index != view`;
- `mapResources`' `/events/:id` dies at `:407-409` — `array_intersect_key` cannot find the routed key `id`, because the url carries the id as positional element `0`;
- Cake's `/:controller` (`Cake/Config/routes.php:72`) dies on the action;
- `/:plugin/:controller/:action/*` survives to the options `preg_match` at `:469-475` and fails on the null plugin;
- the match finally lands on `/:controller/:action/*`, `Cake/Config/routes.php:73` — the **last** route.

Route table size: 11 `Router::connect` in `app/Config/routes.php:28-41`, plus 12 from `Router::mapResources(array('events','attributes'))` at `routes.php:45` (6 entries of `Router::$_resourceMap`, `Router.php:167-174`), plus 10 from Cake's own `Config/routes.php:57-73` under `Routing.prefixes = array('admin')` (`app/Config/core.default.php:117`). That is **≥33**, and higher in practice: `app/Config/bootstrap.default.php:132-134` loads SysLog / Assets / SysLogLogable unconditionally, so `CakePlugin::routes()` and the plugin block at `Cake/Config/routes.php:45-64` add ~6 more.

For the magnifier, `app/Config/routes.php:48` calls `connectNamed` with an array, so `greedy` stays `true` (`Router.php:484`) and the `'value'` key is accepted as a named param — the match still lands on the last route, emitting `/attributes/search/value:<rawurlencode>`.

The only memoisation on this path is the **per-route compiled regex**: `CakeRoute::compiled()` (`CakeRoute.php:102-104`) returns early on `$this->_compiledRoute`, `compile()` (`:114-120`) builds it once, and `match()` (`:399-401`) compiles at most once per route object. That removes the regex *construction* and nothing else — not the loop, not `persistParams`, not the `array_flip` / `array_intersect_key` / `array_diff_key` work at `CakeRoute.php:407-415`, not the `Router::namedConfig()` / `Router::prefixes()` calls at `:418-420`, and not `_writeUrl`.

`Router::url` itself has **no result cache**. Reading `Router.php:829-919` end to end: no `Cache::read`, no Redis, no static memo. The only early-out in the whole function is the string-url branch at `:920-923` (one `preg_match` on the scheme, then a `substr`), which an array url never reaches.

The `View::element` cache does not help either: the element is rendered with no `'cache'` option (`row_attribute.ctp:210`), so `View::_elementCache` (`View/View.php:409-414`) never fires.

And nothing is being cached upstream — the codebase already builds these exact hrefs by hand elsewhere on the same page (`row_attribute.ctp:104`, `:335`, `PivotHelper.php:54`/`:56`), which is what this PR aligns with.

## Speedup

**MEASURED: 9.68x per link** (80.10 µs → 8.27 µs).

Benchmark method — the "original" side is **not** a hand-mirrored strawman, it is the real CakePHP stack. The script bootstraps Cake from `app/Lib/cakephp/lib` with a scratch APP carrying MISP's own `Config/core.php` (copied from `core.default.php`, so `Routing.prefixes = array('admin')`), MISP's own `Config/routes.php` and MISP's own `View/Helper/AppHelper.php`, then instantiates a real `Controller`/`View`/`HtmlHelper` and calls `$Html->link()` with the exact array arguments the unpatched template used. The "patched" side is a verbatim copy of the expressions now in the template. Timing shape: 60 attribute rows × (1 magnifier + 2 correlation links) = 180 links per page render; 20 warmup renders, then 300 timed renders per side, with Router state static so neither side pays per-iteration setup.

Raw output:

```
route table size: 26
correctness cases: 6500
correctness mismatches: 0
orig   : 4.3256s for 300 page-renders (180 links each) = 80.10 us/link
patched: 0.4467s for 300 page-renders (180 links each) = 8.27 us/link
SPEEDUP: 9.68x
RESULT: OK
```

(An earlier run of the same script with a smaller correctness corpus: 1780 cases, 0 mismatches, 71.27 → 6.97 µs/link, 10.23x. A second independent rerun of the same script gave 9.04x.)

**Correctness assertion: 6500 inputs, 0 mismatches**, comparing the *full rendered `<a>` markup byte-for-byte* against the real Router output. 300 ids (ints and digit-strings, as the DB yields) × {withPivot on/off} × {same-org/other-org} × {red/white} for the correlation link, plus 100 repetitions over a 41-value attribute-value corpus for the magnifier: IPv4/IPv6, md5/sha256, urls with `?&=#`, spaces, backslash paths, `a/b/c`, `<script>alert(1)</script>`, `">`, single quotes, `&`, `=`, `:`, `;`, `|`, `%20`, `100%`, `%2F`, `~-_.`, `+`, tab, newline, latin-1 accents, CJK, emoji, empty string, `'0'`, whitespace-only, a 4KB value and a ~4.6KB long value with spaces.

Two normalisations, disclosed so the comparison is honest: `$request->base = ''` and `$baseurl = ''`, so Router's base prefix and the concat's base prefix are byte-comparable; and the bench's route table is **26** because no plugins are loaded, versus **≥33** in production. Fewer routes means a shorter walk, so **9.68x is a floor** — production is worse for the original side, not better. The 4KB/4.6KB values in the timing corpus also inflate the patched side's per-link cost (`rawurlencode`/`h` over 4KB dominates a `sprintf`), so the figure is conservative in that direction too; an isolated measurement on realistic attribute values gave 44x for the correlation link and 56x for the magnifier.

**DERIVED** (the pre-benchmark estimate, for reference): ~455 ops per `Html->link` with an array url, of which ~420 are the route walk (33 iterations × a conservative ~12 ops average, most routes bailing early), versus ~25 ops for the concat — 18x nominal, discounted hard to a ≥5x honest floor. The measured 9.68x sits between the discounted floor and the nominal ratio, as expected.

**Scope of the win, stated plainly:** 9.68x is per **link**, not per page. `View::element` path resolution and `_render`/`_evaluate` overhead are untouched, so a whole event-view render improves by low single-digit percent — a 60-row page saves ~180 route walks, real but a few milliseconds against a page costing hundreds. No page-level claim is intended.

## Risk

- **Named-param reproduction (the finding's stated trap).** The magnifier carries a named param whose value is arbitrary attribute content, so the replacement must reproduce `CakeRoute::_writeUrl` exactly — `$key . ':' . rawurlencode($value)` (`CakeRoute.php:507-521`, separator from `Router::$_namedConfig:151`). It does. The outer `h()` is a byte-identical no-op in practice, because `rawurlencode`'s output alphabet (`A-Za-z0-9-_.~` plus `%XX`) contains no HTML-special character — it is kept anyway so the XSS-safety argument does not rest on that reasoning. Verified byte-identical over the hostile corpus above, including `<script>` and `">`.
- **Pivot variant.** The original finding's text suggested `'/true/'`; that is wrong. `CakeRoute::_writeUrl:500-502` does `rawurlencode((string)$param)` over pass args and `(string)true === '1'`, so the real href is `/events/view/<id>/1/<eventId>`. `'/1/'` is implemented, matching what `PivotHelper.php:56` already builds by hand.
- **Intended, disclosed non-identity.** `Router::url` prepends `$request->base` (a path, usually `''`), while `$baseurl` is `h(Configure::read('MISP.baseurl'))` set at `AppController.php:1061`. On a default instance the href therefore changes from `/events/view/7` to `https://host/events/view/7`. The navigation target is identical, and this is exactly the href `row_attribute.ctp:104` / `:335` and `PivotHelper.php:54`/`:56` already emit for the same targets on the same page — no new base-path assumption is introduced. The only combination that would break — `MISP.baseurl` empty **and** a subdirectory install, where Router emits `/misp/...` and the concat emits `/...` — already breaks those sibling links today, so no new failure mode.
- **`$baseurl` availability.** Guaranteed: `AppController::_setupBaseurl()` is called from `beforeFilter` (`AppController.php:128`) and sets it via `$this->set('baseurl', h($baseurl))`; `View::_renderElement` merges viewVars into element data (`View/View.php:1224`), so all three callers see it.
- **Divergences found by a separate edge-case harness, both unreachable with real data.** (a) magnifier with `$object['value'] === null`/`false`: `CakeRoute::match` drops the named param entirely (`/attributes/search`) while the patch emits `/attributes/search/value:`. Unreachable — `attributes.value1`/`value2` are `NOT NULL` (`INSTALL/MYSQL.sql:104`), the composed `value` is always a string, and all three callers pass a full attribute/shadow-attribute row; the unpatched code already dereferenced `$object['value']` unconditionally, so a caller lacking it would emit a notice today. No guard added. (b) correlation link with a non-numeric `$relevantId` containing url-special characters: the original `rawurlencode`s, the patch only `h()`s. Unreachable — `$relevantId` is `$relatedAttribute['Event']['id']` or `['id']`, a DB integer. Not an XSS surface even hypothetically: `h()` is `ENT_QUOTES`, so attribute breakout is impossible (`a"b` → `href="/events/view/a&quot;b"`).
- **Deliberately narrower than the original suggestion.** `value_field.ctp` is **not** touched. `:95` (cveurl), `:105` (cweurl) and `:117` (link) pass **string** urls that short-circuit at `Router.php:920-923` after one `preg_match` — patching them buys nothing. `:129` passes arbitrary attribute text as a **positional** pass arg, where Router applies `rawurlencode` but the codebase idiom applies `h()`; those are not equivalent for non-alphanumeric input, so a "same treatment" patch there would be a real behaviour change. `:76` (attachment download) is an array url but is gated on attribute type and carries several params; left alone for minimality. This keeps the diff to two hunks, 10 insertions / 14 deletions, one file, no reformatting.

## Testing

- **Lint:** `php -l` on the patched template — `No syntax errors detected`.
- **Benchmark:** the script described above, run under a lock so no concurrent work polluted the timings. 9.68x; a second independent rerun of the same script gave 9.04x.
- **Correctness assertion:** 6500 inputs, **0 mismatches**, full `<a>` markup compared byte-for-byte against real `Router::url` output — including a hostile attribute-value corpus and 4KB values.
- **Reviewed:** the mechanism chain (`HtmlHelper::link` → `AppHelper::url` → `Helper::url` → `Router::url` → `CakeRoute::match`/`_writeUrl`) was re-read end to end in a second pass, and the bench's scratch `Config/routes.php`, `Config/core.php` and `View/Helper/AppHelper.php` were diffed byte-for-byte against MISP's real `routes.php`, `core.default.php` and `AppHelper.php`.
- **Not tested:** **no live MISP instance was available**, so there is no end-to-end verification — no rendered event view was clicked through in a browser, and no functional/integration suite was run against a running server. The equivalence argument rests entirely on the byte-for-byte markup comparison against the real Router and on the source reading above. A reviewer with an instance should sanity-check one event view with correlations, one with the pivot active, and one attribute whose value contains url-special characters.

### DIFF
--- a/app/View/Elements/Events/View/attribute_correlations.ctp
+++ b/app/View/Elements/Events/View/attribute_correlations.ctp
@@ -33,12 +33,11 @@
                 $popover .= '<b class="black">' . h($k) . '</b>: <span class="blue">' . h($v) . '</span><br>';
             }
             $relevantId = !isset($relatedAttribute['attribute_id']) ? $relatedAttribute['Event']['id'] : $relatedAttribute['id'];
-            $link = $this->Html->link(
-                $relevantId,
-                    $withPivot ?
-                            ['controller' => 'events', 'action' => 'view', $relevantId, true, $event['Event']['id']] :
-                            ['controller' => 'events', 'action' => 'view', $relevantId],
-                ['class' => ($relatedAttribute['org_id'] == $me['org_id']) ? $linkColour : 'blue']
+            $link = sprintf(
+                '<a href="%s" class="%s">%s</a>',
+                $baseurl . '/events/view/' . h($relevantId) . ($withPivot ? '/1/' . h($event['Event']['id']) : ''),
+                ($relatedAttribute['org_id'] == $me['org_id']) ? $linkColour : 'blue',
+                h($relevantId)
             );
             echo sprintf(
                 '<li class="no-side-padding %s" %s data-toggle="popover" data-content="%s" data-trigger="hover">%s&nbsp;</li>',
@@ -71,12 +70,9 @@
             __('Too many correlations.')
         );
     }
-    echo $this->Html->link(
-        '',
-        ['controller' => 'attributes', 'action' => 'search', 'value' => $object['value']],
-        [
-            'class' => 'fa fa-search black',
-            'title' => __('Search for value'),
-            'aria-label' => __('Search for value')
-        ]
+    echo sprintf(
+        '<a href="%s" class="fa fa-search black" title="%s" aria-label="%s"></a>',
+        $baseurl . '/attributes/search/value:' . h(rawurlencode($object['value'])),
+        h(__('Search for value')),
+        h(__('Search for value'))
     );
--- a/app/View/Elements/Events/View/attribute_correlations.ctp
+++ b/app/View/Elements/Events/View/attribute_correlations.ctp
@@ -35,7 +35,7 @@
             $relevantId = !isset($relatedAttribute['attribute_id']) ? $relatedAttribute['Event']['id'] : $relatedAttribute['id'];
             $link = sprintf(
                 '<a href="%s" class="%s">%s</a>',
-                $baseurl . '/events/view/' . h($relevantId) . ($withPivot ? '/1/' . h($event['Event']['id']) : ''),
+                h($baseurl) . '/events/view/' . h($relevantId) . ($withPivot ? '/1/' . h($event['Event']['id']) : ''),
                 ($relatedAttribute['org_id'] == $me['org_id']) ? $linkColour : 'blue',
                 h($relevantId)
             );
@@ -70,9 +70,10 @@
             __('Too many correlations.')
         );
     }
+    $searchTitle = __('Search for value');
     echo sprintf(
         '<a href="%s" class="fa fa-search black" title="%s" aria-label="%s"></a>',
-        $baseurl . '/attributes/search/value:' . h(rawurlencode($object['value'])),
-        h(__('Search for value')),
-        h(__('Search for value'))
+        h($baseurl) . '/attributes/search/value:' . rawurlencode($object['value']),
+        $searchTitle,
+        $searchTitle
     );


app/View/Elements/Events/View/attribute_correlations.ctp
```diff
@@ -71,12 +70,9 @@
...
-        ['controller' => 'attributes', 'action' => 'search', 'value' => $object['value']],
-        [
-            'class' => 'fa fa-search black',
-            'title' => __('Search for value'),
-            'aria-label' => __('Search for value')
-        ]
+    echo sprintf(
+        '<a href="%s" class="fa fa-search black" title="%s" aria-label="%s"></a>',
+        $baseurl . '/attributes/search/value:' . h(rawurlencode($object['value'])),
+        h(__('Search for value')),
```
iglocska: Calling h() on static strings (even when translated and taken from the pot) is a waste.

app/View/Elements/Events/View/attribute_correlations.ctp
```diff
@@ -33,12 +33,11 @@
...
             $relevantId = !isset($relatedAttribute['attribute_id']) ? $relatedAttribute['Event']['id'] : $relatedAttribute['id'];
-            $link = $this->Html->link(
-                $relevantId,
-                    $withPivot ?
-                            ['controller' => 'events', 'action' => 'view', $relevantId, true, $event['Event']['id']] :
-                            ['controller' => 'events', 'action' => 'view', $relevantId],
-                ['class' => ($relatedAttribute['org_id'] == $me['org_id']) ? $linkColour : 'blue']
+            $link = sprintf(
+                '<a href="%s" class="%s">%s</a>',
+                $baseurl . '/events/view/' . h($relevantId) . ($withPivot ? '/1/' . h($event['Event']['id']) : ''),
```
iglocska: I am not 100% sure, can't recall it accurately, but I think h($baseurl) isn't a bad idea in case of a compromised admin account changes it, but this is a highly unlikely xss angle. Either way, link() had this baked in.

app/View/Elements/Events/View/attribute_correlations.ctp
elhoim: Agreed, fixed in 7e35ae81e2.

Worse than a single wasted call, actually — it was translated *and* escaped twice, once for `title` and once for `aria-label`. Now hoisted into `$searchTitle` and used unescaped.

The codebase backs you up: `.ctp` views contain 14960 plain `__()` calls against 181 wrapped in `h()`, so `h(__(...))` was against house style as well as wasteful.

I also dropped the `h()` around `rawurlencode($object['value'])` in the same line for the same reason — `rawurlencode` only ever emits URL-safe characters, so escaping its output is a guaranteed no-op.

app/View/Elements/Events/View/attribute_correlations.ctp
elhoim: You are right, and I think the last sentence is the important half: `link()` had this baked in. Restored in 7e35ae81e2 on both links.

I agree the XSS angle is thin — `$baseurl` is admin-set config, so reaching it means the instance is already owned. But the regression is real regardless of exploitability: this PR advertises itself as a pure performance change and it silently dropped an escaping guarantee the helper was providing. That is not a trade worth making for one `h()` call per link, especially against the `Router::url` route walk being removed. The performance claim is unaffected.

While here, one thing I should have flagged in the description rather than leaving implicit, since it is the real fragility in this approach. The second link hand-builds a **named-parameter** URL:

```php
h($baseurl) . '/attributes/search/value:' . rawurlencode($object['value'])
```

The original handed Cake `['controller' => 'attributes', 'action' => 'search', 'value' => $object['value']]` and let the Router decide the shape. Hardcoding `value:` bakes in both the current route and named-parameter syntax, so if either changes this link breaks silently — wrong URL, no error. The first link (`/events/view/...`) is a plain default route and is far safer. Happy to revert the second one to `Html->link()` and keep the optimisation only on the first, which is the one inside the per-correlation loop and therefore where the cost actually is — say the word.

PR #11103, opened 2026-09-10
fix: [eventReport] correctly allow pasting of pictures from clipboard…

Wachizungu:
… into event reports, as non site-admin

#### What does it do?

Fixes a regression issue introduced in https://github.com/MISP/MISP/commit/ea0e1e8227535201d7e8b60855de222b09f14c37 

Makes it possible for non site-admins to paste images into event reports again.

image

#### Questions


iglocska:
Thank you!

PR #11109, opened 2026-09-11
fix: [sync] fix custom galaxy cluster sync on Server:push.

Wachizungu:
Note: sync after event and cluster publication actions was working. It's only the Server:push function that was affected. This means this issue only popped up if initial sync didn't work for some reason (for example due to missing connection at time of publication).

Referring to https://github.com/MISP/MISP/pull/10991 for initial report. It required some extra modification to fully fix and test, hence the new PR. Sorry @elhoim !

#### Questions


iglocska:
Thanks a lot to you both!

PR #11110, opened 2026-09-11
fix: [UI] correct broken sync buttons on servers index (default theme)

Wachizungu:
#### What does it do?

Fixes regression from https://github.com/MISP/MISP/commit/6a991702f71a3cac76341b4923577c54a0190cc0

Buttons on the index page didn't have the correct extra params anymore. So clicking the push button ended up giving a 500 for example (due to missing /full).

#### Questions


iglocska:
Thanks a lot!

PR #11132, opened 2026-09-16
[medium] chg: [galaxy] share the edit-galaxies action across two controllers

elhoim:
## BLUF

- **Priority: medium.**
- `EventsController::editEventGalaxies()` and `AttributesController::editAttributeGalaxies()` were the same ~160-line action, duplicated down to the inline comments.
- The copies had already silently diverged: the attribute response sends `check_publish`, the event response does not.
- The shared body now lives once, as `AppController::__editGalaxiesForScope()`, parameterised by target scope.
- Both actions shrink to fetch + ACL + one delegating call; request/response behaviour is unchanged for both scopes.
- Pure move, no behaviour change, no DB change, no API change. Verified with `php -l` and a diff review.

#### What does it do?

**What was wrong.** `app/Controller/EventsController.php:3130-3294` and `app/Controller/AttributesController.php:3142-3306` held the same controller action twice: the same galaxy-tag harvesting, the same `GalaxyCluster::getClustersByTags()` lookup and `array_column()` indexing, the same entry building with `GalaxyColour::hue()`, the same POST block diffing `global_ids`/`local_ids` into add/remove sets, the same "detach before attaching" workaround, and the same GET tail building `$galaxyList`. The inline comments were duplicated too.

**Concrete impact.** Every change to galaxy editing had to be made twice, and a change applied to only one side would go unnoticed. That has already happened: the attribute POST success payload carries `'check_publish' => true` (which makes `handleGenericAjaxResponse()` in `app/webroot/js/misp.js` call `checkAndSetPublishedInfo()`), while the event payload does not.

**What this changes.**

- Adds `protected function __editGalaxiesForScope($scope, $scopeId, array $tags, $mayModify, $invalidMessage, array $extraSuccessFields = [])` to `AppController`, next to the existing `__canModifyEvent()` / `__canModifyTag()` helpers. Its body is the event copy verbatim, with the scope string, the record id, the tag associations, the "invalid target" message and any extra success-payload keys injected as parameters.
- `editEventGalaxies()` keeps `fetchSimpleEvent()`, the `NotFoundException`, `__canModifyEvent()` and `$this->set('eventId', ...)`, then delegates.
- `editAttributeGalaxies()` keeps its `$id === null` guard, `fetchAttributeSimple()`, the `NotFoundException`, `__canModifyTag()` and `$this->set('attributeId', ...)`, then delegates, passing `['check_publish' => true]`.
- `App::uses('GalaxyColour', 'Tools')` moves to `AppController` (it is now used there); it is dropped from `AttributesController`, where it became unused, and kept in `EventsController`, which still uses it in `searchGalaxyClusters()`.

**Deliberately preserved, not changed.**

- The `check_publish` asymmetry stays as it is today, carried through `$extraSuccessFields`. Whether the event scope should send it too is a separate question and a separate PR.
- The attributes-only `$id === null` guard stays; events continue to rely on `fetchSimpleEvent()` returning empty.
- The two ACL predicates stay different (`__canModifyEvent()` vs `__canModifyTag()`).
- No ACL surface change: `ACLComponent.php` keys on the action names (lines 71 and 369), which are unchanged, and the helper is `protected`, so it is not routable and does not appear in `queryACL()`.
- The Overmind templates (`Events/edit_event_galaxies.ctp`, `Attributes/edit_attribute_galaxies.ctp`) are untouched — the view variables they read are identical. This keeps out of the way of the Overmind WIP branch.

**How it was verified.**

- `php -l` clean on all three modified files.
- The patch was applied mechanically to fresh copies of the current default branch, each replacement asserted to match exactly once.
- Diff review confirms the moved body is byte-identical to the event copy apart from the parameterised names, so both scopes produce the same JSON on POST and the same five view variables (`currentGlobalClusters`, `currentLocalClusters`, `galaxyList`, the scope id, `mayModify`) on GET.
- No unit test is included: `app/Test/` is pure PHPUnit with no CakePHP bootstrap and no database, and exercising this helper would need the full stack (Controller base, `loadModel()`, Auth, request, two models, `CakeResponse`). A live instance should confirm that both modals still open, list the current clusters, and save.

Local check:

```
php -l app/Controller/AppController.php && \
php -l app/Controller/EventsController.php && \
php -l app/Controller/AttributesController.php && \
grep -c __editGalaxiesForScope app/Controller/AppController.php \
    app/Controller/EventsController.php app/Controller/AttributesController.php
```

Expected: 1 definition in `AppController`, 1 call in `EventsController`, 2 hits in `AttributesController` (the call plus the docblock reference).

#### Questions

- [ ] Does it require a DB change? — No.
- [ ] Are you using it in production? — No, not yet.
- [ ] Does it require a change in the API (PyMISP for example)? — No. Both endpoints keep their URLs, parameters and JSON payloads.


elhoim:
Superseded by #11147.

The Overmind work merged in #11118 rewrote both actions (Bootstrap 5 modal, TomSelect pickers, `fetchAttributeSimple` + `__idToConditions` on the attribute side), so this diff no longer applies in any useful form — rebasing it would be a rewrite anyway.

The premise held up, and got stronger: the two copies have now drifted a *third* way. `AppController::_clustersByTagId()` was extracted and only the attribute copy was updated to use it — the event copy still inlines its own `getClustersByTags()` + `array_column()` harvesting. That is precisely the failure mode this PR was opened about. The event docblock has also gone stale (it advertises `cluster_ids`, both bodies read `global_ids`/`local_ids`).

#11147 is the same change written against current `2.5`, and fixes both of those as a side effect of sharing the body. Closing this one.

PR #11133, opened 2026-09-16
[medium] chg: [acl] extract the duplicated distribution ACL block into AppModel

elhoim:
## BLUF

- **Priority: medium.**
- The owning-organisation subquery plus the distribution-level ACL `OR` block is written out inline **five times** — `Event::fetchPaginatedAttributes()`, `Event::fetchPaginatedObjects()` (twice), `Event::fetchEvent()` (three model aliases in one block) and `MispObject::fetchObjects()`.
- The `MispObject` copy has **no PostgreSQL branch**, so on PostgreSQL it emits an unquoted, unqualified `events` subquery where every `Event` copy emits the schema-qualified quoted form.
- This PR adds `AppModel::distributionAclConditions($alias, $user, $sgids)` and calls it from all five sites.
- The six `Event` copies are reproduced **element for element on both drivers** — no behaviour change there.
- The `MispObject` copy is folded in too: same rule, differently encoded, and it gains the PostgreSQL branch it was missing.
- No DB change, no API change, no PyMISP change, no config change.

---

#### What does it do?

**What was wrong**

`(SELECT events.org_id FROM events WHERE events.id = <Alias>.event_id)`, combined with the "distributed, or in one of my sharing groups, or my org owns the event" `OR` block, is the core sharing ACL of the application. It is currently spelled out at:

| File | Method |
| --- | --- |
| `app/Model/Event.php` | `fetchPaginatedAttributes()` |
| `app/Model/Event.php` | `fetchPaginatedObjects()` — objects |
| `app/Model/Event.php` | `fetchPaginatedObjects()` — contained attributes |
| `app/Model/Event.php` | `fetchEvent()` — `Attribute`, `Object`, `EventReport` |
| `app/Model/MispObject.php` | `fetchObjects()` |

The four `Event.php` blocks are identical modulo the model alias. The `MispObject` block is the same rule written differently: it encodes the distribution test as `distribution IN (1, 2, 3, 5)` instead of `distribution > 0 AND distribution != 4`, nests the `OR` one level deeper, and — the part that actually matters — never calls `isMysql()`, so it has no PostgreSQL branch.

**Concrete impact**

Six copies of a security-relevant condition that must be changed in lockstep, and one of them that already drifted. `$attributeConditions` in `fetchObjects()` is used on both branches of the method (the `checkDbSupport('reverseJoin')` params and the plain `else` params), so it is evaluated whatever the driver. In practice the unqualified subquery still resolves on PostgreSQL under a default `search_path`; it diverges from the `Event` copies when the MISP schema is not on the search path. So the practical blast radius is narrow, but object attribute visibility is genuinely built by different SQL from event attribute visibility on PostgreSQL, and that is not something this condition should be allowed to do.

**What this changes**

- `app/Model/AppModel.php` — new `protected function distributionAclConditions($alias, array $user, array $sgids)` placed directly after `isMysql()`, its only dependency. It returns the three `OR` alternatives in the existing order (distribution range, sharing-group match, owning org), with the owning-org subquery in whichever form the driver needs.
- `app/Model/Event.php` — the six inline copies become six calls. Every `SharingGroup->authorizedIds()` call stays exactly where it was (in `fetchPaginatedObjects()` it is computed outside the `!$isSiteAdmin` guard and reused by the second block — unchanged).
- `app/Model/MispObject.php` — the fifth copy becomes `'OR' => $this->distributionAclConditions('Attribute', $user, $sgids)`.

Net: **-96 lines, +32**, one implementation.

**Deliberately not touched**

- `Event::fetchEvent()`'s soft-delete subquery (`sprintf('(SELECT events.org_id …)', $softDeletable)`, under `if (!$user['Role']['perm_sync'])`) and the `ShadowAttribute` proposal subquery. Both are org-only conditions with no distribution block — not instances of this duplication. Routing them through the helper would give them the quoted PostgreSQL form and change generated SQL, which is out of scope for a cleanup.
- `MispObject::buildConditions()`, which uses the joined `Event.org_id` directly rather than a subquery.
- The `distribution IN (1, 2, 3, 5)` encoding in `ShadowAttribute`, `EventReport` and `MispAttribute`. This PR normalises the one site the duplication covers; it does not normalise the codebase.

**How it was verified**

- `php -l` clean on all three changed files.
- A standalone PHP probe loads the patched `AppModel` under minimal stubs and asserts with `===` that the helper reproduces the pre-refactor condition array for `Attribute`, `Object` and `EventReport`, on the MySQL branch and on the PostgreSQL branch — element order included, so CakePHP builds identical SQL. All six pass; the same probe fails with "undefined method" before the patch. The probe is in the PR discussion below and needs nothing but `php`.
- The one intentional change, `MispObject`: `distribution IN (1, 2, 3, 5)` and `distribution > 0 AND distribution != 4` select the same rows over MISP's distribution domain 0-5, and flattening `OR(org, OR(dist, sg))` into `OR(dist, sg, org)` is a no-op. On PostgreSQL that site now emits the quoted schema-qualified subquery, matching the `Event` copies.

**On tests**

No new file under `app/Test/`. That suite runs in a single PHPUnit process and several existing tests (`CollectionCaptureTest`, `UserCanSeeEmailsTest`, …) declare their own stub `class AppModel`; a test that loads the real `app/Model/AppModel.php` would either fatal on the duplicate class name or, if it won the race, suppress those stubs and break the tests that rely on them. The verification probe above is that assertion run standalone. Happy to land it as a test if the maintainers would rather introduce per-test process isolation.

#### Questions

- **Does it require a DB change?** No.
- **Are you currently using it in production?** No — reviewed against `2.5` at `0c9f682`.
- **How can we test it?** Run the probe above, or exercise the event view, the paginated attribute/object views and `restSearch` as a non-site-admin user of a non-owning organisation and confirm visibility is unchanged. On PostgreSQL, additionally confirm `MispObject::fetchObjects()` returns the same contained attributes as before.
- **Does it change the API / PyMISP?** No.


elhoim:
Closing: superseded upstream.

Commit `acf97eac` ("fix: [db] Spell the correlated owner subqueries so both engines quote them", 2026-09-15) introduced `AppModel::correlatedLookup()` and `Event::eventOwnerSubquery()`, which already land the part of this PR that mattered:

- All six `Event.php` sites now build the owning-org subquery through `eventOwnerSubquery($alias)`.
- `MispObject::fetchObjects()` now calls `correlatedLookup('events', 'org_id', 'Attribute', 'event_id')`, so the PostgreSQL divergence this PR was opened for no longer exists.
- Upstream also routed the soft-delete subquery (`Event.php:4376`) and the ShadowAttribute proposal subquery (`Event.php:4404`) through the same helper — two sites this PR deliberately left alone.

What remains duplicated is only the distribution `OR` block itself (six near-identical copies in `Event.php`, plus the `distribution IN (1, 2, 3, 5)` encoding in `MispObject`). That is a readability matter with no correctness or driver-divergence argument behind it, so it does not justify rebasing this diff. If it is wanted, it belongs in a separate, much smaller PR.

Thanks — the underlying issue is fixed.

PR #11141, opened 2026-09-16
[medium] perf: [correlation] count over-correlation occurrences one chunk of values per query

elhoim:
## BLUF

- **Priority: medium** — performance only, no behaviour change, no DB change, no API change.
- `OverCorrelatingValue::generateOccurrences()` issues **one COUNT query per row** of `over_correlating_values`, each carrying two `LIKE` predicates and a join on `events`.
- On an instance with N blocked values that is N passes over `attributes` — the *Generate occurrences* button and the `jobGenerateOccurrences` worker scale linearly with a table that grows on its own.
- This change sums **one match expression per value over a single pass, 100 values per statement** — 100× fewer statements for the same work.
- The matching is untouched: the summed expression is the exact condition the per-value COUNT queries were built from.
- Adds a pure-PHPUnit regression test (no DB, no CakePHP bootstrap) that fails on 2.5 and passes with the change.

#### What does it do?

**What was wrong**

`app/Model/OverCorrelatingValue.php:140` loops over every row returned by an unbounded `find('all')` and runs one `find('count')` per row:

```php
foreach ($overCorrelations as &$overCorrelation) {
    $value = $overCorrelation['OverCorrelatingValue']['value'] . '%';
    $count = $this->Attribute->find('count', [ ... 'Attribute.value1 LIKE' => $value ... ]);
}
```

Each of those counts joins `events` and ORs two `LIKE '<value>%'` predicates, which the optimiser cannot satisfy from the `value1`/`value2` prefix indexes when combined that way. Row-by-row counting where one aggregation would do.

**Concrete impact**

`generateOccurrences()` is live, not dead code: it is reachable from `CorrelationsController::generateOccurrences()` (the button in `Correlations/over_correlations.ctp` and its Overmind copy) and from `AdminShell::jobGenerateOccurrences()`. `over_correlating_values` holds one row per value that exceeded `MISP.correlation_limit`, so on a busy instance it holds thousands to tens of thousands of rows — that many separate scans of `attributes` per run of the job.

**What this changes**

- The values are processed in chunks of 100. For each chunk one statement is sent, selecting `SUM(<match expression>) AS occurrence_<k>` once per value, with the shared filters (`type NOT IN`, `disable_correlation`, `Event.disable_correlation`, `deleted`) in the WHERE, exactly as before.
- The WHERE also carries the union of the per-value expressions, so the statement never considers a row that could not contribute to any sum.
- `saveMany()` is still called once, with the same rows.
- Nothing else in the file changes.

**Why not one grouped query**

A stored value is truncated to 191 characters and is therefore matched as a *prefix* of the attribute values (`LIKE '<value>%'`). Two stored values can match the same attribute row, so the per-value conditions do not partition the rows and cannot become a `GROUP BY` without changing what is counted. Summing one expression per value over a single pass is the aggregation that preserves the current matching exactly — the prefix semantics are deliberately left as they are.

**Why the counts are identical**

1. The LIKE patterns are quoted through the same `DboSource::value($v, 'string')` (PDO::quote) that Cake uses for `'field LIKE' => $v`, so the predicate text is byte-identical.
2. `NOT (type IN (...))` is rendered as `type NOT IN (...)` — equivalent on a `NOT NULL` column.
3. Containable's `LEFT JOIN events` plus `Event.disable_correlation = 0` in the WHERE is equivalent to the explicit `INNER JOIN` used here.
4. The WHERE is the union of the summed expressions, so every row it drops contributes 0 to every sum; a row matching a value through both `value1` and `value2` is still counted once.

**How it was verified**

- `php -l` on both files.
- `app/Test/OverCorrelatingValueOccurrencesTest.php` — 150 values produce 2 statements instead of 150, every value receives the count of its own expression, a value with no match counts as 0, and the generated condition (including quoting of a value containing an apostrophe, the join and the four shared filters) is asserted literally. All four cases fail on 2.5 and pass with the change.
- The occurrence values themselves need a database; the review environment had none, so the numbers produced by the new statement should be spot-checked against the previous ones on a real instance before release.

#### Questions

- [ ] Does it require a DB change? — No.
- [ ] Are you using it in production? — No, not yet.
- [ ] Does it require a change in the API (PyMISP for example)? — No.

---

### Local verification

- `./app/Vendor/bin/parallel-lint --exclude app/Lib/cakephp/ --exclude app/Vendor/ -e php,ctp app/` — no syntax error found.
- `./app/Vendor/bin/phpunit app/Test/` (PHPUnit 8.5.54, the version pinned in `app/composer.json`) — **596 tests**, no new failures or errors.
- Baseline on unmodified `2.5` @ `0c9f682` is 592 tests with 2 errors and 1 failure (`AttributeValidationToolTest::testDomainModify`, a missing local IDN capability). Those same 3 are present before and after this change and are unrelated to it.


elhoim:
### CI note

This PR's CI has been run three times. Each leg has passed at least once, and each failure was a different, already-known-flaky test:

| Run | ubuntu-22.04 / PHP 8.1 | ubuntu-24.04 / PHP 8.3 |
|---|---|---|
| 1 | pass | `GpgToolTest::testSignAndVerify` — `Crypt_GPG_BadPassphraseException: Cannot sign data. No passphrase provided.` |
| 2 | pass | `testlive_comprehensive.py::test_attribute` — `requests.exceptions.ConnectionError: RemoteDisconnected` |
| 3 | `testlive_comprehensive.py::test_user_settings` — `AssertionError: 4 != 3` (`len(mail_logs)`) | pass |

None of these is reachable from this change. `OverCorrelatingValue::generateOccurrences()` has exactly two callers — `AdminShell` (`cake Admin jobGenerateOccurrences`) and `CorrelationsController::generateOccurrences()` (the *Generate occurrences* button) — and neither is exercised by the failing tests. The PHP diff is confined to that one method plus a new test file. I also checked the server `error.log` on the run that closed the connection: no PHP fatal and no segfault, so nothing crashed.

Local verification of this branch, using the versions CI pins:

- `./app/Vendor/bin/parallel-lint --exclude app/Lib/cakephp/ --exclude app/Vendor/ -e php,ctp app/` — no syntax error found.
- `./app/Vendor/bin/phpunit app/Test/` (PHPUnit 8.5.54) — 596 tests, no new failures or errors versus the unmodified `2.5` @ `0c9f682` baseline.

I have stopped re-running CI rather than keep retrying a flake. Happy to rebase or rerun if you would like another data point.

PR #11142, opened 2026-09-17
fix: defer collection view initialization until CSRF token is available

microblag:
## What does it do?

Fixes collection event lookups failing with “The request has been black-holed” after the REST security changes.

The collection view started its AJAX request before the layout assigned `csrfToken`, sending an empty `X-CSRF-Token` header. Run collection
initialisation on DOM ready so the token is available before requesting `/events/restSearch.json`.

## Questions


iglocska:
Thanks a lot!

PR #11143, opened 2026-09-17
fix: use CSRF header validation for TAXII discovery requests

microblag:
## What does it do?

Fixes TAXII discovery requests failing with “‘_Token’ was not found in request data” after session-authenticated REST requests stopped bypassing form
security.

Use the existing CSRF-header handling for `getRoot` and `getCollections`. This skips CakePHP’s form-field hash validation for these AJAX requests while enabling CSRF validation. Add the CSRF header to the classic UI’s dropdown-population requests; the Overmind caller already sends it.

## Questions


iglocska:
Thanks a lot, much appreciated!

PR #11144, opened 2026-09-17
fix: restore session AJAX requests under CSRF validation

microblag:
## What does it do?

Fixes browser AJAX requests failing with “The request has been black-holed” after the REST security changes.

- Defer Insights collection view and event collection-chip initialisation until DOM ready, so the layout has assigned the CSRF token before requests
run.
- Send the CSRF header when loading report galaxy matrices and event timelines.
- Skip form-field hash validation for event timeline requests while retaining CSRF protection.
- Replace the TAXII discovery CSRF bypass with header-based validation and add the header to the classic UI’s dropdown-population requests.

## Questions


iglocska:
Thanks a lot!

PR #11148, opened 2026-09-20
[medium] perf: [authkey] indexed HMAC lookup before the bcrypt candidate loop

elhoim:
## BLUF

- **Priority: medium.** Replaces #11081, which was written before the schema-migration ledger landed and can no longer be applied.
- **Problem** — with `Security.advanced_authkeys` enabled (the shipped default), `AuthKey::getAuthUserByAuthKey()` selects every row sharing the 4-character prefix pair and runs a **bcrypt verify per candidate**: ~105 ms of pure hashing on *each* authenticated REST call, unamortised.
- **Fix** — an indexed `auth_keys.authkey_hmac` column, keyed by the existing `hmac_key.php` secret, tried as one indexed `SELECT` first. The bcrypt path is untouched and stays as the fallback, lazily backfilling the column.
- **Cost** — **one file** under `app/Lib/Migration/Migrations/`. `AppModel`, `INSTALL/MYSQL.sql` and `db_schema.json` are untouched. No external behaviour change.
- **CI** — will be red on `testlive_comprehensive::test_search_publish_timestamp`, *because the optimisation works*. Fix is MISP/PyMISP#1482, still open. See **CI status**. Opened as a draft for that reason.

---

## What changed since #11081

#11081 carried its schema change as legacy update `case 160` in `AppModel::updateDatabase()`, plus hand-edits to `INSTALL/MYSQL.sql` and `db_schema.json`. Since then the ledger landed (`0265de92`, `8293ed43`, `e2e93726`), `AppModel::DB_CHANGES_FREEZE` is `159`, and `findUpgrades()` enforces it at runtime. That approach is now not merely stale but **dead code, and red on two counts**: `LegacyCorpusFreezeTest::testTheCorpusDoesNotReachPastTheFreeze()` fails outright on a `160 => false` entry.

This branch is the same change with the schema half rewritten as `Migration_20260920_131500_auth_keys_authkey_hmac` — a guarded `addColumn` plus its index, declared once and rendered for whichever engine is in front of it.

`INSTALL/MYSQL.sql` and `db_schema.json` are deliberately **not** touched, following `bf69455a` (the `roles.perm_ai_tools` port), which touches only its migration file:

- `BaselineGenerator` describes `INSTALL/<ENGINE>.sql` as "the baseline a fresh instance loads **before the upgrade system carries it the rest of the way**", seeded at 159. `MigrationManager::pending()` is discovered ids minus applied ids, so on a fresh install the ledger is empty and every migration runs. The column therefore arrives the same way on a fresh install as on an upgrade.
- The migration stub states plainly that `db_schema.json` "is not yours to update — it is regenerated from a clean build before a release".

As a side effect this **removes the one wart #11081 had to apologise for**: that its `ALTER` would log a duplicate-column error on a fresh install because `MYSQL.sql` already carried the column.

## The change itself

`auth_keys.authkey_hmac` holds `hash_hmac('sha512', $authkey, $hmac_key)` under the **same** `app/Config/hmac_key.php` secret MISP already uses for its Redis fast-authkey cache (`AppController.php:1505-1517`). `getAuthUserByAuthKey()` tries a single indexed equality SELECT on it before falling back to the unchanged bcrypt candidate loop.

- On a successful bcrypt verify the HMAC is written back via `updateAll` — callbacks bypassed, so no spurious audit rows. Lazy migration, because the plaintext key is in hand exactly at that moment and existing bcrypt rows cannot be converted offline.
- `beforeValidate` populates the column for newly created keys.
- Both the lookup and the `beforeValidate` write are gated on `hasField('authkey_hmac')`, so code deployed ahead of the migration falls back silently instead of 500-ing every REST auth request. `MigrationRunner` calls `cleanCacheFiles()`, so `hasField` flips to true once the migration lands.
- A missing, unwritable or **rotated** `hmac_key.php` produces a miss, never a false accept: `getHmacKey()` returns null and every caller takes the original path. A rotated key re-migrates each row on its next slow-path hit.

## Why it is hot

Tier **T3**. `AppController::beforeFilter` → `__loginByAuthKey` → `_checkAuthUser` → `AuthKey::getAuthUserByAuthKey`, once for **every** REST/automation request.

The gate: `Security.advanced_authkeys` must be true — its *setting-definition* default is `false` (`Server.php`), but `app/Config/config.default.php:11` ships it as `true`, so a stock install of recent MISP is on this path. `Security.api_key_quick_lookup` must be false (the default). And `Security.authkey_keep_session` defaults to false, with `AppController` explicitly destroying the session after every authed REST request — so even a cookie-carrying client (PyMISP uses a `requests.Session`) gets no reuse. One bcrypt verify per request, forever.

The Redis fast-lookup that already exists is gated on `api_key_quick_lookup`, which is `SETTING_CRITICAL`, defaults to false, and trades up to `api_key_quick_lookup_expiration` seconds (default 180) of **revocation lag**. This change has zero revocation lag: deleting the row invalidates the key immediately.

## Speedup

**MEASURED**, against seeded tables carrying genuine `password_hash(..., PASSWORD_BCRYPT)` values with the real column/index set, 1000 rows, 10% expired, 5% deliberately colliding on both the 4-char start and the 4-char end. The user SELECT with its three `belongsTo` LEFT JOINs is inside **both** denominators; the fallback branch and the backfill UPDATE are modelled on the patched side.

```
phase A (authkey_hmac all NULL): mismatches=0  fastpath=0   slowpath=1600
phase B (fully migrated):        mismatches=0  fastpath=900 slowpath=700
TOTAL CORRECTNESS MISMATCHES: 0   (1600 mixed inputs x 2 states = 3200 comparisons)

ORIGINAL: 146.174 ms/resolution
PATCHED :   2.194 ms/resolution
SPEEDUP :  66.61x
bcrypt cost in use: {"algo":"2y","algoName":"bcrypt","options":{"cost":10}}
single password_verify: 104.51 ms
```

**DERIVED.** That container's bcrypt is pathologically slow (104.5 ms/verify, ~10-20x a server-class CPU) and the bench is a raw-PDO mirror, so ORM overhead is absent from both sides. Against a server-class 50 ms bcrypt: `(50 + 2.19) / 2.19` = **~24x**. Against a deliberately lowballed 5.3 ms bcrypt: **~3.4x**. The replacement work — one HMAC-SHA512 over a 40-byte key — is under 5 microseconds. No claim is made about the end-to-end HTTP request.

## Verification

- **The migration renders correctly on both engines.** Driven through the shipped `MysqlGrammar` / `PostgresGrammar` with the offline datasources, no database:

```
== mysql ==
ALTER TABLE `auth_keys` ADD `authkey_hmac` varchar(128) CHARACTER SET ascii COLLATE ascii_general_ci DEFAULT NULL AFTER `authkey_end`;
ALTER TABLE `auth_keys` ADD INDEX `authkey_hmac` (`authkey_hmac`);

== pgsql ==
ALTER TABLE "auth_keys" ADD "authkey_hmac" varchar(128) DEFAULT NULL;
CREATE INDEX "idx_auth_keys_authkey_hmac" ON "auth_keys" ("authkey_hmac");
```

  The MySQL render is byte-identical to the hand-written `ALTER` #11081 carried — which is the point: the same statement, now declared once and portable. `Console/cake Admin migrationApply --dry-run --id 20260920_131500_auth_keys_authkey_hmac` gives the same thing on an instance with a database; it needs one because `up()` guards on `inspector()->hasColumn()`, so it could not be run here.
- `php -l` clean on every changed file — what CI's `parallel-lint` step covers.
- Correctness assertion as above: 0 mismatches, both migration states. The 700 phase-B slow-path hits are exactly the expired and unknown keys, which correctly still resolve to `false`.
- Behaviour preservation reviewed: identical field list and expiration predicate on both paths; the extracted `authUserForAuthKey()` reproduces the original body verbatim including the `if ($user)` guard; no in-place rewrite of `authkey` exists anywhere (`resetAuthKey` expires the old row and creates a new one) so `authkey_hmac` cannot go stale against it; CRUD `edit` whitelists only `comment/allowed_ips/expiration/read_only`; all four `AuthKeysController` `afterFind` sites unset it beside `authkey`, so REST responses are unchanged.
- **The migration is exercised by CI, against a real upgrade.** The `Legacy upgrade path` step is **green on both legs**: it builds a database from `v2.5.45:INSTALL/MYSQL.sql`, runs `Admin runUpdates` (applying this migration), then `schemaDiagnostics`, asserts `migrationStatus` reports *"Everything on disk has been applied."*, and finally runs `verifyInstallBaseline --connection legacy --database misp`, which reports **"106 tables compared between \"legacy\" and the reference \"misp\": 0 findings"**. That last check is what confirms leaving `INSTALL/MYSQL.sql` alone is correct: the upgraded schema and the fresh-install schema agree.
- **One expected diagnostic.** `schemaDiagnostics` prints ``Column `authkey_hmac` is indexed but should not``. That is `db_schema.json` being stale, which the migration stub calls out as expected — it "is regenerated from a clean build before a release, so schemaDiagnostics reporting this change as a difference until then is expected". The step still passes; no action needed, but worth knowing before it is read as a defect.
- **No HTTP-level REST authentication test was performed by hand**, beyond what the live PyMISP suite exercises on every authenticated call.

## Risk

**MEDIUM.**

- **`app/Config/hmac_key.php` becomes load-bearing for lookups.** Losing or rotating it silently degrades every key back to bcrypt until re-migrated — correct, just slow. The fallback is therefore *kept*, not deleted.
- **New side effect:** `getHmacKey()` now creates `hmac_key.php` on any advanced-authkey resolution or key creation, whereas previously it was only created when `Security.api_key_quick_lookup` was enabled. Unlike `AppController`'s writer, this one is guarded by `is_writable()` and writes **atomically**: a `tempnam()` file in the same directory, then `rename()`. Two concurrent first-requests cannot interleave a partial file; the loser discards its key and re-reads the winner's. The file is created **0644**, matching what `AppController`'s writer produces under the usual umask — the CLI and the web server commonly run as different users, and whichever creates it first, the other has to be able to read it. Tightening that is a real improvement but has to be done to both writers at once, so it is not in scope here. The read side is guarded by `is_readable()`, so an existing but unreadable key file means no fast path rather than a warning on every request. The file is gitignored, so it does not dirty a working tree, and it is not externally observable in any response.
- **One extra UPDATE per key** on its first post-deploy use (the lazy backfill), via `updateAll`, so no audit/syslog rows. Expired rows are never backfilled and keep taking the correct `false`-returning path.
- **Theoretical row-selection difference:** the fast path is a `find('first')` on the `authkey_hmac` index, the fallback a loop over the start/end candidate set. If two non-expired rows ever hold the *same* plaintext authkey (possible via caller-supplied keys), the two paths could resolve to a different `auth_keys.id`. Both are valid, authorised rows and `$user` derives from `user_id`, so no ACL or response-shape change. The ordering among equally-valid duplicates was unspecified before.
- **Security posture is unchanged.** Storing a keyed hash is weaker than bcrypt only against an attacker holding both the DB *and* the key file — who already has the Redis fast-lookup surface MISP ships today, under the very same secret. Authkeys are 40 chars of `[a-zA-Z0-9]` (~2^238), so a slow hash buys nothing against offline brute force. Comparison is a DB equality on a hash, not on user input.
- **Follow-up, not in this diff:** the backstop sanitise lists in `app/Model/AuditLog.php` and `app/Model/Log.php` still enumerate `['password','api_key','authkey',...]` without `authkey_hmac`. No demonstrated leak — `AuditLogBehavior` and `SysLogLogableBehavior` (both patched here) mask the value before it reaches either, and the HMAC is not a usable credential since authentication still requires the plaintext — but a future direct writer of an `AuthKey` change array would trip over the inconsistency.

## CI status — unblocked, pending a submodule bump

MISP/PyMISP#1482 **merged on 2026-09-20** and is now PyMISP `main` HEAD (`e99bc200`).

That does not turn this PR green on its own, because MISP vendors PyMISP as a **submodule** and CI runs `testlive_comprehensive.py` from the pinned checkout. `2.5` pins `c6bde820`, 15 commits behind, which predates the fix. So this PR stays red until the pin moves — a maintainer action, and not something that belongs inside this diff, since bumping drags in every other PyMISP change in between.

**Verified that the bump is sufficient.** On a scratch branch carrying these two commits plus the submodule at `e99bc200`, MISP's own workflow was run and passed on both legs ([run](https://github.com/elhoim/MISP/actions/runs/35701368817)):

```
test_search_publish_timestamp PASSED
84 passed, 2 skipped, 8 warnings in 168.14s
```

Zero failures. So the only thing between this PR and green is the pin.

**What the bump costs.** `tests/check_requirements.py` requires `PyMISP/pyproject.toml`, `requirements.txt` and `AppController::$pyMispVersion` to agree, so moving the submodule to `e99bc200` also means `2.5.34.2` → `2.5.34.4` in the latter two. Three lines total.

## Questions

- **Does it require a DB change?** Yes — one ledger migration, `20260920_131500_auth_keys_authkey_hmac`.
- **Are you currently using it in production?** No.
- **Does it change the API / PyMISP?** No.


elhoim:
Pushed `b7076cc`, correcting a regression I introduced in the first commit.

I had the new `hmac_key.php` writer create the file `0600`. The CLI and the web server commonly run as different users, so whichever process created it first left the other unable to read it — and since this path now runs on **every** authenticated REST request, `include()` emitted a warning per request. With `debug` on, that warning is rendered into the response body, so CI did not fail the one live test I expected: it failed *every* authenticated call with "Authentication failed".

Fixed two ways:

- The file is created `0644`, matching what `AppController`'s existing writer produces under the usual umask. `tempnam()` creates at `0600`, so this has to be set explicitly. Tightening the permissions is worth doing, but it has to be applied to both writers at once and handle the two-user case, so it is not this PR's business.
- The read is guarded by `is_readable()` rather than `file_exists()`, so an existing but unreadable key file is silently no fast path instead of a per-request warning. The bcrypt fallback resolves the key correctly either way. The creation branch is guarded on `!file_exists()` so an unreadable file is never mistaken for an absent one and overwritten.

The atomic `tempnam`/`rename` write is kept — that is what closes the concurrent-creation race, and it was not the problem.

Verified by lifting `getHmacKey()` out of the file mechanically and running it against real files: missing key → created `0644`, no warnings; existing readable key → read, no warnings; existing unreadable key → null, no warnings, no stray temporary files.

Expectation for this run: green on lint and `phpunit app/Test/`, and red only on `testlive_comprehensive::test_search_publish_timestamp`, which stays blocked on MISP/PyMISP#1482.

elhoim:
CI re-run after `b7076cc` is now exactly the expected shape on both legs:

```
ubuntu-24.04 / 8.3:  1 failed, 83 passed, 2 skipped, 8 warnings in 147.31s
ubuntu-22.04 / 8.1:  1 failed, 83 passed, 2 skipped              in 145.37s

FAILED tests/testlive_comprehensive.py::TestComprehensive::test_search_publish_timestamp - AssertionError: 0 != 1
```

The `hmac_key.php` permission warnings are gone entirely — zero occurrences in either log, where the previous run had them on every authenticated request. Everything else passes.

The one remaining failure is the known race, and the runtime is the evidence for why it is this PR that trips it: **145-147s against the 190-216s the same 85 tests take on other branches**. `AssertionError: 0 != 1` is the `publish_timestamp='5s'` search matching nothing because it now runs before the background publish worker has ticked.

Still blocked on MISP/PyMISP#1482, which makes that test wait for the worker instead of relying on incidental latency. Staying a draft until it lands.

elhoim:
MISP/PyMISP#1482 merged on 2026-09-20, so the blocker is gone — but not automatically, and I have measured rather than assumed the rest.

**The merge alone does not turn this green.** MISP vendors PyMISP as a submodule and CI runs `testlive_comprehensive.py` from the pinned checkout. `2.5` pins `c6bde820`; the fix is in `e99bc200` (PyMISP main HEAD), 15 commits later. Until the pin moves, this PR keeps failing that one test.

**The bump is sufficient — verified.** I put these two commits on a scratch branch with the submodule at `e99bc200` and ran MISP's own workflow on it: https://github.com/elhoim/MISP/actions/runs/35701368817

```
test_search_publish_timestamp PASSED
84 passed, 2 skipped, 8 warnings in 168.14s
```

Both legs green, zero failures. So the pin is the only thing left.

**What the bump costs:** `tests/check_requirements.py` requires `PyMISP/pyproject.toml`, `requirements.txt` and `AppController::$pyMispVersion` to agree, so moving the submodule also means `2.5.34.2` → `2.5.34.4` in the latter two. Three lines. I have deliberately kept all of that **out** of this PR — bumping the submodule pulls in unrelated PyMISP changes and is yours to decide, not a perf patch's business.

**On the migration**, which I had previously flagged as untested: it is not. The `Legacy upgrade path` step is green on both legs here, and it is a real test — it builds a database from `v2.5.45:INSTALL/MYSQL.sql`, runs `Admin runUpdates`, `schemaDiagnostics`, asserts `migrationStatus` says "Everything on disk has been applied.", then runs `verifyInstallBaseline`, which reports **"106 tables compared between 'legacy' and the reference 'misp': 0 findings"**. That confirms leaving `INSTALL/MYSQL.sql` untouched is right: the upgraded and fresh-install schemas agree.

One diagnostic worth pre-empting: `schemaDiagnostics` prints `Column \`authkey_hmac\` is indexed but should not`. That is `db_schema.json` being stale, which the migration stub names as expected until it is regenerated before a release. The step passes.

Marking ready for review — the code is done and proven; what remains is your call on the pin.

